PR #14, opened 2020-03-07
Merging development (with logging) into master

Felienne:


app.py
```diff
@@ -1,30 +1,23 @@
 import hedy
 import json
+import os
+import requests
```
rix0rrr: This import can be removed

app.py
```diff
@@ -1,30 +1,23 @@
 import hedy
 import json
+import os
```
rix0rrr: This one too

PR #22, opened 2020-03-17
Correct grammar in demo code for Level 2 Ask, en/es 

dragondave:
["color is ask" is correct grammar, not "ask color is"]


rickbeerendonk:
The English version is fixed (in the meantime) by commit 026fb34.
The Spanish version is not fixed yet.

PR #27, opened 2020-03-18
Spanish level 3 now matches latest changes in NL/EN.

mivalencia:


Felienne:
Thanks, that was quick!

PR #28, opened 2020-03-19
Update README.md

arjenpdevries:
Fixed markdown with what I think you meant.


Felienne:
Thanks that is what I meant (the readme was exported from a Latex file and apparently not exactly in the right way)

PR #29, opened 2020-03-19
Word "Level" can now be translated into other languages.

rickbeerendonk:
Screenshot 2020-03-19 at 17 09 38


Felienne:
Cool! Because it is Level in Dutch and English I did not consider it needed translation, oops.

PR #32, opened 2020-03-19
Fix typos in NL and ES.

rickbeerendonk:


Felienne:
Thanks for all the fixes again!!

PR #58, opened 2020-04-06
Extend support for if statements; allow in list checks

### LINKED ISSUE #41: Request: additional operators in `if` condition
I would like to have used `or` and maybe also having `in` would be good (one kid had a program that tested if the input is in a range she put in a list).

arjank:
Updated grammars for levels 4 - 7 to accept both `foo is bar` and `foo in bar` as conditions for the if-statement. In the latter case `bar` should be a list, because generating lists in place is not supported.

Also the implementation of `ifelse` has been moved so it is closer to the implementation of `if`.

This fixes #41, at least partly. I have opted not to implement `or`, because supporting both `and` and `or` should imo also require support for parentheses.

### DIFF
--- a/grammars/level4.txt
+++ b/grammars/level4.txt
@@ -18,10 +18,12 @@ assign : var " is " textwithspaces
 invalid: /^(?!(print|if|and )).+/ //everything that is not the start of a regular command
 
 // new commands for level 4
-ifelse : "if " equality_check " " command " else " command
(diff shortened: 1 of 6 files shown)


Felienne:
Cool, thanks!
I thought about doing or, but indeed then you need brackets which we clearly do not want yet. This night turn out to be hard for kids too, but now at least we can try it and see if it leads to many errors :)

PR #63, opened 2020-04-20
Add support for repeat variable number of times

### LINKED ISSUE #49: Lesson 5: repeat
`repeat` does not take a variable as parameter?

```
n is ask Hoe vaak?
repeat n times print 'Zo vaak!'
```

arjank:
Repeat (which is introduced in Level 5) now supports both numbers and variables. Grammar has been updated for levels 5 - 7, and necessary code changes are also present. Includes tests for the updated behaviour.

Fixes #49


Felienne:
Awesome work, thanks for this!!

PR #66, opened 2020-04-22
Added 404 error page

moorlag:
(needs work, .htacces needs to be updated)

### DIFF
--- a/app.py
+++ b/app.py
@@ -20,7 +20,7 @@ from flask import Flask, request, jsonify, render_template, session, abort
 from flask_compress import Compress
 
 ALL_LANGUAGES = {
-    'en': '🇺🇸',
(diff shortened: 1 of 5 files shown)


Felienne:
Hi Ramon! 

Are you planning to finish this? You will also need to update app.pt to serve the error page when needed. If not also fine! Then I can merge this and update myself.

moorlag:
> Hi Ramon!
> 
> Are you planning to finish this? You will also need to update app.pt to serve the error page when needed. If not also fine! Then I can merge this and update myself.
Out of my league
Suggestion:
line 229 @app.route(/static/404/index.html)

Felienne:
This is too far behind, we can add the 404 separately later.

PR #67, opened 2020-04-24
 fixed small textual typos and mistakes in existing docs

jeannegoossens:
Some small Dutch and English language fixes, some apostrophes fixed, etc. 
For example in docs/en-level1.md: '...right? Then, after practicing...': using the word then here is a bit Dunglish. Or if you want to keep the sentence structure, perhaps use 'if so' ?
And in docs/nl-level2.md: Henk is used in all examples except Bert here, so it seems more consistent to work with Henk in this example as well.

// (this is my first ever contribution to open source so if I'm doing any of that wrong, with the fork and pull request etc, please let me know!)


Felienne:
Thanks!!

PR #71, opened 2020-05-20
Add French translations

sportebois:
Hello

After discovering the Hedy introduction yesterday, I found the approach so nice for children that I wanted to make it available for children around here too. 

This PR adds French translations (the Hedy grammar is still in English though, like it is for other languages too)

A second commit also fixes a few typos I spot in the English texts.

Hope this helps, thank you for building this 🙏

### DIFF
--- a/app.py
+++ b/app.py
@@ -29,7 +29,7 @@ ALL_LANGUAGES = {
     'en': '🇺🇸',
     'nl': '🇳🇱',
     'es': '🇪🇸',
-    'fr': 'ғʀ',
+    'fr': '🇫🇷',
 }
 
 LEVEL_DEFAULTS = collections.defaultdict(courses.NoSuchDefaults)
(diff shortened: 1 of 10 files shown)


app.py
```diff
@@ -29,6 +29,7 @@
     'en': '🇺🇸',
     'nl': '🇳🇱',
     'es': '🇪🇸',
+    'fr': 'ғʀ',
```
MecuSorin, line 32: Can you give me a link, or a link on how to type that symbol?

sportebois: I typed it, and to tell the truth, it was the most difficult part of the translation.
I used an online font converter (probably this one if I remember correctly https://manytools.org/facebook-twitter/smallcaps-text-tool/ ) to get the text to small caps. 

I’m not sure it’s exactly the intended formatting @Felienne expected, but when I run Hedy, the dropdown menu with the languages looks fine with this, so I guess it’s correct.

Felienne: Yeah it all works fine, that a lot, it just misses the flag, but I can add that myself!

Felienne:
btw, what I do is I google "French flag emoji" and then I copy it into the code :)

PR #73, opened 2020-06-05
Fix contact and add GitHub link

stroborobo:
Hi!

Just watched your presentation at fsharpConf, that was awesome, thank you!

Noticed a broken link and included a link to the project's GitHub page, the dutch part is Google translated and might be bogus. Please let me know if I need to change it or you like to have it somewhere else maybe.


Felienne:
Thanks for this! The Dutch in correct, so thanks Google translate too :)

PR #91, opened 2020-08-22
Update es.yaml

nacholopex:
Minor fix in some sentences. 😄
At level 2 I wrote wrong the print, so when the user run the command the console showed him` Your favorite blue is blue` because the variable color was wrote 2 times.
Need to study again first level 😂 

Also I added `¿` in two sentences that I didn't see


Felienne:
Haha, yeah Hedy can be a bit tricky some times :)

PR #93, opened 2020-10-02
Working Translation to Brazilian Portuguese

brunopessanha:
🇧🇷


Felienne:
Thanks @brunopessanha we are going to publish the Portugese version today! My knowledge of Portugese is not much :) How different is Br-pt from Pt-pt? Will Portugese kids be able to use this too?

brunopessanha:
> Thanks @brunopessanha we are going to publish the Portugese version today! My knowledge of Portugese is not much :) How different is Br-pt from Pt-pt? Will Portugese kids be able to use this too?

Yes, I think they will be able to understand too. In written form, the two versions shouldn't differ much. They may find strange the use of `você` instead of `tu` for `you` or the use of "americanisms", which are more common in Brazil, like `printar` for the verb `to print`, but they should be able to uderstand.

PR #102, opened 2020-11-07
Changed all Lark grammar files to have .lark extension

HenryLeC:
Lark Grammar files are expected to have a .lark file extension

[Lark Docs](https://github.com/lark-parser/lark/blob/master/README.md)

ps. This is mainly because I was super confused about how the level files were written until I looked through the code and saw the Lark parser being used.


HenryLeC:
Oops Unit Tests need to be rewritten :)

HenryLeC:
Cool, it works now

Felienne:
Thanks that is a better idea!

PR #107, opened 2020-12-01
Use POST /parse instead of GET /parse to submit programs.

fpereiro:
Currently we use a GET request to submit a program from the client to the server for parsing (`GET /parse`). GET requests are recommended to not exceed 2048 characters, because of limited support for long URLs on some browsers (such as IE11).

By switching to POST, we can sidestep this problem. It is also somewhat more RESTful to use POST rather than GET for this type of operation, since we're in effect submitting data.

This SO answer has further info on this subject: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

### DIFF
--- a/app.py
+++ b/app.py
@@ -72,12 +72,22 @@ if not os.getenv('HEROKU_RELEASE_CREATED_AT'):
     logging.warning('Cannot determine release; enable Dyno metadata by running "heroku labs:enable runtime-dyno-metadata -a <APP_NAME>"')
 
 
-@app.route('/parse/', methods=['GET'])
+@app.route('/parse', methods=['POST'])
 def parse():
-    # Retrieve the name from url parameter
-    code = request.args.get("code", None)
-    level = int(request.args.get("level", None))
-    email = request.args.get("email", None)
+    body = request.json
+    if not body:
+        return "body must be an object", 400
+    if 'code' not in body:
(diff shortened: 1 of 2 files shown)


Felienne:
Hi @fpereiro!

Can you rebase this one or otherwise mix the merge issues? I want to test this locally extensively before merging because it might impact everything.

Felienne

fpereiro:
@Felienne  Just merged the latest master onto the branch; there should be no conflicts now :)

app.py
```diff
@@ -72,11 +72,18 @@
...
 
-@app.route('/parse/', methods=['GET'])
+@app.route('/parse', methods=['POST'])
 def parse():
-    # Retrieve the name from url parameter
-    code = request.args.get("code", None)
-    level = int(request.args.get("level", None))
+    body = request.json
+    if not body:
+        return "body must be an object", 400
```
rix0rrr, line 79: This looks weird. Are you sure this is the way to return a 400 response?

fpereiro: Good question. I just checked and it does seem to return the string as the body and the integer as the response code. I found this line in this official doc (http://www.flaskapi.org/api-guide/status-codes/): `return content, status.HTTP_404_NOT_FOUND`

PR #121, opened 2021-02-03
Add some missing translations

fpereiro:
- Adds missing translations for navigation bar.
- Link to try Hedy on the right language.
- Missing press releases in other languages.


fpereiro:
@Felienne I added some missing articles in the Press section so that all languages are even - if that's out of scope or not desired, I can remove it from the PR.

PR #132, opened 2021-02-11
Scroll editor on error showing and close error box

fpereiro:
Resize editor when showing warning or error; add X to close warning or error.


Felienne:
Seems to work in alpha, merging!

image

PR #133, opened 2021-02-12
signup.html: Minor Usability improvements

jdittrich:
* Birth Year as integer (+placeholder) (Out of habit I gave it a date (did not work) and then an ISO-8601 (which worked)) 
* Mail as appropriate type to call up correct soft keyboard on phones.

### DIFF
--- a/templates/signup.html
+++ b/templates/signup.html
@@ -18,7 +18,7 @@
     <input id="password_repeat" type=password>
     <br><br>
     <label>{{auth.birth_year}}</label>
-    <input id="birth_year" placeholder="e.g. 1995" type="number">
+    <input id="birth_year" placeholder="1995" type="number">
     <br><br>
     <label>{{auth.gender}}</label>
     <select id="gender">
--- a/templates/signup.html
+++ b/templates/signup.html
@@ -9,7 +9,7 @@
     <input id="username">
     <br><br>
     <label>{{auth.email}}</label>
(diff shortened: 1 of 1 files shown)


fpereiro:
Hi @jdittrich ! This looks great. We apply `parseInt` to the input in client side js, so that in case an ISO date is entered we only keep the year. But it's better to catch the error earlier.

@Felienne do you think that adding "e.g." would be confusing for kids? Perhaps we can simply put a year as placeholder.

templates/signup.html
```diff
@@ -18,7 +18,7 @@ <h2>{{auth.create_account}}</h2>
     <input id="password_repeat" type=password>
     <br><br>
     <label>{{auth.birth_year}}</label>
-    <input id="birth_year">
+    <input id="birth_year" placeholder="e.g. 1995" type="number">
```
fpereiro, line 21: Maybe we can omit the "e.g" in case it confuses some kids?

fpereiro (approved):

Felienne:
> do you think that adding "e.g." would be confusing for kids? Perhaps we can simply put a year as placeholder.

Year I think e.g. is hard. An example placeholder would be better

jdittrich:
deleted the e.g.

fpereiro:
Merged. Thanks again @jdittrich !

PR #154, opened 2021-02-17
Output field should also have a scroll bar

### LINKED ISSUE #144: Output field should also have a scroll bar
Not really pretty that the output field is enlarged when a lot of lines are printed:

image

fpereiro:


Felienne:
Wow that was quick! Right in time for Friday's final lesson.

Issue #146, opened 2021-02-15
UI improvements in profile view


fpereiro:
- Make all clickable elements as buttons (delete account, log out).
- Use style of buttons from other parts of the app.

Issue #145, opened 2021-02-15
Add suggestion to reset password on entry of wrong password


Felienne:
![image](https://user-images.githubusercontent.com/1003685/107954060-451c5c80-6f9c-11eb-84eb-bec5a1cd4ba8.png)

Maybe in the red bar?

fpereiro:
@Felienne Good idea! What about adding by default a "Forgot password" link below "Nog geen account", so that it is always visible in this view?

Felienne:
It think that's the way we have it now, always visible?

![image](https://user-images.githubusercontent.com/1003685/107958659-9d565d00-6fa2-11eb-806f-cb23833b2f81.png)

The issue with that is:

1. it is in the way and distracts kids that do not need it
2. when kids do need it, it is not in the place that they are reading (in the red box)

I propose:
not visible initially, when they enter a wrong ww, if shows in the red box.

fpereiro:
Let me add it to the red box, but still keep the other one. Compromise :).

If we take the existing link away, the only way that someone could find the "forgot password" would be to try with wrong credentials, which is not necessarily intuitive for everyone.

Virtually every application with login/signup has a link to "forgot password" at the bottom of the form, because users have learned to expect it there.

Would that be OK?

Felienne:
Fine by me to have both, I think it will also all be a bit clearer if we use the same style for everything (now it "hangs" a bit loose as a link)

fpereiro:
Agreed, I'll convert that link into a button as well!

PR #155, opened 2021-02-17
Auth UI improvements.

### LINKED ISSUE #143: When a user is not logged in, and they have made a program, they are always send to Register
This is confusing to kids that already have an account! Can we change the message to say something like: ...create an account or login if you already have an account?

image

fpereiro:
- Closes #143, #145 and #146.
   - Adds consistent style (buttons) on all five auth views: /my-profile, /signup, /login, /recover and /reset.
   - Redirects users with unsaved programs to /login instead of /signup. The message shown here is also different, to reflect that.
   - Improves error/success message visibility in /my-profile.
   - Shows "make an account" button when submitting invalid credentials in /login.
   - After invalid login, if user chooses to create an account, that username (or email) is preserved in the /signup form.
   - Adds internationalized "quit without saving progress" for future use.
   - Implementation note: buttons that function as links are embedded within forms to perform redirects - that explains the extra forms added to the templates :).


Felienne:
> * Implementation note: buttons that function as links are embedded within forms to perform redirects - that explains the extra forms added to the templates :).

Maybe we should keep something like a list of design decisions somewhere? We could start a new .MD where we keep track of this is a lightweight way?

Issue #160, opened 2021-02-18
Double echo in level-1


MarinusVanDijk:
![image](https://user-images.githubusercontent.com/37830964/108326890-10421c80-71cb-11eb-80a5-d55cbe861a72.png)

Then hit the run it button not the enter button.

![image](https://user-images.githubusercontent.com/37830964/108327076-454e6f00-71cb-11eb-8870-9a64af63ef02.png)

Then hit the enter button after typing Hallo instead of Hello

Results in:

![image](https://user-images.githubusercontent.com/37830964/108327226-6f079600-71cb-11eb-8515-357999815641.png)

I expected only one Hallo.

Felienne:
Not sure what happens here and how we could prevent it, but this again is not a thing we have seen kids try, so it is not high prio at the moment.

MarinusVanDijk:
Perhaps the run code button must be suppressed when a question is answered? But is it better to fix the underlaying error. This error can pop up in no time somewhere else. I think the answer is stored twice because the answer variable is not made empty.

fpereiro:
Hi @MarinusVanDijk . This should be fixed by fixing this issue: https://github.com/Felienne/hedy/issues/147 . I'll post an update here when the fix is committed.

Felienne:
Yeah in essence this bug is a duplicate of #147, it describes a fix to this issue, so I think we can close this one.

PR #172, opened 2021-02-19
Disable code running when prompt is being shown

### LINKED ISSUE #147: Disable "run code" button when showing a prompt

fpereiro:
- Closes #147 and #160.
- Blocks running the code through the button and also through the keyboard shorcut (ALT+ENTER).
- Also puts global variable `unsaved_programs` into `window.State`, our holder for state global variables in the browser.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -30,7 +30,7 @@
     // If prompt is shown and user enters text in the editor, hide the prompt.
     editor.on('change', function () {
       if ($('#inline-modal').is (':visible')) $('#inline-modal').hide();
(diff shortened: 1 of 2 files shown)


Felienne:
Nice! 

image

Maybe as a nice low-prio addition we could show a small popup when it is disabled the first time to tell them that is happening?

fpereiro:
Hi @Felienne ! Good suggestion! But what if we do it every time instead of just the first? I think this might be useful not just the first time but every time, since once they understand the flow they'll likely stop attempting to run the code while waiting for input (and if not, the popup will remember them).

PR #188, opened 2021-02-25
fix: `parse` always uses browser language (instead of selected language)

### LINKED ISSUE #179: If Hedy is in Another language than Dutch and a mistake is made, the error message still appears in Dutch
The error message recommending print instead of prit for example still appears to be Dutch.
This also occurs for every other available language.
image

rix0rrr:
The `/parse` endpoint code used `requested_lang()`, which looks at the
query string in the URL (or the browser language) to determine the
language.

Instead, `/parse` is POSTed to and the language is in the POST body of
that request. It should be read from there.

Fixes #179.

### DIFF
--- a/app.py
+++ b/app.py
@@ -96,6 +96,10 @@ def parse():
 
     code = body ['code']
     level = int(body ['level'])
+    # Language should come principally from the request body,
+    # but we'll fall back to browser default if it's missing for whatever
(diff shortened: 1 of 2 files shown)


fpereiro:
@rix0rrr thanks for catching that!

PR #189, opened 2021-02-26
Beautifies error messages, addressing issue 184

### LINKED ISSUE #184: Catch ugly parser errors
The very least we can do is hide them and replace by "oops", but maybe with a bot of string magic we can make them less scary and more readable? They do sometimes help me in debugging with kids.

Felienne:
Closes #184


fpereiro:
Hi @Felienne 

Exciting to have this level of detail in the errors! I added some translations for Spanish and Portuguese. I also added English placeholders in French so that we can remember to translate them later.

I took the liberty of adding a conditional to the code: https://github.com/Felienne/hedy/pull/189/commits/f7b1c4c0a80c4eba46eb0e0f4559f18bff3a6622 . This prevents a 500 error if that key is not present. To test this, try running `print 'hello world` (note the lack of quote at the end).

Felienne:
What a great catch @fpereiro! Can you believe I ran this new code over all 250.000 programs in the database to check what might go wrong and I did not see this error! Will check out your patch and see if I can come up with other scenarios for this error.

fpereiro:
It all looks good, merging now!

PR #192, opened 2021-02-27
Update hedy.py

### LINKED ISSUE #162: Suggestion for misspelling commands prefers echo above ask command level-1
Nice error message when typing askquestion without space:

image

But then when typing askechoquestion without space:

image

Error message suggestion preference is the command echo.

I expect a suggestion for the ask command, because the command line starts with ask command?

And perhaps also give a suggestion for the echo command?

MarinusVanDijk:
Replaced is the closest_command() function with a new version to resolve #162

The old closest_command() function is renamed to keep the code for now.

Unit tests not added.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -6,7 +6,10 @@ import sys
 
 reserved_words = ['and','except','lambda','with','as','finally','nonlocal','while','assert','false','None','yield','break','for','not','class','from','or','continue','global','pass','def','if','raise','del','import','return','elif','in','True','else','is','try']
 
-# commands are used to suggest the right command when kids make a mistake
+#
+# Commands per Hedy level which are used to suggest the closest command when kids make a mistake
+#
+
 commands_per_level = {1: ['print', 'ask', 'echo'] ,
                       2: ['print', 'ask', 'echo', 'is'],
                       3: ['print', 'ask', 'is'],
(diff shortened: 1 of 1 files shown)


MarinusVanDijk:
It seems that an invalid command 'abc' should be suggest the command 'ask' to the kid in level-1.

My advise would be to remove this test until better algorithm is placed? For example a phonetic comparison algorithm?

Felienne:
Hi @MarinusVanDijk,

Thanks for the PR, some issues that you still need to address:

> The old closest_command() function is renamed to keep the code for now.

There is no need to keep the old code in the code base, because we have source control, we can always revert to an old version should issues arise in the future. Can you take the old code out?

> Unit tests not added.

Please add at least a few tests for new functionality. Not only to make sure your code works as intended, but also to communicate to the rest of the maintainers how the code is supposed to work, and... maybe most importantly, tests help "protect" this functionality when we make further changes. Would be a pity if new updates accidentally revert your hard work.

> It seems that an invalid command 'abc' should be suggest the command 'ask' to the kid in level-1.

When a kid types `abc` in level 1, we now suggest `ask`, which was a test I put in to test the rudimental string distance repair code. I think it is a sensible suggestion? I did not run your code locally, but it seems the new code suggests an empty command if you enter `abc`, which is not a great fallback, it would create a weird error message. Can you make sure your code outputs `ask` also?

MarinusVanDijk:
Hi @Felienne,

Thanks for your feedback.

I will close this PR and make a new one.

I improved the distance repair method and intend to integrate this with the closest_command() function.

Please see details in the next PR.

--------

Still studying how to resolve abc -> ask. 

The distance is 2 chars different, or 67% which is perhaps too much to have other repairs done correctly. 

Perhaps the solution is to replace the empty return of closest_command() with the first listed command available per level in the list of all its commands?

Or replace this unit test with the tests for the repair method?

PR #193, opened 2021-02-28
add german language

kilian-funk:
Hi Felienne,
I read about Hedy in the Hello World magazin. I would like to try it in my class. Therefore I started translating into German.
I will continue with course text as soon as I find the time.
Looking forward to seeing more levels.
Kilian


fpereiro:
Hi @kilian-funk ! Thank you for your great contribution!!

Could you add the following content into a new file at `coursedata/course/hedy`? This is necessary for hedy to run properly. With this file, I confirm that I could start the app and use the German version :).

```
---
# This file seems rather empty but the assignments all use the defaults from 'level-defaults'
course:
  - level: "1" 
  - level: "2" 
  - level: "3" 
  - level: "4" 
  - level: "5" 
  - level: "6" 
  - level: "7" 
```

Thank you again!

kilian-funk:
I missed the last file. I hope that works now.

fpereiro:
Just tested it again and it's working well, including the account section. We'll merge it shortly!

PR #200, opened 2021-03-04
Move explanation & videos to an expandable section within the main code section

### LINKED ISSUE #150: Move explanation & videos to an expandable section within the main code section
- Join videos & explanation into one page.
- Move explanation & videos to an expandable section within the main code section
- Remove second level tabs to switch between views since they're no longer necessary.

fpereiro:
- Closes #150.

- Join videos & explanation into one page.
- Move explanation & videos to an expandable section within the main code section
- Remove second level tabs to switch between views since they're no longer necessary.


fpereiro:
Todo: update non-Hedy tracks and remove old render function and old template.

Felienne:
@fpereiro I guess with the latest change we are ready to merge this?

fpereiro:
@Felienne just removed the links to the old doc pages, the function that rendered them, and the old template. I think we should ready to test & merge!

I also checked and the alternate paths (onlinemasters and space_eu) were not using the removed functionality, only Hedy.

Felienne:
Just tried it out, and it seems to work well!! A few things that can be improved:

* The Dutch translation is weird :) It should be imperative, which in Dutch is the verb - en: toon uitleg en verberg uitleg 
* I'd first show the video and then the text. Kids hate reading :) 
* Not sure I love the green, it is a bit much with the "bands" we already have. Maybe we can just use the lightblue that we have above?

fpereiro:
Hi @Felienne ! Changes pushed:

- Thanks for correcting my Dutch :)
- I tried the bar in light blue but it looks too "washed-up", so I went with the stronger blue. Let me know if that works well. The "hide explanation" bars remain in the same gray as before.
- Added logic so that videos are always shown first.

Felienne:
Looks great like this @fpereiro !!

Only the Regenwormen dice mage is missing? It seems that issue already existed, but maybe we want to fix is now because presumably more kids will read it?

image

fpereiro:
Fixed :)

PR #208, opened 2021-03-09
Set English lang query parameter when going to the Hedy editor

### LINKED ISSUE #206: In Chrome, on MacOs, translation of English does not stick between FP and Hedy
A fun mystery for you front end kinf @fpereiro :D

Here's what happens, I set the front page to English:

image

And then I click try it, I go to the Dutch version:

image

This does *not* happen for any of the other langs....?! My understanding of the website part of the app is too shallow to know what's up here!

fpereiro:
- Set English lang query parameter: useful for when browser has a non-English supported language as default and user changes the language to English.
- Closes #206


Felienne:
Yeah I guess the issue here is my browser being set to Dutch which in fact I did because I wanted to test how it would work for most Dutch kids.

PR #213, opened 2021-03-11
Block Hedy from running in Python < 3.6; update CONTRIBUTING.md

### LINKED ISSUE #210: Detect Python version and avoid running Hedy if the version is too low
Thanks to @tazaria for pointing this out!

fpereiro:
- Closes #210.
- Exit with message.
- Modify docs.

Todo: Test with pyenv that 3.6 is indeed the minimal version required


Felienne:
That was quick!!

PR #218, opened 2021-03-11
Revert "Block Hedy from running in Python < 3.6; update CONTRIBUTING.md"

Felienne:
Reverts Felienne/hedy#213

It seems to break on Heroku (not 100% sure this was the cause but rolling it back fixed it) It could also be generic "lots of traffic" because in addition to rolling back, I also upgraded our Heroku account.

Locally it worked for me, so we need to dive into what the issue is.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ Contributing to Hedy
 Help Hedy with translations
 ------------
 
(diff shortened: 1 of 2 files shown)


fpereiro:
I think this should fix it: https://github.com/Felienne/hedy/pull/219 :|

PR #226, opened 2021-03-16
Email checks on signup

### LINKED ISSUE #223: Verify email address in signup to prevent some typing errors
- Add confirm email field that hopefully doesn't allow copy-pasting from the first.
   - Mark when initial email sending fails.
   - Add functionality in the admin to change email and send again the verification email.

fpereiro:
Given the large number of invalid email addresses we receive (mostly from typos), the idea of this PR is to make them less common. 

- Add repeat email field on signup.
- Don't allow copy pasting into email fields to prevent typos. Disallowing pasting is heavy-handed and might not be a good idea - maybe we can review this empirically. Disallowing copy-cut seems to be a good idea, since we want kids to either 1) retype their email address; or 2) copy it from a good source.
- Closes #223 

Checking whether emails bounce through AWS SES is very cumbersome and hasn't been tried. If the above is not enough we might add other palliative measures in the future as needed.


fpereiro:
It'd be really good to test this especially since it might break signups. I did test it but never trust the tests from the person submitting the feature!

Felienne:
Hi @fpereiro 

I tried the branch but I see two issues.

1) it seems to allow me to paste into the second field

2) After clicking make an account, I get redirected and then get an error: 
image

Felienne:
ow I found one more, with password 123456, I get an error on the password (not sure why) and the error box is on the email address:

image

Felienne:
(maybe I miss a library or something because these are a suspicious number of things that go wrong)

This is in Chrome on Mac, btw

fpereiro:
HI @Felienne ! Just fixed two bugs you spotted:
- Typo that allowed to copy-paste on the second field, fixed now.
- There was a bug in the nl translation for letting you know that your *email* was wrong (not your password).

I cannot replicate the error you have when creating your account. Could you try again and, if you get it, when you refresh the page let me know whether you're logged in or not? I don't think you're missing libraries - if you're able at all to login/logout locally, this should also work!

Felienne:
Tested this in alpha, seems to work well now!

Felienne:
@fpereiro Sorry this needs to be merged with the new tests, I cannot deploy in on Heroky otherwise :( I though that it would run the old tests as in GH but that is not the case.

Can you 

1) accept my pr with the new tests
2) merge the new master in
3) make a new pr?

PR #229, opened 2021-03-18
Count programs per user

### LINKED ISSUE #216: Add #programs per user
Can we have the number of programs behind each users (to quickly find interesting users to explore)

fpereiro:
- Count programs per user.
- Contains script to add current count to all existing users. The script is idempotent although we should remove it soon to avoid doing a lot of calls to DynamoDB.
- Closes #216

### DIFF
--- a/app.py
+++ b/app.py
@@ -436,6 +436,10 @@ def delete_program (user, program_id):
     if not result or result ['username'] != user ['username']:
         return "", 404
     db_del ('programs', {'id': program_id})
+    program_count = 0
+    if 'program_count' in user:
+        program_count = user ['program_count']
+    db_set ('users', {'username': user ['username'], 'program_count': program_count - 1})
     return redirect ('/programs')
 
 @app.route('/programs', methods=['POST'])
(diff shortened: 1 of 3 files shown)


Felienne:
> The script is idempotent although we should remove it soon to avoid doing a lot of calls to DynamoDB.

I am not sure what you mean here, do you mean we don't cache the values and run an (expensive) query each time we load the page?

fpereiro:
Hi! It means that every time that Hedy is started, the script to load the count will run - after the first time, we can remove it. But if the script runs multiple times, apart from a few extra DB reads, everything will be consistent.

We store the program count in each user so that each time we open the admin we don't have to count the programs for each user - the reasons being both performance and saving on DB queries.

Felienne:
I am still not sure what you mean (esp with "when Hedy starts", do you mean: when it is deployed?")

Let me know when I can accept this, because the way I understand it is that right after you want to remove the init script for the counts?

fpereiro:
Sorry for the lack of clarity :) I meant that this script to update the db will run every time that Hedy is deployed - that is, when the app starts for the first time.

I can do a separate PR afterwards to remove this script, once we check that the values have been updated in the DB.

Felienne:
Yes, got it now! I'll see the other one in a new branch. Can you merge this one yourself, after the new version of the tests? Because I fear it will not work otherwise because the new tests will run and are broken.

PR #236, opened 2021-03-22
Revert "Update fr.yaml"

Felienne:
Reverts Felienne/hedy#231

Seems to be an error in this yaml, need to be checked


Felienne:
@fpereiro maybe you can fix the yaml here, at one point?

PR #239, opened 2021-03-23
A/B testing

### LINKED ISSUE #203: A/B testing support
- Set up environment.
- Set up deterministic redirect in code.
- Check if we can make HTTPS and account logic work in this way.

fpereiro:
- Reverse proxies requests to a test environment. This avoids issues with cross-domain, HTTPS and showing a different address on the URL bar.
- Users/sessions are redirected deterministically according to a function that can control the percentage of users/sessions to be redirected.
- Closes #203


app.py
```diff
@@ -70,6 +71,44 @@
 
 app = Flask(__name__, static_url_path='')
 
+def into_number (string):
```
rix0rrr, line 74: You're looking for a hash function, and this not a great one because it will not distribute across the available key space very randomly.

Fortunately, Python has hash functions built in. You can get specific ones like md5 or sha256, but `hash()` provides a hashcode that Python itself uses for dictionaries:

```python
>>> hash('asdf')
6285382739108471212
```

The range is `-sys.maxint..sys.maxint`.

Note: this hash function will be different every time the Python program is restarted. That may or may not be what you want. Think about that, because your user may alternatingly be in or out of the treatment group every time the server is redeployed.

Use one of the hashes from `hashlib` or use a more common hash function if you need that.

app.py
```diff
@@ -70,6 +71,44 @@
...
+    for char in string:
+        total += ord (char)
+    return total
+
+def redirect_ab (request, session):
+    # If the user is logged in, we use their username as identifier, otherwise we use the session id
+    user_identifier = current_user(request) ['username'] or str (session_id)
+
+    # This will send 50% of the requests into redirect.
+    redirect_flag = (into_number (user_identifier) % 100) < 50
```
rix0rrr, line 85: The % of how many users you want to test should probably be a parameter. Also, 50% seems extremely high.

app.py
```diff
@@ -70,6 +71,44 @@
...
+        # If we enter this block, we will reverse proxy the request to the REDIRECT_AB_TEST environment.
+        if (redirect_ab (request, session)):
+            url = request.url.replace (os.getenv ('HEROKU_APP_NAME'), os.getenv ('REDIRECT_AB_TEST'))
+
+            request_headers = {}
+            for header in request.headers:
+                if (header [0].lower () in ['host']):
+                    continue
+                request_headers [header [0]] = header [1]
+            r = getattr (requests, request.method.lower ()) (url, headers=request_headers, data=request.data)
```
rix0rrr, line 102: This feels like something Python is not made for, and makes me feel a little uneasy.

Why is making a network request from Python to a second copy of the application preferable to:

* Just having the test behavior INSIDE the one application? (simple)
* If you must, front two different instances of the Flask app with a tool designed to be a reverse proxy, like nginx? (more complex but definitely robust, and imo preferably because the two Flasks are symmetric)

Also, how is the second copy of the application deployed?

rix0rrr: Okay, I've now understood this was a conscious choice to keep the implementation burden on the second copy of the code base low. 

Doing the proxying in Python still makes me feel a little uneasy. Using nginx seems complicated but I'm going to give it a shot nonetheless.

rix0rrr: Okay I managed to get it to work in https://github.com/Felienne/hedy/pull/252

I think the solution is mostly solid, but it has as big downside that treatment *cannot*  be based on username (which is what you have now).

app.py
```diff
@@ -70,6 +71,44 @@
...
 
+def into_number (string):
+    total = 0
+    for char in string:
+        total += ord (char)
+    return total
+
+def redirect_ab (request, session):
+    # If the user is logged in, we use their username as identifier, otherwise we use the session id
+    user_identifier = current_user(request) ['username'] or str (session_id)
```
rix0rrr, line 82: After thinking through this some more: session_id is a bit of a dangerous hashing key as well.

If the kid closes and reopens their browser, this may move them from inside the treatment group to outside the treatment group.

It may be disorienting if the UI of Hedy keeps changing.

Falling back to IP address instead might be safer...?

fpereiro:
Hi @rix0rrr ! I'll address your comments & contributions in here rather than inline, to unify the work forward in regards to A/B testing.

**Hashing**

My criteria was to use a hashing function that is 1) deterministic; 2) cheap. You make a good point about using a standard hashing function that has a more even distribution than what I did, which is just defaulting to the ASCII values of characters. Would md5 be a good candidate?

I'll also parametrize the `50` so it's more clear that it's the percentage of users/sessions sent to the test environment. Thanks for the suggestion!

**Implementation**

The implementation offered in this PR wasn't arrived at lightly. Before it I went through two options that seemed simpler:

1) _Modifying only certain parts of the code and having a switch that enabled/disabled them_: this option was quickly discarded, because the requirement is that anything in Hedy can and should be tested: Lark grammars, Hedy code for transpiling, documentation for users, UI changes, etc. These changes can affect all sorts of things (Lark files, Python code, templates) and can only be dealt reasonably within a self-contained branch. So the test environment needs to be in its own branch and should not duplicate code.

2) _An external proxying mechanism_: I considered redirecting traffic from nginx but I discarded that approach for four reasons: 1) avoid the complexity of having an extra moving part; 2) have the redirection logic in Python code, so we can quickly and flexibly change which users/sessions are redirected; Python has access to the context of the username and the session and Nginx doesn't; 3) the fact that the shown URL is different, so that attentive users will immediately know they are in a different subdomain; 4) the need to change our cookies to work at the top-level domain since they have to be shared by different subdomains.

The solution offered here is a reverse proxy function that uses a standard Python HTTP library (Requests) to redirect traffic in a way that 1) adds no moving parts, just a function; 2) uses Python code that is readily understandable to decide which users/sessions to send to the test environment; 3) is transparent to users/sessions and 4) requires absolutely no changes to our cookies. This solution also enables us to have a separate test environment set up that doesn't even have to know that it is a test environment.

I understand that Python + Requests is not a standard way to implement a high performance reverse proxy; but at the levels of traffic we currently have I believe that the performance of this arrangement shouldn't be an issue. If necessary, I could perform stress tests to see if this reverse proxying would add a significant performance impact on both our servers and the response time for users. Let me know if you think this to be necessary.

For now, and after carefully considering the nginx approach you propose (https://github.com/Felienne/hedy/pull/252), I still prefer this solution.

**Consistent redirecting**

Any system that performs A/B testing does so at the expense of presenting an inconsistent experience to its users. When username information is available, this inconsistency can be mitigated by always making the decision to put a certain user X to be inside or outside the control group. When the user is not logged in, which is the most common case with Hedy, we can only resort to workarounds.

Right now, as you mention, sessions are loosely correlated with an user's experience over time. It might be better to do this by IP or by device, although both are also error-prone. It'd be good to discuss this further in two regards: 1) how much information we want to collect from sessions in order to identify them? (privacy); 2) how important is this in terms of implementation time?

Thank you for your review and feedback.

rix0rrr:
> Would md5 be a good candidate?

Anything in the hashlib would be good. `md5` would be fine.

> external proxying mechanism: [...] ) 3) the fact that the shown URL is different [...]; 4) the need to change our cookies to work at the top-level domain [...]

These I don't get. As long as we proxy (and not redirect) the URL won't actually be different?

> at the levels of traffic we currently have I believe that the performance of this arrangement shouldn't be an issue

No, and I agree. I guess I can't look at things anymore and *not* see scalability problems 🙄. You are right that this solution will be a lot simpler to maintain, and the additional CPU and memory cost is probably negligible. Also not sure how much this will eat into your max concurrent connections... but I guess monitoring will tell you when this solution stops being good enough.

--------

*HOWEVER* while reading your last comment, I just though of something else: *sessions*... aren't they in-memory objects? So once you get forwarded based on your session ID, you will get forwarded to a server that doesn't have a record of that session ID!

I wonder what happens, but I would imagine the second Flask is not just going to create that session out of thin air? Or will it?

We have two different Flask stacks with two different in-memory session stores:

```
                   ┌──────────────┐                      ┌──────────────┐           
                   │              │                      │              │           
                   │  hedy-beta   │                      │  hedy-test   │           
                   │              │                      │              │           
                   └──────────────┘                      └──────────────┘           
                           │                                     │                  
                           │                                     │                  
           user request    │                                     │                  
          ────────────────▶│                                     │                  
                           │────┐ create                         │                  
                           │    │ session                        │                  
                           │    │   123                          │                  
  Set-Cookie: session=123  │◀───┘                                │                  
         ◀─────────────────│                                     │                  
                           │                                     │                  
                           │                                     │                  
                           │                                     │                  
                           │                                     │                  
      Cookie: session=123  │                                     │                  
          ────────────────▶│                                     │                  
                           │────┐                                │                  
                           │    │forward 123?                    │                  
                           │    │     yes                        │                  
                           │◀───┘                                │                  
                           │                                     │                  
                           │                                     │                  
                           │         Cookie: session=123         │                  
                           ├───────────────────────────────────▶ │                  
                           │                                     │────┐             
                           │                                     │    │session 123??
                           │                                     │    │   WTF???    
                           │                                     │◀───┘             
                           │                                     │                  
                           │                                     │                  
```

I see you've made the login session store shared, so that at least this can't happen with logins (that would be even worse!), but curious what the behavior will be with sessions.

fpereiro:
Hi @rix0rrr ! Replying in order:

- Just pushed a change using md5, hopefully this should make the distribution quite even.
- Because of the way we're proxying traffic, the user's browser only interacts with the main server (beta); if the main server decides to use the test server for a particular user/session, the main server does a background request to the test server on behalf of that user; when the main server gets the response from the test server, it passes it along to the user's browser. So there's no direct communication between the user's browser and the test environment. Unless someone goes directly to the test environment by going to hedy-test.herokuapp.com, all traffic arriving there should come only from background requests from hedy-beta.
- We have an e2e script. Perhaps I could add some logic to run it N times concurrently and measure delays and perhaps unresponded requests, which would be a makeshift stress testing - I have experience doing this with node, not python, but the principle should be the same.
-  _Wonder if running ab on hedy-alpha with two different configs might just be enough?_ I'm not sure I understand what you mean here. The reason for running test on a different environment is that we can have a 1:1 relationship between a test branch and an environment; if we use alpha for testing other features, we lose that relationship.
- I will definitely make a point of explaining further the main decisions in PRs like this - duly noted!
- Interesting scenario regarding sessions! As it stands now, any requests that arrive to test are, from the perspective of the test server, indistinguishable from a request coming from outside. So the test server will create a new session (if needed) for the incoming request. The session created by beta is only relevant for redirection purposes and inaccessible to the test server. I'm not clear on the lifecycle of Flask sessions. In my (limited) local tests, the behavior seems to stick, so probably some information from the browser headers is being used to create a session of a certain level of persistence. As long as this is the case for logged-out users, we should be good :). I will see more how Flask creates sessions to figure out if this will work, but I'm pretty sure that as long as we redirect the right headers, it will.
- Regarding user sessions, because the database is the single source of truth for cookies, and because both beta and test should use the same table for both users & cookies, the determination of the user from the cookie is deterministic and correct.

rix0rrr:
> Because of the way we're proxying traffic, the user's browser only interacts with the main server (beta);

I get it. My confusion was about you originally saying that the nginx solution did not have the behavior you describe here: you were saying points your points (3) and (4) give the Python `requests` solution an edge over the nginx solution. However I thought the solutions would be identical, so therefore points 3 and 4 would not be a reason to prefer one over the other?

image

>> Wonder if running ab on hedy-alpha with two different configs might just be enough? 
>
> I'm not sure I understand what you mean here. The reason for running test on a different environment is that we can have a 1:1 relationship between a test branch and an environment; if we use alpha for testing other features, we lose that relationship.

Okay, sure. Sorry, I skipped a beat.

I mean:

* Run [`ab`](https://httpd.apache.org/docs/2.4/programs/ab.html) against an environment WITHOUT experiments enabled. See what the numbers are.
* Then turn on 100% traffic redirection for experiments and run `ab` again, see what the difference is.

Then I mean, we'd do this kind of testing against the `hedy-alpha` environment, instead of `hedy-beta`, because we wouldn't want to disturb production traffic in case we happen to do something that accidentally kills the server during all of this testing.

> In my (limited) local tests, the behavior seems to stick, so probably some information from the browser headers is being used to create a session of a certain level of persistence.

Flask sets a cookie like this once you start a session:

```
Set-Cookie: session=eyJzZXNzaW9uX2lkIjoiYWU5ZDYxODMxMTAyNDcyYzllOTFkYjFiOGY2ODE0MDMifQ.YGHWcg.RGl-1bZ7CzsQ3ZJ8Qen1k0HtESI; HttpOnly; Path=/
```

The session identifier is mapped to an in-memory hash table.

Since the forwarded requests will contain the session cookie, it's rather important what happens when Flask sees a session cookie it didn't itself set.

* If it goes "okay fine that'll be my session then" everything is mostly fine: except your session object may be empty (for example it won't have a session ID if Flask followed that behavior!). That's why I assume Flask will NOT choose this behavior.
* I think it will the Flask at `hedy-test` will disregard the session cookie, create a new session and send back a NEW `Set-Cookie` header. Next time the browser comes around to `hedy-beta`, that session cookie will be unknown and it will be dropped and a new one created, and so on and so forth.

In fact, the more I think about it, the `hedy-beta`'s `Set-Cookie` header will never be sent, right? Because the request will be immediately forwarded to `hedy-test`, and ITS `Set-Cookie` will be sent instead?

So normally this is supposed to happen between page loads:

image

But I'm afraid this is going to happen instead (because the session Cookies from `hedy-beta` will be dropped on the floor):

image

And the hedy-beta frontend is going to generate new sessions until the generated session ID *doesn't* have experiments applied, after which point the session is going to stop being redirected.

It's either that, OR the session cookie from `hedy-beta` sticks, and the session from `hedy-test` is constantly regenerated (making the analysis of multiple programs submitted in a single session impossible because session information is not retained).

It's probably worthwhile to dive into confirming/disconfirming this behavior. It's subtle but may have big implications for the correctness of the data collected during the experiment.

fpereiro:
Hi @rix0rrr !

- Will add a task for being able to run the E2E tests concurrently against an environment and track performance. We can then do exactly what you propose, test on `alpha` without redirection and then test on `alpha` with redirection to `test`.

- *However I thought the solutions would be identical, so therefore points 3 and 4 would not be a reason to prefer one over the other?* I stand corrected, you're absolutely right: nginx and (as specified here) Python/requests operate as a reverse proxy, so nginx would show the same URL and the cookies would be the same too - I was confusing here a 302 redirect on nginx with a proxy_pass, which would make it a reverse proxy.

- You are 100% right about the inconsistency in sessions between environment, great catch! Both environments create sessions & cookies constantly, so the logging of programs is going to be inconsistent for non-logged in users. I'm working on a fix now that will also keep the auth routes and the cookie setting exclusively on the main server.

app.py
```diff
@@ -70,6 +71,51 @@
...
+
+# If present, REDIRECT_AB_TEST should be the name of the target environment (that is, environment B).
+if os.getenv ('REDIRECT_AB_TEST') and os.getenv ('REDIRECT_AB_TEST') != os.getenv ('HEROKU_APP_NAME'):
+    @app.before_request
+    def before_request():
+        # If it is an auth route, we do not reverse proxy it to the REDIRECT_AB_TEST environment, with the exception of /auth/texts
+        # We want to keep all cookie setting in the main environment, not the test one.
+        if (re.match ('.*/auth/.*', request.url) and not re.match ('.*/auth/texts', request.url)):
+            # Dummy value. We do this in a separate block so that the block below doesn't depend on a very long conditional.
+            0
```
rix0rrr, line 95:
```suggestion
            pass
```

🤣

app.py
```diff
@@ -152,23 +198,37 @@ def parse():
         'server_error': response.get('Error'),
         'version': version(),
         'username': username
-    })
+    }
+
+    if os.getenv ('REDIRECT_AB_TEST') == os.getenv ('HEROKU_APP_NAME') and os.getenv ('HEROKU_APP_NAME'):
+        log_object ['is_test'] = 1
+        if 'x-session_id' in request.headers:
+            log_object ['session'] = request.headers ['x-session_id']
```
rix0rrr, line 206: Doesn't it make more sense to update the global `session` object, rather than just do it here? 

(For example, this logic could have been put into `pre_request`)

Doing it like this makes it so that `log_object['session']` contains something different than `session_id()` returns, which feels like a recipe for confusion later on if we now have 2 different notions of session ids.

app.py
```diff
@@ -152,23 +198,37 @@ def parse():
...
         'date': str(datetime.datetime.now()),
         'level': post_body.get('level'),
         'code': post_body.get('code'),
         'client_error': post_body.get('client_error'),
         'version': version(),
         'username': current_user(request) ['username'] or None
-    })
+    }
+
+    if os.getenv ('REDIRECT_AB_TEST') == os.getenv ('HEROKU_APP_NAME') and os.getenv ('HEROKU_APP_NAME'):
```
rix0rrr, line 226: This is duplicated with the previous block.

Should have been put into a separate function for deduplication purposes, or go into the `pre_request` trick.

app.py
```diff
@@ -152,23 +198,37 @@ def parse():
         'server_error': response.get('Error'),
         'version': version(),
         'username': username
-    })
+    }
+
+    if os.getenv ('REDIRECT_AB_TEST') == os.getenv ('HEROKU_APP_NAME') and os.getenv ('HEROKU_APP_NAME'):
```
rix0rrr, line 203: How do you see this being used?

Does the `REDIRECT_AB_TEST` variable need to be set on BOTH `hedy-beta` and `hedy-test` ?

Feels like an `HEDY_IS_TEST_ENV` variable might be more appropriate.

app.py
```diff
@@ -70,6 +71,51 @@
...
+            url = request.url.replace (os.getenv ('HEROKU_APP_NAME'), os.getenv ('REDIRECT_AB_TEST'))
+
+            request_headers = {}
+            for header in request.headers:
+                if (header [0].lower () in ['host']):
+                    continue
+                request_headers [header [0]] = header [1]
+
+            # Send the session_id to the test environment for logging purposes
+            request_headers ['x-session_id'] = session_id ()
```
rix0rrr, line 107: I like this! But please do document the private protocol that is now being used between the two Hedy instances and how cookies are handled somewhere.

Bottom line:

- Cookies are only set on the primary Hedy instance (both for sessions and login)
- Session IDs are forwarded in the reverse proxy protocol, and taken as gospel in the test instance
- Login cookies are forwarded and looked up the database by both instances
- Session cookies are still uselessly forwarded though, right? We should probably disable session on the Hedy test instance? Or something? Or have it not try to start sessions on its own, for no purpose?

app.py
```diff
@@ -70,6 +71,51 @@
...
+            r = getattr (requests, request.method.lower ()) (url, headers=request_headers, data=request.data)
+
+            response = make_response (r.content)
+            for header in r.headers:
+                # With great help from https://medium.com/customorchestrator/simple-reverse-proxy-server-using-flask-936087ce0afb
+                # We ignore the set-cookie header
+                if (header.lower () in ['content-encoding', 'content-length', 'transfer-encoding', 'connection', 'set-cookie']):
+                    continue
+                response.headers [header] = r.headers [header]
+            return response, r.status_code
```
rix0rrr, line 117: How does this inject the `set-cookie` header from the frontend Hedy session though? Does Flask just do that automatically?

fpereiro:
Hi @rix0rrr ! I just implemented your suggestions:

- Use `pass` instead of a dummy value. 
- Use a `IS_TEST_ENV` variable to make things more explicit. Also renamed the other variable to `PROXY_TO_TEST_ENV`.
- On a test environment, modify the `session_id` function to return the one passed by the main environment (but only if present, so the test environment can be tested directly without having to go through alpha/beta).
- Add debugging logs so we make sure that session ids are what we expect them to be.
- Document the proxy pattern in `doc/backend.md` and add variables to `CONFIGURATION.md`.

I don't think there's any harm in ignoring the session cookies set by the test environment. The main environment can ignore them, but if you access test directly, you can still have sessions as needed. Also, the main advantage of this is that we don't have to pepper our code with conditionals to cover the case of the test environment.

rix0rrr (approved):
Alright, looks good! Let's go!

PR #253, opened 2021-03-30
Refactors tests using textwrap and proper order of arguments in assertEqual

Felienne:
Improves layout of tests for better readability.


Felienne:
Hi @fpereiro Can you check this out when you have time and merge? Make it easier to communicate want we want new tests to look like also

fpereiro (approved):
Looking good! I tend to favor single line assertions, but those get messy very quickly, so I agree with the approach. And big +1 to the addition of `import random` in one place only!

PR #258, opened 2021-03-31
chore: deploy to 'alpha' only if the unit test workflow succeeds

rix0rrr:
Right now, the "unit test" workflow and the "deploy to alpha" workflow
run in parallel.

This means that the code will get deployed to alpha, even if it doesn't
pass the tests, which is kinda dangerous (on the other hand, it's
alpha so we might not care).

This change links the two workflows together, so that "deploy to alpha"
only runs if the unit tests on master ran successfully.

Feel free to take or close this if you don't care about this!

Two additional notes:

* The tests also run on the PR build, but that's pre-merge and it's not
  necessarily guaranteed that the tests will pass after the merge.
* This has no impact on the "deploy to prod" workflow; it's still
  very well possible to deploy a `master` to production that doesn't
  pass tests. We need to write the workflows slightly differently if
  we also want to protect against that.

Let me know what you'd prefer.

### DIFF
--- a/.github/workflows/deploy-to-alpha.yml
+++ b/.github/workflows/deploy-to-alpha.yml
@@ -1,9 +1,16 @@
 name: Automatically deploy to Alpha
 on:
-  push:
+  # Run this AFTER the unit test job finishes
+  workflow_run:
+    # Must be the same as in unittests.yml
+    workflows: ["Unit tests"]
     branches: [master]
+    types:
+      - completed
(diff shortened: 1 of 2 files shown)


fpereiro:
HI @rix0rrr ! I think it's a good idea to not deploy (even to alpha) if the tests don't pass. It could be easy to overlook this if we're tired or in a hurry to fix something.

PR #275, opened 2021-04-03
Fixes issue 274 allowing quotes in level 2

Felienne:


fpereiro:
This is fantastic. What's not forbidden should be allowed!

PR #276, opened 2021-04-05
fixes #185 by allowing unquoted strings in the grammar for print

### LINKED ISSUE #185: Never suggest the command that is found
This makes kids hate computers :)

image

Felienne:
In levels 1 upto 5. I saw in the collected data that after level 4 the error is not common anymore, so I did not allow it in higher levels.

Also refactors the tests a bit for the distance algo


fpereiro:
Great changes! The specialized quoting error is going to be super useful :)

PR #281, opened 2021-04-08
Fix repeated function names

fpereiro:
I think the code still works well but it's wrong to have these function names repeated. All functions with repeated names are of my authorship :+1:


Felienne (approved):
Nice find, love that you are helping to keep the code base maintainable!

PR #282, opened 2021-04-08
Update third example to be consistent with others

### LINKED ISSUE #157: Change third example code in all languages for level 6
In level 6, it would be good to change the third example (https://github.com/Felienne/hedy/blob/master/coursedata/level-defaults/en.yaml#L119) to also use a mathematical problem (sum/multiplication) instead of a color. This would increase consistency among all three examples.

This should be done in all five languages. English can be used as a placeholder for the languages not known by whoever tackles the issue and then we'll finish the translations before merging.

HeroicHitesh:
Fixes #157

### DIFF
--- a/coursedata/level-defaults/en.yaml
+++ b/coursedata/level-defaults/en.yaml
@@ -116,8 +116,8 @@
       -
         name: "repeat"
         explanation: "`repeat` and `if` combined"
-        example: "Example: if color is green repeat 3 times print 'pretty!' else repeat 5 times print 'meh'"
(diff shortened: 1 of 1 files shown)


Felienne:
This is a great fix, but I want to also fix it in the other languages, maybe @fpereiro can you pick up this for Sp and Pt? I can do Dutch and German.

fpereiro:
Hi @HeroicHitesh ! Thank you for this!

@Felienne agreed! I can also pick up Italian and French. I suggest we merge it and then immediately I can open a branch on which we can add the missing translations. Until then, we can keep the issue open. Would that work?

fpereiro:
Thanks @HeroicHitesh !

PR #284, opened 2021-04-08
Level11_12_13

LauraTSD:
Includes:
tests for level 11,12,13
texts (yaml) for level 11,12,13 in English and Dutch
Grammar for level 11,12,13 (lark)
Edits to hedy.py


hedy.py
```diff
@@ -578,7 +591,34 @@ def elifs(self, args):
         all_lines = [indent(x) for x in args[1:]]
         return "\nelif " + args[0] + ":\n" + "\n".join(all_lines)
 
+class ConvertToPython_11(ConvertToPython_9):
+    def input(self, args):
+        var = args[0]
+        all_parameters = [a for a in args[1:]]
+        return f'{var} = input(' + '+'.join(all_parameters) + ")"
+
+class ConvertToPython_12(ConvertToPython_11):
```
fpereiro, line 600: Hi! Since this function is useful for levels 12 and 13, I suggest renaming it to something that makes that explicit (for example: `ConvertToPython_12_13`).

fpereiro:
Hi @LauraTSD !

I played around with the new levels and it all looks great! The only thing that tripped me up a bit - and might be worth explaining in the examples - is that in level 12, the elements of lists must also be quoted. In the explanation for level 12, the brackets are explained, but not the quotes.

Congrats on adding three more levels!

Felienne:
I have deployed this to alpha and tested and it all looks good to me! Maybe @fpereiro you can add the translations of these levels into at least Spanish? That would be nice! And indeed @LauraTSD maybe before we really release into production, you can add a bit of explanation of the quotes? 

I envision in the future there will be smaller, in between, steps (1 for brackets and 1 for quotation marks) but for now it is ok if we just explain it together.

fpereiro:
Hi @Felienne ! I can definitely add the translations to Sp. I just updated the issue (and also the related language issues).

PR #289, opened 2021-04-11
feat: add a page showing version and unreleased changes

rix0rrr:
Generates a GitHub link which shows the changes that are on
`master` that haven't been deployed to the current page yet.


fpereiro:
This is very handy, thank you @rix0rrr !

PR #294, opened 2021-04-13
Implementation of the Gradual Feedback Integration (GFI) Model

TiBiBa:


fpereiro:
Hi @TiBiBa ! Great work, I'm excited to try this :)

Some comments that emerged while reviewing the code:

- I suggest removing pandas, and using the native csv Python library for parsing the csv (https://docs.python.org/3/library/csv.html). In this way we eliminate a dependency.
- In the payloads between the client and the server, it'd be great to change the camel cased variables to underscore separated names to be consistent with the rest of the app. I personally prefer camel case but the app already uses underscores.
- It'd be more readable to merge all the print statements in lines 322-327 of app.py into a single one.
- Please rename `runit` to `run_button` in static/app.js (lines 39, 381, 393) so that we don't lose existing functionality.

I'll try it out and give you some more feedback!

Felienne:
Great work Timon! 

I have one suggestion:

* Can we swap the order here? So first the blue one with the arrow (then) pointing to the code and then the error message in red below?
image

Also, I did not succeed in getting more information to be shown than this, what is a path to test getting even more help?

TiBiBa:
That's weird! When submitting the same (wrong) code the layout should look like the image below. When pressing the arrow the box should expand and provide additional information on why submitting the exact same code doesn't work. The model doesn't give new information until a different mistake is made. A good way to test the model is by making consecutive different mistakes. I just tried it locally (using Safari on MacOS) with the newest version of the code and it should look like this:

Schermafbeelding 2021-04-20 om 13 41 51
Schermafbeelding 2021-04-20 om 13 42 02

requirements.txt
```diff
@@ -9,3 +9,5 @@ attrs==19.3.0
 Flask-Commonmark==0.8
 bcrypt==3.2.0
 boto3==1.16.50
+pandas==1.2.4
```
fpereiro, line 12: Hi @TiBiBa ! Could we remove this dependency?

TiBiBa:
It seems the CSS I wrote in `additional.css` is not read correctly when you run it. Because the arrow should be bold and in the upper-right corner and the closing X is also missing. The styling of the box itself is inherited from the `generated.css` file so that could explain why that is shown correctly.

requirements.txt
TiBiBa: Completely forgot about that one, will remove it right away!

static/js/app.js
```diff
@@ -34,7 +41,7 @@
     editor.on('change', function () {
       if ($('#inline-modal').is (':visible')) $('#inline-modal').hide();
       window.State.disable_run = false;
-      $ ('#runit').css('background-color', '');
+      //$ ('#runit').css('background-color', ''); Delete to prevent changing color within time-out
```
fpereiro, line 44: This is being done now from inside the timeout, right? If so, could we just remove the line from here? As long as the default color of the button is shown when the user enters text, we should be alright.

TiBiBa: Yes I think this line can be deleted!

fpereiro:
@TiBiBa merged!

PR #298, opened 2021-04-15
Omid syntax highlighter lvl 1

omedtjeee:
I think I succesfully adapted syntax highlighting for the first level.


fpereiro:
Great work @omedtjeee !

PR #299, opened 2021-04-16
Adventure mode!

### LINKED ISSUE #151: Adventure mode
Add a Hedy mode where a particular challenge/problem is the axis of work ("challenge mode"), instead of the levels ("level mode", the currently implemented one).

Principles:
- Make it incremental. Don't ruin what already works!
- No forks in the road, no decisions to be imposed on first time or early users.
- Always be able to go back to "level mode" without hassle.
- A lot of similarity between "level mode" and "challenge mode".

Starting point for implementation:
- Make a list of problems.
- Add a button/switcher between levels and challenges.
- Show a list of Hedy-style drawn tiles (@Felienne, I need you here), one for each problem.
- Jump to a particular problem.
- Add data/hints for each combination of problem & level.

Project: https://github.com/Felienne/hedy/projects/1

fpereiro:
- Allows for a parallel "adventure mode" where a task/adventure is the axis of progress, in contrast with the default "level mode".
- Lists all adventures.
- Can save and load code for a combination of adventure + level.
- Automatically restore saved code for adventure + level.
- If no level specified for an adventure, defaults to highest level where the user saved a program.
- Program loading now distinguishes between programs belonging to adventures and those belonging to plain levels.
- Closes #151; part of https://github.com/Felienne/hedy/projects/1

Includes the following additional improvements:
- We now check if the required language exists; if it doesn't, we return 404 instead of 500 in multiple routes (not just those of adventures).
- Use of a localize_link helper to add query strings with language to links.
- Ignore trailing slashes, so links such as `/hedy/` and `/hedy/adventures/` work.


fpereiro:
@Felienne I think it's ready for a spin :)

Felienne:
Hi @fpereiro! This looks great!!! There are a few things I think we can add that are not all too complex to be added to this phase:

* Images for the starting page. I am not just I can just add a link in the yaml file?
image
* Buttons per level as we have in the regular Hedy. I think by default we can just inherit from the regular Hedy. But then *if* there are buttons in the adventure mode, these override the normal ones. That we we can easily give hints on how to start.

fpereiro:
Hi @Felienne ! Glad you like this :)

Just added your images. I settled on a semi-arbitrary size of 225x225 for images.

I also added an example command to the Story adventure, in level 1, for both languages.

Felienne (changes requested):
Great work @fpereiro!

A few small issues:

* Misspelling adventures (f.e. adventure) gives a nasty error:
image

* Shall we center images?

* Do we want to add a button to go back to Adventures next to "Go to levels"?

image

Felienne:
Thanks for the update @fpereiro! I am translating all to Dutch now, and updating the Uitleg a bit to be more aligned with the new system.

There is one thing missing, the buttons:

image

I'd like the buttons to be there as default, *unless* I override them in the adventure mode (f.e. with thematic demo code)

So if level 2 has no commands (as it does not):
image

I want the regular buttons to be there

fpereiro:
Great idea, just implemented it!

Felienne:
The markdown in the yaml works but it means we need to use 2 newlines for one newline which creates a really long file. This is the spec (https://yaml-multiline.info/) 

What do you think @fpereiro? It is a bit of a hassle to type (I kept forgetting) but it does work. But if you can think of a fix, let me know!

fpereiro:
> The markdown in the yaml works but it means we need to use 2 newlines for one newline which creates a really long file. This is the spec (https://yaml-multiline.info/)
> 
> What do you think @fpereiro? It is a bit of a hassle to type (I kept forgetting) but it does work. But if you can think of a fix, let me know!

Hi @Felienne ! Reading that resource (https://yaml-multiline.info/) I notice there's a *block style* available that treats single newlines as newlines - it requires using `|` instead of `>` at the outset of the string. Would that perhaps solve the problem? Not sure if this type of quote would break other things.

Felienne:
ok @fpereiro I think I did it :) There is an issue in the story, I really cannot figure out what, the rest seems to be working. I also removed the descriptions from docs, we do not need them there and at one point we can remove all docs and have all this info in adventure mode.

If you fix & translate we should be good to go for En and Nl (and maybe Sp and Pt also if you feel like it?)

Felienne:
Ok @fpereiro, I think we are now really almost good to go!! 🤣 

Somehow I did break the individual adventures, the overview page works but I get a 404 if I click on one, and I don't understand why. Since we don't link to it, that is not a big issue, but if you could look into it, that'd be nice.

Here's an overview of changes and my thinking on them:

* Ultimately, I want all adventures visible in a level. I totally changed my mind on that, sorry. I figured out I don't want kids rushing through levels in one adventure, I want them to try multiple adventures in one level (or move on if they think they can do it)
* Therefore I also removed the buttons to switch adventures and to switch to regular mode, because I don't want them to switch to one adventure.
* I do want to keep the adventures page, because it is perfect for one day workshops where kids so want to reach higher levels quickly. So the overview page was not for nothing! Maybe we need to make some changes to the templates for the separate adventures (maybe that's what I messed up?)
* The pallet is great, I love that, but I removed the example from the template because it makes it a bit long. Not sure I am ready to remove it form the yaml to, we might want to use it elsewhere.

Work left to be done by @fpereiro:
* Fix individual adventure issue
* Rework and merge En and Sp translations from the adventures
* Test it out a bit more?
* (optional in this PR, I am not sure this is hard. If so, can you make a separate issue?) Log to jsonbin which adventure tab is open when code is ran, that is great feedback for us to have.

Felienne:
I see there's a build issue in my translation, will fix in the next push

PR #305, opened 2021-04-20
Update example level 6; closes #157.

### LINKED ISSUE #157: Change third example code in all languages for level 6

fpereiro:
@Felienne in this branch you can add the changes to the third example of level 6 in nl and de; I still need to add the changes to it, pt_br and fr.

### DIFF
--- a/coursedata/level-defaults/es.yaml
+++ b/coursedata/level-defaults/es.yaml
@@ -117,8 +117,8 @@
       -
         name: "repeat"
         explanation: "`repeat` e `if` combinados"
-        example: "Ejemplo: if color is verde repeat 3 times print '¡Qué bonito!' else repeat 5 times print 'A mi me gusta otro'"
-        demo_code: "color is ask ¿Cuál es tu color favorito?\nif color is verde repeat 3 times print '¡Qué bonito!' else repeat 5 times print 'A mi me gusta otro'"
+        example: "Ejemplo: if producto is 50 repeat 3 times print 'correcto!' else repeat 5 times print 'incorrecto!'"
+        demo_code: "producto is ask Cuanto es 10 veces 5?\nif producto is 50 repeat 3 times print 'correcto!' else repeat 5 times print 'incorrecto!'"
   7:
     start_code: "repeat 5 times\n    print 'Hola amigos'\n    print 'Esto se va a mostrar 5 veces'"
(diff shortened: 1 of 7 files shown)


Felienne:
I think we can close this Fede, but I'd be good if you'd check my latest adds

fpereiro:
Great idea, just implemented it!

fpereiro:
@Felienne ready to merge!

coursedata/level-defaults/es.yaml
```diff
@@ -117,8 +117,8 @@
       -
         name: "repeat"
         explanation: "`repeat` e `if` combinados"
-        example: "Ejemplo: if color is verde repeat 3 times print '¡Qué bonito!' else repeat 5 times print 'A mi me gusta otro'"
-        demo_code: "color is ask ¿Cuál es tu color favorito?\nif color is verde repeat 3 times print '¡Qué bonito!' else repeat 5 times print 'A mi me gusta otro'"
+        example: "Ejemplo: if producto is 50 repeat 3 times print 'correcto!' else repeat 5 times print 'incorrecto!'"
+        demo_code: "producto is ask Cuanto es 10 veces 5?\nif producto is 50 repeat 3 times print 'correcto!' else repeat 5 times print 'incorrecto!'"
```
lpmi-13, line 121: this should probably be `10 por 5` if you mean "ten times five" (multiplication).

fpereiro: Hi @lpmi-13 ! Native Spanish speaker here :) Both "por" and "veces" are used in school settings for multiplication, but "veces" seems to be more prevalent than "por" in educational settings, since it's more intuitive - usually adults use more "por" because it's shorter. I think it's best to leave it as is, but I appreciate the review!

lpmi-13: oh really? I apologize, I had no idea! Thanks for the thorough explanation.

lpmi-13: ...this is probably an appropriate place for an "estoy embarazado" joke

fpereiro: Hahaha, no worries! In fact, your question made me look deeper into this, to make sure that "veces" was the right word - I had to look it up myself, because I didn't really think about it when I wrote the translation!

PR #314, opened 2021-04-22
fix: "ongetiteld" -> "naamloos"

rix0rrr:
Traditionally, untitled documents are called "naamloos" in Dutch.


fpereiro:
Thanks @rix0rrr !

PR #315, opened 2021-04-22
redesign: move command palette to the left

rix0rrr:
This has the following advantages:

* more vertical space becomes available, so the commands that are still
  valid from previous levels can be displayed as well.
* it frees up vertical space in the main content area, so that can be
  filled with more instructional content.

This PR is not complete, and should not be merged yet! There is a lot
of work yet to do:

* the header on the left side of the palette should be a translatable
  string, instead of hardcoded
* all command instructions in all languages need to be rewritten to
  fit the new layout (mostly: they need to be made a lot shorter)
* the "Try this" button probably needs a shorter string in all languages
  as well (ex: got rid of the "this/dit" for "en/nl" already).
* the css should be validted to work properly (or at least fine) on
  a variety of screen sizes including mobile

I will not be in a position to pick up those remaining tasks myself. Help to finish this out would be very welcome.

Example of new layout:

image


fpereiro:
Hi @rix0rrr ! Thanks for this contributio -- the new layout is a bold change!

Should I pick up the remaining tasks? I can do that as soon as I finish Adventure Mode (https://github.com/Felienne/hedy/pull/299).

tailwind/styles.css
```diff
@@ -182,18 +182,8 @@ table.users td {
   border: solid 1px gray;
 }
 
-/* RESTORE OLD COLORS */
```
fpereiro, line 185: Hi @rix0rrr ! Could we leave the old colors? They changed with a Tailwind update and @Felienne wanted them to stay the same.

rix0rrr: Interesting. We probably can, but this is not the right way to do it. Adding these here just breaks tailwind because there are now conflicts between built-in classes and new classes.

The right way to do it would be either:

* Introduce a `package.json` to fix to a specific version of Tailwind with the right colors.
* Change the Tailwind config file to restore exactly the colors we want.

I can take this as an action, should be a separate PR though.

Do you have a reference or link to the upstream Tailwind change?

rix0rrr:
> Should I pick up the remaining tasks?

That would be great, thanks!

tailwind/styles.css
fpereiro: Hi @rix0rrr ! Good to know, thanks for catching this. I like the second option, so that we can keep on updating Tailwind. Unfortunately I don't know which version has the old colors, but perhaps we don't need to know it if we go with the second solution?

I can take up this task as I'm working on the PR, if you prefer.

Thanks!

fpereiro:
> > Should I pick up the remaining tasks?
> 
> That would be great, thanks!

Will do! I'll pick this up next week. Thanks again!

tailwind/styles.css
rix0rrr: You can take it, but I do strongly feel it should be a separate PR to put proper versioning on Tailwind though, so I don't think it should go into this one.

The only reason I found out is because my Tailwind broke when I tried to rebuild the styles.

The reason we probably need the upstream change is because I believe the way to configure it in Tailwind is to give one base color, and the `-100`, `-200`, etc shades are (I think??) automatically generated from that one.

So we do need the original hex code to put back into our config.

rix0rrr: > automatically generated from that one.

Ah actually not true:

> Bad news, color is complicated and despite trying dozens of different tools, we've yet to find one that does a good job generating these sorts of color palettes. We picked all of Tailwind's default colors by hand, meticulously balancing them by eye and testing them in real designs to make sure we were happy with them.

https://tailwindcss.com/docs/customizing-colors

rix0rrr:
We talked about it ourselves and I'll continue on this quickly after all, so you can focus on more important work.

PR #316, opened 2021-04-22
add Dockerfile for local development without python

lpmi-13:
- Add a Dockerfile to run the application without needing to install any python dependencies (or even python itself!)
- Update app.py to bind to 0.0.0.0 in DEV mode so it can respond inside a container
- Update Contributing.md with instructions for building and running a containerized version of the webapp

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.github
+*.md
+.dockerignore
+Dockerfile
+License.txt
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,14 @@ you can wrap the command in a loop to restart the server quickly).
 (.env)$ env FLASK_ENV=development python app.py
 # or in a loop if it stops too often
 (.env)$ while true; do env FLASK_ENV=development python app.py; sleep 1; done
+
+If you want to run the website locally, but would prefer to use Docker instead of installing python, you can build a container and run it like so:
+```bash
+docker build -t NAME_OF_CONTAINER .
+```
+and then
+```bash
+docker run -it --rm -p 5000:5000 NAME_OF_CONTAINER
 ```
 
 If you don't want to use a website, you can run the code locally without the need of a website. To create a file use:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,13 +81,13 @@ you can wrap the command in a loop to restart the server quickly).
 # or in a loop if it stops too often
 (.env)$ while true; do env FLASK_ENV=development python app.py; sleep 1; done
 
-If you want to run the website locally, but would prefer to use Docker instead of installing python, you can build a container and run it like so:
+If you want to run the website locally, but would prefer to use Docker instead of installing python, you can build a container image and run it like so:
 ```bash
-docker build -t NAME_OF_CONTAINER .
+docker build -t hedy .
 ```
 and then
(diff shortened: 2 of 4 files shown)


.dockerignore
```diff
@@ -0,0 +1,6 @@
+.git
+.github
+*.md
+.dockerignore
```
lpmi-13, line 4: I will forever lol at the dockerignore needing to ignore itself

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:20.04
```
lpmi-13, line 1: I expected this to be a lot heavier, but the final image size, even with the full application, is only 485MB. I'm open to trying to slim that down if it's still too big.

fpereiro:
Hi @lpmi-13 ! Thanks for this - I'll review early next week!

CONTRIBUTING.md
```diff
@@ -56,6 +56,15 @@ If you want to run the website version locally, run:
 (.env)$ python app.py
 ```
 
+If you want to run the website locally, but would prefer to use Docker instead of installing python, you can build a container and run it like so:
+```bash
+docker build -t NAME_OF_CONTAINER .
```
tiborvass, line 61: technically it's the name of the image not the name of the container. I would instead just name the image `hedy` instead.

.dockerignore
tiborvass: I agree it's a bit weird. It was modeled after gitignore which does have different design considerations but oh well.

Dockerfile
tiborvass: You can reduce the image to 200MB by remove the `apt install` and doing `FROM python:3.8-slim` (might as well update to 3.9 or 3.10), as well as using `pip3 install --no-cache-dir -r /tmp/requirements.txt`

fpereiro:
Hi @lpmi-13 ! Haven't forgotten about this! We're hard at work finishing https://github.com/Felienne/hedy/pull/299; when that's cleared out I will take a look at this. I'll keep you updated.

lpmi-13:
No rush at all. I'll update this PR with @tiborvass excellent suggestions in the meantime!

Dockerfile
lpmi-13: couldn't find any 3.10 slim images that had gcc, so going with `3.9-slim`, which "just works" (:tm: )

CONTRIBUTING.md
lpmi-13: this is much simpler, and very sensible. thanks!

lpmi-13:
Have updated based on feedback, and do please let me know if you would like me to rebase at some point or if one of the maintainers would prefer to.

fpereiro (approved):

fpereiro:
Hi @lpmi-13 ! I finally got around to testing it. It works brilliantly!

There's a small conflict in CONTRIBUTING.md - if you can fix that (by adding both texts, yours and the updated version from master), we'll be good to merge.

lpmi-13:
ah darnit...I tried to be too clever and fix it without needing to rebase, but I couldn't figure out how to do it, so I'm just gonna rebase and then update.

lpmi-13:
alright, should be all good now!

fpereiro (approved):

fpereiro:
@lpmi-13 merged! Thank you again for this!

PR #317, opened 2021-04-23
feat: palette buttons append code instead of replacing

rix0rrr:
Right now, clicking a palette button complete replaces whatever
code is in the editor, destroying whatever they might have typed
already and losing their work.

This change will **append** the example code to whatever is in the code
editor, so the user can get the benefit of trying out the code without
destroying what they already have.  Probably goes well together with
PR #315, which really turns the palette into a palette of options to mix
and match.

I can also see a downside: this feature makes it possible to finish
exercises by clicking the right button+lightly editing the example.
Kids might use this to skip having to type out the entire syntax of
the command, thereby not getting the full learning experience.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -128,6 +128,25 @@ function runit(level, lang, cb) {
   }
 }
 
+/**
+ * Called when the user clicks the "Try" button in one of the palette buttons
+ */
+function tryPaletteCode(exampleCode) {
+  var editor = ace.edit("editor");
+  var currentUnsaved = window.State.unsaved_changes;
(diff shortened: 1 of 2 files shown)


fpereiro:
Hi @rix0rrr -- looks great, just tested it.

Felienne:
Hi @rix0rrr! I love this idea as an idea, and I think it would be a cool project for a student to try out in a study to see what the effect is, but my guess is the same as yours that it would get kids to just click and not pay deliberate attention to the code which ultimately harms learning, so I am closing this for now.

PR #318, opened 2021-04-23
Add placeholders for GFI texts

fpereiro:
Adds missing text fields required by GFI to other languages, as English placeholders. If/when we merge to master, we can then translate them.


fpereiro:
Note: this PR is against `test`, not `master`.

PR #324, opened 2021-04-26
Fix int variables

LauraTSD:
Problem/bug 
When doing the following: var is 5. var + 5, it would give 55. Now it gives 10.


LauraTSD:
not necessary for now

PR #328, opened 2021-04-28
update course data-level defaults-zh

Grace666-code:
I translated this material into Chinese again。 
I have one question on line 217, demo_code: "if 5+3 == 8:\n    print('5+3 is indeed 8')\nelse:\n  print('This will not be printed because 5+3 is 8!')", I think a NOT missing in this sentence, it should be print('This will not be printed because 5+3 is NOT 8!')

### DIFF
--- a/coursedata/level-defaults/zh.yaml
+++ b/coursedata/level-defaults/zh.yaml
@@ -1,13 +1,12 @@
-
 ---
 	  1:
-	    intro_text: "在第1级目标你可以使用这些命令"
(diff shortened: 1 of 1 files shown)


Felienne:
Thanks @Grace666-code!

The line you mention in fact is correct, 'This will not be printed because 5+3 is 8!' is correct because it will be be printed since 5 + 3 = 8 is True, so the false branch is not executed, But maybe it is confusing for kids too!

PR #333, opened 2021-04-28
Update contact-zh.md

Grace666-code:
I translated several words in part of do you like Hedy into Chinese. But I don't know where I should add these words in part of subscribing Hedy newsletter, I translated these words into Chinese as following, please use then to replace the right words in this part, thanks. 
* indicates required 必填项
Email Address * 电子邮箱
First Name 名
Last Name 姓
Country 国家


Felienne:
Ah great point @Grace666-code about the mail letter terms. I guess we should change these in MailChimp, I will look into it.

PR #341, opened 2021-05-01
add central security settings for cookies

lpmi-13:
We aren't currenty setting samesite values for cookies, and that would be a good thing to do (https://flask.palletsprojects.com/en/1.1.x/security/).

This commit sets the cookie values for samesite, httponly, and secure in a central place so we don't need to worry about it if we add more cookie values.

### DIFF
--- a/app.py
+++ b/app.py
@@ -137,6 +137,14 @@ if os.getenv ('REDIRECT_HTTP_TO_HTTPS'):
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
 app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
 
+# Set security attributes for cookies in a central place
+app.config.update(
+    SESSION_COOKIE_SECURE=True,
+    SESSION_COOKIE_HTTPONLY=True,
+    SESSION_COOKIE_SAMESITE='Lax',
+)
+
+
 Compress(app)
 Commonmark(app)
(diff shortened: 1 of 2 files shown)


lpmi-13:
I'm happy to add additional information around this PR, as necessary.

As a short summary, there are a number of these issues found when running the `owasp/zap2docker-stable` container baseline scan, and I'm hoping to start addressing them.

Please let me know if any additional information would be helpful (I can open a specific issue, or comment more here, or whatever you think).

lpmi-13:
This could also be addressed with a more comprehensive central update like https://github.com/GoogleCloudPlatform/flask-talisman, though would require an additional dependency. On the other hand, it would deal with security header issues at the same time. Happy to delve deeper into details, if that would be helpful!

fpereiro:
Hi @lpmi-13 ! This PR is very welcome :). From what I see, the extra header is a protection against CSRF - we currently don't have CSRF tokens, so this definitely improves our security.

To get things going, I can merge this. But Talisman looks intriguing. I don't like the extra dependency, but it might be quite worth it. If you want to open a separate PR with a quick implementation of it that would be awesome!

lpmi-13:
Yes, its CSRF and also some XSS stuff, and we could also use the HSTS header to remove our manual conversion from http to https (though this last bit might be more easily done in heroku anyway).

I'll investigate a bit and open an additional PR for discussion.

fpereiro:
That'd be awesome, thank you. Really appreciated!

PR #343, opened 2021-05-01
feat: add a translation tool

rix0rrr:
Hopefully makes it (slightly) easier to translate all Hedy content.

Visit:

```
http://<hedy>/translate/en/nl
```

(Or whatever pair of languages you want to translate)

This PR uses content from the "adventure-mode" branch (without actually
having the code that uses the data yet).

Change Python YAML dependency from **PyYAML** to **ruamel.yaml** because
it has better support for leaving the original file mostly in place.

I am not proud of the code but it works and I wanted to get it out
quickly.


fpereiro:
@rix0rrr this is fantastic work and it's going to be incredibly helpful. Thank you!!

PR #352, opened 2021-05-03
polish existing German translations

dpprdan:
This PR polishes some of the translations from [Google Translate](https://github.com/Felienne/hedy/commit/782060ee168b6917ef7dee8bacdc253a8f01397f). Still not perfect, but hopefully better.

While working on this I was wondering:

1. Should we add translations of commands? E.g. "`ask` (dat betekent 'vragen')".
2. Should commands always be marked as code (e.g. "`print` still works as in Level 1", not "[Print still works as in Level 1](https://github.com/Felienne/hedy/blob/16fec31f95b750198781cab8ead900c48e1c12fd/coursedata/level-defaults/en.yaml#L37)").
3. Should we mark output in text in quotation marks? (e.g. "This code prints 'nice' if you enter Hedy as a name" instead of "[This code prints nice if you enter Hedy as a name](https://github.com/Felienne/hedy/blob/16fec31f95b750198781cab8ead900c48e1c12fd/coursedata/level-defaults/en.yaml#L102)")

NB: Google Translate also translated some commands. 😮


Felienne:
Thank for helping out!! 

To answer your quesions:

1. Should we add translations of commands? E.g. "ask (dat betekent 'vragen')".

Lovely idea! We don't for Dutch now, but feel free to do that for German

2. Should commands always be marked as code (e.g. "print still works as in Level 1", not "Print still works as in Level 1").

Preferably, yes! It might get a bit weird at the beginning of a sentence (which is why we have not done it everywhere) but if you can work around that, go for it!

3. Should we mark output in text in quotation marks? (e.g. "This code prints 'nice' if you enter Hedy as a name" instead of "This code prints nice if you enter Hedy as a name")

This is deliberately not done, we found it does not help children, it rather confuses them when the code print nice and not 'nice' if you catch my drift?

dpprdan:
@Felienne I followed you suggestions. I also polished the texts/de.yaml where some translations [where missing](https://github.com/Felienne/hedy/issues/205#issuecomment-811052097). 

This is ready for review as far I am concerned.

Felienne:
Thank you so much @dpprdan! Apart form a tiny misplaced space in the yaml, all looks fine!

If you want to continue translation, the lesson plans (adventures) could use a German version too: https://github.com/Felienne/hedy/tree/master/coursedata/adventures

Also, if you want to, you can also use our new translation UI https://www.hedycode.com/translate/en/de, adventures yaml shown at the end), than you do not have to fight the yamls :D

PR #356, opened 2021-05-04
adds teachers page in Nl and En

Felienne:
Also makes some changes to the css to make layout of large md files nicer


Felienne:
@fpereiro: There is just one superweird issue that I cannot fix, and has be to addressed before merge. In code blocks, lines after the second one are indented:

image

This could *not* happen in the code blocks coming from the adventure more, only in these md files. Fiddled around with the css but I think the origin lies elsewhere!

PR #358, opened 2021-05-04
Adds additional security headers to auth cookies

fpereiro:
- Adds additional security headers to auth cookies
- Updates/fixes first test of the E2E suite.

@lpmi-13 it seems that the change from the PR only affects session cookies, not the auth cookies. I added the cookie properties manually here.


lpmi-13:
Oh! Thats not at all what I would've expected. Good catch

fpereiro:
Ha, me neither! But I think that I might not be setting the cookies in a Flask-esque way, so probably that has to do with it :).

PR #370, opened 2021-05-04
Patch 11

Grace666-code:
I found a problem on the homepage of Hedy， I added the Chinese words  "简体中文” which means "Simplified Chinese.


Felienne:
@fpereiro @Grace666-code This too has a merge conflict, can someone address it? 

Thanks!

PR #372, opened 2021-05-05
Adventure mode improvements

### LINKED ISSUE #354: Improvements & fixes to load/save with adventures
Upcoming work here: https://github.com/Felienne/hedy/tree/adventure-mode-improvements

   - For logged in users, load adventure saved programs together with adventures, so they are available.
   - Remove render_adventure in hedy.py, just use new unified level+adventure view to render adventures.
   - Cleanup templates from the standalone adventure view.
   - Add adventure_name to window.State, so it's available to js. 
   - Refactor saveit to simplify the way in which adventure_name is passed.
   - When in an adventure, hide the input to specify the saved program name.
   - When clicking on tabs, set or clear window.State.adventure_name for properly saving programs.
   - When loading an adventure (through a link or through a saved game that belongs to it), open the proper tab.
   - Load saved adventures (if any) when changing tabs
   - When saving a program for an adventure+level, overwrite the old one if it exists
   - Rework:
      - Allow multiple programs for a given adventure+level to be saved (in a global name namespace).
      - Include program name again on saveit.
      - Always show the input in adventure.
      - Load programs: use ids always, but if id belongs to adventure_name show the right adventure tab!
      - Improvement: put the name of the loaded program on the input
      - Improvement: use adventure_name + level or Untitled + level as default program names

fpereiro:
- List of fixes & improvements:
   - For logged in users, load adventure saved programs together with adventures, so they are available.
   - Remove render_adventure in hedy.py, just use new unified level+adventure view to render adventures.
   - Cleanup templates from the standalone adventure view.
   - Add adventure_name and loaded program variables, so the are available to js. 
   - When loading an adventure (through a link or through a saved game that belongs to it), open the proper tab.
   - Load saved adventures (if any) when changing tabs
   - Allow multiple programs for a given adventure+level to be saved (in a global name namespace).
   - Warn before changing tabs if unsaved code can be lost.
   - Put the name of the loaded program on the input
   - Update the name of the saved program on the input after save if the server modified the name to avoid a name collision
   - Use adventure_name + level or Untitled + level as default program names
   - Load default programs for adventures when changing tabs. Also for levels.
   - Warn when switching tabs in case there are uncommitted changes.

- Closes #354


Felienne:
Hi @fpereiro!

Great work. Two tiny things (tiny to point out, maybe 2. is not tiny to fix though....)

* The name bar is too small, so the level number is cropped off :( 
image

* I am hating how much localization is infecting *ALL THE THINGS* but can we localize the file name too? I tis not pretty we have "naamloos" in Dutch but "dice - Level 1" in English.  That does mean the adventure name (from the adventures yaml) plus the translation of the word level from, I guess, the texts yaml?

Felienne (changes requested):
See my reply just yet!

PR #374, opened 2021-05-05
Do not use security headers on session cookie when running locally

fpereiro:
Do not use security headers on session cookie when running locally, because it disrupts session variables somehow.

@tibiba this fixes the problem you had yesterday.


Felienne (approved):
This seems to run, but from the the address has changed from 127.0.0.0 to 0.0.0.0:5000?

Not sure why that is needed but it works!

fpereiro:
I believe the `0.0.0.0` is a requirement for running it in Docker. And it doesn't affect things locally either, so we're all good :)

PR #377, opened 2021-05-05
Make sure lines in double nested loops are correctly indented

karel1980:


Felienne (approved):
Wow what a great catch, embarrassing our demo code has a mistake in it! Lovely of you also to fix it in all languages @karel1980.

karel1980:
@Felienne I can't guarantee that I caught them all, but it seems like that one error got copy pasted around. I just checked all occurrences of "for j"

PR #378, opened 2021-05-06
Sublevels

Tazaria:
Added framework for creating sublevels and added sublevels 7-1 and 7-2.


Felienne:
Hi @Tazaria!

I think my new PR (https://github.com/Felienne/hedy/pull/384) fixes your issue. Can you merge my branch into your and see if that is true? You can also wait until Laura merges my PR in the master and check afterwards.

fpereiro:
Hi @Tazaria ! The fix is now merged so it shouldn't be a problem. There's only a few merge conflicts left in hedy.py - I took a look to fix them but there's some parser specific things that I think it's best I don't touch. If you need help merging let me know.

PR #381, opened 2021-05-06
refactor: cache Lark parsers

rix0rrr:
Instead of creating a Lark parser over and over again
every time we go to parse code, cache and reuse the level
parser to get a speed boost.

Reduces parse time from ~60ms to ~10ms (not the best but still helpful!)


Felienne:
Great work @rix0rrr!!

PR #382, opened 2021-05-06
translated again  zh.yaml

Grace666-code:
I found a lot of problems on the Hedy website (Chinese version), so I has translated most parts of the document again. I can not find the branch of Chinese translations so I pull the request here now.

### DIFF
--- a/coursedata/level-defaults/zh.yaml
+++ b/coursedata/level-defaults/zh.yaml
@@ -42,7 +42,7 @@
         你可以在左边的命令栏里面找到第2级学习内容的所有命令. 还有作业题目栏.
     start_code: "print 你好世界!"
     commands:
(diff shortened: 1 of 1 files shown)


fpereiro:
Hi @Grace666-code ! Thank you for this! Could you change the PR base to master please? Thank you!

fpereiro:
Hi @Grace666-code . Thank you for the updated translations!

There's a couple of syntax issues on the yaml file, but they should be easily fixable:
- On line 45, you use double quotes in the middle of the line, so you should escape them like this: `     -   explanation: "`is`的意思是\"是\", 用来给一个词命名以便在程序中使用"`
- On line 177, you should add a quote at the end
- Also, on coursedata/course/hedy/zh.yaml, could you add these four lines so that your translations of levels 14-17 will be available?
```  
  - level: "14"
  - level: "15"
  - level: "16"
  - level: "17"
```

Thank you!

PR #385, opened 2021-05-07
Revert "refactor: cache Lark parsers"

Felienne:
Reverts Felienne/hedy#381


Felienne:
Gives a weird error on alpha, reverting

PR #386, opened 2021-05-07
refactor: cache Lark parsers, take 2

rix0rrr:
Instead of creating a Lark parser over and over again
every time we go to parse code, cache and reuse the level
parser to get a speed boost.

Reduces parse time from ~60ms to ~10ms (not the best but still helpful!)

------

Re-submission of #381, removing the dependence on an implicit Flask context to obtain the debugging configuration: just pass the `app` object we have.


Felienne:
Yes, this works!

PR #395, opened 2021-05-10
Session variable support for A/B testing.

fpereiro:
- Implements the setting of session variables from a test environment to the main environment in an A/B testing context.
- Implementation is done through a special HTTP header. This entails extra code but sidesteps the problems of 1) sharing sessions; or 2) being able for one environment to read another one's session.
- In this commit on another branch, it can be seen how the function for setting session variables works: https://github.com/Felienne/hedy/commit/87198cb8d21bdbf94fec3bf27d53866af32d62f6#diff-568470d013cd12e4f388206520da39ab9a4e4c3c6b95846cbc281abc1ba3c959R240

### DIFF
--- a/app.py
+++ b/app.py
@@ -170,6 +170,13 @@ if os.getenv ('PROXY_TO_TEST_ENV') and not os.getenv ('IS_TEST_ENV'):
                     continue
                 request_headers [header [0]] = header [1]
 
+            # Sets session variables in a header x-session_vars
+            session_vars = {}
+            for var in session:
+                if var not in ['session_id']:
+                    session_vars [var] = session [var]
+            request_headers ['x-session_vars'] = json.dumps (session_vars)
+
             # Send the session_id to the test environment for logging purposes
             request_headers ['x-session_id'] = session_id ()
             r = getattr (requests, request.method.lower ()) (url, headers=request_headers, data=request.data)
@@ -180,6 +187,12 @@ if os.getenv ('PROXY_TO_TEST_ENV') and not os.getenv ('IS_TEST_ENV'):
                 # We ignore the set-cookie header
                 if (header.lower () in ['content-encoding', 'content-length', 'transfer-encoding', 'connection', 'set-cookie']):
                     continue
+                # We set the session vars returned by the test server
+                if (header.lower () == 'x-session_vars'):
+                    session_vars = json.loads (r.headers ['x-session_vars'])
+                    for var in session_vars:
+                        print ('DEBUG RECEIVING SESSION VAR FROM TEST', var, session_vars [var])
+                        session [var] = session_vars [var]
                 response.headers [header] = r.headers [header]
             return response, r.status_code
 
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -170,6 +170,13 @@ def before_request_proxy():
                     continue
                 request_headers [header [0]] = header [1]
 
+            # Sets session variables in a header x-session_vars
+            session_vars = {}
```
rix0rrr, line 174: Honestly, we probably need to get rid of session_vars that are only readable by a single instance anyway.

Sessions are currently stored on a single box, and this will never scale up to multiple Dynos (if requests get routed to different dynos then the session won't be there), same problem as with A/B testing.

So we need to either:

* Drop sessions and transport all the data via cookies; or
* Store session data in DynamoDB anyway.

Since our sessions are pretty small, we can probably do the first? Although if we have an  `is_admin` flag in the session, that might be trickier and we need something to verify the integrity of that cookie if we go that route...

fpereiro: Good point. I didn't think of that either. I would rather store the sessions in DynamoDB so they can be retrieved everywhere. I'd much rather not have to trust cookies - in my experience, it's always best to give as little to the client in terms of state and let the server manage it.

In any case, the `is_admin` flag pertains to logged in users, and those only use sessions that are stored in DynamoDB. Those cookies are only identifiers and map to a username; the username then maps to its information and privileges.

I'm going to close the PR and implement logic for storing and retrieving session variables in DynamoDB.

rix0rrr: > in my experience, it's always best to give as little to the client in terms of state and let the server manage it.

I mean, of course, but consider this:

If we're going to store sessions in DynamoDB, every page request is going to require a GET to DynamoDB. This will increase latency and worker pool pressure (apparently GETs from outside EC2 aren't as fast as I thought they should be, looks like a DDB operation is routinely ~25-50ms), and it will increase DDB cost (we get charged for data as well as per-request). We will also accumulate garbage (every anonymous visitor will accrue a new record in DDB every time they visit the page... what will clean those up? Yes we can use TTL, now we need to set DDB TTL to match session lifetime...)

Now, those are the downsides. What are the upsides? What are the important pieces of information we store in the session that we are willing to pay these downsides for?

As far as I know, it's:

* A unique ID we use to tie programs together from anonymous users

Is there anything else I'm unaware of? Because if it's only the unique ID... it seems silly to send the users a unique ID in a cookie only to use that to retrieve a DIFFERENT unique ID from DynamoDB.

If the `is_admin` flag comes from DynamoDB anyway, and is not stored in a session, then there's no need for anything else.

So maybe the best solution is just to get rid of Flask session storage altogether, and just send the user a unique ID in a per-session cookie, say that is their "session id", and be done with it.

------------

> I'd much rather not have to trust cookies

Sure, but there are solutions for this. We can *sign* the cookies with a shared secret that's only known to the servers. That way, we can be sure that the server originally created the data we can detect if the user tried to mess with it (let's say they tried to maliciously flip their `is_admin` flag to `True`).

Keeping the server as stateless as possible is a good idea (for scaling), and having the client carry around a signed payload is a tried way of doing that without exposing yourself to risk.

fpereiro: I'm aware that cookies can be signed and it's possible thus to trust client input.

Currently we don't need to store anything in the session, but this PR is meant to do that; an upcoming feature to be tested relies on session variables for non logged-in users. So we do potentially (and unfortunately) need to store session variables.

Regarding the performance, we currently already do a request to DynamoDB per every request for logged in users, to resolve their username from their session. Storing the sessions for anonymous requests would multiply the load by `1 / proportion of requests done by logged in users`. This could be quite significant, you're right about that. However, I believe we chose DynamoDB because it's fast and cheap, so we shouldn't be deterred from using it.

Maybe it's best, as you suggested, to store the session id (and other possible variables) in a signed cookie. We could furthermore delete that session cookie when the user logs out, to make sure that it doesn't remain behind in case sensitive data is stored in there. I'll start working in that direction.

rix0rrr: > we chose DynamoDB because it's fast and cheap

Not as fast as I'm used to, at least not from Heroku 😅

fpereiro:
I think that after #443 this will work again. I'm reopening it since we'll need this mechanism anyway once session cookies can be shared by multiple machines.

PR #400, opened 2021-05-10
Improved GFM implementation with new similar code model

TiBiBa:
In this PR the newest implementation of the similar code model is implemented in app.py as well as new data files are used being free of swear words and possible personal information. The model itself is improved and some minor bugs are fixed.


fpereiro:
CLosing temporarily to reopen, the diff doesn't look properly computed.

PR #401, opened 2021-05-10
Lvl10 highlighter

omedtjeee:


fpereiro:
Hi @omedtjeee  -- thank you for this!

Could you please join all the PRs into one that contains all the changes from your `master` branch? Alternatively we could look at multiple PRs, but if each PR contains only the changes for an individual level. In this way, we avoid having to review the code for the lower levels multiple times.

Thank you again!

Felienne:
Hi @fpereiro! I agree that it is a lot better to make PR's for each level separately (so surely do that for the next ones) but I will pick up this one and its siblings so we keep the list of PR's a bit manageable.

Felienne:
Nice job @omedtjeee!!

I will merge this now, but I do have one request. In ranges can we highlight `in` and `range` and `to` also? I think they too are parts of the syntax.

image

PR #404, opened 2021-05-12
Adjust bcrypt rounds of hashed password on login.

### LINKED ISSUE #387: bcrypt work factor is killing our poor Heroku box
We have `bcrypt` at the default work factor of 12, which is a bit steep for our single-core Heroku box. 

Right now, any bcrypt hash operation takes about ~`300`ms:

```
$ time python -c 'import bcrypt; bcrypt.hashpw(bytes("asdf1234", "utf-8"), bcrypt.gensalt())'

real	0m0.354s
user	0m0.340s
sys	0m0.012s
```

Keeping in mind that the box has only a single processor and so can only do one thing at a time, 3 logins/signups/password resets/any bcrypt operation will block anyone else from submitting a program to the web server or doing anything else for a full second [0]. This is already leading to noticeable load and latency spikes on the metrics dashboard.

Yes, we could spring for more cores, but we're on a bit of a budget and would like to stretch our single-core machine as far as it'll go.

We should reduce the work factor of the `bcrypt` operations to something more reasonable, like 9:

```
~ $ time python -c 'import bcrypt; bcrypt.hashpw(bytes("asdf1234", "utf-8"), bcrypt.gensalt(rounds=9))'

real	0m0.084s
user	0m0.080s
sys	0m0.000s
```

This is going to require some code to check the load factor of a user's existing salt and regenerate, re-hash and re-store their password if their current salt does not match our current desired work factor.

> [0] Not entirely of course, because we have a couple of threads running and not all work is purely CPU-bound, some of the work is I/O bound like reading files from disk or reading things from DynamoDB... but the core work of the Hedy server is compiling code which is *also* CPU bound so we're definitely contending with the main purpose of the web app.

fpereiro:
- Closes #387

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -3,7 +3,7 @@ import bcrypt
 import re
 import urllib
 from flask import request, make_response, jsonify, redirect, render_template
-from utils import type_check, object_check, timems, times, db_get, db_set, db_del, db_del_many, db_scan, db_describe, db_get_many
+from utils import type_check, object_check, timems, times, db_get, db_set, db_del, db_del_many, db_scan, db_describe, db_get_many, extract_bcrypt_rounds
 import datetime
 from functools import wraps
 from config import config
@@ -98,7 +98,7 @@ def routes (app, requested_lang):
 
         # If the number of bcrypt rounds has changed, create a new hash.
         new_hash = None
-        if config ['bcrypt_rounds'] != int (re.match ('\$2b\$\d+', user ['password']) [0].replace ('$2b$', '')):
+        if config ['bcrypt_rounds'] != extract_bcrypt_rounds (user ['password']):
             new_hash = hash (body ['password'], make_salt ())
 
         cookie = make_salt ()
--- a/auth.py
+++ b/auth.py
@@ -21,7 +21,7 @@ def check_password (password, hash):
     return bcrypt.checkpw (bytes (password, 'utf-8'), bytes (hash, 'utf-8'))
 
 def make_salt ():
-    return bcrypt.gensalt ().decode ('utf-8')
+    return bcrypt.gensalt (rounds=config ['bcrypt_rounds']).decode ('utf-8')
 
 def hash (password, salt):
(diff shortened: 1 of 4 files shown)


auth.py
```diff
@@ -96,9 +96,17 @@ def login ():
         if not check_password (body ['password'], user ['password']):
             return 'invalid username/password', 403
 
+        # If the number of bcrypt rounds has changed, create a new hash.
+        new_hash = None
+        if config ['bcrypt_rounds'] != int (re.match ('\$2b\$\d+', user ['password']) [0].replace ('$2b$', '')):
```
rix0rrr, line 101: How about extracting this out to a function:

```python
def extract_bcrypt_rounds(hash):
  ...
```

To calculate this value, and add a couple of unit tests on it. Specifically, one we should have:

```python
def test_extract_rounds():
  salt = bcrypt.gensalt(rounds=7)
  self.assertEquals(7, extract_bcrypt_rounds(salt))
```

Other than that, looks good

rix0rrr (changes requested):

auth.py
fpereiro: I wondered about that too. But because we use the function only in one place, I thought it was overkill to extract it, especially since the bcrypt hash format is so unchanging (single digit rounts are prepended with a 0, but the regex I use allows one or more digits, so it covers both cases anyway). But perhaps, for the sake of having unit tests, it could make sense. I'll amend the PR.

rix0rrr: > we use the function only in one place

It's not just about deduping. It's also very useful to be able to give a piece of code a *name*, and abstract away over the mechanics of what's begin done.

Instead of `random_string_manipulation_here(x)` it's a lot easier to be able to read WHAT the random string manipulation is supposed to do: `extract_meaningful_value_from_string(x)`

rix0rrr: And also testing, of course

fpereiro:
Changes added :)

rix0rrr (approved):

PR #406, opened 2021-05-13
feat: add 'query logging', structured logging per request

rix0rrr:
This introduces structured request logging, with a way to log
counters, arbitrary values, and timers within each request.

Scopes are automatically opened and closed matching Flask requests,
by using `@app.before_request` and `@app.teardown_request`.

Currently annotates a number of functions such as `db_get()`,
`render_template()`, etc etc with `@querylog.timed` which automatically
records the time taken inside those functions, and how often they are
called. Some counters and values are recorded manually with
`log_counter` and `log_value`.

I'd recommend everyone log as many things as they can that may impact
performance or stability of the app, to help with ease of debugging
later on.

The mechanism to add a log sink are in place, but I'm not sure what
the current preferred log target is (I see something has been done
involving Papertrail but I'm not sure what that is or how that works),
and I investigated logging these to JSONBin but we may run into issues
there (it's not strictly optimized for that).

S3 is probably an easy and cheap log target, but I'm open to other
suggestions.

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -2,7 +2,8 @@ import os
 import bcrypt
 import re
 import urllib
-from flask import request, make_response, jsonify, redirect, render_template
+from flask import request, make_response, jsonify, redirect
+from flask_helpers import render_template
 from utils import type_check, object_check, timems, times, db_get, db_set, db_del, db_del_many, db_scan, db_describe, db_get_many
 import datetime
 from functools import wraps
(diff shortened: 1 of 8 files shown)


auth.py
```diff
@@ -307,7 +312,7 @@ def update_profile (user):
                    # If on local environment, we return email verification token directly instead of emailing it, for test purposes.
                    resp = {'username': user ['username'], 'token': hashed_token}
                 else:
-                    send_email_template ('welcome_verify', email, requested_lang (), os.getenv ('BASE_URL') + '/auth/verify?username=' + urllib.parse.quote_plus (username) + '&token=' + urllib.parse.quote_plus (hashed_token))
+                    send_email_template ('welcome_verify', email, requested_lang (), os.getenv ('BASE_URL') + '/auth/verify?username=' + urllib.parse.quote_plus (user['username']) + '&token=' + urllib.parse.quote_plus (hashed_token))
```
fpereiro, line 315: Wow, great catch. I don't know how this could have ever worked, although I distinctly remember having tested it! Oh well.

fpereiro:
Looks great! And thank you for the fix in auth.py with the email template when verifying a changed password.

I still have to review the actual querylog.py. The rest looks good. A possible suggestion is, when logging performance in the db functions, to also pass the name of the table, to make it easier to know what we're querying.

Will finish reviewing on Monday! But in the meantime you want to merge, feel free!

rix0rrr:
> when logging performance in the db functions, to also pass the name of the table

Good call! Will add that

rix0rrr:
@fpereiro do you have any suggestions on where we could best send these logs. I'm not up-to-date on Papertrail, how do we access the logs there?

fpereiro:
Good question. Papertrail is accessible through Heroku (I can add you as a collaborator). But we're in the free tier and we don't keep much data (I think it's max 7 days or 50MB). JSONbin might be a better option.

rix0rrr:
From all of these I think S3 might just be simpler and cheaper. I'll add that then.

fpereiro:
Brilliant, even better. S3 will do very well and can stay around for as long as we want.

PR #408, opened 2021-05-14
fix typos in main/for-teachers-en.md

dpprdan:
This fixes some typos in https://github.com/Felienne/hedy/blob/master/main/for-teachers-en.md, mainly typographic quotation marks in code, i.e. `’` are now `'`. There still are typographic quotation marks in non-code.


Felienne:
What a great catch @dpprdan! I think it would be best to change them everywhere (made an issue for that: #419)

Issue #205, opened 2021-03-05
Add missing German translations


Felienne:
![image](https://user-images.githubusercontent.com/1003685/110076252-79599080-7d84-11eb-9b2c-de01df11726a.png)

I think step this might also be missing from our translation instructions?

fpereiro:
We're also missing translations for lines 32-42 of coursedata/texts/de.yaml to improve user errors

fpereiro:
Summary of missing translations:j
- Main page: https://github.com/Felienne/hedy/blob/master/main/start-en.md
- Press section and contact section: https://github.com/Felienne/hedy/blob/master/main/contact-en.md and https://github.com/Felienne/hedy/blob/master/main/press-en.md
- Hedy docs: https://github.com/Felienne/hedy/tree/master/coursedata/course/hedy/docs-en
- Lines 34-44 of  https://github.com/Felienne/hedy/tree/master/coursedata/texts/de.yaml

dpprdan:
Current status:
- [x] Main folder 
	- [x] Start section https://github.com/Felienne/hedy/blob/master/main/start-en.md
	- [x] Press section https://github.com/Felienne/hedy/blob/master/main/press-en.md
	- [x] Contact section: https://github.com/Felienne/hedy/blob/master/main/contact-en.md
	- [x] Teachers section: https://github.com/Felienne/hedy/blob/master/main/for-teachers-en.md
	- [x] Learn more section: https://github.com/Felienne/hedy/blob/master/main/learn-more-en.md
- [x] Lines 34-44 of https://github.com/Felienne/hedy/tree/master/coursedata/texts/de.yaml
- [ ] Code examples for levels 8-17: https://github.com/Felienne/hedy/tree/master/coursedata/level-defaults/en.yaml
- [ ] Adventures: https://github.com/Felienne/hedy/blob/master/coursedata/adventures/en.yaml

🤔 Has this been moved?
- [ ] Hedy docs: https://github.com/Felienne/hedy/tree/master/coursedata/course/hedy/docs-en

Felienne:
It is no longer feasible to manage translations with issues, moving this to a dedicated mailing list

PR #409, opened 2021-05-14
translate texts in ./main to German

dpprdan:
This translates texts in the https://github.com/Felienne/hedy/tree/master/main folder to German. 

I couldn't find instructions for the [haunted house assignment](https://github.com/Felienne/hedy/blob/65994e714a9490d95c8a1df841adcd29868c3b69/main/for-teachers-en.md#L241) on the teachers page. Is that correct?

Ref #205

### DIFF
--- a/main/contact-de.md
+++ b/main/contact-de.md
@@ -1,0 +1,56 @@
+page_title: Kontakt — Hedy
+---
+# Abonniere den Hedy-Newsletter
+
+<!-- Begin Mailchimp Signup Form -->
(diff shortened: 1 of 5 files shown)


Felienne:
Thanks again! Do reach out to hedy@felienne.com if you want us to send you some Hedy swag (we have laptop stickers and mugs!)

Felienne:
Ow and yes we have not made an adventure for the haunted house, that now only lives in our Dutch Word docs.

dpprdan:
> we have not made an adventure for the haunted house, that now only lives in our Dutch Word docs.

Do we need an issue for that?

PR #410, opened 2021-05-14
Modify e2e test suite to be able to run concurrent/stress tests

### LINKED ISSUE #264: Modify E2E script to make it a stress tester
- Pass target environment & concurrent users.
- Use a certain domain that won't trigger transactional emails.
- Collect & print statistics.

fpereiro:
- Modify e2e test suite to be able to run concurrent/stress tests.
- Allows to run tests against `alpha` or `test`, not just locally.
- Allow certain domains to be considered as e2e test domains, which allows test environments (but not prod) to not send transactional emails to requests originating on an e2e test, and in some cases returning tokens directly instead of emailing them.
- Closes #264 

Notes:
- We need to enable the IS_PRODUCTION flag in production as soon as we deploy this.
- Please don't test against `alpha` or `test` until the code is deployed there, otherwise a lot of transactional emails will be sent to non-existing email addresses.


e2e_tests.py
```diff
@@ -4,19 +4,24 @@
 from utils import type_check, timems
 import urllib.parse
 from config import config
+import sys
+import threading
 
 host = 'http://localhost:' + str (config ['port']) + '/'
+hosts = {'alpha': 'https://hedy-alpha.herokuapp.com/', 'test': 'https://hedy-alpha.herokuapp.com/'}
```
rix0rrr, line 11: Typo in the `test` field?

e2e_tests.py
```diff
@@ -93,95 +109,120 @@ def invalidMap(tag, method, path, bodies):
...
-    ['reset password, invalid password', 'post', '/auth/reset', {}, lambda state: {'username': username, 'token': state ['token'], 'password': 'short'}, 400],
-    ['reset password', 'post', '/auth/reset', {}, lambda state: {'username': username, 'token': state ['token'], 'password': 'foobar3'}, 200],
-    ['login again after reset', 'post', '/auth/login', {}, {'username': username, 'password': 'foobar3'}, 200],
-    ['destroy account', 'post', '/auth/destroy', {}, {}, 200]
-]
-
-run_suite (suite)
+def suite (username):
+    return [
+        ['get root', 'get', '/', {}, '', 200],
```
rix0rrr, line 164: This is pretty unreadable, don't you think? Isn't it better to write these in unit test format, and have a unit test runner run them?

If you don't like the built-in `unittest`, I have good experiences with [`pytest`](https://docs.pytest.org/en/6.2.x/).

So, the `pytest` runner has built-in mechanisms for concurrent running of the tests, timing them, recording the output, etc. etc., so you don't need to build all that machinery yourself. This will improve the code readability by a lot and we'll get more flexibility out of using a well-tested tool.

e2e_tests.py
rix0rrr: Why is the full hostname not just `sys.argv[1]` or something?

e2e_tests.py
```diff
@@ -93,95 +109,120 @@ def invalidMap(tag, method, path, bodies):
...
+        invalidMap ('reset password', 'post', '/auth/reset', ['', [], {}, {'username': 1}, {'username': 'foobar', 'token': 1}, {'username': 'foobar', 'token': 'some'}, {'username': 'foobar', 'token': 'some', 'password': 1}, {'username': 'foobar', 'token': 'some', 'password': 'short'}]),
+        ['reset password, invalid username', 'post', '/auth/reset', {}, lambda state: {'username': 'foobar', 'token': state ['token'], 'password': 'foobar'}, 403],
+        ['reset password, invalid token', 'post', '/auth/reset', {}, lambda state: {'username': username, 'token': 'foobar', 'password': 'foobar'}, 403],
+        ['reset password, invalid password', 'post', '/auth/reset', {}, lambda state: {'username': username, 'token': state ['token'], 'password': 'short'}, 400],
+        ['reset password', 'post', '/auth/reset', {}, lambda state: {'username': username, 'token': state ['token'], 'password': 'foobar3'}, 200],
+        ['login again after reset', 'post', '/auth/login', {}, {'username': username, 'password': 'foobar3'}, 200],
+        ['destroy account', 'post', '/auth/destroy', {}, {}, 200]
+    ]
+
+if len (sys.argv) == 1:
```
rix0rrr, line 201: Switching to [`argparse`](https://docs.python.org/3/library/argparse.html) might be more readable, maintainable *and* usable rather than having unexplained command line behaviors.

auth.py
```diff
@@ -15,7 +15,11 @@
 cookie_name     = config ['session'] ['cookie_name']
 session_length  = config ['session'] ['session_length'] * 60
 
-env = os.getenv ('HEROKU_APP_NAME')
+def is_e2e_test_request (email):
```
rix0rrr, line 18: How about a slightly different approach that doesn't rely on magic email addresses?

What if we put a special header:

```
X-Testing: 1
```

In the request we send from the test?

Then in the `app.py` we can do:

```python
if os.getenv('IS_PRODUCTION'):
  @app.before_request
  def reject_tests():
    if is_testing_request():
      # This probably needs to return a response instead of throwing, but you get the idea :)
      raise RuntimeError('No testing in production!')
```

And if we want to skip certain behavior like sending email for tests:

```python
def is_testing_request():
  return request.headers.get('X-Testing')
```

e2e_tests.py
fpereiro: Fixed the typo, thanks for the catch. The full hostname would be cumbersome to type. I'd rather have a friendly and meaningful alias for the environment, such as `alpha` or `test`.

e2e_tests.py
fpereiro: I respectfully disagree. To me, this format is eminently readable, maps very clearly to the structure of an HTTP request, and the abstraction layer is minimal (it is within the same file and fits in 100 lines). I have extensive experience writing tests in this way without relying on testing libraries which almost always involve a lot of accidental complexity.

e2e_tests.py
fpereiro: Again, I'd rather not add dependencies for very simple functions. But I could document within the tester what parameters it expects so its usage is crystal clear.

auth.py
fpereiro: This is a great idea! I will use a header instead.

e2e_tests.py
Felienne: Chiming in here but I do agree with Rico and I think it is better to use a more traditional way of testing, with unittest. Not only is that more readable for newcomers, it will also make it easier (as Rico says) integrate with existing tooling. I am thinking also of future students or other newcomers, it is very good if they too can easily run all tests, incl. these ones.

fpereiro: While I believe that the existing test runner is extremely easy to run (it's run as a mere python program), it is clear that I hold a minority view. What I suggest is that, once I work on the other suggestions, we merge this and then I open another issue where I will rewrite the e2e suite using unittest or another established tool.

rix0rrr: I would again recommend `pytest` over `unittest`. Trust me, you'll love it :).

e2e_tests.py
rix0rrr: `argparse` is built into the standard library so it's not much of an additional dependency.

e2e_tests.py
rix0rrr: > This format is eminently readable, maps very clearly to the structure of an HTTP request,

Maybe, but wouldn't you rather map to the structure of a *scenario* you're testing, rather than HTTP semantics? Code can already do all of that!

Compare:

```python
def test_create_account_and_login():
  username, password = create_new_account()
  do_login(username, password)

  # We could write an assertion here to assert something, but the above would presumably throw if something
  # fails which would also fail the test.


# 5 concurrent workers
$ py.test -n 5 test_e2e.py
```

vs

```python
        ['login again', 'post', '/auth/login', {}, {'username': username, 'password': 'foobar2'}, 200, successfulLogin],
        ['valid signup', 'post', '/auth/signup', {}, {'username': username, 'password': 'foobar', 'email': username + '@e2e-testing.com'}, 200, successfulSignup],
```

fpereiro: I understand your point but I cannot share it because I am using different criteria to judge readability and elegance. Probably a conversation to have over a beer rather than over a screen :).

e2e_tests.py
fpereiro: OK, I will use argparse.

e2e_tests.py
Felienne: > I would again recommend `pytest` over `unittest`. Trust me, you'll love it :).

I am not against pytest but just in the interest of keeping it all the same I prefer unittest for now

rix0rrr: > I cannot share it because I am using different criteria to judge readability and elegance.

I understand, and it's your right to. Lord knows I've been in your exact situation a lot of times already, where something that is perfectly clear to me wasn't clear to my colleagues 😅.

I guess it's important to consciously make the trade-off when you're in this kind of situation: how important is it that "this thing is super optimized but it's hard to read and modify for other people so effectively I'm the only one who can work on it" vs "boring ugly code that anyone can read and modify but might take longer to maintain". And then decide whether the costs are worth the benefits. That trade-off will be different on a case-by-case basis...

auth.py
Felienne: I guess this then is the only change I am awaiting before merge?

fpereiro:
Sounds good!

The changes required are:
- Use `argparse` and document issue.
- Use a header instead of a magic domain.

I'll drop a mention here once the changes are implemented.

fpereiro:
@rix0rrr @Felienne changes implemented!

e2e_tests.py
```diff
@@ -4,19 +4,34 @@
 from utils import type_check, timems
 import urllib.parse
 from config import config
+import sys
+import threading
 
-host = 'http://localhost:' + str (config ['port']) + '/'
+# USAGE: python e2e_tests.py [CONCURRENT_TESTS] [alpha|test]
```
rix0rrr, line 10: If all is well, `argparse` should actually be telling you this if you forget.

So the other advantage of doing that is that your tool becomes self-documenting, and also responds to `e2e_tests.py --help`.

rix0rrr: See and it does:

```
$  python3 e2e_tests.py --help
usage: e2e_tests.py [-h] [--concurrent CONCURRENT] [--host HOST]

optional arguments:
  -h, --help            show this help message and exit
  --concurrent CONCURRENT
                        how many tests to run at the same time (stress
                        testing), default is 1
  --host HOST           Host against which to run the tests (optional),
                        default is localhost

```

rix0rrr (approved):

e2e_tests.py
fpereiro: True! But I wanted to leave a comment in the code with the usage, particularly for those not acquainted with `argparse`.

I have to say, adding `argparse` was a great idea!

fpereiro:
Just added the `IS_PRODUCTION` flag to hedy-beta in Heroku.

PR #413, opened 2021-05-15
Greek translation of adventures and level-defaults

irisdiakoumi:
Hello again Felienne and team! 🙂 I went on to translate the Adventures section to Greek. Hopefully it's ok!


irisdiakoumi:
PS. I started to used the live translation platform, but as I was about to finish up, it threw a 500 error and I lost all the progress I had made.. I was hoping it would save to local storage but unfortunately that wasn't the case. I thought I should mention it!

Felienne:
Thanks for this @irisdiakoumi We noticed the error on the translation site too and we are supersorry you lost your work there. We have now fixed the error, so it should work but it might be wise to test with one line first :)

PR #417, opened 2021-05-16
add levels 8-17 to ./coursedata/level-defaults/de.yaml

dpprdan:
This adds the German translations of levels 8-17 to ./coursedata/level-defaults/de.yaml.


Felienne:
Wow thank you so much!!

PR #423, opened 2021-05-17
remove typographic quotes

### LINKED ISSUE #419: Remove all typograpic quotes from text to not confuse kids
This [PR]( https://github.com/Felienne/hedy/pull/408) changed all quotes in code, but I think it would be best to also do that in the text (in all langs) to be less confusing.

dpprdan:
This removes all typographic (single) quotes from the repo (with the exception of https://github.com/Felienne/hedy/blob/master/License.txt), i.e. replaces all `‘` and `’` with `'`.

Closes #419


fpereiro:
Looks great, merged!

PR #425, opened 2021-05-18
Fix timing issue with signups

fpereiro:
When logging in a just signed up user, if the timing was very slow, we would overwrite most of the fields of a freshly created user with just `username` and `last_login`. This happened twice.

The fix also reduces an unnecessary call to DynamoDB.

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -192,7 +192,8 @@ def routes (app, requested_lang):
             'password': hashed,
             'email':    email,
             'created':  timems (),
-            'verification_pending': hashed_token
+            'verification_pending': hashed_token,
+            'last_login': timems ()
         }
 
         if 'country' in body:
@@ -207,7 +208,6 @@ def routes (app, requested_lang):
         # We automatically login the user
         cookie = make_salt ()
         db_set ('tokens', {'id': cookie, 'username': user ['username'], 'ttl': times () + session_length})
-        db_set ('users', {'username': user ['username'], 'last_login': timems ()})
 
         # If on local environment, we return email verification token directly instead of emailing it, for test purposes.
         if not env:


auth.py
```diff
@@ -207,7 +208,6 @@ def signup ():
         # We automatically login the user
         cookie = make_salt ()
         db_set ('tokens', {'id': cookie, 'username': user ['username'], 'ttl': times () + session_length})
-        db_set ('users', {'username': user ['username'], 'last_login': timems ()})
```
rix0rrr, line 210: This is fine as a fix for now, but there is a deeper underlying problem that `db_set` is inherently risky with its `get+write` behavior.

It's better to make a distinction between "put record" and "update record" from the get-go, and not have it switch between those dynamically (and, as proven here, unpredictably).

The advantage of not having `db_set` do a `db_get` first is also that we'd remove unnecessary calls to DynamoDB on every write.

------

> When logging in a just signed up user, if the timing was very slow, we would overwrite most of the fields of a freshly created user with just username and last_login. 

It sounds like the problem was that a user would log in before their signup record had been written. How could this happen though? I would think that if a user `POST`s to the `/signup` page, we would first do a `db_set` creating the user record and AFTERWARDS redirects with a cookie to the page that does the login?

Or was there a different issue?

> This happened twice.

That explains why we were getting `KeyNotFound` `'email'` errors. Did you remove the broken records from the database?

rix0rrr (approved):

auth.py
Felienne: Thanks for the analysis @rix0rrr! Do you think we can merge this for now or is it better to implement your more sustainable fix separating put and update?

fpereiro:
Hi @rix0rrr , thank you for the review.

After briefly reviewing our usage of `db_set`, it seems to be very clear in all cases whether we're creating or updating. So it does make sense to split `db_set` into `db_set` and `db_update`. I'd actually prefer to name the former `db_create` to make it even clearer (to me `set` has an upsert-y meaning). I'll open an issue and work on that right away.

Regarding this bug, although I'm not sure about this 100%, this is what I believe happened:
- It all happens during creation of the user (`/signup`).w
- The user is created and written to DynamoDB.
- The code then creates a session for the user (our signup automatically logs you in) and that code either goes very fast or the write to DynamoDB goes very slowly.
- When the user is updated with `last_login`, the server retrieves the existing user from DynamoDB and the user is not present yet. So instead of updating the user, it overwrites it completely. Apparently, DynamoDB must let this happen, because the username is exactly the same on both cases.
- Except for the two times this actually happened, for the rest of the users, the user is merely updated.

In any case, we should have never have made two calls to create and then update the user - this happened because I copied the `/login` logic to `/signup` and didn't think about joining those two calls.

I removed the two users from the DB (one when it happened weeks ago, the second one today). None of them had any programs (because they couldn't really use the app at all). It dawned on me that the session for the last user created like this is still active so I just deleted it.

@Felienne I believe this code should be merged and the separation of the db functions done in a separate issue & PR.

rix0rrr:
> * The code then creates a session for the user (our signup automatically logs you in) and that code either goes very fast or the write to DynamoDB goes very slowly.

See this is the part that gets me.

It feels like it shouldn't be possible, because all our code is synchronous. So how could the login even happen before the write to DynamoDB has finished?

fpereiro:
The code gets executed synchronously and the first write is received by Dynamo. However, by the time the read happens, Dynamo still replies that there's nothing there. My guess is that it is a race condition within DynamoDB - because if you try to create things twice, it wouldn't work (if I remember correctly). But because we're not using it transactionally, perhaps therein lies the issue.

PR #428, opened 2021-05-20
typo

StephanEggermont:
cobe -> code


fpereiro:
Thanks for spotting it @StephanEggermont !

PR #430, opened 2021-05-20
Level 18 19

LauraTSD:
Level 18 and 19 are done. 18 doesn't change a lot but explains how to get a value from a list. Level 19 learns len() (length()) and how to use the for loop to loop through a list.


Felienne:
Ziet er weer goed uit Laura!!

PR #433, opened 2021-05-20
add talisman for security configuration

lpmi-13:
By adding in an overall security config in one central place, it's easier to make sure all the basic stuff is covered. Talisman is mostly for headers, and we can add something like seasurf to deal with the CSRF stuff later.

### DIFF
--- a/app.py
+++ b/app.py
@@ -192,6 +192,16 @@ def before_request_proxy_testing():
         if os.getenv ('IS_TEST_ENV'):
             session ['test_session'] = 'test'
 
+# HTTP -> HTTPS redirect
+# https://stackoverflow.com/questions/32237379/python-flask-redirect-to-https-from-http/32238093
+if os.getenv ('REDIRECT_HTTP_TO_HTTPS'):
+    @app.before_request
+    def before_request_https():
+        if request.url.startswith('http://'):
+            url = request.url.replace('http://', 'https://', 1)
+            # We use a 302 in case we need to revert the redirect.
+            return redirect(url, code=302)
+
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
 if utils.is_production():
@@ -854,8 +864,8 @@ import auth
 auth.routes (app, requested_lang)
 
 # *** START SERVER ***
-
 if __name__ == '__main__':
+<<<<<<< HEAD
     # Start the server on a developer machine. Flask is initialized in DEBUG mode, so it
     # hot-reloads files. We also flip our own internal "debug mode" flag to True, so our
     # own file loading routines also hot-reload.
@@ -865,3 +875,7 @@ if __name__ == '__main__':
     app.run(threaded=True, debug=True, port=config ['port'], host="0.0.0.0")
 
     # See `Procfile` for how the server is started on Heroku.
+=======
+    # Only for local development, production server is started via `gunicorn`
+    app.run(threaded=True, debug=True, port=config['port'], host="0.0.0.0")
+>>>>>>> 71d64f8 (revert the https redirect block and simplify local run)
--- a/app.py
+++ b/app.py
@@ -27,6 +27,7 @@ import utils
 from flask import Flask, request, jsonify, session, abort, g, redirect, Response
 from flask_helpers import render_template
 from flask_compress import Compress
+from flask_talisman import Talisman
 
 # Hedy-specific modules
 import courses
@@ -134,6 +135,40 @@ logging.basicConfig(
 
 
 app = Flask(__name__, static_url_path='')
+
+# set the security stuff https://github.com/GoogleCloudPlatform/flask-talisman
+content_security_policy = {
+    'default-src': [
+        '\'self\'',
+        '*.cloudfront.net',
+        'youtube.com',
+    ],
+    'font-src': [
+        'fonts.googleapis.com',
+        'fonts.gstatic.com',
+    ],
+    'frame-src': [
+        'www.youtube.com',
+    ],
+    'img-src': [
+        '\'self\'',
+        'data:',
+    ],
+    'script-src': [
+        '\'self\'',
+        's3.amazonaws.com',
+        '\'unsafe-inline\'',
+        '\'unsafe-eval\'',
+    ],
+    'style-src': [
+        '\'self\'',
+        '\'unsafe-inline\'',
+        'cdn-images.mailchimp.com',
+        'fonts.googleapis.com',
+    ],
+}
+Talisman(app, content_security_policy=content_security_policy)
+
 # Ignore trailing slashes in URLs
 app.url_map.strict_slashes = False
 
@@ -156,16 +191,6 @@ def before_request_proxy_testing():
     if utils.is_testing_request (request):
         if os.getenv ('IS_TEST_ENV'):
             session ['test_session'] = 'test'
-
-# HTTP -> HTTPS redirect
-# https://stackoverflow.com/questions/32237379/python-flask-redirect-to-https-from-http/32238093
-if os.getenv ('REDIRECT_HTTP_TO_HTTPS'):
-    @app.before_request
-    def before_request_https():
-        if request.url.startswith('http://'):
-            url = request.url.replace('http://', 'https://', 1)
-            # We use a 302 in case we need to revert the redirect.
-            return redirect(url, code=302)
 
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
--- a/app.py
+++ b/app.py
@@ -865,7 +865,6 @@ auth.routes (app, requested_lang)
 
 # *** START SERVER ***
 if __name__ == '__main__':
-<<<<<<< HEAD
     # Start the server on a developer machine. Flask is initialized in DEBUG mode, so it
     # hot-reloads files. We also flip our own internal "debug mode" flag to True, so our
     # own file loading routines also hot-reload.
@@ -875,7 +874,3 @@ if __name__ == '__main__':
     app.run(threaded=True, debug=True, port=config ['port'], host="0.0.0.0")
 
     # See `Procfile` for how the server is started on Heroku.
-=======
-    # Only for local development, production server is started via `gunicorn`
-    app.run(threaded=True, debug=True, port=config['port'], host="0.0.0.0")
->>>>>>> 71d64f8 (revert the https redirect block and simplify local run)
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 Flask==1.1.4
+flask-talisman==0.7.0
 lark==0.10.1
 lark-parser==0.7.8
 gunicorn==19.9.0


lpmi-13:
this is very much still in the discussion phase, I think. But the idea was to get an example implementation down so we could be looking at the same thing. I'll make some specific comments in the code diff itself.

app.py
```diff
@@ -104,6 +105,40 @@ def load_adventure_assignments_per_level(lang, level):
...
+    ],
+    'img-src': [
+        '\'self\'',
+        'data:',
+    ],
+    'script-src': [
+        '\'self\'',
+        's3.amazonaws.com',
+        '\'unsafe-inline\'',
+        '\'unsafe-eval\'',
```
lpmi-13, line 131: these two are generally considered bad practice, and it might be better to try and get rid of the inline script stuff before doing this talisman addition anyway. On the other hand, I could easily add some TODO comments or something to the effect of "these two lines probably shouldn't be here, and we've created issues #XXX and #YYY to address them".

app.py
```diff
@@ -104,6 +105,40 @@ def load_adventure_assignments_per_level(lang, level):
...
+    ],
+    'style-src': [
+        '\'self\'',
+        '\'unsafe-inline\'',
```
lpmi-13, line 135: this is another insecure practice, and the comments are similar to those above.

app.py
```diff
@@ -188,16 +223,6 @@ def before_request_proxy():
...
 
-# HTTP -> HTTPS redirect
-# https://stackoverflow.com/questions/32237379/python-flask-redirect-to-https-from-http/32238093
-if os.getenv ('REDIRECT_HTTP_TO_HTTPS'):
-    @app.before_request
-    def before_request_https():
-        if request.url.startswith('http://'):
-            url = request.url.replace('http://', 'https://', 1)
-            # We use a 302 in case we need to revert the redirect.
-            return redirect(url, code=302)
```
lpmi-13, line 199: I'm curious to know the reason behind adding this initially. Doesn't heroku do redirects from http => https automatically anyway? Since Talisman forces HTTPS, I don't think this section is needed anymore, and actually, I'm not even sure it was needed previously, but there's very likely to be some context that I'm missing here.

app.py
```diff
@@ -808,6 +833,6 @@ def update_yaml():
 if __name__ == '__main__':
     # Threaded option to enable multiple instances for multiple user access support
     if version() == 'DEV':
-        app.run(threaded=True, port=config ['port'], host="0.0.0.0")
+        app.run(threaded=True, debug=True, port=config ['port'], host="0.0.0.0")
```
lpmi-13, line 836: I'm also curious as to why we're not currently running in debug mode locally. I added this, since talisman won't force HTTPS when in debug mode, so this should allow us to use talisman and still run fine locally.

app.py
```diff
@@ -104,6 +105,40 @@ def load_adventure_assignments_per_level(lang, level):
...
+    'font-src': [
+        'fonts.googleapis.com',
+        'fonts.gstatic.com',
+    ],
+    'frame-src': [
+        'www.youtube.com',
+    ],
+    'img-src': [
+        '\'self\'',
+        'data:',
```
lpmi-13, line 125: this seems to have mixed reviews online. If we're using inline base64 images to save on network requests, it might be just as easy to use the original png and optimize it with something like https://squoosh.app/. It technically opens up a XSS vulnerability.

app.py
rix0rrr: You've probably added the `unsafe-eval` because you've determined that it is necessary. The only place I can imagine this is coming from is probably going to be `skulpt`, which is not in power to change. So if that library needs `eval()`, then we need `eval()` and that's not likely to change.

As for `unsafe-inline`... we technically could work those, but they're just so gosh-darn convenient :).

app.py
rix0rrr: Can you elaborate on the unsafety of inline styles? Keen to be educated.

app.py
rix0rrr: Good question. I'll leave this to @fpereiro to answer.

app.py
rix0rrr: Oh that's a good call. I generally start the server with `FLASK_ENV=development`, but you're right.

In fact, I think this whole block can be simplified a lot, because it will only ever be executed for local setups anyway. For production setups, we'll go through the `Procfile` which doesn't execute this block. So yeah, always defaulting to `debug=True` makes sense.

Not sure why we do the check of `version()` -- @fpereiro, do you think this will ever take the alternative branch/

rix0rrr:
Modulo some small discussions, I'm broadly okay with the change.

I would encourage you to discuss the motivation, benefits and costs in more detail when submitting a PR. Basically, sell us on why we should want this 🙂 (and in a little more detail than "because security" please).

lpmi-13:
> Modulo some small discussions, I'm broadly okay with the change.
> 
> I would encourage you to discuss the motivation, benefits and costs in more detail when submitting a PR. Basically, sell us on why we should want this 🙂 (and in a little more detail than "because security" please).

Excellent suggestion (which I will remember for future PRs)! I'll draft something a bit more in-depth. 👍

app.py
lpmi-13: in investigating, it is indeed skulpt (which I hadn't appreciated was giving us python parsing capability in the browser) that requires that `unsafe-eval`. In terms of the `unsafe-inline`, that is also very convenient, and I think I'm going to recommend we keep it, to be elaborated in a comment below.

app.py
lpmi-13: This is definitely one of the more exotic ones, and while most modern browsers won't execute any javascript in CSS attributes (like `style=background-image:url('javascript:alert("hi")')`), it _could_ be an issue for super old browsers, but I'm also going to recommend we keep it, for similar reasons as the `script-src` ones, to be expanded on below.

lpmi-13:
To expand a bit on the motivation behind this PR, I'll add some further comments here.

Given that we don't currently set any security headers on https://hedycode.com, there is a good opportunity to address some low-hanging fruit with a comprehensive approach like Talisman.
- Forcing connections to use https by default over http goes a long way towards mitigating a man-in-the-middle attack, and is good internet hygiene these days.
- Forcing all cookies to use `secure` and `httpOnly` helps protect user sessions, and although we already technically do that in lines 241-242 of `app.py`, giving that responsibility to talisman would let us delete those lines.
- Setting the `X-Frame-Options` header helps with clickjacking, and the `X-XSS-Protection` header, while ignored by Chrome and Firefox in favor of a strong Content Security Policy, would still help out with Safari and IE.

Content-Security-Policy is the tricky one, so I'll give it its own section. Since this implementation uses Talisman, we would need to either specify a completely empty CSP (which seems a bit silly), or define a custom one, as has been done in this PR. While code execution would be allowed via the `unsafe-inline` and `unsafe-eval` directives, this would still need to be first injected using capabilities of our application code itself, which after some inspection, doesn't seem to be possible. We could tighten it up a bit using some `unsafe-hashes` in lieu of `unsafe-inline`, but this would probably end up being more trouble than it's worth, given that every single code change would require updating the hashes, and we don't have anything that would automatically update those for us (there would also be a LOT of them).

So to summarize, the main benefits are:
- establish a sensible baseline for security headers
- put the configuration for these headers in one central place
- improve our security posture without affecting our current local development flow

and the main drawbacks are:
- introduces an additional dependency (we could use meta tags for the CSP stuff, but things like `X-Frame-Options` aren't possible to do via meta tags)
- we might need to tweak the CSP as we add more things (probably not, given the permissiveness of the `script-src` and `style-src`, but it's a possibility to call out)
- This approach doesn't address any Cross-Site Request Forgery issues, and those will have to be handled with something like https://github.com/maxcountryman/flask-seasurf.

As a note, in more testing, it looks like the `X-Content-Type-Options` header is going to mess with `templates/error_messages.js` , since the mimetype isn't _technically_ `text/javascript`, so we might want to look at disabling that header for the three routes that use it. I can update this PR to reflect that if it seems like a sensible approach.

app.py
fpereiro: Hi @lpmi-13 ! If I remember correctly, Heroku doesn't do the redirect from http to https automatically, so that's why I added that block. I distinctly remember testing this in `alpha` and seeing how this block enforced the redirect. I just tested `wget --save-headers --max-redirect=0 --output-document - http://hedycode.com` and it replies with a 302 redirecting to `https`, so I'm now pretty sure we need this :)

app.py
fpereiro: HI @lpmi-13 & @rix0rrr -- this block would be necessary unless `__name__` was not equivalent to `'__main__'` - if we remove it, wouldn't it be executed in Heroku as well?

:+1: to run debug=True locally.

fpereiro:
Hi @lpmi-13 -- thank you for the code and the comprehensive explanation!

If I understand correctly, this PR will allow us chiefly to reduce the risk of XSS by adding security headers.

- Regarding http -> https, I believe that the 302 redirect we have above already takes care of this in a more basic way (also more universal since it supports old browsers), but there's no harm in using the header for upgrading the transport.
- It's good that Talisman sets `secure` and `httpOnly`, although I believe this wouldn't save us any code since we need to put those lines anyway for setting the cookie. I am curious: does Talisman adds these cookie properties on every call to `set_cookie` that we perform in the application?
- :+1: regarding X-Frame-Options.
- CSP is indeed tricky. My main concern is that we might inadvertently add further security issues by bypassing certain things. My secondary concern is that modifying other parts of the code might become trickier. I wonder:  would a faulty CSP add further risks than having no CSP at all? Or the worst-case scenario of a bad CSP is the same as not having one? Because if not, we can start with a CSP that keeps the code as simple as possible and then, as need arises, we can tighten it up. My choosing of simplicity here is also aimed at lowering the barrier for non-core contributors that might want to try different things in the UI/JS side of Hedy.
- :+1: regarding dealing with CSRF later. From what I've seen, in modern browsers, this can be solved with headers and without recourse to CSRF tokens (here's a discussion about this topic: https://github.com/fpereiro/backendlore/issues/12).
- `so we might want to look at disabling that header for the three routes that use it. I can update this PR to reflect that if it seems like a sensible approach.` sounds good, let's give it a try

Thank you again for your work - it's very appreciated!

app.py
rix0rrr: I meant the `if version() == 'DEV'` condition:

https://github.com/Felienne/hedy/blob/9e7a2e7f3fcaf360f774d429e2f42b9bec18bb80/app.py#L877

In Heroku, the command that is executed is:

```
$ gunicorn app:app
```

(Which you can see in the `Procfile`).

`if __name__ == '__main__'` is only true if the script in question is the one that is started directly, which in the case of `gunicorn` is not true.

So I think we could simplify this whole block to:

```python
if __name__ == '__main__':
    app.run(threaded=True, debug=True, port=config['port'], host="0.0.0.0")
```

app.py
rix0rrr: Ack.

app.py
fpereiro: Ok, then we can definitely simplify it :+1:

app.py
lpmi-13: ah okay...so even after enabled the HSTS header via talisman, on the first connection from a unique client, we'd still want this manual redirect. I suppose we could also use the preload list, though this would require registering and IIRC can't be easily undone if it happens to break anything.

Not sure how old this page's info is, but basically looks like what we had is the only way to do this. https://help.heroku.com/J2R1S4T8/can-heroku-force-an-application-to-use-ssl-tls

I'll go ahead and revert the removal of this block for this particular PR.

app.py
lpmi-13: will update as suggested!

app.py
lpmi-13: actually...I got curious and deployed a very basic flask app with the default talisman config, and it looks like a request using http redirects automatically with a 302 to https.

code is at https://github.com/lpmi-13/flask-talisman-demo/blob/main/app.py
and deployed at http://flask-talisman-demo.herokuapp.com/

It looks like in the docs that talisman _both_ sets the strict transport header as well as returns a 302 to https on all requests. Is there a way we could test out talisman, but without that explicit redirect block?

lpmi-13: ha...and actually, the code in talisman looks like it's basically doing the exact same thing our current code does
https://github.com/GoogleCloudPlatform/flask-talisman/blob/master/flask_talisman/talisman.py#L211

lpmi-13:
arrrggg...so in terms of the X-Content-Type-Options, it looks like there are only a few options that can be disabled on a per-view basis, and that's not one of them. So lemme see if I can rework the way that `templates/error_messages.js` is working its magic. If we can't update that, then talisman might be a no-go.

fpereiro:
@lpmi-13 thanks for your further work in this. I'll review on Monday!

fpereiro:
@lpmi-13 just reviewed your latest work. It seems that Talisman might be a bit unwieldy for the things we're needing.

I was thinking: what if we could group the security headers we want to apply in a single object, and them apply them to each outgoing response with an `after_request` hook? I did this for other headers in a PR that wasn't eventually needed. If you like the approach and give me a list of headers, we could give this a try. The upside for this solution would be that 1) we'd select what headers we need; 2) we'd add most of the security without external dependencies. What do you think?

lpmi-13:
I think you're right, particularly since the main benefits we're gaining with Talisman would be mostly the headers (we already have the redirect and cookie security stuff, plus Talisman wouldn't necessarily give us CSRF protection anyway).

The only big thing that we might not be ready to add yet is the CSP header. It might still be worth implementing, since anything we don't explicitly specify would default to `self`, and there's effectively no difference between specifying `unsafe-inline` for scripts/styles and just not having any CSP, but we can have that conversation in another issue/PR.

The headers we'd want to add would be:
```
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
```

- The strict transport header will tell the browser that on every request after the first, it should auto-upgrade to https without needing to first connect to our server to receive an explicit redirect. The setting above is for 1 year, and although we don't have any subdomains at the moment, there isn't any reason not to include subdomains.
- The X-Frame deny keeps other sites from spoofing ours by loading it inside an iframe, which is probably not very likely as an attack vector, but no reason not to implement this.
- The XSS-Protection isn't used by a lot of the more modern browsers, but gives us a bit of protection for the older browsers (https://caniuse.com/?search=xss-protection), and is a stopgap until we can decide on a solid Content Security Policy, which is the best-practice for addressing XSS these days.

In terms of the only update that we did want to keep from this PR (the simplification of the local run in `Debug` mode), would it be easier to raise a separate PR for that? I'm assuming that the context of this PR (and the discussion) might be confusing later if it gets mingled with a simple change for locally running the server.

fpereiro:
Hi @lpmi-13 ! Thank you for the (as usual) thorough explanation :).

Let's then add those three headers to all outgoing responses! If you want, I could implement it, but I'd rather you have the glory of finally following up on all your effort done in the research. A way we could do it is with an `after_request` hook. Here's the example I mentioned earlier, instead of using `set_cookie` we would set some flask headers directly on the response: https://github.com/Felienne/hedy/pull/443/files#diff-568470d013cd12e4f388206520da39ab9a4e4c3c6b95846cbc281abc1ba3c959R888

Agreed on leaving the CSP for a bit later, or a separate PR.

Regarding your change to simplify local/debug, we already appropiated during a refactor recently. I just checked and it seems to be incorporated already (thank you for that!). Let me know if I missed something.

lpmi-13:
Sounds good! I'll sort that out in the next few days. 💁‍♂️ 🎩

fpereiro:
> Sounds good! I'll sort that out in the next few days.

Thank you @lpmi-13 !

lpmi-13:
The agreed upon approach is now in https://github.com/Felienne/hedy/pull/488, so closing this particular PR.

PR #441, opened 2021-05-23
Corrects a mistake with level numbering

irisdiakoumi:
Corrects a mistake with level numbering (and a couple minor errors)


Felienne:
Hi @irisdiakoumi,

I think the numbering is correct because calculations are only introduced in level 6. The restaurant adventure at this point does not have a level 5, we might make one but it is not needed for all adventures to have steps at each level.

The other small changes are useful, maybe you can update this PR or make a new one with only these changes?

irisdiakoumi:
Hi @Felienne ! Yes I see what you mean with the level numbering! No worries, I'll be making a PR with the rest of the changes!

PR #442, opened 2021-05-23
Level 20, 21 and 22 implemented

LauraTSD:
Levels were the same as before but at a different position now. Implements = and ==, != and <= and >=.


Felienne (approved):
Looks good Laura! I fixed the tiny space issue and now it all runs and all demo code runs too! Well done.

PR #443, opened 2021-05-24
Serialize, sign & deserialize session cookie 

### LINKED ISSUE #429: Re-implement sessions so that they can be stored/retrieved as cookies
Currently sessions for anonymous users only scale to a single machine. Work should be done to ensure that sessions can be held by a cookie in the client, so that multiple machines can share sessions.

- The sessions need to be signed by a shared key.
- There should be utility functions for setting and retrieving variables from the session besides `session_id`. This is required by #353

fpereiro:
- We use a cookie named `hedy-session` to store the entire session. We name it `hedy-session` and not plainly `session` to not mess with Flask's `session` cookie which is automatically set.
- The session is serialized and then signed using the app's secret key.
- This allows to share sessions between multiple machines.
- Closes #429

### DIFF
--- a/app.py
+++ b/app.py
@@ -24,6 +24,7 @@ from auth import auth_templates, current_user, requires_login, is_admin, is_teac
 from utils import db_get, db_get_many, db_set, timems, type_check, object_check, db_del, load_yaml, load_yaml_rt, dump_yaml_rt
 import utils
 import hashlib
+from itsdangerous.url_safe import URLSafeSerializer
 
 # app.py
 from flask import Flask, request, jsonify, session, abort, g, redirect, make_response, Response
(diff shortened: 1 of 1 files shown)


fpereiro:
So... this seems to be a solution to a non-problem. By default, Flask stores the entire session (not just an id) into a signed cookie. So there's no need to add any code after all. I just tested this locally by doing the following:

- Hardcoding the app secret so that it doesn't change between restarts.
- Reading the session before modifying it.
- Adding a certain key value combination to the session.
- Making sure it's there in subsequent requests *after* the server has been restarted (which would erase any trace of the session in memory).
- Doing that again, but with a longer key value combination and making sure that the cookie actually gets longer (it does).
- Changing the secret key and making sure that the old session data is ignored (which it is).

Thank you Flask! Sorry to have doubted you.

PR #444, opened 2021-05-24
Session variable support for A/B testing

fpereiro:
After #443 is merged this will be necessary and also possible.

Original PR here: https://github.com/Felienne/hedy/pull/395


rix0rrr:
I added a script called `ab_test.sh` to start up 2 instances side-by-side ready for testing the A/B testing infrastructure.

To make it work I had to change `PROXY_TO_TEST_ENV` (which would do a search/replace on the current `HEROKU_APP_NAME`) to `PROXY_TO_TEST_HOST`, which is supposed to look like either `http://localhost:5050` or `https://hedy-test.herokuapp.com` (whichever applies).

rix0rrr:
Also started doing the extractions that I mentioned to show how it works and how we can organize the code a bit better that way.

rix0rrr (approved):

PR #446, opened 2021-05-31
Small organizational improvements to `app.py`

fpereiro:
- Do not allow starting a non-dev environment without a SECRET_KEY
- Always change to the root Hedy folder when running the code
- Simplify logic for launching the app.

### DIFF
--- a/app.py
+++ b/app.py
@@ -879,7 +879,5 @@ auth.routes (app, requested_lang)
 # *** START SERVER ***
 
 # Threaded option enables multiple instances for multiple user access support
-if version() == 'DEV':
+if __name__ == '__main__':
     app.run(threaded=True, port=config ['port'], host="0.0.0.0")
-else:
-    app.run(threaded=True, port=config ['port'])
--- a/app.py
+++ b/app.py
@@ -37,6 +37,9 @@ import translating
 import querylog
 import aws_helpers
 
+# Set the current directory to the root Hedy folder
+os.chdir(os.path.join (os.getcwd (), __file__.replace (os.path.basename (__file__), '')))
+
 # Define and load all available language data
 ALL_LANGUAGES = {
     'en': 'English',
@@ -135,7 +138,6 @@ app = Flask(__name__, static_url_path='')
 app.url_map.strict_slashes = False
 utils.set_debug_mode_based_on_flask(app)
 
-
 CDN_PREFIX = os.getenv('CDN_PREFIX', None)
 STATIC_PREFIX = '/'
 if CDN_PREFIX:
@@ -231,9 +233,26 @@ if os.getenv ('REDIRECT_HTTP_TO_HTTPS'):
             # We use a 302 in case we need to revert the redirect.
             return redirect(url, code=302)
 
+def version():
+    """Get the version from the Heroku environment variables."""
+    if not os.getenv('DYNO'):
+        # Not on Heroku
+        return 'DEV'
+
+    vrz = os.getenv('HEROKU_RELEASE_CREATED_AT')
+    the_date = datetime.date.fromisoformat(vrz[:10]) if vrz else datetime.date.today()
+
+    commit = os.getenv('HEROKU_SLUG_COMMIT', '????')[0:6]
+    return the_date.strftime('%b %d') + f' ({commit})'
+
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
-app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+if version() == 'DEV':
+    app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+else:
+    if not app.config['SECRET_KEY']:
+        raise RuntimeError('The SECRET KEY must be provided for non-dev environments.')
+    app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY')
 
 # Set security attributes for cookies in a central place - but not when running locally, so that session cookies work well without HTTPS
 if os.getenv ('HEROKU_APP_NAME'):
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -872,9 +878,8 @@ def update_yaml():
...
 # *** START SERVER ***
 
-if __name__ == '__main__':
-    # Threaded option to enable multiple instances for multiple user access support
-    if version() == 'DEV':
-        app.run(threaded=True, port=config ['port'], host="0.0.0.0")
-    else:
-        app.run(threaded=True, port=config ['port'])
+# Threaded option enables multiple instances for multiple user access support
+if version() == 'DEV':
```
rix0rrr, line 882: But the check on `version()` here can go. If the `if __name__ == '__main__':` block is taken, we will be in development mode by default and so we don't actually need to check `version()` anymore.

app.py
```diff
@@ -231,9 +233,26 @@ def before_request_https():
...
+    commit = os.getenv('HEROKU_SLUG_COMMIT', '????')[0:6]
+    return the_date.strftime('%b %d') + f' ({commit})'
+
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
-app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+if version() == 'DEV':
+    app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+else:
+    if not app.config['SECRET_KEY']:
```
rix0rrr, line 253: Did you mean to check `os.getenv()` here?

app.py
```diff
@@ -231,9 +233,26 @@ def before_request_https():
...
+    the_date = datetime.date.fromisoformat(vrz[:10]) if vrz else datetime.date.today()
+
+    commit = os.getenv('HEROKU_SLUG_COMMIT', '????')[0:6]
+    return the_date.strftime('%b %d') + f' ({commit})'
+
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
-app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+if version() == 'DEV':
+    app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
```
rix0rrr, line 251: `os.getenv()` takes a second argument, which is what should be returned if there is no environment variable with the given name. So this could simplify (a little) to:

```suggestion
    app.config['SECRET_KEY'] = os.getenv('SECRET_KEY', uuid.uuid4().hex)
```

app.py
```diff
@@ -231,9 +233,26 @@ def before_request_https():
...
+    vrz = os.getenv('HEROKU_RELEASE_CREATED_AT')
+    the_date = datetime.date.fromisoformat(vrz[:10]) if vrz else datetime.date.today()
+
+    commit = os.getenv('HEROKU_SLUG_COMMIT', '????')[0:6]
+    return the_date.strftime('%b %d') + f' ({commit})'
+
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
-app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+if version() == 'DEV':
```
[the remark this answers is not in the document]
fpereiro: Agree with everything you said. We should use `IS_PRODUCTION` to see if we're in `beta`, and distinguish between Heroku and non-Heroku.

I would suggest using a helper method (perhaps in `utils.py`) to determine if we're on Heroku, that returns the result of `__name__ == '__main__'`, for readability purposes. Maybe we can call it `is_heroku`?

In terms of which thing to enforce here, I'd make it so that if we're in Heroku, there should be a `SECRET KEY` already.

app.py
fpereiro: Good catch, I meant to check whether the environment variable was there or not. Fixed now.

app.py
fpereiro: Great, just added both your suggestions. Thanks for catching that!

app.py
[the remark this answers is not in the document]
fpereiro: @rix0rrr I went ahead and added `utils.is_heroku`, but we can of course revert to `__name__ == '__main__'`.

utils.py
```diff
@@ -283,3 +283,6 @@ def isoformat(timestamp):
     """Turn a timestamp into an ISO formatted string."""
     dt = datetime.datetime.utcfromtimestamp(timestamp)
     return dt.isoformat() + 'Z'
+
+def is_heroku():
+    return __name__ == '__main__'
```
rix0rrr, line 288: This won't work. `__name__` has a meaning local to a file, and it means "the module name by which the *current* file was imported." For this file, it will always be `"utils"`.

That's why it works as a trick in the entrypoint file: since the entrypoint file is NOT actually imported, it has a magic special value there (`"__main__"`).

See:

* https://docs.python.org/3/library/__main__.html
* https://stackoverflow.com/questions/419163/what-does-if-name-main-do

I have an idea though. Give me a moment.

fpereiro: Oh, great catch. That's on me for not digging a bit deeper on what that means.

rix0rrr:
I changed it a little to trigger most of the behavior off of the main `if __name__ == '__main__':` block, added an `is_production()` method and clarified the `is_heroku()` method a bit (which I think we should avoid using, say, 99% of the time).

What do you think?

rix0rrr (approved):

app.py
```diff
@@ -169,15 +169,21 @@ def before_request_https():
 
 # Unique random key for sessions.
 # For settings with multiple workers, an environment variable is required, otherwise cookies will be constantly removed and re-set by different workers.
-app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY') or uuid.uuid4().hex
+if utils.is_production():
```
fpereiro, line 172: As long as we have a single machine on alpha or test, we don't need this check, so I'm ok with the change.

app.py
```diff
@@ -169,15 +169,21 @@ def before_request_https():
...
+    if not os.getenv ('SECRET_KEY'):
+        raise RuntimeError('The SECRET KEY must be provided for non-dev environments.')
+
+    app.config['SECRET_KEY'] = os.getenv ('SECRET_KEY')
 
-# Set security attributes for cookies in a central place - but not when running locally, so that session cookies work well without HTTPS
-if os.getenv ('HEROKU_APP_NAME'):
     app.config.update(
```
fpereiro, line 178: I think it'd be prudent to have the same cookie attributes in `alpha` and `test`, if only to make sure that there's no unexpected behaviors that can only be tested in `beta`.

fpereiro:
The changes look great @rix0rrr ! The only thing is that I'd prefer that the security attributes be present in `alpha` and `test` as well. If you agree, I can make the change myself on that condition, from `is_production` to `is_heroku`.

rix0rrr:
> If you agree, I can make the change myself on that condition, from `is_production` to `is_heroku`.

Sounds good!

PR #448, opened 2021-06-01
Split db_set into db_create and db_update 

### LINKED ISSUE #426: Split `db_set` into `db_create` and `db_update`
Benefits:
- Avoid race conditions like the one fixed in #425 
- Performance: avoid doing a `db_get` every time we invoke `db_set`.

fpereiro:
Closes #426


rix0rrr (commented):
With my limited understanding, looks good to me.

It's a little scary that we don't have tests on any of this though.

How will we know that we're not accidentally going to `db_create` when we MEANT to do `db_update` and blow away the data that's in the database?

utils.py
```diff
@@ -216,15 +216,17 @@ def db_get_many (table, data, *not_primary):
...
-    return result
+def db_create (table, data):
+    querylog.log_counter('db_create:' + table)
+    return db.put_item (TableName = db_prefix + '-' + table, Item = db_encode (data))
+
+# Updates an item by primary key.
+@querylog.timed
+def db_update (table, data):
+    querylog.log_counter('db_update:' + table)
+    return db.update_item (TableName = db_prefix + '-' + table, Key = db_encode (db_key (table, data)), AttributeUpdates = db_encode (db_key (table, data, True), True))
```
rix0rrr, line 229: > AttributeUpdates = db_encode (db_key (table, data, True), True))

Wait... isn't this supposed to be `AttributeUpdates=db_encode(data)` ?

rix0rrr: Maybe an APi that looks like:

```python
def db_update(table, key, updates):
  ...
```

Is clearer and safer?

rix0rrr: In the definition of `db_key()`, the optional 3rd argument `True` is throwing me off:

```python
def db_key (table, data, *remove):
  ...
```

This can also be called as:

```python
db_key(table, data, {})
db_key(table, data, 'asdf')
db_key(table, data, 5)
```

Wouldn't a default value for the 3rd parameter be more readable:

```python
def db_key (table, data, remove=False):
```

?

rix0rrr (commented):
Ah:

```python
def db_encode (data, *args):
    # args may contain a boolean flag which we use to detect whether we should format the payload for update_item
    # when len (args) is truthy, we know we're dealing with update_item
```

Isn't it simpler to spit into:

```python
def db_encode_for_putitem(data):
   ...

def db_encode_for_updateitem(data):
  ...
```

?

fpereiro:
Hi @rix0rrr -- thank you for the review! Just committed further changes.

> How will we know that we're not accidentally going to db_create when we MEANT to do db_update and blow away the data that's in the database?

Good question. We have the e2e tests which test all the auth endpoints quite thoroughly - 90% of our database use is related to auth. I took this opportunity to add some basic testing of creating, deleting and listing programs.

I also did the following:
- Created an object defining the main keys for each table, so it's more explicit and not buried in a function.
- Replaced optional arguments with default arguments. Great suggestion!

Regarding `db_encode`, I'd rather not split it into two since both versions are used for `db_update` and the non-update mode is used also in `db_get` and `db_delete` functions, so the naming would be tricky.

rix0rrr:
> Regarding `db_encode`, I'd rather not split it into two since both versions are used for `db_update` and the non-update mode is used also in `db_get` and `db_delete` functions, so the naming would be tricky.

Then `db_encode` and `db_encode_updates` perhaps?

fpereiro:
> > Regarding `db_encode`, I'd rather not split it into two since both versions are used for `db_update` and the non-update mode is used also in `db_get` and `db_delete` functions, so the naming would be tricky.
> 
> Then `db_encode` and `db_encode_updates` perhaps?

Done :)

utils.py
```diff
@@ -161,30 +166,30 @@ def db_decode (data):
             raise ValueError ('Unsupported type passed to db_put')
     return processed_data
 
+db_main_indexes = {
```
rix0rrr, line 169: This is excellent!

rix0rrr (approved):

PR #450, opened 2021-06-02
Add functionality for changing an user email from the admin

fpereiro:
- Add functionality for changing an user email from the admin.
- Also refactors the email validation in both Python and js to a single regex.


Felienne (approved):
Tried out the email validation and it still seems to work

PR #454, opened 2021-06-03
Add yaml validation for levels

### LINKED ISSUE #421: Add yaml validation to the rest of the YAMLs
Instead of just the Adventure mode YAMLs. The idea is to leverage the code we already have.

fpereiro:
- Closes #421
- This allowed us to find and fix #453 
- This doesn't include validation for the yamls in `coursedata/texts`. @Felienne @rix0rrr do you think that's also necessary? If so, I can open a new PR. I didn't add it because there's a ton of properties and I wonder whether we could take a different approach of having a canonic file (`en.yaml`)and then comparing from it.


fpereiro:
Tests should pass once #453 is merged

Felienne (approved):

PR #455, opened 2021-06-03
Move ribbon downwards so that it will not overlap with the languages dropdown

### LINKED ISSUE #451: Move the Fork Us on Github ribbon to the left of the page
Move the Fork Us on Github ribbon to the left of the page so that it doesn't conflict with the language bar.

fpereiro:
- Closes #451 
- Ribbon is moved downwards.
- We give higher precedence (z-index) to the language selection dropdown.
- Moving the ribbon to the left didn't look nice *and* it blocked the Hedy logo, so this approach was tried instead.


Felienne:
Haha guess O was not clear about the requirement here! I did not mean to shift it down a bit, I meant to move it to a lower corner, so it is like this:
-----
|      |
---/-  <-- banner here

And not like this
---\-
|      |
-----

Felienne (changes requested):
Haha guess I was not clear about the requirement here! I did not mean to shift it down a bit, I meant to move it to a lower corner, so it is like this:
```
-----
|    |
---/-  <-- banner here
```

And not like this
```
---\-
|    |
----- 

```

PR #457, opened 2021-06-04
Refactoring grammars

Felienne:
This much needed refactoring changes the way we define grammars. Instead of copy-pasting all rules, we can now simply add and override rules in later levels. This is done in the -Partial files. The code then merges all into -Merged files.

This could be made more efficient by preprocessing the merging and only loading the merged files (but since grammars are cached it is not done all the time anyway)


LauraTSD (commented):
It is very easy to find an existing part as it is alphabetically but it is a bit more difficult to see what belongs to with part for example, if elif and elses are separated now. Why are the names Merged and Partial used? The name Partial is not really clear to me, are these the changes to each level? could

Felienne:
> It is very easy to find an existing part as it is alphabetically but it is a bit more difficult to see what belongs to with part for example, if elif and elses are separated now. Why are the names Merged and Partial used? The name Partial is not really clear to me, are these the changes to each level?

Ah that is a great point! With partial I mean it is not the real complete grammar but just a part. But I am open to other names absolutely. What do you think is better?

LauraTSD:
> > It is very easy to find an existing part as it is alphabetically but it is a bit more difficult to see what belongs to with part for example, if elif and elses are separated now. Why are the names Merged and Partial used? The name Partial is not really clear to me, are these the changes to each level?
> 
> Ah that is a great point! With partial I mean it is not the real complete grammar but just a part. But I am open to other names absolutely. What do you think is better?

I thought about something along the lines of Changes / Additions. As the Changes/Additions are in that specific file

Felienne:
> > > It is very easy to find an existing part as it is alphabetically but it is a bit more difficult to see what belongs to with part for example, if elif and elses are separated now. Why are the names Merged and Partial used? The name Partial is not really clear to me, are these the changes to each level?
> > 
> > 
> > Ah that is a great point! With partial I mean it is not the real complete grammar but just a part. But I am open to other names absolutely. What do you think is better?
> 
> I thought about something along the lines of Changes / Additions. As the Changes/Additions are in that specific file

I like Additions, I will update that. Shall we keep "Merged" or do you think another word would be better. Maybe total?

LauraTSD:
I think Total or Complete is more obvious than Merged, merged doesn't say what is merged together. With Total or Complete, you immediately know that it is the entire grammar for that level. So I like those two better. :)

Felienne:
Great I will update those!

PR #460, opened 2021-06-04
fix: save and restore querylog across restarts

rix0rrr:
Sometimes our request handling takes more than 30 seconds, at
which point in time `gunicorn` shoots our poor worker process in the
head. This means the logs never get written and we never get to figure
out what was taking so long.

When this happens, use a gunicorn shutdown hook to save pending log
records and flush the queue to disk, so the next run of the program
can try to submit it.

Also in this change, recording some more values that might come in
useful when debugging (performance) issues:

- Record the PID in the records
- Record the load average in the records


querylog.py
```diff
@@ -189,6 +226,76 @@ def transmitter(timestamp, records) -> Boolean:
...
+        There may be multiple LogQueues trying to load the same files at the
+        same time, so we need to be careful that only one of them actually
+        loads a single file (in order to avoid record duplication).
+
+        We use the atomicity of renaming the file as a way of claiming ownership of it.
+        """
+        candidates = glob.glob('querylog_dump.*.jsonl')
+        for candidate in candidates:
+            try:
+                claim_name = candidate + '.claimed'
```
fpereiro, line 260: I really like this approach.

querylog.py
```diff
@@ -121,6 +142,13 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+def emergency_shutdown():
```
fpereiro, line 145: If I understand correctly, when Python receives SIGKILL (or SIGTERM?), it still waits for operations like this to end gracefully, right?

rix0rrr: In this specific case it's not Python-specific, it's a `gunicorn` feature. See here: https://docs.gunicorn.org/en/stable/settings.html#worker-exit

The documentation on it is a bit sparse, but by inserting a `time.sleep(40)` in a request handler I've been able to observe that it works and gets called. It doesn't work *exactly* the way I expected -- for example, it seems to be run in a different thread so the thread-local below doesn't actually refer to the in-progress `LogRecord`.

Nevertheless, Flask requests seem to be successfully killed which causes the requests to finish and the LogRecord to be written as well.

We'll have to see how this behaves in production and potentially do something more drastic to track down and write out LogRecords in flight. But for now, maybe this is already enough to get us the information that we need.

fpereiro (approved):

PR #469, opened 2021-06-06
feat: also log programs to S3

rix0rrr:
To get rid of speed limits of jsonbin (and to stop having to pay for
it), also log parse log files to S3.

Factor out the `LogQueue` class and reuse it for this logging.

JSON Bin logging has not been turned off yet: it's safer to log to
both destinations in parallel until we're convinced that everythings
work the way we want it to.

Add a script called `tools/download-logs` which can be used to download
all programs from the S3 bucket (and add the AWS CLI dependency to
`requirements.txt`).


fpereiro (approved):
Looks good. Really like the idea of having full ownership of the logs by storing them into S3.

fpereiro:
@rix0rrr let me know if you want me to fix the merge conflicts.

rix0rrr:
Yes, I’ll do it. I have some more TODOs here anyway

rix0rrr:
Ready for review!

Issue #466, opened 2021-06-05
"Restaurant" adventure needs a Felienne original


rix0rrr:
It's missing an image!

Felienne:
Closed by https://github.com/Felienne/hedy/pull/479

PR #470, opened 2021-06-07
Update en.yaml

MarleenGilsing:
Added level 1-6 Fortune teller adventure


fpereiro:
@Felienne yaml fixed. All we need is an image for this adventure! See also #466

Felienne:
Yes I will add one for the fortune teller tomorrow (I already drew it!)

PR #471, opened 2021-06-07
Allow a proportion of 0% in the ab testing redirect

fpereiro:
Currently, if we pass a proportion of 0% to `PROXY_TO_TEST_PROPORTION`, we default to 50%. Besides this being tricky, it is also useful to be able to set the proportion to 0% quickly. This change allows for this.

### DIFF
--- a/ab_proxying.py
+++ b/ab_proxying.py
@@ -68,7 +68,10 @@ def redirect_ab (request):
     user_identifier = current_user(request) ['username'] or str (session['session_id'])
 
     # This will send either % PROXY_TO_TEST_PROPORTION of the requests into redirect, or 50% if that variable is not specified.
-    redirect_proportion = int (os.getenv ('PROXY_TO_TEST_PROPORTION', '50'))
+    if os.getenv ('PROXY_TO_TEST_PROPORTION') == None:
+        redirect_proportion = 50
+    else:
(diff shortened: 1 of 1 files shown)


ab_proxying.py
```diff
@@ -68,7 +68,10 @@ def redirect_ab (request):
     user_identifier = current_user(request) ['username'] or str (session['session_id'])
 
     # This will send either % PROXY_TO_TEST_PROPORTION of the requests into redirect, or 50% if that variable is not specified.
-    redirect_proportion = int (os.getenv ('PROXY_TO_TEST_PROPORTION', '50'))
```
rix0rrr, line 71: I don't see how setting `PROXY_TO_TEST_PROPORTION=0` would cause this whole expression to return 50?

Testing shows it doesn't:

```
$ env PROXY_TO_TEST_PROPORTION=0 python3 -c 'import os; print(int(os.getenv("PROXY_TO_TEST_PROPORTION", "50")))'
0
```

fpereiro: You're right. I had tested it but I think I applied the `int` operation in the wrong way. I just re-tested and you're right. Closing the PR now, thank you for spotting this!

rix0rrr: I think you might have mistyped the variable name?

fpereiro:
I tried with a `FOO` variable to avoid exactly that problem. Seems that, once again, the problem seems to be between the keyboard and the chair.

PR #473, opened 2021-06-08
Allow to share a program publicly

### LINKED ISSUE #452: Allow sharing programs publicly
- Add Share button next to Save
- Allow any user (or non-logged in users) to access public programs.
- Mark the program as public in the Programs page and add functionality to unshare it as well.

fpereiro:
- Functionality for marking a program as public from the Programs page.
- Closes #452 

@Felienne this still doesn't include the green "Share program" button because I'd first like to discuss a few UI matters:
- If the user is not logged in, should we show this button? Should we show it disabled?
- If the user is logged in but the loaded program is from someone else, should we require the user to save the program as their own first and then show the share button? Or should we save+share the program?
- Same as the above for when trying to share a program that is not saved.
- What if the program is already shared? Do we still show the share button? Or do we show an unshare button.

Besides these design issues, the implementation of that button on the main page is non-trivial, because of all the ways in which a program can get there (default program from a level/adventure, loaded program, newly written program) combined with the not-logged vs logged in user. Before I go ahead and implement this, it'd be good to define the scope of this green button.

The existing implementation allows for sharing/unsharing from the Programs page.

### DIFF
--- a/app.py
+++ b/app.py
@@ -369,7 +369,7 @@ def programs_page (request):
 
             date = round (date / 24)
 
-        programs.append ({'id': item ['id'], 'code': item ['code'], 'date': texts ['ago-1'] + ' ' + str (date) + ' ' + measure + ' ' + texts ['ago-2'], 'level': item ['level'], 'name': item ['name'], 'adventure_name': item.get ('adventure_name')})
+        programs.append ({'id': item ['id'], 'code': item ['code'], 'date': texts ['ago-1'] + ' ' + str (date) + ' ' + measure + ' ' + texts ['ago-2'], 'level': item ['level'], 'name': item ['name'], 'adventure_name': item.get ('adventure_name'), 'public': item.get ('public')})
 
     return render_template('programs.html', lang=requested_lang(), menu=render_main_menu('programs'), texts=texts, ui=ui, auth=TRANSLATIONS.data [requested_lang ()] ['Auth'], programs=programs, username=username, current_page='programs', from_user=from_user, adventures=adventures)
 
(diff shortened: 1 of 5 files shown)


fpereiro:
Build failed because of a missing Flask dependency? https://github.com/Felienne/hedy/pull/473/checks?check_run_id=2772761563 - well, this is odd.

Felienne:
> Build failed because of a missing Flask dependency? https://github.com/Felienne/hedy/pull/473/checks?check_run_id=2772761563 - well, this is odd.

Yeah weird you other PR did tot have that! Are you looking into this?

Felienne:
On the UI matters:

> * If the user is not logged in, should we show this button? Should we show it disabled?
I would say we do not show it.

> * If the user is logged in but the loaded program is from someone else, should we require the user to save the program as their own first and then show the share button? Or should we save+share the program?

Oew interesting, I did not think about that! We might need a fork there HAHAHAHA and have kids push their changes to the original creator. (not being serious!) I would show both button. Save means they save it, share means they save + share.

> * Same as the above for when trying to share a program that is not saved.

Here you mean their own program right? Then I would do save + share too

> * What if the program is already shared? Do we still show the share button? Or do we show an unshare button.

Unshare seems perfect for that

> Besides these design issues, the implementation of that button on the main page is non-trivial, because of all the ways in which a program can get there (default program from a level/adventure, loaded program, newly written program) combined with the not-logged vs logged in user. Before I go ahead and implement this, it'd be good to define the scope of this green button.

Ah I did not anticipate all these complex issues!

> The existing implementation allows for sharing/unsharing from the Programs page.

I can also live with just the one on the Programs page for now so we can focus on other issues!!!

Felienne (approved):
Seems ok, merging to test in alpha

Felienne:
It seems to work fine in alpha, perfect!

But... form the perspective of a kid, nothing happens. Maybe we can at the very least tell them that the program is now shared  in a different way than just the word changing? Maybe a nice popup?

Also, we could place the url on the clipboard and tell them in the popup? Like "Your program is now shared and everyone with the link can see it. You can now simply paste the link with ctrl-c to send it to your friend or teacher!"

PR #474, opened 2021-06-08
Fix loading of program belonging to a level

### LINKED ISSUE #452: Allow sharing programs publicly

fpereiro:
Fix loading of program belonging to a level that is not the latest program for that level.

- Fixes a bug that overrode the loaded program for a logged in user if 1) the program belonged to a level and not an adventure; 2) the program was not the latest saved program for that level. In this case, the latest program for the level saved by the user was loaded instead.
 - Closes #452

### DIFF
--- a/static/js/tabs.js
+++ b/static/js/tabs.js
@@ -50,7 +50,7 @@ $(function() {
     });
 
     // If the loaded program (directly requested by link with id) matches the currently selected tab, use that, overriding the loaded program that came in the adventure or level.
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Seems to work although I did not see the bug myself so I was not able to check if it is fixed.

PR #478, opened 2021-06-08
chore: support '+' in email addresses

rix0rrr:
This is a valid and commonly used character in email addresses.


fpereiro:
This is great, thank you for catching it!

fpereiro (approved):

PR #483, opened 2021-06-09
Merges contact and press into learn more for all languages

Felienne:
Cause the menu bar was getting a bit crowded. Also adds the crowd funder to the home page.

Would be nice if the reviewer of this thoroughly checks that all languages still work!


fpereiro (approved):
Great idea!

Just tested locally on all languages and it works well. We're missing a few translations, but that's the same as before - none were lost in this process.

PR #484, opened 2021-06-10
Basic improvements to sharing/unsharing

### LINKED ISSUE #480: Missing things for deploying basic program sharing
- Prompt to let kids know that their sharing/making public was successful.
- Copy share link to clipboard.
- Internationalize texts (either translations or placeholders) so that they are available everywhere.

fpereiro:
- Alerts to give feedback when sharing/unsharing
- Add share in other languages
- Copy program link to clipboard.
- Closes #480


Felienne:
@fpereiro Seems to be working ok! (merged so I can test at alpha)

Maybe in the message "Program shared successfully" we can make more kid friendly and more helpful into: 

"You can now show this program to the work! We have link to share copied to clipboard that you can paste."

fpereiro:
Hi @Felienne ! Besides changing the message, do we want to copy the link to the clipboard already?

Felienne:
Ah I thought we already did but if we don't yet, on clicking ok in "do you really want to share" we can place the link on the clipboard immediately. Saves kids a click!

fpereiro:
> Ah I thought we already did but if we don't yet, on clicking ok in "do you really want to share" we can place the link on the clipboard immediately. Saves kids a click!

WIll do!

PR #485, opened 2021-06-10
Add script for validating texts yamls

fpereiro:
This node script creates a schema on the fly, taking the en version as the canonical one. It then validates all of the other yaml files using ajv and stops if one of them is either invalid yaml or it doesn't match the keys that the English file has.

The script is called by the bash script for validating yamls, so it will become part of our build process.

As you can see, it already found a few missing keys in several languages!


Felienne (approved):
Good idea to validate all other langs based on En!

PR #486, opened 2021-06-10
Preparsing of newlines

fpereiro:
Recenty we noticed that a program with many newlines in a row is very hard to parse. Since we don't add any semantic value to multiple newlines in succession, we can simply condense them into two consecutive newlines.

We also add a newline at the end of each program, since this makes it easier to parse if there's no newline present.

This was almost too easy. My main concern is breaking existing programs, although all the tests are passing.

### DIFF
--- a/app.py
+++ b/app.py
@@ -244,6 +244,9 @@ def parse():
     level = int(body ['level'])
     sublevel = body.get ('sublevel') or 0
 
+    code = re.sub(r'\n{2,}', '\n\n', code)
+    code += '\n';
+
     # Language should come principally from the request body,
     # but we'll fall back to browser default if it's missing for whatever
     # reason.


app.py
```diff
@@ -244,6 +244,9 @@ def parse():
     level = int(body ['level'])
     sublevel = body.get ('sublevel') or 0
 
+    code = re.sub(r'\n{2,}', '\n\n', code)
```
rix0rrr, line 247: Why retain even 2 sequential newlines? Wouldn't `\n{1,} -> \n` be even fewer newlines? (And therefore moar better?)

Felienne:
Hi @fpereiro 

I am afraid it is indeed too easy 🤣  because it messes up error messages, see:

image

The error is in line 5 of the Hedy program, but the parser gets the shortened program and thus says line 4.

This can be done but then we need a mapping or org line numbers to shortened ones and that will be more involved (and I guess we want to do it recursively also so all empty lines are removed?)

rix0rrr (commented):
> We also add a newline at the end of each program, since this makes it easier to parse if there's no newline present.

I don't understand this sentence. Can you explain why it's easier to parse with a newline at the end?

>  we don't add any semantic value to multiple newlines in succession

No semantic value to the execution, that's true. But parser failure messages will typically indicate a line number where a mistake was encountered, and those line numbers will now be with respect to `program_without_newlines.hedy`, whereas the user will want to interpret those line numbers with respect to `original_program.hedy`. 

So you have to retain a mapping somewhere so you can map line numbers between those 2 programs.

Felienne:
> > We also add a newline at the end of each program, since this makes it easier to parse if there's no newline present.
> 
> I don't understand this sentence. Can you explain why it's easier to parse with a newline at the end?
> 
> > we don't add any semantic value to multiple newlines in succession
> 
> No semantic value to the execution, that's true. But parser failure messages will typically indicate a line number where a mistake was encountered, and those line numbers will now be with respect to `program_without_newlines.hedy`, whereas the user will want to interpret those line numbers with respect to `original_program.hedy`.
> 
> So you have to retain a mapping somewhere so you can map line numbers between those 2 programs.

Are you repeating exactly what I said but in a lower voice?! :D

PR #489, opened 2021-06-12
chore: pickle YAML files when we're in production

rix0rrr:
Pickling is ~400x faster than loading the YAML file directly,
so this saves a bunch of seconds once on every restart.

Could be even better if we pre-process the files once before
deploy, but this is a nice start.

This also fixes some otherwise potentially broken unit tests...


utils.py
```diff
@@ -325,3 +362,23 @@ def version():
...
+def atomic_write_file(filename, mode='wb'):
+    """Write to a filename atomically.
+
+    First write to a unique tempfile, then rename the tempfile into
+    place. Use as a context manager:
+
+        with atomic_write_file('file.txt') as f:
+            f.write('hello')
+
+    THIS WON'T WORK ON WINDOWS -- atomic file renames don't overwrite
```
fpereiro, line 377: Would this limitation break only production environments on Windows? Or also dev environments? We currently have a few contributors working from Windows.

fpereiro (approved):
Anything that can bring down our load times at the expense of little complexity is very welcome. Thank you for this!

utils.py
rix0rrr: This will only be used in production, and all of our production is Linux... but good point. I was being lazy. I will add safeguards just to be sure.

fpereiro: Thank you!

fpereiro (approved):

Issue #375, opened 2021-05-05
Exception found in server error log


rix0rrr:
Needs to be investigated:

```
2021-05-05T13:35:15.574265+00:00 app[web.1]: [2021-05-05 13:35:15,573] ERROR   : Exception on /parse [POST]
2021-05-05T13:35:15.574277+00:00 app[web.1]: Traceback (most recent call last):
2021-05-05T13:35:15.574278+00:00 app[web.1]:   File "/app/hedy.py", line 763, in transpile
2021-05-05T13:35:15.574278+00:00 app[web.1]:     result = transpile_inner(input_string, new_level)
2021-05-05T13:35:15.574279+00:00 app[web.1]:   File "/app/hedy.py", line 906, in transpile_inner
2021-05-05T13:35:15.574279+00:00 app[web.1]:     raise HedyException('Parse', level=level, location=["?", "?"], characters_expected="?")
2021-05-05T13:35:15.574281+00:00 app[web.1]: hedy.HedyException: Parse
2021-05-05T13:35:15.574281+00:00 app[web.1]: 
2021-05-05T13:35:15.574282+00:00 app[web.1]: During handling of the above exception, another exception occurred:
2021-05-05T13:35:15.574282+00:00 app[web.1]: 
2021-05-05T13:35:15.574282+00:00 app[web.1]: Traceback (most recent call last):
2021-05-05T13:35:15.574283+00:00 app[web.1]:   File "/app/app.py", line 228, in parse
2021-05-05T13:35:15.574283+00:00 app[web.1]:     result = hedy.transpile(code, level)
2021-05-05T13:35:15.574283+00:00 app[web.1]:   File "/app/hedy.py", line 766, in transpile
2021-05-05T13:35:15.574284+00:00 app[web.1]:     raise E
2021-05-05T13:35:15.574285+00:00 app[web.1]:   File "/app/hedy.py", line 753, in transpile
2021-05-05T13:35:15.574285+00:00 app[web.1]:     return transpile_inner(input_string, level)
2021-05-05T13:35:15.574285+00:00 app[web.1]:   File "/app/hedy.py", line 906, in transpile_inner
2021-05-05T13:35:15.574286+00:00 app[web.1]:     raise HedyException('Parse', level=level, location=["?", "?"], characters_expected="?")
2021-05-05T13:35:15.574286+00:00 app[web.1]: hedy.HedyException: Parse
2021-05-05T13:35:15.574286+00:00 app[web.1]: 
[...]

PR #491, opened 2021-06-14
adding code to log in case parse fails unexpectedly

Felienne:
I don't understand how #375 can happen so now when there is an exception in processing parse we log so we can see what goes on.

### DIFF
--- a/app.py
+++ b/app.py
@@ -292,7 +292,7 @@ def parse():
                         E.arguments['character_found'] = hedy_errors[E.arguments['character_found']]
                     response["Error"] = error_template.format(**E.arguments)
                 except:
-                    print(code) #TEMP!! We need to find the programs that cause weird issue #375
+                    print('DEBUG ISSUE 375', code) #TEMP!! We need to find the programs that cause weird issue #375
                     raise E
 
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -285,11 +285,16 @@ def parse():
...
-                    E.arguments['character_found'] = hedy_errors[E.arguments['character_found']]
-                response["Error"] = error_template.format(**E.arguments)
+                try:
+                    error_template = hedy_errors[E.error_code]
+                    # Localize the names of characters
+                    if 'character_found' in E.arguments:
+                        E.arguments['character_found'] = hedy_errors[E.arguments['character_found']]
+                    response["Error"] = error_template.format(**E.arguments)
+                except:
+                    print(code) #TEMP!! We need to find the programs that cause weird issue #375
```
fpereiro, line 295: Could you add a prefix here to the `print` saying `DEBUG ISSUE 375` (or something of the like) so we can easily identify it? Something like `print ('DEBUG ISSUE 375', code)`

Felienne: > print ('DEBUG ISSUE 375', code)

Good idea, added!

fpereiro (approved):

PR #492, opened 2021-06-14
Final implementation of the Gradual Feedback Model

TiBiBa:
The final implementation of the Gradual Feedback Model containing a small fix where a `KeyError` would occur due to a missing variable being passed when using a POST request on the logger.


fpereiro (approved):
Merged!

PR #493, opened 2021-06-14
fix: improve errors when hitting the parse endpoint

rix0rrr:
We used to dump the whole XHR object in the error box. Now recognize two
cases:

- HTTP request didn't complete: show a message about checking their
  internet connection.
- HTTP request completed but was unsuccessful: tell the user they
  legitimately did something we didn't expect and ask them to send us
  an email.

Also fixes an exception that may occur when the Flask app is reloaded in debug mode.


fpereiro (approved):
Yes! This was long overdue. Thank you for picking it up!

PR #494, opened 2021-06-14
chore: offline database and run e2e tests during build

rix0rrr:
Add the notion of an offline database (for use during development).
This also makes it so the end-to-end tests can be run during build.

We abstract away from the cloud infrastructure necessary to run
all the tests (specifically: DynamoDB), and replace it with an in-memory
database.

This will also make it easier to--during development--test features that
require having an account and being logged in, without having to test
against a real production database.

----

The database code has moved to `dynamo.py` -- there is the concept of
a `Table` with 2 storage backends: `AwsDynamoStorage` and `MemoryStorage`.
`AwsDynamoStorage` is now the host of all the old `db_*` functions, and `MemoryStorage`
has in-memory implementations of all, that serialize themselves to a JSON file
(for easy offline hacking of the database and persistence between restarts).

To abstract from the specific data/table layout of the database, introduced a layer
called `Database` which contains all database operations. This will allow us to change
data layout and indexes without having to change the application code in the future.

Also replaced custom type checking function with Pythonic built-ins (most of
the time it looks like replacing `type_check(x, 'str')` with `isinstance(x, str)`--
notice the quotes).

Sorry about that, I thought the `type_check` function was only used in the Dynamo code and so
removed them, and when I discovered they were being used all over the code base
I just replaced the idiom everywhere.

Also replaced the custom DynamoDB serialization and deserialization routines
with the [built-ins from `boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html).


fpereiro:
Hi @rix0rrr -- I applaud the heroic effort to get the E2E tests running during build!

I want to understand the main difficulty: is it that the build time server cannot perform requests to Dynamo because it has no connectivity? Or because it is tricky to get the environment variables passed to it? If we could somehow make it hit Dynamo (using alpha), we'd also make the E2E truly real.

I have to admit I'm not happy about the complete rewrite on the functions for manipulating users for the sake of running the E2E at build time. My concerns are: 1) more code (though I have to say what you have written is elegant and not bloated at all); 2) inconsistency, since the other DB entities are not written like this - and if we decide to do so, well, that's another debate! 3) it's not a true E2E if it uses a code path that is different from the one used in reality.

I think it all boils down to understanding whether there's a possibility to hit Dynamo from the build process. If so, I think it would be totally worth it. And of course I'm glad to help with that!

rix0rrr:
> I want to understand the main difficulty: is it that the build time server cannot perform requests to Dynamo because it has no connectivity? 

Yes. Preferably we should have a suite of tests that will give us good confidence that "the website" works correctly, which we can run swiftly and in a lot of places. I'm not just thinking "the build", I'm also thinking "my desktop". I definitely don't want Dynamo credentials on my desktop, but I do want the ability to add more tests, and run all of them to (a) make sure my new feature doesn't regress and (b) I haven't broken anyone else.

Testing against an in-memory copy of the database is a quick way to have more dependency-free testing. Plus, apart from the E2E test suite--if you're developing on any feature that requires you to be logged in, an in-memory database will be an easy way to test the functionality (without, again, having to have production credentials on your workstation, which should be a big no-no).

> If we could somehow make it hit Dynamo (using alpha), we'd also make the E2E truly real. [...]  it's not a true E2E if it uses a code path that is different from the one used in reality.

For sure, and we should totally do that as well! Running against an in-memory database is not a replacement, it's an addition!

The ultimate goal here is to develop a comprehensive enough test suite that we feel comfortable saying: "if it passes this test suite, it's good to go". That means: no more manual promotion from alpha to prod: every new commit gets deployed to alpha (if it's not obviously broken), and if it passes the tests then it can automatically go to prod! 

That will get rid of the built-up inventory problems we occasionally have on alpha.

However, that requires a chain of things:

* We need to have confidence in the completeness of the test set
* WHICH REQUIRES a lot of tests to be written
* WHICH REQUIRES people to be comfortable writing and running tests
* WHICH REQUIRES few dependencies and the tests not taking a long time to run

So view this as the first step towards the ultimate goal of Continuous Delivery.

> more code

I'm of two minds on this still. I'm in doubt between the API being functional (`register_user`, `update_password`, `make_program_public`, `login_user` etc), and the API just being an slight abstraction over DynamoDB that just happens to have an in-memory implementation (basically: `db_update` can work in-memory).

The advantage of the first is that we can hide some concerns. For example, the `login_user` does a couple of things with the `users` and `tokens` table, which are together conceptually one operation and can be represented as such. HOW it's implemented is just a detail. In fact, we could change the datamodel easily with the first API (collapse some tables, move columns and indexes around) and none of the application code would need to change.

The downside of the first is that there are a lot of operations we might want to do, and at some point I just gave up and added `update_user` (because there are many individual user fields that we might want to update). In that sense, the 2nd one will be less code overall (but also allow less opportunity for improvements and refactoring).

---

In any case, one feature definitely missing atm is writing the in-memory database to local disk; especially in debug mode hitting `Ctrl-S` on a Python file will restart the project and clear your "in-memory" database, which is probably a bad experience if you're trying to work on some issue that requires you to be logged in 😅

fpereiro:
Hi @rix0rrr ! Thank you for your comprehensive explanation. To summarize my points after your explanation:

- I completely agree with having full coverage of E2E tests - this is always a good thing because it raises the bar on code quality and makes processes more automatable.
- I usually am wary of tests that don't use the same code paths as actual production use. In this case, things that work on the in-memory database may not work on Dynamo, or viceversa. However, it is indeed a heavy requirement to require Dynamo access to be able to run the E2E tests, so something's got to give.
- I like the approach of having the API functions use low-level DB functions directly since that approach has worked out empirically for me to minimize lines of code and eliminate an extra layer of abstraction. My starting point is to minimize lines of code and layers of abstractions. But that's a stylistic decision that can be indeed debatable, particularly if the starting points are different.

rix0rrr:
> * I usually am wary of tests that don't use the same code paths as actual production use. In this case, things that work on the in-memory database may not work on Dynamo, or viceversa.

I think I want to say one thing more to this, to clarify my reasoning about the trade-offs here.

I hope we can take the following as a truism:

**The code most likely to break (contain bugs) is code that's changed**

It stands to reason that this is true: if code isn't changed, after a while we'll have discovered and fixed all the bugs in there, and since we're not changing the code it also means that no new bugs are being introduced. 

On the other hand if we introduce a change, there's no telling if we left any bugs in there that we accidentally didn't detect (because of lack of testing, because we plain forgot).

With that established, we can now reason about "likelihood to contain bugs" as a function of "how often is this code changed".

Looking at these two components:

```
┌───────────────┐           ┌───────────────┐
│               │           │               │
│               │   uses    │   Database    │
│    Website    │──────────▶│  abstraction  │
│               │           │               │
│               │           │               │
└───────────────┘           └───────────────┘
        ▲                           │        
        │      helps with testing   │        
        └───────────────────────────┘        
```

The question is: is it worth investing time into the **database abstraction** to improve testability of the **website** ? Aren't we more likely to introduce bugs in the abstraction, negating the benefits?

I would propose we answer that question by looking at the rate of change:

1. If the **website** code is more likely to change than the **database** code: it becomes reasonable to assume that we'll have eliminated the bugs from the **database** code after a while, at which point it becomes a solid and reliable tool which can help us test the **website** code (which is a good thing, since that needs the most tests as it's the thing that's more likely to change).
2. If the **database** code is more likely to change than the **website** code: it's probably not reasonable to invest effort into the abstraction, as we now have to do double work every time we make a change there: we have to implement the change for the "real" database and for the "fake" database... and the reason for doing that would be to gain more confidence in a component that doesn't change so often, and therefore can be reasonably expected to not need as much of test coverage.


*(Now obviously there's room in the middle between two extremes: we can discuss nuances, and it also depends on our tolerance to bugs--could a bug in a component potentially be life-threatening?--and the likelihood of introducing bugs... we can argue about the numbers but this is the basic framework I'm reasoning with)*

I would argue that we are more likely in scenario (1) than in scenario (2), which is why I think it's worth investing in a double code path (and even dealing with the bugs that come from that strategy) if it makes testing the component that's seeing the highest rate of change, and increasing our confidence in that code, easier.

rix0rrr:
> * My starting point is to minimize [...] layers of abstractions.

The thing is, well-chosen layers of abstraction are benefits, not liabilities. To trot out a cliché example: **filesystems**.

Every time you open a file, do you think the following:

* Is this disk FAT32 formatted? If so, I need to read the File Allocation Table at disk offset `0xc700` to find the location of the disk entry, then read it to find the location on disk, and now I can start reading the file data!
* Oh no, wait, it's NTFS formatted. I now read the BPB and EBPB at slightly different offsets with different data to find the file locations...
* I also need to not forget to add support for common Linux filesystems, is it `ext3` or `ext4` or ...

You don't do that--you rely on an abstraction which the Operating System provides to you where you can think only about the files, and don't have to think about how they're laid out on disk. The Operating System can even switch those implementations out from under you and your application code doesn't need to know, because it's coded to the abstraction and the abstraction stays the same!

Everyday work computers are doing is built upon dozens of these abstraction layers, and this all comes out to our benefit.

It's not abstractions themselves that are the problem--it's probably **poorly chosen** abstractions that you have bad experiences with. 

It's challenging to design them correctly, and there's surely an art there: you want to hide irrelevant details but not to the point that the consumer of your abstraction can't get their work done anymore. You want to be able to swap out implementations without the consumer noticing, and while doing that not accidentally detect that you've tied yourself to one concrete implementation because you made too many assumptions in your abstraction.

So yeah, it's *hard*, but I would say ultimately worthwhile (when done correctly).

fpereiro:
> With that established, we can now reason about "likelihood to contain bugs" as a function of "how often is this code changed".

I agree that bugs are much more likely in newer code; also in untested code. But for me, the main point is that bugs are a function of lines of code - the more code, the more bugs. This is one of the findings of Fred Brooks in the Mythical Man-Month.

The approach that I see here adds an extra layer between the database and the application code. Right now, the auth functions (which don't change often at all and can be considered complete at the momebt), directly call the DB primitives we have already defined - to me, that's also a conceptual win, because it makes it easy to understand the data flow. Adding an extra layer here 1) adds 10s to 100s of lines of code; 2) all of it new; 3) makes the E2E funnel go through a different code path than the one used in production (which introduces the posibility of false positives and false negatives, from the perspective of production code).

> The thing is, well-chosen layers of abstraction are benefits, not liabilities.

You're right and I wholeheartedly agree. If we minimize layers of abstractions, we'd just be writing reams of pages with opcodes. What I meant to say is: minimize the number of abstraction layers while also minimizing the overall complexity of the implementation. To me, it again boils down to lines of code: what you can do with tens of thousands of opcodes, you can do with a line of Python. The latter is more maintainable and easier to reason about.

The ideal system, to me, is the one that reduces overall complexity, at every level of abstraction. What's the combination of abstraction layers that generates the simplest implementation, at every level? For me, empirically, this turned me on the path of 1) using a high level language; 2) using very few and carefully selected dependencies; 3) minimizing abstraction layers done at the app level; 4) minimizing the lines of code of the application. I however understand this is a minority opinion and one that is not backed by either peer-reviewed papers or the standard practices of the industry leaders in software. It has the following downsides:

- It makes your code more bound to the decisions you already made.
- It puts a very high bar on the introduction of new features or patterns. In other words, it is conservative.
- It discourages contribution because, in some sense, it reinforces the cathedral approach, only that you're building something way smaller than a cathedral. Perhaps a model train would be a better analogy.
- While it reduces the overall complexity of it all, it probably increases the required understanding for changing any part of it, so it essentially rises the required average understanding for making changes to the code.

Felienne (approved):

PR #497, opened 2021-06-15
Overwrite save

### LINKED ISSUE #482: Change behavior of save
- Overwrite already saved programs when saving them again, instead of saving them with another name.
- Namespace is the name of the program, not its context (level plus mode (level mode vs adventure x).
- Auto-overwrite without confirmation.
- Remove the "Untitled" from titles and simply have the level + [optional] adventure name.

fpereiro:
- Closes #482


Felienne (approved):
Merging so I can test in alpha (I still can't test DB stuff locally)

PR #498, opened 2021-06-15
Add endpoint for logging uncaught client exceptions

fpereiro:
When the js client throws an uncaught exception in the client, it'd be good to have a trace of this. I added an endpoint in the server dedicated just for this, but we could repurpose the one we have for catching exceptions when running Python code returned by Hedy.

Possibilities to consider:
- Merge two endpoints into one and use a 'X' level to represent general browser exceptions.
- Make the server always return 500 (or a 5xx) so we get alerts immediately if the client js breaks. I'd like these errors to trigger alerts so we don't have to go look for them after the fact.


static/js/app.js
```diff
@@ -321,6 +321,19 @@ function reportClientError(level, code, client_error) {
   });
 }
 
+window.onerror = function reportClientException() {
+
+  $.ajax({
+    type: 'POST',
+    url: '/client_exception',
+    data: JSON.stringify({
+      error: [].slice.call(arguments)
```
rix0rrr, line 330: This data type is going to be an array. Is that desirable?

The arguments to `window.onerror`s handler are probably pretty well-defined, so we can be more specific than this.

app.py
```diff
@@ -338,6 +338,21 @@ def report_error():
 
     return 'logged'
 
+@app.route('/client_exception', methods=['POST'])
+def report_client_exception():
+    post_body = request.json
+
+    parse_logger.log ({
```
rix0rrr, line 345: Why is this going into the `parse_logger`? That's the log stream Felienne does her program analysis on--is it the right place?

I would log it into the querylog logger.

static/js/app.js
```diff
@@ -321,6 +321,19 @@ function reportClientError(level, code, client_error) {
   });
 }
 
+window.onerror = function reportClientException() {
```
rix0rrr, line 324: Do you have a sense for the exceptions that would be caught by this handler? Have you seen any in practice? Or is this more of a "defense in depth" kind of thing?

I was under the impression we have pretty good coverage with `try/catch` statements (and/or Promise handlers), so I'm guessing not a lot of exceptions are going to be uncaught.

static/js/app.js
fpereiro: Good point; the arguments are different in old browsers that we're not supporting anyway. I'll change it to be more specific.

app.py
fpereiro: Great catch!

static/js/app.js
fpereiro: Good question. The main use case for this is to catch js programmatic errors, usually introduced in new features, or that blow up because the user has a set of circumstances that we haven't tested. It is a catch-all error handler for those circumstances, and particularly useful after deploys.

I haven't seen this pattern in other apps but I've used it myself in a few webapps with good results.

fpereiro:
@rix0rrr ready for re-review :)

rix0rrr (approved):

PR #500, opened 2021-06-15
Fixed the error where an automatic error was given in level 10 with indents

LauraTSD:
The error was in the start code which resulted in a lot of errors in user programs.

### DIFF
--- a/coursedata/level-defaults/en.yaml
+++ b/coursedata/level-defaults/en.yaml
@@ -328,7 +328,7 @@
     start_code: |-
         for i in range 1 to 3:
             for j in range 1 to 5:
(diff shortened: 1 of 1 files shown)


LauraTSD:
Fixes error https://github.com/Felienne/hedy/issues/501

Felienne:
Well done @LauraTSD! Did you find this error through the logs?

LauraTSD:
> Well done @LauraTSD! Did you find this error through the logs?

Yes I did! 30.53% of the errors made of level 10 were caused by this error.

Felienne:
Great catch!! And I was wondering why I did not see it but that is because I only test in Dutch... So I learned a good lesson too!!!

PR #502, opened 2021-06-15
Fix error level 10 indents other languages

LauraTSD:
Another fix for https://github.com/Felienne/hedy/issues/501 but in different languages

Also found an error in level 11. A = was used in level 11


Felienne (approved):
Ah yeah translators of course did their work based on the En version!

PR #507, opened 2021-06-17
Tool for batch test/benchmark of the transpiling

JGobeil:
batchhedy.py is a tool to do batch testing and benchmarking of the Hedy transpiling

I can summit optimisation in another pull request, or in this one.


Felienne:
Thanks @JGobeil!! I will look at this soon!

rix0rrr:
Hey @JGobeil, thanks!

It looks like `pylint` is complaining about this code:

```
batchhedy.py:27:0: E0611: No name 'BaseModel' in module 'pydantic' (no-name-in-module)
batchhedy.py:27:0: E0611: No name 'FilePath' in module 'pydantic' (no-name-in-module)
batchhedy.py:174:4: E0102: method already defined line 66 (function-redefined)
batchhedy.py:192:4: E0102: method already defined line 67 (function-redefined)
```

The final 2 error messages are probably because

```py
    @lazy
    def jobs(self):
```

conflicts with:

```py
    jobs: ClassVar[None]
```

And the same for `checkdata`. 

I'm a bit more mystified about missing definitions in `pydantic`. Got any thoughts on those? 

If you think these issues can't or don't need to be addressed, can you add suppression comments? https://pylint.readthedocs.io/en/latest/user_guide/message-control.html

batchhedy.py
```diff
@@ -0,0 +1,389 @@
+#!/bin/env python3
+""" Usage: batchhedy.py [--help] [--level <level>]... [--output <output>]
```
rix0rrr, line 2: This says `--output` is optional, but if I forget it I'm greeted with:

```
  File "batchhedy.py", line 381, in <module>
    RunHedy(**{
  File "batchhedy.py", line 72, in run
    self.output.mkdir(parents=True, exist_ok=True)
AttributeError: 'NoneType' object has no attribute 'mkdir'
```

🙂

rix0rrr:
Okay the linter errors on missing symbols can be fixed by adding the following file:

`.pylintrc`

with:

```
[MASTER]
extension-pkg-whitelist=pydantic
```

Felienne:
Superseded by #537 thanks for this @JGobeil!!

PR #511, opened 2021-06-18
Avoid using the resource module on querylog.py if using Windows

fpereiro:
As reported by @omedtjeee .

- Avoid using the resource module on querylog.py on Windows to maintain Windows compatibility.
- The solution loses some functionality for Windows, but it should still do most of the logging. Non-windows functionality is untouched.

@rix0rrr do you think it is worth it to explore a Windows supported alternative to `resource`?

### DIFF
--- a/website/querylog.py
+++ b/website/querylog.py
@@ -44,8 +44,8 @@ class LogRecord:
 
         self.set(
             end_time=dtfmt(end_time),
-            user_ms=user_ms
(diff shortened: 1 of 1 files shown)


rix0rrr:
> @rix0rrr do you think it is worth it to explore a Windows supported alternative to resource?

No. Totally not important.

rix0rrr (approved):

PR #527, opened 2021-06-22
fix: add caching on static content

rix0rrr:
About 30% of the total request volume to the server is to retrieve
static error messages to use in the JavaScript (`/error_messages.js` and
`/auth/texts`). Because they are not cached, they are currently loaded
upon every page load (every time a user hits **Next level**).

They could *easily* be cached though, bringing down the total request
volume to the server by a lot.

Add a caching header to make the browser cache these responses for an
hour; we could think about routing them through the CDN as well,
but that's perhaps a bit much for now. Let's just see how this performs.

### DIFF
--- a/app.py
+++ b/app.py
@@ -23,7 +23,7 @@ from utils import db_get, db_get_many, db_create, db_update, timems, type_check,
 import utils
 
 # app.py
-from flask import Flask, request, jsonify, session, abort, g, redirect, Response
+from flask import Flask, request, jsonify, session, abort, g, redirect, Response, make_response
 from flask_helpers import render_template
 from flask_compress import Compress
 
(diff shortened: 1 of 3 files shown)


fpereiro:
In general, I prefer cache mechanisms that live as long as the content doesn't change. But given that 1) these assets rarely change; and 2) we only cache them for an hour, I think this solution should work well.

The only issue I think would be with local development. Perhaps we should avoid setting the max_age when in `DEV`, otherwise we should force refresh when developing changes to these assets and it is easy to forget to do that; even more so for first-time contributors.

fpereiro (approved):

rix0rrr:
> The only issue I think would be with local development.

Good call!

PR #529, opened 2021-06-24
Add validations to prevent 500s with malformed URLs

fpereiro:
Currently, `/space_eu/1%3Flang=en` and `/hedy/1?lang=en?lang=en` trigger 500s. This code fixes both classes of malformed URLs by adding validations.

### DIFF
--- a/app.py
+++ b/app.py
@@ -525,7 +525,10 @@ def index(level, step):
 @app.route('/onlinemasters/<level>', methods=['GET'], defaults={'step': 1})
 @app.route('/onlinemasters/<level>/<step>', methods=['GET'])
 def onlinemasters(level, step):
-    g.level = level = int(level)
+    try:
+        g.level = level = int(level)
+    except:
+        return 'No such Hedy level!', 404
     g.lang = lang = requested_lang()
     g.prefix = '/onlinemasters'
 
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -525,7 +525,10 @@ def index(level, step):
 @app.route('/onlinemasters/<level>', methods=['GET'], defaults={'step': 1})
 @app.route('/onlinemasters/<level>/<step>', methods=['GET'])
 def onlinemasters(level, step):
-    g.level = level = int(level)
```
rix0rrr, line 528: I wonder if changing the route to `/onlinemasters/<int:level>/<int:step>` wouldn't do the exact same thing, but in a built-in way?

https://flask.palletsprojects.com/en/2.0.x/api/#url-route-registrations

fpereiro: @rix0rrr good idea. Just committed the change you suggested. The behavior is slightly different (we get a generic 404 since the route is outright not matched), but I think it should do.

Felienne (approved):

website/querylog.py
```diff
@@ -140,7 +140,9 @@ def finish_global_log_record(exc=None):
 
 def log_value(**kwargs):
     """Log values into the currently globally active Log Record."""
-    THREAD_LOCAL.current_log_record.set(**kwargs)
+    if hasattr(THREAD_LOCAL, 'current_log_record'):
+        # For some malformed URLs, the records are not initialized, so we check whether there's a current_log_record
```
rix0rrr, line 144: Nice find!

PR #530, opened 2021-06-24
Share code button

### LINKED ISSUE #481: Share button in main view
- After #480 is closed, add a "Share Button".
- Figure out all the corner cases that may be non-trivial of when to show it, when to not, and when to do a save+share.
- If necessary, clean up the data flow from the server to both the templates and js, which has been complicated since the adventures + autoload programs features, especially the parts implemented by yours truly.

fpereiro:
- Closes #481 

- After quite some tinkering and finding of corner cases, I found that the easiest approach to implement this is to make the "Share" button a "Save & Share" button. This takes care of cases where 1) there's no saved program, just a template text; 2) there's a saved program that doesn't belong to the user; 3) the user modifies the saved program just before sharing it.

I will do a bit more of testing tomorrow with fresh eyes, but I think this should be ready, barring some bug I haven't caught yet.


Felienne:
Looks good! But I guess we need localization of some sentences still (at least in Dutch?)

image

fpereiro:
@Felienne I gave it another spin and I think it's working well. It's even now with `main`. There's a few missing NL translations, but I added one for the green button.

Felienne (approved):

PR #534, opened 2021-06-27
feat: add a page for exclusively viewing of a shared program 

rix0rrr:
Add a page at `/hedy/<id>/view` which allows looking at a program
and running it without being in the full editor environment.


static/js/app.js
```diff
@@ -1,14 +1,25 @@
 (function() {
+  // A bunch of code expects a global "State" object. Set it here if not 
+  // set yet.
+  if (!window.State) {
```
fpereiro, line 4: We do add this in the script block of each template that needs it. But it might be a good idea to add it there, for those templates that don't receive window.State directly from the template.

static/js/app.js
```diff
@@ -217,7 +228,7 @@ window.saveit = function saveit(level, lang, name, code, cb) {
     if (! window.auth.profile) {
        if (! confirm (window.auth.texts.save_prompt)) return;
        // If there's an adventure_name, we store it together with the level, because it won't be available otherwise after signup/login.
-       if (window.State.adventure_name) level = [level, window.State.adventure_name];
+       if (window.State && window.State.adventure_name) level = [level, window.State.adventure_name];
```
fpereiro, line 231: Isn't `saveit` only called from the code page, where `window.State` is always defined?

static/js/app.js
```diff
@@ -284,6 +295,10 @@ window.saveit = function saveit(level, lang, name, code, cb) {
   }
 }
 
+function viewProgramLink(programId) {
```
fpereiro, line 298: Brilliant, thank you for this!

fpereiro (approved):
Looks good!

static/js/app.js
rix0rrr: 🤷‍♂️ I had this fail on me

PR #540, opened 2021-07-01
fix(logging): race condition in S3 logging

rix0rrr:
The 'boto3.client' method is not thread safe: https://github.com/boto/boto3/issues/1592

Put a lock around initializing an S3 client, so the two threads that
might do this in parallel don't stomp on each other.


fpereiro (approved):
Great catch!

PR #541, opened 2021-07-01
Ask for programming experience on signup & profile

### LINKED ISSUE #532: Add question of programming experience when creating/updating account
Allow the user to provide optional information about their programming expertise, with two inputs:

- Do you have any experience with programming? (yes/no/unanswered).
- If so, with what? 1) Scratch; 2) another block language; 3) Python; 4) another text language. All of these options can be selected independently of the others.

These options are shown on account creation and on account update. They are to be stored in a column, for each of the users who respond it.

fpereiro:
- Closes #532
- Includes visibility in the admin.


fpereiro:
Once we agree on the texts & the look & feel I will add the text entries to the translation files so that the tests can pass.

Felienne:
Seems to work ok! Just one thing, we now show the checkmarks always, I think it is more intuitive if we first show the radio buttons only, and when yes is selected, the check marks appear.

image

Felienne:
Also I think the text is nicer if it reads:

Which of these programming languages have you used before?

instead of 

Have you used any of these programming languages?

fpereiro:
@Felienne great suggestions; I totally forgot about hiding the languages checkboxes conditionally. I also changed the text and added placeholders for all other languages (except ES, which is translated already).

Felienne (approved):

PR #549, opened 2021-07-05
feat: Adds the option to read the output aloud in all supported languages

Felienne:
Based on user studies, this is a feature that kids love!

It is probably good if one of the web experts checks this cause I do not know what I am doing :D I am also not sure the tick mark is pretty enough, a bit of CSS might be good there but that is surely out of my reach!

### DIFF
--- a/coursedata/texts/el.yaml
+++ b/coursedata/texts/el.yaml
@@ -6,6 +6,7 @@ ui:
     contact: 'Επικοινωνία'
     video_title: 'Βίντεο'
     run_code_button: 'Εκτέλεσε κώδικα'
+    read_code_label: 'Εκφωνούν'
     save_code_button: 'Αποθήκευσε κώδικα'
     edit_code_button: "Edit code"
     share_code_button: "Save & share code"
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -1,5 +1,5 @@
 (function() {
-  // A bunch of code expects a global "State" object. Set it here if not 
+  // A bunch of code expects a global "State" object. Set it here if not
   // set yet.
   if (!window.State) {
     window.State = {};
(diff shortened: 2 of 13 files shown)


static/js/app.js
```diff
@@ -401,6 +401,46 @@ window.onerror = function reportClientException(message, source, line_number, co
...
+    case 'es':
+        return 'es-ES';
+    default:
+        return 'en-US';
+
+}
+}
+
+function speak(text) {
+    speak_enabled = $('#speak').is(":checked")
```
rix0rrr, line 433:
```suggestion
    var speak_enabled = $('#speak').is(":checked")
```

static/js/app.js
```diff
@@ -401,6 +401,46 @@ window.onerror = function reportClientException(message, source, line_number, co
...
+    console.log('speak?', speak_enabled);
```
rix0rrr, line 434: Debugging? :)

static/js/app.js
```diff
@@ -401,6 +401,46 @@ window.onerror = function reportClientException(message, source, line_number, co
...
+    if (speak_enabled){
+        let utterance = new SpeechSynthesisUtterance(text);
+        lang = window.State.lang;
```
rix0rrr, line 437:
```suggestion
        var lang = window.State.lang;
```

static/js/app.js
```diff
@@ -401,6 +401,46 @@ window.onerror = function reportClientException(message, source, line_number, co
...
+        utterance.lang = convert_lang_to_ISO(lang);
```
rix0rrr, line 438: Some testing finds that this doesn't work on Safari, unfortunately.

We're supposed to use the `utterance.voice = ` field there, apparently.

How about:

```ts
function findVoice(langWithCountry) {
  // If the feature doesn't exist in the browser, return null
  if (!window.speechSynthesis) { return undefined; }
  const ourVoices = window.speechSynthesis.getVoices().filter(voice => voice.lang === langWithCountry);
  
  // Give preference to local voices to not send Chrome users a big data bill (otherwise
  // Google voices will be the first in the list)
  const localVoices = ourVoices.filter(voice => voice.localService);
  if (localVoices.length > 0) { return localVoices[0]; }

  // Otherwise any voice we can get our hands on
  return ourVoices.length > 0 ? ourVoices[0] : undefined;
}
```

static/js/app.js
```diff
@@ -401,6 +401,46 @@ window.onerror = function reportClientException(message, source, line_number, co
...
+    default:
+        return 'en-US';
+
+}
+}
+
+function speak(text) {
+    speak_enabled = $('#speak').is(":checked")
+    console.log('speak?', speak_enabled);
+    if (speak_enabled){
```
rix0rrr, line 435:
```suggestion
    if (speak_enabled && window.SpeechSynthesisUtterance) {
```

Otherwise this will crash with a `ReferenceError` if run on a browser that doesn't have the speech API.

coursedata/texts/el.yaml
```diff
@@ -6,6 +6,7 @@ ui:
     contact: 'Επικοινωνία'
     video_title: 'Βίντεο'
     run_code_button: 'Εκτέλεσε κώδικα'
+    read_code_label: 'Εκφωνούν'
```
rix0rrr, line 9: ❤️ voor de zaak ;)

rix0rrr:
```
print ha ha ha ha 
print (a dri jaan)
```

rix0rrr (approved):

PR #551, opened 2021-07-06
feat: show a language dropdown with all available voices

rix0rrr:
Avoids having to map from a language to a country code,
and gives the kids choice.

image

### DIFF
--- a/static/js/tabs.js
+++ b/static/js/tabs.js
@@ -79,5 +79,5 @@ $(function() {
   // If we're opening an adventure from the beginning (either through a link to /hedy/adventures or through a saved program for an adventure), we click on the relevant tab.
   // We click on `level` to load a program associated with level, if any.
   const adventureName = window.State && window.State.adventure_name;
-  $('[data-tab="t-' + (adventureName || 'level') + '"]').click ();
+  $('*[data-tab="t-' + (adventureName || 'level') + '"]').click ();
 });
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -595,15 +595,18 @@ function buildUrl(url, params) {
 
       clearInterval(timer);
     }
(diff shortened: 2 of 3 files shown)


static/js/tabs.js
```diff
@@ -79,5 +79,5 @@ $(function() {
   // If we're opening an adventure from the beginning (either through a link to /hedy/adventures or through a saved program for an adventure), we click on the relevant tab.
   // We click on `level` to load a program associated with level, if any.
   const adventureName = window.State && window.State.adventure_name;
-  $('[data-tab="t-' + (adventureName || 'level') + '"]').click ();
+  $('*[data-tab="t-' + (adventureName || 'level') + '"]').click ();
```
fpereiro, line 82: Question: why do we need a wildcard here?

rix0rrr: Not sure. I was getting weird errors and adding this seemed to fix it. It might be a red herring.

fpereiro (approved):
I don't seem to have any speech synthesis currently installed in my OS, so I can't test. But the code looks good!

rix0rrr:
I might still have made a mistake in there. If you go the JavaScript console in a page, and then run:

```
window.speechSynthesis.getVoices()
```

And then again after a couple of seconds:

```
window.speechSynthesis.getVoices()
```

Does it show anything or not?

What is your OS and browser?

rix0rrr:
For reference, for me it usually looks like this:

image


(Yes, programming for the web is 🤬)

fpereiro:
window.speechSynthesis.getVoices() shows me an empty array, both after loading and if I wait a number of seconds. That's how I came to the conclusion that I have no support for that.

I'm running Chromium 91.0 on Xubuntu 20.04, that's why my hopes are quite tempered...

PR #552, opened 2021-07-07
fix: user_ms and sys_ms are logged as lists

rix0rrr:
In the query log, the resource consumption entries are logged as
`user_ms: [15]` instead of `user_ms: 15` because of a trailing `,`
which makes the values tuples.


fpereiro (approved):
Good catch! I didn't even know this was syntactically possible.

PR #553, opened 2021-07-07
fix: script errors aren't reported properly in prod

rix0rrr:
Because `app.js` is loaded from the CDN in prod, if there's an error in
the script the browser hides the details of the error for security
reasons (different domain = you don't get to see the details).

Add a cross-origin header so the browser will treat the external
resources as safe.

### DIFF
--- a/app.py
+++ b/app.py
@@ -357,6 +357,7 @@ def report_client_exception():
         is_test=1 if os.getenv ('IS_TEST_ENV') else None
     )
 
+    # Return a 500 so the HTTP status codes will stand out in our monitoring/logging
     return 'logged', 500
 
 @app.route('/version', methods=['GET'])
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -357,6 +357,7 @@ def report_client_exception():
         is_test=1 if os.getenv ('IS_TEST_ENV') else None
     )
 
+    # Return a 500 so the HTTP status codes will stand out in our monitoring/logging
```
fpereiro, line 360: I like this clarification, thank you for this.

fpereiro (approved):
Thank you for this, we definitely need it. I didn't realize we need it because for some reason I thought we were not CDN caching our own js, but of course that has been the case for a while already :|

PR #564, opened 2021-07-18
feat: atomic set updates for DynamoDB

rix0rrr:
DynamoDB supports atomic set updates (addition and removal) by passing
`{ "SS": ["x", "y", "z"] }` to `UpdateItem`.

Add support for adding and removing elements in string sets in both the
DynamoDB emulation layer and the real DynamoDB layer, by using
`DynamoAddToStringSet` and `DynamoRemoveFromStringSet` classes.

The set must initially have been stored as a `set(['a', 'b', 'c'])`, if
it has been stored as a `['a', 'b', 'c']` list the set manipulation
won't work anymore.

See unit tests for a description of the API.


fpereiro:
Hi @rix0rrr ! I reviewed the code and it definitely works. Thank you for this!

I merged the changes onto a branch with the teacher backend, to try it out: https://github.com/Felienne/hedy/tree/teacher-backend-dynamoset . Currently, this passes all tests.

I did some tweaking (or rather, crude hacking) on the Dynamo layer, for support of empty sets (see 1) below) and to deserialize the set into a list. The deserialization works well, but I don't think I put it in the proper place - it works, but perhaps it should go somewhere else.

I am having issues with two things:
1) Apparently there's no way to create empty sets (at the Dynamo level), so I'm using the workaround of initializing the set with an empty string and then removing it in the decoding function if it's present. I've marked the workarounds (there are two, one in the route layer when creating the item, one in the DB layer when reading it back).
2) Currently I'm accessing the Dynamo methods for adding/removing sets from the top database layer. I think this will break the E2E tests because those don't have Dynamo available. Should we add these methods into the top database layer instead?

rix0rrr:
> Apparently there's no way to create empty sets (at the Dynamo level).

Hmm, that might well be true. Bit annoying, I'll give you that.

Isn't a solution to just NOT set the `students` set at all when you are creating a class, and use `.get()` when you are reading the value?

```py
  Database.store_class({
            'id': uuid.uuid4().hex,
            'name': body ['name'],
            # ...no 'students' here
        }
```

And later:

```py
# Use .get to substitute an empty set if the 'students' key is not set
for student_name in class.get('students', set()):
  print(student_name)
```

?

rix0rrr:
> Currently I'm accessing the Dynamo methods for adding/removing sets from the top database layer. I think this will break the E2E tests because those don't have Dynamo available. Should we add these methods into the top database layer instead?

I'm not sure I understand. Can you point me to the code you're talking about?

fpereiro:
> > Apparently there's no way to create empty sets (at the Dynamo level).
> 
> Hmm, that might well be true. Bit annoying, I'll give you that.
> 
> Isn't a solution to just NOT set the `students` set at all when you are creating a class, and use `.get()` when you are reading the value?
> 
> ```python
>   Database.store_class({
>             'id': uuid.uuid4().hex,
>             'name': body ['name'],
>             # ...no 'students' here
>         }
> ```
> 
> And later:
> 
> ```python
> # Use .get to substitute an empty set if the 'students' key is not set
> for student_name in klass.get('students', set()):
>   print(student_name)
> ```
> 
> ?
> 
> If you want to hide this concern, that's exactly what the `Database` class could be for:
> 
> ```python
> Database.store_class({
>             'id': uuid.uuid4().hex,
>             'name': body ['name'],
>             'students': set()
>         }
> 
> class Database:
>   def store_class(self, klass):
>     if klass['students'] == set():
>       del klass['students']
>     CLASSES.put(klass)
> 
>   def load_class(...):
>      klass = CLASSES.get(...)
>      if 'students' not in klass:
>        klass['sudents'] = set()
>     return klass
> ```

What a great idea, I hadn't thought of that! I will try that.

rix0rrr:
According to this news item DynamoDB *does* support empty sets:

https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-dynamodb-now-supports-empty-values-for-non-key-string-and-binary-attributes-in-dynamodb-tables/

fpereiro:
> > Currently I'm accessing the Dynamo methods for adding/removing sets from the top database layer. I think this will break the E2E tests because those don't have Dynamo available. Should we add these methods into the top database layer instead?
> 
> I'm not sure I understand. Can you point me to the code you're talking about?
> 
> EDIT: if it's this:
> 
> ```python
>     def add_student_to_class(self, class_id, student_id):
>         """Adds a student to a class."""
> 
>         CLASSES.update ({'id': class_id}, {'students': dynamo.DynamoAddToStringSet (student_id)})
>         USERS.update({'username': student_id}, {'classes': dynamo.DynamoAddToStringSet (class_id)})
> ```
> 
> That's exactly what I had in mind! `dynamo.DynamoAddToStringSet` will work on the offline database as well. (Should be a simple verification otherwise, if you don't set the `DYNAMODB_` environment variables it will automatically fall back to an offline database)
> 
> (EDIT2: or run `build-tools/validate-e2e`)

Fantastic! I had thought that `dynamo` was a direct reference to the Dynamo layer, not the general database layer that can default to in-memory. Then we're all good!

I will try the changes you suggest in the other comment and let you know how it goes.

fpereiro:
> According to this news item DynamoDB _does_ support empty sets:
> 
> https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-dynamodb-now-supports-empty-values-for-non-key-string-and-binary-attributes-in-dynamodb-tables/
> 
> What kind of error are you seeing?

I just tried it again. Here's what I get: `An error occurred (ValidationException) when calling the PutItem operation: One or more parameter values were invalid: An number set  may not be empty`

I get the same error with both `'students': set ([])` and 'students': set ()

fpereiro:
@rix0rrr success! With a couple of workarounds, we can avoid empty sets altogether.

All I would need for the teacher backend to work with this is to add two lines to `dynamo.py` at line 441, to deserialize the set into a JSON array:
```
    elif isinstance(obj, set):
        return list(obj)
```

fpereiro (approved):
As soon as we add the little bit of code to deserialize sets into JSON arrays, we should be good to go!

rix0rrr:
I'm not sure that replacing sets with lists in that way is safe, or at least I don't understand what difference it makes.

* If we keep the representation in Python as a `list`, but leave it in DynamoDB as a `set`, that doesn't fundamentally change the proposition: empty sets in DynamoDB will still not be supported.
* If this changes the representation in DynamoDB to be a `list` instead of a `set`, I'm not sure the `AddToStringSet` operation is going to work on the combination of `list` and `set`, and I also think that `list` has dangerous semantics: a list can contain duplicates, so we could potentially end up with a list like `[johnny, elly, elly, johnny, samantha, elly]`, which doesn't make a lot of sense.

All in all, I'd feel more for just avoiding storing the empty set if Dynamo doesn't support it, and hiding the distinction in the `Database` layer (or potentially even the `AwsDynamoStorage` layer).

fpereiro:
> > I'm not sure that replacing sets with lists in that way is safe, or at least I don't understand what difference it makes.
> > 
> > * If we keep the representation in Python as a `list`, but leave it in DynamoDB as a `set`, that doesn't fundamentally change the proposition: empty sets in DynamoDB will still not be supported.
> > * If this changes the representation in DynamoDB to be a `list` instead of a `set`, I'm not sure the `AddToStringSet` operation is going to still work (can it work on the combination of `list` and `set`?), and I also think that `list` has dangerous semantics: a list can contain duplicates, so we could potentially end up with a list like `[johnny, elly, elly, johnny, samantha, elly]`, which doesn't make a lot of sense.
> > 
> > All in all, I'd feel more for just avoiding storing the empty set if Dynamo doesn't support it, and hiding the distinction in the `Database` layer (or potentially even the `AwsDynamoStorage` layer).
> 
> I am not really happy with that mapping either, but we need to be able to pull a set out as JSON, and the closest candidate would be an array. An object, having both keys and values, seems to be even more removed. The change I propose is strictly for deserialization purposes (without that, I was getting an error about sets not deserializable to JSON).

Never mind, I understood you now. The alternative would be to deserialize it at the route level, and there specifically choose to read the set as a list. I think I'll go ahead with that. Sorry for the misunderstanding.

Issue #572, opened 2021-07-19
is ask is not highlighted properly in level 3


Felienne:
![image](https://user-images.githubusercontent.com/1003685/126149824-74abccd8-9c0a-4d05-8c0c-dafc813cde03.png)

In some other cases, it is:

![image](https://user-images.githubusercontent.com/1003685/126150128-13fe8f12-450b-4e07-a762-c4a783fcbc1a.png)

Tazaria:
This always happens when a print statement with only a string is followed by another statement. When the print statement has a variable, it does highlight correctly
![Highlight error](https://user-images.githubusercontent.com/47422769/126153992-34038d8c-4f5b-4457-b843-9d2786ba2190.png)

Tazaria:
However, adding a space at the very end of the line somehow fixes the highlighting
![Highlight fix](https://user-images.githubusercontent.com/47422769/126154249-9135ec14-4048-41eb-8025-953a3bb137ca.png)

Felienne:
Closed by #579

Issue #571, opened 2021-07-19
at random should be highlighted in level 3 and up


Felienne:
Level 1 and 2 werk fine, but after that `at random` is no longer highlighted

![image](https://user-images.githubusercontent.com/1003685/126149428-a611dd46-3b4d-4415-b031-0b003e0d86f8.png)

Felienne:
Closed for other levels by #584

PR #579, opened 2021-07-19
fix: level 3 syntax highlighting

rix0rrr:
The regexes in the tokenizer will only ever see one line at a time,
without the closing `\n`. As soon as all characters in a line have
been consumed, the tokenizer moves to the next line, remembering
the current state.

This means that consuming 0 characters at the end of line to move
back to the start state is not possible.

Instead, we take advantage of the ability of the tokenizer to capture
multiple groups (characters between `(...)` in the regex) and assign all
of them token classes at once by specifying them in arrays. This allows
us to finish regexes with `(.*)` to consume anything till the end of the
line (if present), so we know that we can move back to the `start` state
after having done so.

### DIFF
--- a/static/js/syntaxModesRules.js
+++ b/static/js/syntaxModesRules.js
@@ -184,52 +184,39 @@ define('ace/mode/level3_highlight_rules', [], function(require, exports, module)
   var ExampleHighlightRules = function() {
 
     this.$rules = {
-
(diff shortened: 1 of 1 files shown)


Felienne:
Close #572 and #571

Felienne (approved):

Issue #587, opened 2021-07-20
Highlight issue with multiple strings in print statement


Tazaria:
When there are multiple strings in a print statement from level 3 and up, only the first string is highlighted. Also, any key words in later strings are highlighted as key words while being interpreted as strings. They should just be highlighted as strings and not as key words.
![Highlight is](https://user-images.githubusercontent.com/47422769/126326874-5e59791a-eebf-42ef-b5a3-531a7bfe8e36.png)

Felienne:
Closed by #617

PR #589, opened 2021-07-20
fix: more fixes to lvl3 syntax highlighting

rix0rrr:
- Also highlight quotes after `is ask` and `is`.
- Have two clauses for every expression highlight: one for a highlight, staying
  in `print option`, one for going back to the `start` state if it's the
  last expression in a line.


Felienne (approved):

Felienne:
Closes #587 for level 3 (not for other levels yet)

PR #598, opened 2021-07-21
added test and updated grammar and error message for en and nl

Felienne:


LauraTSD (commented):
Nice addition to catch the error!

PR #601, opened 2021-07-21
loop var should be added to list of vars for proper escaping

### LINKED ISSUE #599: Loop variable in level 8 doesn't work in if statements
See code. The if statement should be true when the counter is at 2, but it still goes to the else statement. It does print the right number however
Counter error

Felienne:
Fixes #599


Tazaria (approved):
This fixes the problem. Loop var now works in if statements.

PR #609, opened 2021-07-23
adds end-blocks to level 7

### LINKED ISSUE #396: Issue with if/else within repeat block in level 7
In the latest master (https://github.com/Felienne/hedy/commit/175ac3156c0fa3f5fbdf334fe26dea9f35ed6f80), the following program fails at level 7:

```
repeat 5 times
    getallen is 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
    getal1 is getallen at random
    getal2 is getallen at random
    goedeantwoord is getal1 * getal2
    print 'Hoeveel is ' getal1 ' keer ' getal2 '?'
    antwoord is ask Vul hier het antwoord in:
    if antwoord is goedeantwoord
        print 'Goedzo'
    else
        print 'Helaas'
```

The error is reported to be in the `else` line, but nothing seems to be wrong there.

If the repeat is taken out and the rest of the program is dedented, it works well. Also, if the else branch of the above program is removed (while keeping the repeat), the program works also well.

This bug was found by Marleen.

Felienne:
Closes #396


Tazaria (approved):
The check that doesn't pass is from a sublevel that has no impact on Hedy itself. This will also be fixed in pull request #600. Apart from that the issue is fixed.

PR #614, opened 2021-07-23
restructured lookup table code

Felienne:
caused it kept giving bugs :)


LauraTSD (commented):
Looks cleaned up. What happened to the def while_loop and the def smaller, bigger, not_equal, equality_check etc? They look removed, are they caught by another function(s)?

Felienne:
Good eye Laura! They will also be caught by the default function. If you
don't define a rule, it will be picked up by __default__ (I will make that
more clear in the comments)

PR #616, opened 2021-07-29
Refactor courses & assignments

Felienne:
- Because we now have adventures to manage courses, old code loading assignments can be removed, in the code and in the templates.

- The Space EU course no longer needs maintenance (now simple redirects to the Hedy course)

- Online masters now uses new "custom" code, set a course to custom and provide a list of adventures and just those adventures will be shown.


app.py
```diff
@@ -693,17 +683,16 @@ def space_eu(level, step):
...
     g.prefix = '/space_eu'
 
-    adventure_assignments = load_adventure_assignments_per_level(g.lang, level)
+    adventures = load_adventures_per_level(g.lang, level)
 
-    return hedyweb.render_assignment_editor(
+    return hedyweb.render_code_editor_with_tabs(
         request=request,
-        course=SPACE_EU_COURSE[g.lang],
+        course=HEDY_COURSE[g.lang],
```
rix0rrr, line 690: What is the goal for the `space_eu` course? The PR description said it would redirect to the regular Hedy course. If you want to do that,  you could replace this whole block with something like:

```py
    return redirect(url_for('index', level=level, step=step), code=302)
```

It would cut down on unnecessary code if we did that.

coursedata/course/online_masters/nl.yaml
```diff
@@ -1,150 +1,12 @@
 ---
 course:
-  -
-    level: "1"
-    assignments:
-      - step: "1"
-        prompt: |
-          <a href="https://youtu.be/LKnQvUvCRqM" target="_blank">Klik hier om de uitlegvideo te bekijken.</a>
```
rix0rrr, line 8: Are you okay losing the YouTube video links? I don't think those are in the regular adventures, are they?

templates/incl-editor-and-output.html
```diff
@@ -1,11 +1,11 @@
 <div class="grid grid-cols-1 lg:grid-cols-2 gap-4 mt-5">
     <div>
       <div class="w-full h-64 flex flex-col">
-        <pre id="editor"
-          style="background: #272822; color: white"
+        <div id="editor"
```
rix0rrr, line 4: Why this change?

rix0rrr (approved):

coursedata/course/online_masters/nl.yaml
Felienne: Yeah we can live without them (they are really old by now anyway)

PR #626, opened 2021-08-12
Unified modal

### LINKED ISSUE #611: Create unified snackbar (for messages/errors) and replace all alerts with snackbar calls
Snackbar should appear at the bottom of page for N seconds, then go away.

Snackbar should be included in a top-level template, positioned absolutely, so it's available on every view without explicit including.

Explore the possibility of also using it for confirms (yes/no) and as prompts (text input).

fpereiro:
- Closes #611 
- Provides three methods within `window.modal`: `alert`, `confirm` and `prompt`. They replace their native counterparts with a nicer UI.
- Both the HTML and the js are loaded as part of `layout.html` to ensure it is everywhere available.
- We make use of the fact that there can only be one alert/confirm/prompt available at the same time in order to set a global callback to continue executing code, in the case of `confirm` and `prompt`.
- Since we load jQuery in `layout.html`, we have removed references to it from other templates.


fpereiro:
@Felienne it should now work when switching levels as well. We can extend this approach wherever necessary to other frequently used links.

Felienne:
Hi @fpereiro 

I did a few hard refreshes but I am still not seeing the modal on level switch:

image

Felienne:
> Hi @fpereiro
> 
> I did a few hard refreshes but I am still not seeing the modal on level switch:
> 
> image

Ow this is interesting.... I am seeing it for English but not for Dutch for the level switch. I am seeing it for adventure switch in Nl though, so it is not the the whole thing is not working in Nl on account of translations or something. Could you test it for Dutch?

fpereiro:
@Felienne it should work now :)

Felienne (approved):
Works now in Dutch too! :D

Issue #623, opened 2021-08-10
Spanish translations inconsistency


Sergih28:
**There isn't a clear way to address to the user in the [Spanish translation](https://github.com/Felienne/hedy/blob/main/coursedata/level-defaults/es.yaml).**

Sometimes it addresses to the user in second person (which is the **normal way** in Spanish), and sometimes does it in third person (which is the **formal way**).

Here are some examples:

- puede**s**. Using _second person_. Line 3:
```
¡Bienvenido a Hedy! En el Nivel 1 puedes usar los comandos `print`,` ask` y `echo`.
```

- **su**. Using _third person_. Line 5:
```
En el lado izquierdo de su pantalla están todos los comandos del Nivel 1.
```

- Presion**e**. Using _third person_. Line 7:
```
¡Presione el botón azul y el código será escrito por ti!
```

- Prueb**a**. Using _second person_. Line 9:
```
Prueba el código tu mismo con el botón verde 'Ejecutar el código' debajo del campo de programación de la izquierda.
```

---

Both are correct, but think **we should stick to one**.
I would suggest to go with the _second person_ form, because it makes sense for what Hedy is.

fpereiro:
Hi @Sergih28 ! That is indeed right - we should use the second person form to be consistent with the style of Hedy. I'll do that on all upcoming translations, and at some later point review the existing ones.

Sergih28:
I can help reviewing the current ones if you want @fpereiro

fpereiro:
> I can help reviewing the current ones if you want @fpereiro

That would be fantastic! Thank you!

fpereiro:
Marking this as closed after merging #629 . @Sergih28 feel free to reopen it if you think something's missing!

PR #629, opened 2021-08-14
Fix Spanish translations

Sergih28:
Related issue: #623

Changing to use the second person in the Spanish translation, in sections 1 to 7.

I did some changes here and there to improve some expressions.

Progress:
- [x] Section 1
- [x] Section 2
- [x] Section 3
- [x] Section 4
- [x] Section 5
- [x] Section 6
- [x] Section 7

Section 4.1 is missing


coursedata/level-defaults/es.yaml
```diff
@@ -51,13 +51,13 @@
...
             print Bienvenido nombre
     -   name: "ask"
-        explanation: "Preguntar algo con `ask`."
-        example: "Ejemplo: color is `ask` ¿Cuál es tu color favorito?"
+        explanation: "Preguntar algo con `ask`. ¡Atención! Necesitas darle un nombre a la respuesta con `is`"
+        example: "Ejemplo: color is ask ¿Cuál es tu color favorito?"
         demo_code: |-
             color is ask ¿Cuál es tu color favorito?
-            print color es tú favorito
+            print El color es tú color favorito
```
fpereiro, line 58: Here I would leave the original line, because both instances of `color` will be replaced by the variable value (so this would read, for example, "el rojo es tu rojo favorito"

fpereiro:
@Sergih28 so far looking great! I left a minor observation there.

coursedata/level-defaults/es.yaml
Sergih28: I see, didn't think about that 🤔..

Reading "rojo es tu favorito" sounds incomplete to me.
Maybe we could rename the variable color so we can have the complete sentence?
For example:
El `nombre_color` es tu color favorito.

I know the variable name has to be as simple as possible, but if it clashes with an actual world you want to use, maybe it's not a bad idea to rename it to `nombre_color` or something similar?

fpereiro: Not a bad idea at all! Although I think @Felienne did that on purpose to create "tension" of sorts, that can be resolved in a later level where things can be quoted.

@Felienne should we aim to have a complete sentence here saying "Color X is your favorite color" or do we want to leave the tension of not being able to say "color" if we use it as a variable name?

Sergih28: @fpereiro After working on section 3, I see that was @Felienne's intention indeed.

So I'd leave it as `El color es tu favorito`

Sergih28:
Things to note about section 6, I changed `veces` for `multiplicado por` as we usually say it like that in Spain when we reference multiplications.

Sergih28:
@fpereiro Also section 4.1 is missing, so it should be done when doing the ones that aren't done yet, below 7.

fpereiro:
@Sergih28 great work, thank you for this!

I just reviewed and everything looks great. The only thing would be to change line 58 to its original "El color es tu favorito" after what we discussed.

Sergih28:
> @Sergih28 great work, thank you for this!
> 
> I just reviewed and everything looks great. The only thing would be to change line 58 to its original "El color es tu favorito" after what we discussed.

Yep, I agree, I'll do that change now.

There's a couple more things for you to check, I forgot to press the "review" button before 😬

Sergih28:
PS: I synchronised the main branch with this one, let me know if I did something wrong as I'm pretty unexperienced doing that 😅

coursedata/level-defaults/es.yaml
```diff
@@ -51,30 +51,30 @@
...
-        explanation: "Seleccionar una palabra aleatoria usando `at` y `random`."
+        explanation: "Seleccionar una palabra aleatoria de un grupo usando `at` y `random`."
         example: "Ejemplo: animales is perro, gato, canguro"
         demo_code: |-
             animales is perro, gato, canguro
             print animales at random
 3:
     intro_text: |
-        `ask` funciona igual en el nivel 3 que en el nivel 2, pero `print` es diferente ahora.
+        En el nivel 3 `ask` y `print` funcionan diferente.
```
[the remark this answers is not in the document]
fpereiro: Perfect, better to keep it consistent!

coursedata/level-defaults/es.yaml
```diff
@@ -125,79 +125,79 @@
...
-         En el nivel 6 se agregan nuevos códigos, ahora puede comenzar a calcular en su código.
+         En el nivel 6 se agregan nuevos códigos, ahora puedes comenzar a calcular en tu código.
 
-         El plus es fácil, lo escribes como con las matemáticas: `5 + 5` por ejemplo. El signo menos también funciona bien, que es `5 - 5`.
+    El signo más es fácil, lo escribes como las matemáticas: `5 + 5` por ejemplo. El signo menos también funciona: `5 - 5`.
 
-         Los tiempos son un poco diferentes, porque no hay tiempos en su teclado. ¡Solo busca, realmente no lo hay!
-         Es por eso que hacemos los tiempos con el asterisco arriba de 8: `5 * 5`. Lea eso como 5 veces 5, entonces lo recordará más fácilmente.
+         La multiplicación es un poco diferente, ya que no hay signo de multiplicación en el teclado. ¡Búscalo, verás que no lo hay!
+         Es por eso que usamos el asterisco para multiplicar: `5 * 5`. Lee esto como 5 veces 5, entonces lo recordarás con mayor facilidad.
```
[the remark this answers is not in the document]
fpereiro: I believe it refers to the key "8" in the keyboard. But in Spanish keyboards it is somewhere else, so better to skip this instruction altogether. Good catch.

fpereiro:
It's looking great and the merge didn't generate any incorrect changes. Should I go ahead and merge? I think we can open a different issue for the missing levels (including 4-1) -- what do you think?

coursedata/level-defaults/es.yaml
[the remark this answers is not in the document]
Sergih28: That's what I though after writing the message, ok, I'll eave it like this then 👍

Sergih28:
> @Sergih28 It's looking great and the merge didn't generate any incorrect changes. Should I go ahead and merge? I think we can open a different issue for the missing levels (including 4-1) -- what do you think?

Just updated again the branch with master, merge ready.

About level 4.1, I think so, let's open a different issue for it.

fpereiro (approved):
Great work, thank you for this!

Issue #630, opened 2021-08-14
"Run code" Button text


Sergih28:
**Run code** button text does not match with level 1 text.

![run_code](https://user-images.githubusercontent.com/16263046/129449194-0ff83ef9-68ff-4d20-96db-6299383e9446.png)

- It says "_Run the code_" above, referencing to the button.
- The button says "_Run code_"

Felienne:
Thanks for the fix @Sergih28!

PR #631, opened 2021-08-14
Fix text that points to 'Run code' button

Sergih28:
Related issue: #630 

Replaced "_Run the code_" for "_Run code_" in the text of section 1, as that's what the button text is.


fpereiro (approved):
Thank you for this!

PR #632, opened 2021-08-15
feat: adds turtle option from level1 to level 7

Felienne:
Includes tests and a new adventure for Nl & En


Felienne:
Calling all lovely translators (@irisdiakoumi, @citharoedus, @adbi13, @dpprdan, @Princekrampah, @Paul-Rutten @rickbeerendonk) to notify there is a new adventure to be translated 🙏 ❤️ )

PR #633, opened 2021-08-16
Warn instead of throw an error if there are missing keys in a translation

### LINKED ISSUE #555: Warn instead of throw an error if there are missing text entries on a language
When validating `coursedata/texts`, if there are missing languages in non-EN languages, warn rather than throw an error.

fpereiro:
- Closes #555 
- Warn instead of throw an error if there are missing translation fields
- Also warn if there are untranslated fields and if there are extraneous fields.


Felienne (changes requested):
Hi @fpereiro 

I tried this code locally, removing, as an example, the intro_text from the Czech translation, but I got an error still:

image

I thought I was supposed to get a warning instead, but maybe I am trying the wrong thing?

fpereiro:
@Felienne thank you for testing with the client! I had tested different cases but never opening the website, just from the command line and E2E tests...

You actually uncovered a larger issue: in many templates, we didn't use the method that defaulted to English translations if the required language translation was missing. I have fixed it throughout and it should now work.

**How to test:** The changes in this commit affect exclusively the files located in `coursedata/texts`. For example, a way to test it is to delete `nl.yaml` and use the website in Dutch. All the texts contained in `coursedata/texts/nl.yaml` should be replaced with their English counterparts throughout the translation. This can be done for any language except English, which is our reference translations.

Deleting lines from other translation files will not work, since we still enforce strict validation on those files. Perhaps we should expand the scope to include *all* YAML files?

Felienne:
> **How to test:** The changes in this commit affect exclusively the files located in `coursedata/texts`. For example, a way to test it is to delete `nl.yaml` and use the website in Dutch. All the texts contained in `coursedata/texts/nl.yaml` should be replaced with their English counterparts throughout the translation. This can be done for any language except English, which is our reference translations.

Defaulting now works perfectly! But... where am I supposed to see the warning? It is not logged in my Python console.

> Deleting lines from other translation files will not work, since we still enforce strict validation on those files. Perhaps we should expand the scope to include _all_ YAML files?

I am not sure, I think I first need to understand what the warning we have now looks like to see how it compares to the pure yaml validation.

Felienne:
Hahaha I accidentally left in a missing line in the Dutch yaml and then I got the error and I realized *that* will be a warning in the future

image

Felienne (approved):

PR #635, opened 2021-08-17
Change repeat email input & type to avoid autocomplete

### LINKED ISSUE #612: Change type of `repeat email` field in signup so users cannot rely on autocomplete

fpereiro:
- Closes #612 
- Changes the input for `repeat email` in the signup to avoid autocomplete.
- **How to test:** attempt to sign up and see that there are no email autocomplete suggestions on the `repeat email` field.


fpereiro:
@Felienne I think I need your approval again after merging with `main`.

Felienne:
For me this still autocompletes in Chrome on Mac but maybe that is the best we can do?

image

fpereiro:
I did a bit more of StackOverflow digging and it sounds like `autocomplete=off` *should* work. There's plenty of corner cases though, apparently. I would merge this because I still think it would prevent at least some autocompletes from triggering. The real proof will be to get our bounce rate originating from typos to be 0 :D

Felienne (approved):

PR #642, opened 2021-08-19
fix(highlight): remove invalid states

rix0rrr:
We recently had 500s, caused by this client error:

```
Uncaught Error: state doesn't exist
```

This is an error thrown by the Ace editor if the syntax
highlighting rules reference a `next` state that doesn't exist.

The state `expression` doesn't exist anymore, but was still referenced.
These days, we recognize most expressions simply from the `start` state.

Remove the invalid references to a state that doesn't exist anymore.


Felienne:
Thanks @rix0rrr! Do you have a path towards this issue so I can test it? Was there a certain level at which this was occurring or a certain program?

PR #643, opened 2021-08-19
chore: client exceptions don't contain source page

rix0rrr:
Very relevant information on how to produce a client error
is missing: the page the error occurred on.

Add that in.


fpereiro (approved):
Great catch!

PR #658, opened 2021-08-22
chore: remove 'grammars-Total' files

rix0rrr:
The files were ignored, but not actually removed yet, so actually
remove them.

Since git only tracks files and not directories, removing all files
would not create the directory on checkout, and writing the `Total`
files would fail because of a missing directory.

So put a single empty placeholder file (`.keepdir`) in the directory
so that git will create the dir.


Felienne (approved):
Not sure how all the gitignore magic works exactly but I works for me locally and tests pass so I trust you :)

PR #660, opened 2021-08-23
fix: ReferenceError: adventureName is not defined

rix0rrr:
Made a mistake an in a previous PR's refactor, left
an uninitialized variable in place.


Felienne (approved):
Not sure how to test this but it seems not to break anything further :)

PR #663, opened 2021-08-23
Fix for `query` method in Dynamo

### LINKED ISSUE #613: Deletion of account with programs throws a 500
When fixed, also add E2E tests to cover this scenario.

Stack trace:

```
  File "/home/hq/Desktop/hedy/website/auth.py", line 74, in inner
    return f (user, *args, **kws)
  File "/home/hq/Desktop/hedy/website/auth.py", line 278, in destroy
    DATABASE.forget_user(user ['username'])
  File "/home/hq/Desktop/hedy/website/database.py", line 91, in forget_user
    PROGRAMS.del_many({'username': username})
  File "/home/hq/Desktop/hedy/website/querylog.py", line 176, in wrapped
    return fn(*args, **kwargs)
  File "/home/hq/Desktop/hedy/website/dynamo.py", line 117, in del_many
    to_delete = self.get_many(key)
  File "/home/hq/Desktop/hedy/website/querylog.py", line 176, in wrapped
    return fn(*args, **kwargs)
  File "/home/hq/Desktop/hedy/website/dynamo.py", line 64, in get_many
    items = self.storage.query(lookup.table_name, key, reverse=reverse)
  File "/home/hq/Desktop/hedy/website/dynamo.py", line 175, in query
    result = self.db.query(
  File "/home/hq/Desktop/hedy/.env/lib/python3.8/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/hq/Desktop/hedy/.env/lib/python3.8/site-packages/botocore/client.py", line 648, in _make_api_call
    request_dict = self._convert_to_request_dict(
  File "/home/hq/Desktop/hedy/.env/lib/python3.8/site-packages/botocore/client.py", line 696, in _convert_to_request_dict
    request_dict = self._serializer.serialize_to_request(
  File "/home/hq/Desktop/hedy/.env/lib/python3.8/site-packages/botocore/validate.py", line 297, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in input: "Key", must be one of: TableName, IndexName, Select, AttributesToGet, Limit, ConsistentRead, KeyConditions, QueryFilter, ConditionalOperator, ScanIndexForward, ExclusiveStartKey, ReturnConsumedCapacity, ProjectionExpression, FilterExpression, KeyConditionExpression, ExpressionAttributeNames, ExpressionAttributeValues
```

fpereiro:
While investigating #613 , it emerged that the `del_many` would throw an error when running on Dynamo (but not when running with the local DB) because of a call to `db.query`. Said call used the `Key` parameter which doesn't seem to be accepted by Dynamo itself.

This fixes `db.query` and, as a consequence, also `del_many`. It is indirectly tested by deleting a user with a program in the E2E tests.

I'm not sure whether the approach I used here is correct. It does work, but it may restrict the generic-ness of the method. @rix0rrr please let me know what you think.

Closes #613

### DIFF
--- a/website/dynamo.py
+++ b/website/dynamo.py
@@ -172,9 +172,13 @@ class AwsDynamoStorage(TableStorage):
         return self._decode(result.get('Item', None))
 
     def query(self, table_name, key, reverse=False):
+        assert len(key.keys ()) == 1
+        key_field = list (key.keys ()) [0]
+        key_value = key [key_field]
         result = self.db.query(
             TableName=self.db_prefix + '-' + table_name,
-            Key = self._encode(key),
+            KeyConditionExpression = key_field + ' = :value',
+            ExpressionAttributeValues = {':value': {'S': key_value}},
             ScanIndexForward = not reverse)
         return list(map(self._decode, result.get('Items', [])))
 
--- a/website/dynamo.py
+++ b/website/dynamo.py
@@ -178,7 +178,7 @@ class AwsDynamoStorage(TableStorage):
         result = self.db.query(
             TableName=self.db_prefix + '-' + table_name,
             KeyConditionExpression = key_field + ' = :value',
-            ExpressionAttributeValues = {':value': {'S': key_value}},
+            ExpressionAttributeValues = {':value': self.SERIALIZER.serialize (key_value)},
             ScanIndexForward = not reverse)
         return list(map(self._decode, result.get('Items', [])))
 
--- a/e2e_tests.py
+++ b/e2e_tests.py
@@ -449,6 +449,7 @@ def suite (username):
         ['unshare program', 'post', '/programs/share', {}, lambda state: {'id': state ['program'] ['id'], 'public': False}, 200],
         ['delete program', 'get', lambda state: '/programs/delete/' + state ['program'] ['id'], {}, {}, 302],
         ['retrieve programs after deleting saved program', 'get', '/programs_list', {}, {}, 200, retrieveProgramsBefore],
+        ['create program before destroying account', 'post', '/programs', {}, {'code': 'print Hello world', 'name': 'Program 1', 'level': 1}, 200],
         ['destroy account', 'post', '/auth/destroy', {}, {}, 200],
         ['get programs without being logged in', 'get', '/programs', {}, {}, 302],
         # Auth: profile (programming experience)


website/dynamo.py
```diff
@@ -172,9 +172,13 @@ def get_item(self, table_name, key):
...
 
     def query(self, table_name, key, reverse=False):
+        assert len(key.keys ()) == 1
+        key_field = list (key.keys ()) [0]
+        key_value = key [key_field]
         result = self.db.query(
             TableName=self.db_prefix + '-' + table_name,
-            Key = self._encode(key),
+            KeyConditionExpression = key_field + ' = :value',
+            ExpressionAttributeValues = {':value': {'S': key_value}},
```
rix0rrr, line 181: The `{'S': key_value }` will only work if the key is a string value. Now, that will probably be true 90% of the time, but it's still making me a little uneasy. We should have a proper DDB encoding function floating around in this module somewhere.

Can you use that instead?

rix0rrr (commented):
Oh wow, that's a ginormous spaz on my part! Sorry about that! 

It should have been [`KeyConditions`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-KeyConditions) instead of `Key`...

Although I suppose what you did works as well and is non-legacy, so probably better :). Thanks for fixing this!

One small remark on encoding the query value, other than that LGTM!

website/dynamo.py
fpereiro: Definitely, great idea. Just replaced it with a call to `self.SERIALIZER.serialize`

fpereiro:
> Oh wow, that's a ginormous spaz on my part! Sorry about that!
> 
> It should have been [`KeyConditions`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-KeyConditions) instead of `Key`...
> 
> Although I suppose what you did works as well and is non-legacy, so probably better :). Thanks for fixing this!
> 
> One small remark on encoding the query value, other than that LGTM!

No problem! Since we were only using this on `del_many`, and we only used `del_many` when deleting multiple programs, we didn't have anything to test this code for us. Now we do :)

rix0rrr (approved):

PR #664, opened 2021-08-23
Fix buggy calculation that shows how old a program is

### LINKED ISSUE #569: Mistake in timestamps in Mijn Programma's
IMG_20210719_111724862

Error: timestamp for previous programs says 3 days ago while it was 1,5 hours ago
Error occurred for user: "Patrik de Panda" and "hoi"

fpereiro:
I really don't know what I was thinking when I wrote the original code that calculated how old a program is. Besides the disputable notion of jumping from `89 minutes ago` to `2 hours ago` and from `35 hours ago` to `2 days ago`, I was carrying out large rounding errors.

The error doesn't affect data on the database or any other view except the **My programs** view.

I believe this should fix #569 

**How to test**: load your own programs view and check that the dates are sensible. Locally, I've tested by changing the dates of a single program to all relevant intervals (< 60 minutes, < 24 hours, > 24 hours) and checked that the reported values are sane.


Felienne:
I must say the error was not showing up for my projects and I can't find the data of the kid that had the issue cause the admin is down, but the code looks ok and does not break anything. Just updated the Dutch wording, for some reason we say "3 uur geleden" and not "3 uren geleden" (but we do say "10 minuten or dagen geleden" 😖 )

Felienne (approved):

fpereiro:
Hi! The bug (I'm almost certain) was triggered by certain amounts of time, but not others. In any case, the new code is definitely better than the old one and this time I did think it through :D

Regarding "uur", we should change that term to "uren", since all the other terms are in plural

Felienne:
> Hi! The bug (I'm almost certain) was triggered by certain amounts of time, but not others. In any case, the new code is definitely better than the old one and this time I did think it through :D
> 
> Regarding "uur", we should change that term to "uren", since all the other terms are in plural

No then we'd get grammatically incorrect Dutch, which is not preferred :) The right Dutch sentence is "3 uur geleden"

fpereiro:
> > Hi! The bug (I'm almost certain) was triggered by certain amounts of time, but not others. In any case, the new code is definitely better than the old one and this time I did think it through :D
> > Regarding "uur", we should change that term to "uren", since all the other terms are in plural
> 
> No then we'd get grammatically incorrect Dutch, which is not preferred :) The right Dutch sentence is "3 uur geleden"

I didn't know that!

PR #669, opened 2021-08-24
Treats _ in code as special case and gives a warning

### LINKED ISSUE #591: "..." in programs should give a special error
So kids know this is a program to be completed.

PS Note to self: the fact that a lang has an explicit "missing" construct for didactic purposes is quite generalizable to other languages!

Felienne:
* Closes #591 
* Replaces ... in texts by _
* Gives a special error when code contains _


Felienne:
@fpereiro I tried to fix the Es adventures too changing dots into "guion bajo" but I might have butchered it, can you check?

PR #671, opened 2021-08-24
Fix 141

### LINKED ISSUE #141: Calc in level 6 is not allowed in print
See: 
image

Not sure if we should enrich the lang or give a better error message?

Felienne:
* Fixes #141 
* Also addresses a small textual issue in regenworm adventure in all langs


fpereiro (approved):
Looks good!

PR #693, opened 2021-08-29
Update grammar level 4 

### LINKED ISSUE #303: Issue with conditional assignment in levels 4 & 5
Starting at level 4, it is possible to assign a variable in a conditional way inside a flat `if`. Take the following program:

```
kleur is 0
if kleur is 0 kleur2 is a else kleur2 is b
print kleur2
```

In both level 4 and 5, it yields `a else kleur2 is b`. But interestingly enough, in level 6 it yields `a`, which to me looks like the expected result. Level 6 only adds calculation, but somehow the precedence rules are different.

I think this is not relevant to levels 7 and onwards, where blocks are present.

Felienne:
Closes #303


fpereiro (approved):
Works great!

PR #734, opened 2021-09-10
fix: load_yaml_pickle test fails if a pickled file is on disk

rix0rrr:
The pickle test that loads `hu.yaml` compares it to loading
`hu.yaml.pickle`; if there was a `.pickle` file remaining on
disk, it may have different contents than the `.yaml` file.

This will not happen on the server since it gets a clean redeploy
every time, but it may happen on dev machines.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -27,7 +27,7 @@ class TestUtils(unittest.TestCase):
     n = 50
 
     # Pick a file with unicode in it so we're sure it gets handled properly
-    file = 'coursedata/adventures/hu.yaml'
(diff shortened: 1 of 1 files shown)


fpereiro:
@rix0rrr thank you for the fix!

Felienne:
> @rix0rrr thank you for the fix!

Beware, this fix is not ready yet! @rix0rrr needs to dive into the issue a bit deeper.

Felienne (changes requested):
This is now in the exact state that is was for me too (locally) with the weird numeric postfix on the filename, no clue why that happens!

Felienne (approved):

PR #748, opened 2021-09-13
chore: log memory usage

rix0rrr:
Log memory usage so that we have something to go on when we go to
investigate memory failures.


Felienne (approved):

fpereiro:
Thank you for this @rix0rrr !

PR #757, opened 2021-09-14
Quiz disable incorrect answers

Fl0werth:
- When clicking an incorrect answer the UI shows a red lined option block with a red background, and the radio button is also disabled.

- Colored pill with 'correct' or 'incorrect' or 'Question x' in the middle of the menu bar


Felienne (changes requested):
Hi Fl0wert!

Looks really good like this. However, the responses in quiz.html are now only in Dutch (juist, onjuist, vraag 1 van de 3 goed, etc). Can we enable localization there?

If you do not know how to do that, we can ask @fpereiro do make that in order.

Felienne:
(also the conflicts need to be fixed before we can merge this)

fpereiro (changes requested):
It looks good! I would just request two small changes:
- Rename `questionFalse` to something like `questionStatus`, because (if I understood correctly), the variable represents whether the question is shown for the first time (start) or a second/third time (wrong answer given last time).
- Remove the print statements in app.py

Fl0werth:
> It looks good! I would just request two small changes:
> 
> * Rename `questionFalse` to something like `questionStatus`, because (if I understood correctly), the variable represents whether the question is shown for the first time (start) or a second/third time (wrong answer given last time).
> * Remove the print statements in app.py

I changed the variablenames to questionStatus and removed the print statements.

fpereiro:
Hi @LotusterHaar ! It looks great, thank you for the changes! One very small thing: could you remove the `quiz-answer-id` variable from `config.py`? I think it's not needed there. Other than that, it looks good!

fpereiro (approved):

Felienne:
> Hi @LotusterHaar ! It looks great, thank you for the changes! One very small thing: could you remove the `quiz-answer-id` variable from `config.py`? I think it's not needed there. Other than that, it looks good!

Hi @LotusterHaar Could you pick this up so we can close the PR? Otherwise we have to keep it up to date with main all the time.

Fl0werth:
> > Hi @LotusterHaar ! It looks great, thank you for the changes! One very small thing: could you remove the `quiz-answer-id` variable from `config.py`? I think it's not needed there. Other than that, it looks good!
> 
> Hi @LotusterHaar Could you pick this up so we can close the PR? Otherwise we have to keep it up to date with main all the time.

The config variable has been removed

Felienne (approved):

PR #789, opened 2021-09-17
Teacher emails & styling of emails

### LINKED ISSUE #738: Send email when teacher is made teacher
I can draft the email! Something with: 

* Welcome to Hedy
* What teachers accounts can do
* How to ask for help

fpereiro:
- Add styling to all emails.
- Add teachers email when a user is made teacher.
- Closes #738 

**How to test**:
- See email being sent to user made into teacher
- See all other transactional emails having style.

Suggestions on styling are welcome! The stylings have to be done inline in the elements because HTML email is extremely limiting in the amount of CSS it allows and how it allows it.


coursedata/texts/en.yaml
```diff
@@ -127,6 +127,17 @@ Auth:
     email_reset_password_body: |-
         Your Hedy password has been reset to a new one. If you did this, all is good.
         If you didn't change your password, please contact us immediately by replying to this email.
+    email_welcome_teacher_subject: "Your Hedy teacher account is ready"
```
Felienne, line 130: hi @fpereiro! 

If we put the text here, it will be translated with the other texts which I guess is not really needed of useful at this time (cause we don't have the notion of a language on an account, we would not know what lang to pick to the email) 

So I guess it is better to place it somewhere else (In a separate file) to not waste people's translation efforts?

What do you think?

Felienne (changes requested):

coursedata/texts/en.yaml
fpereiro: Interesting point! Come to think of it, this happens in *all* the emails we send.

I can indeed move this out of the way by hardcoding the text on the route that sends the email.

However, if you think it might be worth it to send emails in other languages, we could do the following: when doing any of the other transactions that involve an email (signup, changing your email address, change password, recover password, reset password), we send along the preferred language and store it in the user. This allows us to send translated versions of those emails if needed.

In the case of the "upgrade to teacher", we can read that field from the username and send the translated email, if available.

To summarize our options:
- Hardcode the text of the teacher on the route.
- Add preferred language on the transactions that require email, so we can leverage the translated emails. Then use this for the "welcome teacher" email as well.
- Do the above minus the internationalization of the "welcome teacher email".

Felienne: Ah interesting that this happens for all emails! I did not realize. Let's put them all in one file and leave it at that. 

I really appreciate your idea of storing the language of the user, and I think it is great, so let's make an issue for that, but for now we have a few other things to pick up first. For that issue I guess it would be more reliable for us to ask the language(s) at sign-up instead of inferring but we can worry about that when we build it.

fpereiro: I hadn't realized that either... it was only with your observation that I did!

Agreed to make this a separate feature.

By putting all the emails in one file, you mean leaving them where they are? Or a new file altogether? If it's the former, feel free to merge the PR since it should be ready. Otherwise, I'll make the switch to a new file in this PR.

Felienne: > I hadn't realized that either... it was only with your observation that I did!

:)

> By putting all the emails in one file, you mean leaving them where they are? Or a new file altogether? If it's the former, feel free to merge the PR since it should be ready. Otherwise, I'll make the switch to a new file in this PR.

No, we need them in a new file (emails.yaml, for example) so they are not translated (as explained above) Let's do that in this PR so it is all in order.

Felienne (approved):
Looks good from what I can see (can't test emails locally so I will merge and test on alpha)

PR #790, opened 2021-09-18
fix error message for code in wrong level

OnnoEbbens:
When I ran this code in level 2:

`ask Wat is je lievelingskleur?`

I got the error message:

`That was correct Hedy code, but not at the right level. You wrote code for level 2 at level 1.`

The message should be:

`That was correct Hedy code, but not at the right level. You wrote code for level 1 at level 2.`

The proposed change should fix that.


Felienne:
Wow nice catch @OnnoEbbens! Can't believe we had that backwards all that time!!

Issue #794, opened 2021-09-20
Error when using backslash `\`


OnnoEbbens:
When you use a backslash `\` in a string you get the error message:

`bad input`

It would be nice if we could avoid this error or otherwise raise a more descriptive error.

Felienne:
Hi @OnnoEbbens!

I was not able to reproduce the error for level 1 and 2 (pic 1) and 3 (pic 2), could you send me a screenshot + level?

![image](https://user-images.githubusercontent.com/1003685/134172606-b9f6be01-a3bf-448a-b96f-de8bd2816a15.png)

![image](https://user-images.githubusercontent.com/1003685/134172662-b365f870-5c04-422c-a116-a488b1a51d16.png)

OnnoEbbens:
Ah! Now I see that it only occurs at the end of the line (level 1+2) or end of the string (level 3+).

level 1+2:
![image](https://user-images.githubusercontent.com/10562903/134177436-8ef06b4c-f1ad-4aca-a201-9fb9237acd19.png)

level 3+:
![image](https://user-images.githubusercontent.com/10562903/134177310-252bc3e5-55fd-42ea-ab3f-30f392d642f0.png)

Felienne:
Ah great, that helps (this also explains the missing b) will pick this up today, it is a small fix!

Felienne:
Closed by #803

PR #803, opened 2021-09-22
Fixes backslash escape issue #794

Felienne:


Felienne:
I now do a bit of string magic, maybe we would prefer this type of stuff to live in the grammar? We need some sort of guiding principle maybe!

PR #805, opened 2021-09-22
Change [link] into <a>

Felienne:
Made a small fix, the rest is perfect like this!


fpereiro (approved):
Thanks for catching that!

PR #806, opened 2021-09-22
Quiz 2 by grid

Fl0werth:
The option blocks are  shown in alternating colors (green and blue) in a 2x2 grid.  The labels A, B, C, D indices are added to the option blocks. More padding is added.

PR #810, opened 2021-09-23
Quiz tab

### LINKED ISSUE #733: Move quiz to tab
Hi @fpereiro!

I would like a cosmetic change to how we do quizzes. Rather than having the button below, I would like to have the quiz button in a special adventure, called End/Eind, like this:

image

I see this as a normal adventure in the adventure-yaml, but with special command in the story text that means "place quiz button here", f.e. ###QUIZ###

Example:

```
story_text: |
        ## Closing remarks level 1
        Now that you have seen level 1, you know how to make small stories and program's. Tie to test your Hedy 
        knowledge!

        ###QUIZ###
```

If this quiz button is pressed, I want the quiz to be shown in the tab, like this:

image

---
Rationale: This structure helps kids and teachers see the quiz as a step in the teaching sequence which is better didactically. After the quiz, they do the "vooruitblik" and then move on.

fpereiro:
- Closes #733
- Moves quizzes to a special adventure tab called "End/Eind"

Implementation details:
- Because quizzes are implemented as standalone pages, we load them inside an iframe. The quiz templates have been modified, but not extensively, to hide the navigation bar (which is no longer necessary in the context of an iframe).
- Before this change, the button to go to a quiz was shown depending on the available quizzes in `coursedata/quiz`. After this change, this will happen depending on whether there's an `end` adventure for a given level. It is important to set and maintain a 1:1 relationship between these `end` entries in adventures and the quiz content.

**How to test**:
- Quizzes are enabled for levels 1 & 2. The texts are all in Dutch since they haven't been translated (and the quiz code doesn't support internationalization yet).
- By going to the "End/Eind" tab, the whole quiz flow (start quiz button, start quiz screen, questions, quiz end) should be available.
- When looking at a quiz, two sections should be gone: 1) the inputs & button for saving/sharing a program; 2) the editor and output boxes.


Felienne (changes requested):
Love it like this, a lot better!

Two small requests: 

1. Can we keep the light blue bar, but without the save button? I don't like the "flinching" that the site does when going to the quiz where things change places when the final tab is clicked.
2. Can we shrink the whole quiz? It now is a bit big for the small space and I am getting a scroll bar (see pic). I know #806 will help a bit there but the font also is big, can we make the font for "Level" and "Vraag" smaller and for the question itself too? For scale, can we make the question can be more or less the same size as the command text in the side bar. The overall goal is to have no need for a scroll bar.

Thanks!

image

fpereiro:
@Felienne #1 is fixed :)

Felienne:
> @Felienne #1 is fixed :)

Awesome thanks? If needed, we can totally merge this. But could we do a very simple shrink? Something like making the font smaller? Or is that a bigger operation that I think it is?

fpereiro:
> > @Felienne #1 is fixed :)
> 
> Awesome thanks? If needed, we can totally merge this. But could we do a very simple shrink? Something like making the font smaller? Or is that a bigger operation than I think it is? If we make it smaller it will be so much more usable until we can get Lotus to fix the other issues like the 4x4.

@Felienne After some digging, I found the `zoom` CSS property. Just pushed a fix. It's pretty much a hack and I'm not sure it will work well on other screen resolutions, but I think it's an improvement...

Felienne:
Cool let's merge this for now and deploy when the CSS is a bit better. One more question, how is the quiz placed in final tab? Is that based on it coming last or it being named end or...?

Felienne:
> * Quizzes are enabled for levels 1 & 2. The texts are all in Dutch since they haven't been translated (and the quiz code doesn't support internationalization yet).

I think it is only enabled for level 1, but that is ok!Let me know if I missed something and I would see level 2 too.

Felienne (approved):
Not entirely as we want it but ok to merge for now

PR #811, opened 2021-09-23
Do not reload translations every time in local development

fpereiro:
Lately I had been noticing that reloading the page during local development was taking ~5 seconds. After some inspection, I realized that in my (admittedly slow development machine), 3-4 seconds were spent in loading `client_messages.js`. This happens every time any page is reloaded.

The cause for this is that every time the file is requested, all the yamls are loaded from disk and re-processed. While this is very practical in itself, I feel it's not worth the ~3-4 seconds added to each refresh.

I am not 100% about this change, though, because it does require the developer to restart the development server to trigger the reloading of the app to update the translations - this is an inconvenience for someone already acquainted with Hedy development, and a gotcha for those just starting out.

### DIFF
--- a/hedyweb.py
+++ b/hedyweb.py
@@ -20,8 +20,7 @@ class Translations:
 
   @property
   def data(self):
-    # In debug mode, always reload all translations
-    if self._data is None or utils.is_debug_mode():
+    if self._data is None:
       translations = glob.glob('coursedata/texts/*.yaml')
       self._data = {}
       for trans_file in translations:


Felienne (commented):
Hi @fpereiro 

Maybe this is a really naive idea of me but couldn't you just set your own debugging mode locally to False and leave the code as is? or is there a reason that will not work?

fpereiro:
That should definitely work! I'll close the PR without mergin.

Felienne:
Can you add this info to the CONTRIBUTING.md file? I had a student complain about it too when running locally so it might not be clear :)

fpereiro:
@Felienne this required me to add another ENV variable. Here are the changes: https://github.com/Felienne/hedy/pull/817/files

Issue #780, opened 2021-09-16
Cap parsing at x lines


Felienne:
Some kids try to have fun by sending large programs and that upsets Heroku a lot. Can we give an error in js directly for programs over let's say 100 lines (and let's make that number an env var so we can change it easily).

Felienne:
Would be better to implement this server-side, so I will do it myself!

Felienne:
Closed by #821

PR #821, opened 2021-09-28
Cap lines at 100

Felienne:
Caps lines at 100 in the parser. We could make this progressive later on where we allow more and more lines in later levels, but this should help against server overloads.


Felienne:
Addresses #780

rix0rrr (approved):

Felienne:
@OnnoEbbens Good for you to know for in class teaching that we now cap programs at all levels at 100 lines. Do let me know if that leads to issues, we could allow for more lines in higher levels if needed

PR #822, opened 2021-09-28
Fix running on Windows

bjorn3:
`sys.getloadavg()` doesn't exist on Windows.

### DIFF
--- a/website/querylog.py
+++ b/website/querylog.py
@@ -22,17 +22,12 @@ class LogRecord:
             self.start_rusage = resource.getrusage(resource.RUSAGE_SELF)
         self.attributes = kwargs
         self.running_timers = set([])
-        if IS_WINDOWS:
-            self.set(
-                start_time=dtfmt(self.start_time),
-                pid=os.getpid(),
-                fault=0)
-        else:
-            self.set(
-                start_time=dtfmt(self.start_time),
-                pid=os.getpid(),
-                loadavg=os.getloadavg()[0],
-                fault=0)
+        loadavg = os.getloadavg()[0] if not IS_WINDOWS else None
+        self.set(
+            start_time=dtfmt(self.start_time),
+            pid=os.getpid(),
+            loadavg=loadavg,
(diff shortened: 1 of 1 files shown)


website/querylog.py
```diff
@@ -22,11 +22,17 @@ def __init__(self, **kwargs):
             self.start_rusage = resource.getrusage(resource.RUSAGE_SELF)
         self.attributes = kwargs
         self.running_timers = set([])
-        self.set(
-            start_time=dtfmt(self.start_time),
-            pid=os.getpid(),
-            loadavg=os.getloadavg()[0],
```
rix0rrr, line 28: Your PR description says it's about `getrusage`, but it seems to be this is the line you're interested in, right? Is it that `os.getloadavg()` doesn't exist on Windows?

If that is the case, in order to reduce duplication somewhat, can you change it to the following instead:

```py
loadavg = os.getloavg()[0] if not IS_WINDOWS else None
self.set(
    # ...
    loadavg=loadavg
    )
```

Other than that... whoops, sorry. I forgot to check the docs apparently 😅

rix0rrr (changes requested):

website/querylog.py
bjorn3: Indeed. `getloadavg()` is the problem.

bjorn3: Applied your suggestion. Should I squash?

rix0rrr: We'll squash when we merge.

rix0rrr (approved):

rix0rrr:
I think @Felienne needs to merge from `main` now...

Felienne:
> I think @Felienne needs to merge from `main` now...

I can't push on @bjorn3's branch so he will have to merge main on his fork so I can merge

bjorn3:
Rebased

Felienne:
Thanks @bjorn3!!! Congrats on your first Hedy contribution!!

PR #823, opened 2021-09-28
Adds slight pause after turtle forward

### LINKED ISSUE #765: Add a tiny sleep after turtle.forward
So it is more clear the turtle makes multiple steps when repeating forwards

thiagoarrais:
Fixes #765

I _think_ this is my first Python contribution ever (if not, the first one was so long ago as to be obviated from my memory). So please do not refrain to let me know where the code can be made more idiomatic.

I've tried to isolate the code that adds the sleep. I've included the commit here for appreciation, but I'm not sure this really helps with maintainability. We can always revert the commit if we want to.

### DIFF
--- a/app.py
+++ b/app.py
@@ -289,7 +289,7 @@ def parse():
 
             response['has_turtle'] = has_turtle
             if has_turtle:
-                response["Code"] = "# coding=utf8\nimport random\nimport turtle\nt = turtle.Turtle()\nt.forward(0)\n" + python_code
+                response["Code"] = "# coding=utf8\nimport random\nimport time\nimport turtle\nt = turtle.Turtle()\nt.forward(0)\n" + python_code
             else:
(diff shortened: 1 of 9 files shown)


Felienne (approved):
Works well, and thanks for the nice `forward` refactoring too!!

Felienne:
> Fixes #765
> 
> I _think_ this is my first Python contribution ever (if not, the first one was so long ago as to be obviated from my memory). So please do not refrain to let me know where the code can be made more idiomatic.

Awesome, thanks so much for your contribution!

> I've tried to isolate the code that adds the sleep. I've included the commit here for appreciation, but I'm not sure this really helps with maintainability. We can always revert the commit if we want to.

We squash merge all pr's so it is great like it is! In reviewing, separate commits can be really useful!

Felienne:
@thiagoarrais We just deployed so your contribution should be visible on hedycode.com now.

PR #834, opened 2021-09-30
Move all string literals in the parsers to level1.lark

bjorn3:
This reduces the risk of a typo in the string contents, makes it easier to rename the symbols (eg translation) and centralizes token definitions a bit.

I also made the grammar more lenient with regards to multiple spaces where previously at most a single space was expected. This eliminates the existence of an error that is not existing in conventional programming languages either, thus avoiding the need for any program repair for them.

### DIFF
--- a/grammars/level1.lark
+++ b/grammars/level1.lark
@@ -27,6 +27,7 @@ _TURN: "turn"
 
 //level 2
 _IS: "is"
+_AT: "at"
 random : "random" //random needs to appear in the tree for further processing so does not start with _ or is uppercase
 
 //level 4
@@ -34,7 +35,6 @@ _IN: "in"
 _IF: "if"
 _ELSE: "else"
 _AND: "and"
-_AT: "at"
 
 //level 5
 _REPEAT: "repeat"
--- a/grammars/level1.lark
+++ b/grammars/level1.lark
@@ -95,3 +95,6 @@ NAME: ID_START ID_CONTINUE*
 ID_START: /[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}_]+/
 ID_CONTINUE: ID_START | /[\p{Mn}\p{Mc}\p{Nd}\p{Pc}·]+/
 
+// Internal symbol added by the preprocess_blocks function to indicate the end of blocks
+// FIXME ensure that the user can't write this token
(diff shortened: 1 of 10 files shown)


Felienne (changes requested):
Nice work @bjorn3, this will also help the translation team!

Could you add a few tests that describe the error we now no longer have? That serves two purposes:

* Makes sure that future changes do not accidentally revert this
* Helps to maintain a complete description of the language, while we do not yet have a formal semantics

Felienne:
And to be even more nit-picky, this PR in fact does two different things, so it would have been even nicer in 2 PRs. This is totally fine for now but going forward it is best to try to have single focused PRs.

Felienne:
Also also :) The additional leniency in whitepacing might make the parser slower, it would also be good to gain a bit of an understanding on the performance of this change (you could use the batch runner for that, it also times runs)

bjorn3:
I have removed the grammar change from this PR. I will open another PR with the grammar change and tests.

bjorn3:
> you could use the batch runner for that, it also times runs

It crashes for me:

```
$ python runhedy.py batchhedy/input/hallo.hedy
  File "runhedy.py", line 61, in <module>
    main()
  File "runhedy.py", line 52, in main
    hedy.execute(program, level)
  File "~/hedy/hedy.py", line 1362, in execute
    exec(python)
TypeError: exec() arg 1 must be a string, bytes or code object
```

What would be some good benchmarks? Using the examples from the various levels?

Felienne:
> > you could use the batch runner for that, it also times runs
> 
> It crashes for me:

Have you made a folder with *.hedy files and the output files? (First time you want to not pass check because it checks a run against info stored from a first run)

> What would be some good benchmarks? Using the examples from the various levels?

Examples from levels for sure, and maybe also a few larger, handmade programs?

Felienne (approved):
Nice grammar change, thanks!

bjorn3:
> Have you made a folder with *.hedy files and the output files? (First time you want to not pass check because it checks a run against info stored from a first run)

I used the pre-existing https://github.com/Felienne/hedy/blob/main/batchhedy/input/hallo.hedy

Neither `python runhedy.py -h` nor the contributing file mentions anything about output files. It also doesn't write any file to the batchhedy/input directory.

PR #842, opened 2021-10-01
Allow multiple spaces inside lists

bjorn3:
Split out of https://github.com/Felienne/hedy/pull/834.

Level 4 up to level 12 already allowed this, but from level 12 onwards only a single space was allowed previously.


Felienne (approved):
Thanks for the addition of tests, really helpful!

bjorn3:
Rebased

Felienne (approved):

PR #844, opened 2021-10-01
Fix runhedy.py

bjorn3:
Fixes the issue found at https://github.com/Felienne/hedy/pull/834#issuecomment-932143066


Felienne (approved):
Good catch, guess we have not updated the batch runner in a while

PR #847, opened 2021-10-02
Fix a mixup between original_level and working_level

bjorn3:
`original_level` was used to indicate the level for which the code does work, while `working_level` was used for the current level.

Maybe `original_level` should be renamed to `current_level` or something like that to reduce confusion?


Felienne:
Ah I think this is relic the switch of #790!

Felienne (approved):

PR #848, opened 2021-10-02
Unwrap VisitError to allow regular exception raising inside transformers

bjorn3:
This allow the removal of a hack to show a nice error message in case of undefined variables.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -609,10 +609,8 @@ class ConvertToPython_3(ConvertToPython_2):
         if unquoted_in_lookup == [] or all(unquoted_in_lookup):
             return make_f_string(args, self.lookup)
         else:
-            # I would like to raise normally but that is caught by the transformer :(
             first_unquoted_var = unquoted_args[0]
-            return f"HedyException:{first_unquoted_var}"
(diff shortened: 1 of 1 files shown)


hedy.py
```diff
@@ -1306,58 +1304,63 @@ def transpile_inner(input_string, level, sub=0):
...
+            python = ConvertToPython_22(punctuation_symbols, lookup_table).transform(abstract_syntaxtree)
+        else:
+            raise Exception('Levels over 22 are not implemented yet')
+    except visitors.VisitError as E:
+        # Exceptions raised inside visitors are wrapped inside VisitError. Unwrap it if it is a
+        # HedyException to show the intended error message.
+        if isinstance(E.orig_exc, HedyException):
+            raise E.orig_exc
+        else:
+            raise E
```
bjorn3, line 1361: Should I unwrap all exceptions or only `HedyException`?

Felienne: Let's do only HedyException, I don't think we throw (or should throw) anything else

Felienne (approved):
This is making the code a lot more readable, thanks!

PR #850, opened 2021-10-02
Update app.py

oscarkarnalim:
Added Indonesian


Felienne:
Looks good @oscarkarnalim! I will deploy this to hedy-alpha.herokuapp.com so you can test it!

One thing I saw is that you use '' in the level defaults, f.e.:

Tampilkan sesuatu dengan 'print'.

That is not wrong, but if you use `` you will get syntax highlighting, which looks nicer:

Tampilkan sesuatu dengan `print`.

oscarkarnalim:
Hi @Felienne 

Thank you!

Ah you are right. I did not know that yaml works like markdown. Fortunately in the markdown files, I have used the quote you mentioned for keywords.

I will update the yaml files later along with any changes after I will have tested the deployed version.

PR #858, opened 2021-10-03
Fix text, result of switching level 5 and 6

OnnoEbbens:


coursedata/level-defaults/nl.yaml
```diff
@@ -171,7 +171,7 @@
 5:
     start_code: "print '5 keer 5 is ' 5 * 5"
     intro_text: |
-        `ask`, `print`, `if` en `repeat` werken nog zoals in level 4.
+        `ask`, `print` en `if` werken nog zoals in level 4.
```
Felienne, line 174: Nice catch! This same error occurs in all others langs too :( Could you fix these too, or f that is too much work, make an issue so I don't forgot to fix it?

Felienne (approved):

OnnoEbbens:
I added fixes for the other languages but right before I pushed these you merged the other changes into main. I am not sure which changes are added now. Next time I will reply first and then start doing stuff :)

Felienne:
Ow, sorry!! I merged before so I do not have the changes (I made an issue
because I figured you'd want to have this ready tomorrow)

Can you make a new PR based on your latest changes?

OnnoEbbens:
No worries at all. I made a new PR, I think that will do :)

PR #853, opened 2021-10-02
Create id.yaml

oscarkarnalim:
Create the indonesian version of Hedy adventures

PR #863, opened 2021-10-03
Create adventures id.yaml

rix0rrr:
Create the indonesian version of Hedy adventures.

--------------

This is a recreation of #853, which was accidentally merged and had to be recreated.


Felienne:
Will make the needed fixes myself, easier when I can push on the PR.

PR #869, opened 2021-10-04
Switch quiz level 5 and 6

OnnoEbbens:
The quizzes for level 5 and 6 were not swapped yet


Felienne (approved):
Good catch I totally forgot about the quizzes!

Felienne:
Deploying this to prod now, so it is ready for your lessons tomorrow!

PR #877, opened 2021-10-04
Update id.yaml

oscarkarnalim:
Replace single quote with skewed quote for code highlighting in Indonesian texts

### DIFF
--- a/coursedata/texts/id.yaml
+++ b/coursedata/texts/id.yaml
@@ -42,7 +42,7 @@ HedyErrorMessages:
     Parse: Kode yang kamu masukan bukan kode Hedy yang valid. Ada kesalahan pada baris {location[0]}, pada posisi {location[1]}. Kamu menuliskan {character_found}, tapi itu tidak diperbolehkan.
     Unquoted Text: Berhati-hatilah. Jika kamu meminta masukan atau menampilkan sesuatu, teksnya harus diawali dan diakhiri dengan kutip satu. Kamu melupakan salah satu tanda kutip.
     Var Undefined: Kamu mencoba untuk menampilkan variabel {name}, tapi kamu belum mengisinya dengan nilai. Ada kemungkinan bahwa kamu sedang mencoba untuk menampilkan kata {name} tapi melupakan tanda kutip.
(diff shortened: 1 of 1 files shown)


coursedata/texts/id.yaml
```diff
@@ -42,7 +42,7 @@ HedyErrorMessages:
     Parse: Kode yang kamu masukan bukan kode Hedy yang valid. Ada kesalahan pada baris {location[0]}, pada posisi {location[1]}. Kamu menuliskan {character_found}, tapi itu tidak diperbolehkan.
     Unquoted Text: Berhati-hatilah. Jika kamu meminta masukan atau menampilkan sesuatu, teksnya harus diawali dan diakhiri dengan kutip satu. Kamu melupakan salah satu tanda kutip.
     Var Undefined: Kamu mencoba untuk menampilkan variabel {name}, tapi kamu belum mengisinya dengan nilai. Ada kemungkinan bahwa kamu sedang mencoba untuk menampilkan kata {name} tapi melupakan tanda kutip.
-    Lonely Echo: Kamu menggunakan perintah 'echo' sebelum perintah 'ask', atau perintah 'echo' tanpa perintah 'ask'. Pertama tanyakan dengan 'ask' baru lakukan 'echo'.
+    Lonely Echo: Kamu menggunakan perintah `echo` sebelum perintah `ask`, atau perintah `echo` tanpa perintah `ask`. Pertama tanyakan dengan `ask` baru lakukan `echo`.
```
Felienne, line 45: Hi @oscarkarnalim. At this point we don't support syntax highlighting in the error message box yet (but that is a nice idea!) so this is not needed. In the en version we do not use any quotes, so I think we can remove them here too.

Felienne (changes requested):

coursedata/texts/id.yaml
oscarkarnalim: Hi @Felienne,

Ah sorry. Updated now. Thank you

Felienne (approved):

Felienne:
Hi @oscarkarnalim! Can you push 1 more commit to the branch? (It can be as simple as adding a space!)

Github had an outage and therefore the checker did not run properly.

oscarkarnalim:
Hi @Felienne, somehow I cannot edit the file from this branch. Should I resubmit a new request?

PR #878, opened 2021-10-04
Allow reverse ranges

### LINKED ISSUE #606: allow backwards range in level 8
That would make the fishes song easier, we can later add this as a level (when we do step, too)

thiagoarrais:
This PR is a first shot at backward ranges (fixes #606). I'm opening as draft to gather some feedback. The end goal here will be to get a minimal version working for levels 8-10 as well as 11-onwards. Optimizations, such as short-circuiting the conditional clause for literal-only ranges, will be left to later PRs.

The main question I have for now is how do you folks create generated-code-only variables. Was this already needed at some point? You may notice that this first version duplicates the body of the for loop. With an intermediate variable the compiler would be able to conditionally determine the step and use that only once in the for loop.


Felienne:
> The main question I have for now is how do you folks create generated-code-only variables.

You mean creating fresh variables that only exist in Python and not in Hedy? I don't think we ever needed that! It is an interesting question because if we pick a name of course it should not conflict with a variable that exists in the program. You could create a default one (like `j`) and then check if it is in the list? If it is, you pick a second default and then maybe a random one as fallback. I would not stat with a random one because we get more readable Python then which can be handy in debugging. Hope this helps!

Felienne:
> Optimizations, such as short-circuiting the conditional clause for literal-only ranges, will be left to later PRs.

Yea that sounds like a fine choice, first get it working, then make it better!

thiagoarrais:
> You mean creating fresh variables that only exist in Python and not in Hedy?

Yeah, exactly that. I'll try your suggestion. In case anyone got any other ideas on how to avoid the duplicated code problem, I'm always open to other suggestions!

thiagoarrais:
Just added a commit that managed to sidestep the name collision problem by using conditional expressions (learned that bit of Python for this :-).

Felienne:
> > You mean creating fresh variables that only exist in Python and not in Hedy?
> 
> Yeah, exactly that. I'll try your suggestion. In case anyone got any other ideas on how to avoid the duplicated code problem, I'm always open to other suggestions!

I am not sure what exactly the duplicate code problem is but I see you added more commits so I will have a look at this later today!

hedy.py
```diff
@@ -788,7 +788,8 @@ class ConvertToPython_8(ConvertToPython_7):
     def for_loop(self, args):
         args = [a for a in args if a != ""]  # filter out in|dedent tokens
         all_lines = [indent(x) for x in args[3:]]
-        return "for " + args[0] + " in range(" + "int(" + args[1] + ")" + ", " + "int(" + args[2] + ")+1" + "):\n"+"\n".join(all_lines)
+        step = "1 if int(" + args[1] + ")" + " < " + "int(" + args[2] + ") else -1"
```
Felienne, line 791: I know we also don't always use them but an f-string would be a bit more readable here (see: https://realpython.com/python-f-strings/)

Felienne:
> Just added a commit that managed to sidestep the name collision problem by using conditional expressions (learned that bit of Python for this :-).

So the thing I was worried about is when a kids creates a variable called i and then also uses a for loop for which we then use i. Is that covered by the new code?

(Actually, I think this problem is already there in our approach, separate from the reverse ranges! But I have to dive deeper into that later)

Felienne:
> The new code doesn't create any new (Python-only) variables. So, yeah, that is covered.

Ah yeah got it, I was thinking of this issue: #889 but that only happens for `repeat` not for `for`.

> ```
>     for i in range(int(10), int(1)+(1 if int(10) < int(1) else -1), 1 if int(10) < int(1) else -1):
>       print(str(i))
> ```
> 
> which isn't exactly idiomatic Python, but works.

Yeah it is not superpretty but indeed it does work. I would prefer something like i suggested above (`for i in range(0,10,step):`) 

but if you want to finish it like that that is ok, I can make it nicer myself too :)

thiagoarrais:
I would also prefer that, but in that case we'd have to deal with name colisions somehow. I think I'm getting this to work and leave that to a later PR.

Felienne:
> > I would also prefer that, but in that case we'd have to deal with name colisions somehow. I think I'm getting this to work as is and leave that to a later PR.
> 
> As a note to myself, dealing with the collisions should not be all too hard because we have all variables in ConvertToPython_n.lookup. So we could make a function like this:
> 
> ```
> def get_fresh_var(name):
>   while name in self.lookup
>     name = '_' + name 
>   return name
> ```
> 
> So if the name is already in the list, we put a _ before (or whatever other symbol) and keep doing that until we hit a free var. This function would also be needed to fix #889 and could then be used here too (not on this PR of course but on this idea)

Haha, you know what, since I did the thinking already, the typing is easy!

This PR adds the function, if you want to you can use it: https://github.com/Felienne/hedy/pull/892 (If not I will make the fix later)

PR #880, opened 2021-10-05
Stop long running programs in client-side JS

### LINKED ISSUE #560: while loop lvl 17 making Hedy unresponsive
When adjusting the sample level 17 to an infinite while loop, I manage to make Hedy unresponsive.
This may cause a possible data loss when signed in, because the only solution is to exit Hedy.

a possible solution would be: prevent running infinite loops by warning the user and stop the execution.

fpereiro:
- Closes #560
- Stops Python program execution in the client after 3 seconds
- Prints a modal

**How to test**:
- Go to http://localhost:5000/hedy/17
- Run this code:
```
a is True
while a is True:
    b is 1
```
- See that a modal appears after three seconds and that code execution is indeed interrupted.

@Felienne suggestions to the shown text are welcome! Once we agree on the text, I can add the text entry to the yamls.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -402,7 +402,9 @@ function runPythonProgram(code, hasTurtle, cb) {
     read: builtinRead,
     inputfun: inputFromInlineModal,
     inputfunTakesPrompt: true,
-    __future__: Sk.python3
+    __future__: Sk.python3,
+    // Give up after three seconds of execution, there might be an infinite loop.
+    execLimit: 3000
   });
 
   return Sk.misceval.asyncToPromise(function () {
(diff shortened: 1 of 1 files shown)


static/js/app.js
```diff
@@ -402,7 +402,9 @@ function runPythonProgram(code, hasTurtle, cb) {
     read: builtinRead,
     inputfun: inputFromInlineModal,
     inputfunTakesPrompt: true,
-    __future__: Sk.python3
+    __future__: Sk.python3,
+    // Give up after three seconds of execution, there might be an infinite loop.
+    execLimit: 3000
```
Felienne, line 407: Can we put this value in a function that depends on level? For now the function can just be any level returns 3000 but that gives us the flexibility to change it depending on level.

Felienne:
Hi @fpereiro 

Seems to work well, but we get a modal and also the regular error box:

image

 Not sure if that's what I specified but just the regular error box is fine. As for the text, this is fine, if you just put this in the yaml I can make updates myself.

Felienne (changes requested):

PR #881, opened 2021-10-05
replace /n with \n in quiz level 7

OnnoEbbens:
I think the '/n' were supposed to be enters so I replaced them with '\n'

### DIFF
--- a/coursedata/quiz/quiz_questions_lvl5.yaml
+++ b/coursedata/quiz/quiz_questions_lvl5.yaml
@@ -1,4 +1,4 @@
-level: 6
+level: 5
 questions:
     - 1:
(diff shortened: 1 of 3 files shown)


Felienne:
I think the test runner is stuck on this PR because of the outage GitHub had earlier, can you push one more commit so it is activated again?

Felienne:
> I think the test runner is stuck on this PR because of the outage GitHub had earlier, can you push one more commit so it is activated again?

Ah never mind the merge to main kickstarted the tests!

Felienne:
@MarleenGilsing Reminder that a newline is \n and not /n :)

Felienne (approved):

Issue #799, opened 2021-09-21
Exception logic should really be moved to hedy.py


Felienne:
Too much logic resides in app.py (Here: https://github.com/Felienne/hedy/blob/b81dbf931de9499231395237d7e9c703af5c13e4/app.py#L280-L318) This means they are out of view of the tests, for example causing #798 because of the change to ParseResult. 

This has annoyed me forever and needs to be updated.

PR #883, opened 2021-10-05
Refactor error-to-response logic in app.py (#799)

boryanagoncharenko:
Refactor error-to-response logic in app.py (#799)
Split HedyException into multiple exception classes


Felienne:
I think the test runner is stuck on this PR because of the outage GitHub had earlier, can you push one more commit so it is activated again?

Felienne:
Tests are running now but they do show an issue, but we are one step closer I guess :)

Felienne (approved):

PR #926, opened 2021-10-08
Replace AllAssignmentCommandsHashed with hashing the output of AllAssignmentCommands

bjorn3:
This reduces a lot of duplication. I am not sure if it gives an identical result though. I don't fully understand how `AllAssignmentCommands` works.


Felienne:
Sadly this will not work. The reason is that sometimes we store more than the var name in the lookup list, i.e. 

`random.choice(প্রাণী)`

Here we want to hash the variable name only, not the whole string.

Felienne:
I clarified this now in the comments

PR #939, opened 2021-10-09
chore: move to TypeScript

rix0rrr:
This PR moves the current mess of JavaScript to TypeScript. There are more future improvements to be made, but for now this is the first step that should hopefully improve our productivity going forward.

TypeScript is a statically typed variant of JavaScript that will catch many programming errors at compile time, instead of at run time.

To work on the browser code in the future, run this command in a terminal (it will not stop by itself, hit Ctrl-C to stop it):

```
$ build-tools/heroku/generate-typescript --watch
```

And then edit the `.ts` files. Commit the generated JavaScript file (`static/js/appbundle.js`) after you're done so that people that don't need to work on the browser code don't need to do anything special to be able to run Hedy locally.

--------------

Code organization is as follows:

* Packaging starts at `static/js/index.ts`. 
* Everything that needs to be exposed to the HTML templates needs to be `export`ed from `index.ts`. 
* In the HTML templates (global browser scope) it will be available as `hedyApp.yourExportedFunction()`.
* To share functions between TypeScript files, do `import { functionName } from './other-module';`.


bjorn3:
[flow](https://flow.org/) is another typechecker for javascript by facebook. It doesn't require any transpilation if you write all type hints in comments.

rix0rrr:
That's true, and it's a fair suggestion.

I think in the long run it's more  valuable to move to TypeScript, as it has a larger ecosystem and more mindshare, and is steadily becoming the future of writing JavaScript. I don't see the same move towards flow, so TypeScript seems like a better investment.

rix0rrr:
I think this is ready to be merged

Issue #841, opened 2021-10-01
Reorganize test code


Felienne:
The test code is a nice example of code that had no clear architecture and got out of hand :) 

A few issues that could use a fix:

**Multiple levels**

* Some programs need to be tested over multiple levels, this is sometimes done in three ways:
- in the file test_multiple_levels
- in individual levels with max_level = 4
- by copy-pasting in separate files

I like the second strategy most so I want to make that systematic and then also rename these tests to include the mex_level

**Naming**

Tests have super random and non-systematic names, a few suggestions for improvements:

* Remove transpile from test names
* Make it clear in the name if it is a positive (should parse) or negative (should fail) test

PR #958, opened 2021-10-11
Reorganize test code 

Felienne:
Addresses #841, work in progress


Felienne:
Pretty happy with this now, I just want to: 

* Change the order of the tests to be more logical (first old commands, than new one)
* Remove overlap where possible

PR #967, opened 2021-10-11
Frysk translation: coursedata/texts, app.py and menu.json

tjeerdhans:


Felienne:
> @Felienne @rix0rrr Looks like the unit tests fail on missing translations in other languages?

No those are just warning, but the issue was a missing course file, should be fixed now!

Felienne:
@tjeerdhans Are you planning on translating the adventures too or shall I default those to Dutch?

Felienne:
> > @tjeerdhans Are you planning on translating the adventures too or shall I default those to Dutch?
> 
> Definitely planning on doing them, but not this month, so if you can default them to Dutch, that would be great!

Cool will do! I was just making some changes to the adventures so, for now, we default to En but before I release this to the site I will make it fall back to Dutch. This will be live on alpha to test soon if you want to!

PR #979, opened 2021-10-12
Improve the user interface of the login, reset password and signup pages

TiBiBa:
Made some visual adjustments to improve the UI of the login, reset password and signup pages. All can be tested by visiting the corresponding pages. Other then visually, labels are correctly added and assigned to the corresponding inputs and a few small bugs are fixed (incorrect div and header closings). The reset.html page is updates as well, however we are unable to test this page locally and the changes should be tested on Alpha. See screenshots below for clarification of visual changes:

**Old login screen**
Schermafbeelding 2021-10-12 om 13 08 47

**New login screen**
Schermafbeelding 2021-10-12 om 13 09 00

**Old signup screen**
Schermafbeelding 2021-10-12 om 13 09 18

**New signup screen**
Schermafbeelding 2021-10-12 om 13 09 54

**Old reset password screen**
Schermafbeelding 2021-10-12 om 13 10 05

**New reset password screen**
Schermafbeelding 2021-10-12 om 13 10 16


Felienne:
Wow looks awesome Timon!

TiBiBa:
No wait, html should automatically show a browser-language placeholder when there is no starting value. I think we should be able to solve this by resetting the starting value, will have a look later today!

Felienne:
> No wait, html should automatically show a browser-language placeholder when there is no starting value. I think we should be able to solve this by resetting the starting value, will have a look later today!

Cool, hope we can fix it straight away!

TiBiBa:
Fixed! Committing a fix now (we retrieve the `auth.select` value from the gender option and use this for the country option as well)

PR #987, opened 2021-10-12
Show error locations inside the editor itself

bjorn3:
image

image

Keeping it as a draft until I have fixed the glitch on error hover and ~~added squiggly lines (if possible in ace)~~ (can't find how to do this).

### DIFF
--- a/app.py
+++ b/app.py
@@ -508,12 +508,13 @@ def get_quiz(level_source, question_nr, attempt):
                 for options_key, options_value in options.items():
                     for option in options_value:
                         for key, value in option.items():
-                            option_obj[key] = value
+                            if value:
+                                option_obj[key] = value.replace("\n", '\\n')
                             option_obj['char_index'] = char_array[i]
                     i += 1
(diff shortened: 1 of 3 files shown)


bjorn3:
> until I have fixed the glitch on error hover

I have now replaced it with a fixed message.

image

> I have not find squiggles either but you should be able to highlight with a color, see: https://stackoverflow.com/questions/16024721/how-can-i-highlight-multiple-lines-with-ace/16024998#16024998

I will try that.

Felienne:
> > until I have fixed the glitch on error hover
> 
> I have now replaced it with a fixed message.
> 
> image
> 
Do we have to have a message there? This message I think might not add much versus no message at all? If possible could we get the contents of the error there without the position? So: "you typed a space but that is not allowed"?

bjorn3:
It now looks like

image

Felienne:
> I have changed it to an empty error box now.

Cool certainly an improvement over what we have! Will check the final version out later today.

PR #992, opened 2021-10-13
404 & 500 pages

### LINKED ISSUE #65: Accessing a level that is not available gives an ugly error
For example: http://www.hedycode.com/hedy/8?lang=nl

fpereiro:
- Adds nice looking pages (with Felienne originals) for 404 and 500 errors.
- Closes #65 

**- How to test:** 404 errors are in multiple places in the app:
- Quiz requested and quiz not enabled
- No such adventure (http://localhost:5000/hedy/adventures/foo)
- No such level for adventure (http://localhost:5000/hedy/adventures/story/23)
- No such level (http://localhost:5000/hedy/23?lang=en). However, the latest code from main disabled this check, so this now throws a 500 instead of a 404. This should be addressed in a separate issue (see #993)
- No such program (http://localhost:5000/hedy/1/foo)
- No such program in full program view (http://localhost:5000/hedy/foo/view)
- No such class: (http://localhost:5000/class/foo)
- Invalid class links: (http://localhost:5000/class/foo/prejoin/bar, http://localhost:5000/class/foo/join/bar, http://localhost:5000/hedy/l/foo)


Felienne:
Let's close this, the issue was about 404 and 500 and that is fixed! We can do 403 separately.

Congrats on closing our oldest open issue haha, since April 2020!!!

Issue #1000, opened 2021-10-13
Replace mail chimp .css dependency my styling with tailwind library


TiBiBa:
Currently the mailing list form is styled by an external mail chimp .css file. This complicates the process of altering the form style into the desired, similar, style of the Hedy website. I suggest to remove this dependency and style the form with the already existing Tailwind classes.

TiBiBa:
Styled using tailwind instead of mail chimp:

![Schermafbeelding 2021-10-13 om 12 49 19](https://user-images.githubusercontent.com/38155424/137119646-8756b9a5-6812-40e8-9f88-2203c7bed3ab.png)

Felienne:
Good idea, feel free to submit a PR for this!!

Ps also congrats on issue number THOUSAND 🎉

Issue #995, opened 2021-10-13
Update UI of learn-more page to align with the other updated forms


TiBiBa:
Currently, the form for the news letter is outdated and not in style with the rest of the Hedy website. Also the required fields are indicated with a * instead of a string (as is the style in other forms). This should be aligned and the error messages should be language-dependent.

TiBiBa:
Made some small adjustments. However, we currently keep track of 9 different learn-more .md-files. Is there a way to do this more efficiently?

**Old view**
![Schermafbeelding 2021-10-13 om 12 27 49](https://user-images.githubusercontent.com/38155424/137116664-11fdfb36-6abf-4751-bbbf-e7a324c609eb.png)

**New view**
![Schermafbeelding 2021-10-13 om 12 27 58](https://user-images.githubusercontent.com/38155424/137116674-103c826b-ad75-4b1e-aff9-695a5c3fcbde.png)

PR #1001, opened 2021-10-13
Updated the mailing form on the 'learn more' page

TiBiBa:
The old style of the mailing list form was based on a mail chimp .css file. In this PR the dependency of this file is removed and the styling is based on tailwind. Solving both issues #995 and #1000. However, this style should still be altered on all other language .md files. Can be tested by looking at the 'learn more' page in dutch.


TiBiBa:
Just made another PR with a fix!

PR #1006, opened 2021-10-13
Improved 'else' syntax highlighting to be level-dependent

TiBiBa:
In the original fix we replaced the space required of on the 'else' statement. However, on levels 4 till 7 the corresponding code to the else is on the same line as the command, so a space is required. In this PR the 'syntaxModesRules.js' rules are adjusted to provide for this change. Can be tested by using the else command on a level below 7 and on 7 and higher.


Felienne:
Ai I think this will conflict with my pr that renames a bunch of stuff in the highlighter (the one that adds a new level 8)

Felienne:
I am not sure I understand the problem, for me it seems level 7 is already fine?

image

And for lower levels it also seems ok?

image

TiBiBa:
I will try to explain the added value of this PR below. Within the syntax highlighting rules we differentiate between keywords that need a space afterwards to be valid and keywords that don't. For example, print is only valid is it is followed by a space. 

Else is a unique case, because in levels 4, 5 and 6 the else command is introduced as directly being followed by the corresponding code. For example: `if naam is Hedy print 'leuk' else print 'minder leuk!'`. In this case the space after the else command is necessary to be valid. So: It should only by highlighted after the space is typed by the user. However, from level 7 and onwards indentation is introduced and the followed use of the if/else structure is introduced:

```
antwoord is ask 'Hoeveel is 10 plus 10?'
if antwoord is 20
    print 'Goedzo!'
    print 'Het antwoord was inderdaad ' antwoord
else
    print 'Foutje'
    print 'Het antwoord moest zijn 20'
```

In this case the else command is valid (and actually only correct used) **without** a space after the command. So, the command should be highlighted directly after typing the keyword itself, without waiting for an additional space. To solve for this problem we split the if/else highlighting rules into two parts: one for levels 4,5 and 6 and one for levels 7 and onwards. The result is a (very) small visual change for the end user, but makes the highlighting correct and the differentiation might be useful for more cases later on.

rix0rrr:
Level 7 on your branch:

image

TiBiBa:
All should be fixed with the newest commit

PR #1012, opened 2021-10-13
Includes explanation for "in" operator

### LINKED ISSUE #581: Missing explanation level 4
In level 4 there exists the `in` statement to check if a variable exists in a list. This syntax is not explained nor mentioned anywhere in level 4 though. It would be nice if example code was written for this feature, or if a sublevel was introduced to explain it

thiagoarrais:
Trying to fix #581 by adding a short explanation for the "in" operator.

Is this going in the right direction?

P.S.: By the way, please excuse me if someone is already working on this. I'm still getting the hang of what being listed in ["Ongoing work"](https://github.com/Felienne/hedy/projects/5) means.

### DIFF
--- a/coursedata/level-defaults/pt_br.yaml
+++ b/coursedata/level-defaults/pt_br.yaml
@@ -110,6 +110,17 @@
 
          Este código imprime legal se você inserir Hedy como um nome, e não tão legal se você inserir outra coisa.
 
+         Se quisermos conferir se algo está numa lista, agora podemos usar `in` entre a coisa e a lista.
+
+         ## Exemplo de código Hedy
+         ```
+         cores_bonitas is amarelo, verde
+         cor is ask 'Qual sua cor favorita?'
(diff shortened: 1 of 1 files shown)


Felienne:
Awesome, thanks! 

I will add the Dutch translations too so we are ready for Dutch schools. Do you know other languages that you could also add?

We are still also figuring out our process with the boards and the issues but I was not working on this yet. I think it is best if you plan to start working on an issue to assign it to yourself so we know you are taking it up. Does that work?

thiagoarrais:
> we also want to change the syntax highligher for this! Do you want to take a stab at this or shall I do that?

I'm not really sure what or where the change would be. So if you're willing to wait for my learning curve to level out: yeah, I would like to squash that.

thiagoarrais:
I have included the pt-br translation. I'll try updating the highlighter next.

Felienne:
I see you added the highlighting in another PR, cool! We can merge this one now and I will look at the highlighter next (later today)

PR #1028, opened 2021-10-17
Quiz newlines do not work in answers

LterHaar:
Each newline \n  in the option code or option_text of each answer will be replaced first by a \\n substring and when converted to a html python object, the substring will be replaced by a <br /> tag.

### DIFF
--- a/coursedata/texts/cs.yaml
+++ b/coursedata/texts/cs.yaml
@@ -20,6 +20,7 @@ ui:
     show_explanation: Ukázat vysvětlení
     hide_explanation: Skrýt vysvětlení
     hide_commands: "Hide"
+    search: "Search..."
     adventure: Dobrodružství
     switch_to_levels: Opustit dobrodružství
     switch_to_adventures: Otevřít dobrodružství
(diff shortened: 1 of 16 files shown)


Felienne:
> Hi @lotus!
> 
> Two things:
> 
> 1. this PR still has merge conflicts, can you fix these?
> 2. this PR seems to be based of 2 by 2 grid making the commit log a bit confusing. In the future, please go back to the current main and branch from there so it is easier to see what changes you made.

Let me know when you have fixed the conflicts and I can try of the solution works!

Felienne:
I tried the quiz of level 1 and it does not work:

image

Level 2 and higher do seem to work, and the newlines issue indeed is fixed!

image

Felienne:
> > Hi @lotus!
> > Two things:
> > ~1) this PR still has merge conflicts, can you fix these?~
> > Never mind I managed too fix them myself
> > 
> > 1. this PR seems to be based off 2 by 2 grid making the commit log a bit confusing. In the future, please go back to the current main and branch from there so it is easier to see what changes you made.
> 
> Yes, I shall try to use the main branch in the future. I used this branch since the yaml structure is changed here. And in the end I needed to transform the value that is assigned in the loop only in this branch and not in the main.

Great thanks! Have you also seen that there is an issue with the quiz at level 1 in this PR?

PR #1035, opened 2021-10-18
Implements varundefined for higher levels

### LINKED ISSUE #597: VarUndefined needs to be reinstated and improved
We have the error but it is no longer in use since the new grammar system (but it could be!)

Felienne:
Fixes #597 and fixes a surprising number of existing errors in the adventures and some tests!!


Felienne:
@boryanagoncharenko assigned you for a review to gain more experience with grammar/transpiler fixes!

Issue #879, opened 2021-10-05
Language selection with a search box instead of a combo box


oscarkarnalim:
Given that the number of languages accomodated by Hedy is increasing, I think it is better to change the language selection box from a combo box to a search box. As a combobox, the options fill the whole screen when clicked and it is quite difficult to choose languages at the bottom of the list.

Thank you!

Felienne:
I guess this is a quite simple change @fpereiro

Felienne:
Should not be hard to do for @TiBiBa?

PR #1045, opened 2021-10-19
Dropdown menu would disappear after :hover

TiBiBa:
Removed a redundant variable for the hide/show commands functions. Slightly altered the dropdown menu to be more user-friendly. Resolving an issue where the menu would disappear when not moving the mouse quick enough. Preparing for the implementation of #879. Also made sure the dropdown menu can no longer go out-of-window. Lastly the 'level bar' is restructured to be a flex-row, being more suitable for responsive design and no longer has the issue of buttons with long text (for example the Spanish one) getting outside of the window. See screenshot below for examples:

**Old dropdown menu**
Schermafbeelding 2021-10-19 om 11 38 03

**New dropdown menu**
Schermafbeelding 2021-10-19 om 11 37 55

**Old level bar**
Schermafbeelding 2021-10-19 om 11 37 31

**New level bar**
Schermafbeelding 2021-10-19 om 11 37 42


Felienne:
Ah thanks I noticed it was suddenly hard to switch langs!

Issue #1038, opened 2021-10-18
Add scrollbar to adventure explanations


Felienne:
Sometimes explanation gets really long and pushes the code boxes down, for example in 
https://hedy-alpha.herokuapp.com/hedy/7?lang=nl#restaurant

It is also not so pretty that the boxes "hop" between levels. Can we add a scroll bar to the top part so it is always equally long?

PR #1047, opened 2021-10-19
Fix issue #1038 Add scrollbar to adventure

redelaar:
Possible fix for #1038 

Added a scrollbar to adventures with a length longer than 20em. The boxes are not "hopping" anymore between levels unless the adventure contains a short text. I could implement a minimum height too to prevent the "hopping" completely.

See a gif of the result with a minimum height, maximum height and scrollbar: 
[Preview Gif](https://im7.ezgif.com/tmp/ezgif-7-5e9813f146de.gif)

### DIFF
--- a/static/css/additional.css
+++ b/static/css/additional.css
@@ -17,5 +17,5 @@
 
 .bg-white.p-4.mb-8.shadow-md.turn-pre-into-ace {
   max-height: 20em;
-  overflow: scroll;
+  overflow: auto;
 }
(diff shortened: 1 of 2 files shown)


Felienne:
Wow awesome work @fRedelaar! And thanks for the preview gif that saves me some local running. I see you now also place a scroll bar at the bottom of the adventure but that does not seem necessary, could we disable/remove that?

And I think a min height to prevent all hopping would also be nice.

fRedelaar:
Thank you and well spotted @Felienne ! I will remove the horizontal scroll bar, and add a min-height to prevent hopping in the next commit.

PR #1051, opened 2021-10-20
Possible fix #1049 issue quiz tab must not have scrollbar.

### LINKED ISSUE #1049: Quiz tab must not have scrollbar
Oops @fRedelaar, we forgot about quizzes in #1038 

image

Can you (or someone else) make sure that when we are in the quiz tab, we do not get the scrollbar but use the whole height?

redelaar:
Meant for issue #1049 

 Once the green 'go-to quiz' button is clicked, the tab expands in height. Afterwards, all the other tabs still have the same height and are unaffected. When going back to the quiz, it will still appear in maximum height. Also tested for level 2. 

Preview Gif

### DIFF
--- a/templates/endquiz.html
+++ b/templates/endquiz.html
@@ -12,7 +12,7 @@
     <div class="p-10 button-bar border-t-8 border-green-600">
         <div></div>
         <button class="green-btn ml-1"
-                onclick="window.location = '{{ hedy_link(level, next_assignment) }}';">Ga naar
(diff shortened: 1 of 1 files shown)


TiBiBa:
Hi, great fix! One small additional question: could you remove the changes in `additional.css` that where used earlier to set the max-height of the tabs? If you could just commit the change to this PR I think we are good to go! It is best if we keep `additional.css`  as small as possible (as all other .css is automatically generated by Tailwind).

PR #779, opened 2021-09-16
Quiz database

LterHaar:
Stores data in the DynamoDB database in tables QuizAnswer and QuizAttempt.
Data in these tables is stored by a putrequest en retrieved by a getrequest. After submitting each question a put request is done on the QuizAnswer table. And after submitting the quiz the data is submitted by a put request on the QuizAttempt table. 
Data is shown by clicking on a seperate view button for both tables.

PR #1083, opened 2021-10-22
Log quiz attempts to database

LterHaar:
Stores data in the DynamoDB database in tables QuizAnswer and QuizAttempt.
Data in these tables is stored by a putrequest en retrieved by a getrequest. After submitting each question a put request is done on the QuizAnswer table. And after submitting the quiz the data is submitted by a put request on the QuizAttempt table.


fpereiro:
Hi @LotusterHaar ! Just reviewed this PR and I have a couple of questions (we can go through them in a call).

- First of all, it looks like this PR supersedes part, but not completely, the previous PR (#779). We should see how we put all the changes in a single branch for reviewing everything in a coherent way. I can help with that if you want.
- What does `65` stand for in this line? https://github.com/Felienne/hedy/pull/1083/files#diff-568470d013cd12e4f388206520da39ab9a4e4c3c6b95846cbc281abc1ba3c959R630
- If the traditional form submission works well, I'm OK with doing it in this way instead of with an Ajax submission.

rix0rrr:
Hi Lotus,

I'm trying to help out to get this merged. I've made a bunch of changes to your PR that I hope you'll agree with. I'll do my best to summarize them here. The major ones are:

* I changed the database schema. Instead of having 2 tables, one with answers and one with completed attempts, we now have 1 table that is continuously updated every time a user submits a new answer. We record all the attempts in the one record, see the explanation in `database.py`.
* I changed the YAML schema. In a couple of places, we had a dictionary of one element inside a list. This made it awkward to work with, so I changed it back to being either a dictionary with number keys, or a list of objects.
* I changed how pages are linked together. Instead of rendering HTML in response to POST requests, it is considered good form to do a redirect-after-post (see [some](https://en.wikipedia.org/wiki/Post/Redirect/Get) [sources](https://stackoverflow.com/questions/1083357/do-you-always-redirect-after-post-if-yes-how-do-you-manage-it)). The POST handlers change state and write to the database, the GET handlers only produce HTML. This means I had to introduce separate GET handlers to show the HTML pages for `feedback` and `quiz_finish`.

Some smaller changes:

* I removed the handlers for `/quiz/results/<level_source>` and `/quiz/quiz_questions/answer/<level_source>`. I noticed  the HTML files they were rendering didn't exist in this PR, so they didn't work anyway, and appeared to be unused. If they should exist, you can add them again after this PR gets merged.
* I started using [`url_for`](https://flask.palletsprojects.com/en/2.0.x/api/#flask.url_for) to generate links. We should have done this from the start, but we didn't know 😅. Now that we do, let's start doing it the proper way.
* I moved most of the code that handles the quiz YAML data into helper functions in `quiz.py`, to reduce duplication and hide away the internals of those files.
* If the session doesn't have a quiz attempt ID yet, we redirect back to the start page.
* I added a filter called `nl2br` to turn newlines into `<br>` in the Flask template, instead of doing a full replace on the template output.

Felienne (approved):

PR #1107, opened 2021-10-26
Give errors for unrecognized turn and forward arguments instead of using a default

### LINKED ISSUE #1019: Give an error for incorrect turn argument
The `turn` command currently interprets any argument that is not any of: "left", "right", a number or a variable name as a 90 degree right turn. This will be confusing if a variable name was misspelled. It should give an error instead.

### LINKED ISSUE #1020: Give an error for incorrect forward argument
The `forward` command currently interprets any argument that is not a number or a variable name as a 50px forward move. This will be confusing if a variable name was misspelled. It should give an error instead.

bjorn3:
In the process I found that left and right turn were translated inside code of the Spanish version. This would previously default to a right turn, which I can imagine would be confusing for some kids.

Fixes #1019
Fixes #1020


coursedata/texts/nl.yaml
```diff
@@ -57,6 +57,7 @@ HedyErrorMessages:
     Invalid Argument Type: "Je kan {command} niet gebruiken met {invalid_type}{invalid_argument}. Je kan {command} wel gebruiken met {allowed_types}."
     Unsupported Float: "Helaas, Hedy ondersteunt nog geen kommagetallen. Verander {value} in een geheel getal."
     Locked Language Feature: "In jouw programma zit {concept}! Goed gedaan! Maar {concept} is nog niet beschikbaar. Dat komt in een later level."
+    Invalid Argument: "{arg} is niet een geldig argument for het {command} commando."
```
Felienne, line 60:
```suggestion
    Invalid Argument: "{arg} is geen geldig argument voor {command}."
```

coursedata/texts/en.yaml
```diff
@@ -66,6 +66,7 @@ HedyErrorMessages:
     Invalid Argument Type: "You cannot use {command} with {invalid_type}{invalid_argument}. Try using {command} with: {allowed_types}."
     Unsupported Float: "Unfortunately, Hedy does not support non-integer numbers. Try changing {value} to an integer."
     Locked Language Feature: "You are using {concept}! That is awesome, but {concept} is not unlocked yet! It will be unlocked in a later level."
+    Invalid Argument: "{arg} is not a valid argument for the {command} command."
```
Felienne, line 69:
```suggestion
    Invalid Argument: "{arg} is not a valid argument for {command}."
```

Felienne: I prefer to just refer to the command, but also I think this is a complex error message for a kid. What is an argument? Why is it not valid?

hedy.py
```diff
@@ -643,11 +646,13 @@ def ask(self, args):
...
     def forward(self,args):
-        # when a not-number is given, we simply use 50 as default
+        if len(args) == 0:
+            return self.make_forward(50)
+
         try:
             parameter = int(args[0])
         except:
-            parameter = 50
+            raise InvalidArgumentException(command = "forward", arg = args[0])
```
Felienne, line 655: Can we loop in allowed types here for a better error message (see screenshots in discussion)

Felienne:
Hi @bjorn3 (and @boryanagoncharenko)

I think we now have somewhat contradicting ways of communicating issues:

image

image

I really like the top one because it is actionable, it says what to try instead. Can we loop in the expected types in the lower one so we get suggestions too?

coursedata/texts/en.yaml
bjorn3: I removed this error message. Instead I switched it's only remaining use to UndefinedVarException.

bjorn3:
It now gives 
image

Felienne:
> It now gives image

Yeah that is still confusing because the kid did not try to  print lalala. They did not understand what to use after forward.

What I'd like is the error message: You cannot use forward with lalalala. Try using number instead.

I am not sure why that does not happen if you raise InvalidArgumentTypeException? I can look into this to if you don't know either.

Felienne:
Ah I see, we only raise the error in level 1 and I was testing in level 2! I can fix that.

Felienne:
Also... the grammar in level 1 does not allow text behind the forward so first and foremost we get a parse error in level 1.

Felienne (approved):

Issue #922, opened 2021-10-08
Split level1.lark into keywords.lark and level1.lark


HuibSprangers-leiden:
This is so keywords can be more easily overwritten

teundemast:
I think all non-keyword tokens should remain inside level1.lark as they won't be translated.

teundemast:
Missed 4 non-keyword tokens to remain inside level1.lark...

Issue #911, opened 2021-10-08
Lark Grammar merging between languages


HuibSprangers-leiden:
The lark grammar must be able to be changed whenever the language of the site is changed

PR #1110, opened 2021-10-27
Keyword translation first sprint

HuibSprangers-leiden:
Currently the following has been implemented:
- Keywords are separated into their own keywords-(lang).lark files. Resolves issue #922
- When the language of the site is changed, so do the keywords change to their respective language. Resolves issue #911

The only keywords-(lang) files that have been implemented are currently English and Dutch. The other languages are currently unimplemented and use English as the default.

To test this: 
Run the site and switch languages from Dutch to English. The current English command for e.g. 'print' will not work if the language of the site is set to Dutch. Instead Dutch versions of these keywords such as 'drukaf' are used.


Felienne:
According to our meeting this week I am closing this pr and expect you to make a few smaller ones with cherry picking. Let me know if you need help!

Issue #1142, opened 2021-10-31
[BUG] Space at beggining of line is treated as parse error in levels 3, 4, 5,6


jpelay:
**Describe the bug**
In levels 3, 4, 5 and 6 if one starts a line with just a space it will be treated as a parse error, because the invalid_space rule is not added in the grammars for those levels.

**Paste the Hedy code & level**
Add a space to the beginning of any lines in those levels.

**Add a screenshot (optional)**
This one is from level 5
![image](https://user-images.githubusercontent.com/45865185/139601194-5cd2c9c4-6d25-4541-9a22-f4a59f1a8241.png)

**Expected behavior**
InvalidSpaceException should be thrown and the code should be fixed and executed.

PR #1143, opened 2021-10-31
Fixes Issue 1142: space at beginning of line gives error levels 3, 4, 5 and 6 

racertop:
**Description**
Fixes issue #1142, adding `invalid_space` rule to the grammars of those levels. I edited:

* grammars/level4-Additions.lark
* grammars/level4-Additions.lark
* grammars/level6-Additions.lark

To add  `invalid_space` as a rule in the `command` rule of those levels so it can appear later in the tree to be detected and fixed, just as we do in levels 1 and 2.

**Fix for**
#1142 

**How to test**

Just add a space to the beginning of a line in levels 3, 4, 5 and 6. 

**Checklist**


racertop:
Well, this fix might not work because spaces aren't allowed inside ifs in level 4+.

Felienne:
> Well, this fix might not work because spaces aren't allowed inside ifs in level 4+.

Ah yeah true! We could allow them at the start of the line like we do with " print 'hallo'"? But I can also see how we only offer this sort of help until level 3.

Felienne:
ok your error message is not bad:

image

But because this is legal code later on, it is a bit confusing. 

I think my preferred solution for level 5 (and 6, until we have indented blocks) is to allow this and give no warning at all. This might be done by kids (or parents) with a bit of programming knowledge already. What do you think?

racertop:
@Felienne  I do agree this might be executable code. What about adding a new error message? Something like `InvalidMultilineIf`, that tells you "Your trying to write a multiline if, but that is not unlocked until later on", like we do with nested blocks in level 6, 7, etc. I'm not entirely sure how to achieve this, but Ill look into it.

Felienne:
> @Felienne I do agree this might be executable code. What about adding a new error message? Something like `InvalidMultilineIf`, that tells you "You're trying to write a multiline if, but that is not unlocked until later on", like we do with nested blocks in level 6, 7, etc. I'm not entirely sure how to achieve this, but Ill look into it.

That is a lovely idea!! You can do that by throwing a LockedFeature Exception, see how that is done for example here: https://github.com/Felienne/hedy/blob/ba6c6ec1b32d100f8cc5ee88964c0ee9dfcc28e8/hedy.py#L1492

Felienne:
Beware though, I am going to merge #1144 soon so you might want to wait a bit to avoid merge conflicts

racertop:
> Beware though, I am going to merge #1144 soon so you might want to wait a bit to avoid merge conflicts

Perfect! I'm probably not going to work on this until the next few days, so I will pull before working on this new idea!

Felienne:
Making this into draft so I can keep track of open pr's a bit easier

Felienne:
Hi @racertop! What do you think about this one? Is there a path to finishing or is it too far behind main?

racertop:
> Hi @racertop! What do you think about this one? Is there a path to finishing or is it too far behind main?

I think I'm going to do a new PR, since I have to redo some work. There is some more work before this is valid:

- Add a way to detect indented ifs on levels 4, 5, 6 (I have a working prototype for this but involves string hacking and I need to include your start block update).
- Add invalid_space rule to levels 4, 5, 6 (what this PR tries to do).

But maybe we can do better by adding an identer class to Hedy, which is what Im working on right now. It is compatible with the Earley parser if we use the standard lexer, I need to check if it doesn't mess to much with the grammar.


So I'm clossing this PR, I honestly should not have uploaded it before running tests locally, I'm sorry for that!

PR #1146, opened 2021-11-01
Show 404 page when requesting a level that does not exist

### LINKED ISSUE #993: Accessing a level that is not defined throws a 500 error
For example: http://localhost:5000/hedy/23

There is a check for this in `hedyweb.py` but before the execution reaches it, the endpoint at `app.py` throws. We should add tighter validation in the `app.py` endpoint for requesting a level (`defaults = level_defaults_for_lang.get_defaults_for_level(level)`).

fpereiro:
- Closes #993 

**How to test**:
- Go to http://localhost:5000/hedy/29
- See that the app shows a 404 page with the error "No such Hedy level!".
- Note that the app doesn't crash anymore when that page is requested.

### DIFF
--- a/app.py
+++ b/app.py
@@ -766,6 +766,8 @@ def index(level, step):
 
     adventures = load_adventures_per_level(requested_lang(), level)
     level_defaults_for_lang = LEVEL_DEFAULTS[requested_lang()]
+    if level not in level_defaults_for_lang.levels:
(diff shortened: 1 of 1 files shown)


Felienne:
Works!
image

Felienne (approved):

Issue #1064, opened 2021-10-21
More precise error locations


Felienne:
Currently, we highlight a full line, but sometimes we have more information about the location in the line, would be nice to use that but of course that requires changing some parser logic to output begin and end of the issue for parse errors, and even adding begin location for other types of errors.

https://github.com/Felienne/hedy/blob/c778f7c221e7721e725460db3b49a4df006b72bc/static/js/app.js#L194

Felienne:
Similar to #845?

Felienne:
Bumping this one up and assigning it to myself because this will be a lot easier now that https://github.com/Felienne/hedy/pull/3155 is merged and we have access to `meta` in the whole transpiler.

Felienne:
Superseded by #3388

PR #1147, opened 2021-11-01
A more precise error highlighter

dolfein:
These changes mainly target #1064 and tries to improve upon #987. Making the error highlighter more specific.
It works on the 2 error types discussed in #1064.

Before it highlighted the whole line where an invalid character was put, now it just highlights the invalid character and also invalid commands.

Since this is my first contribution to Hedy, all feedback is more than welcome!


Felienne:
Hi @dolfein!

Thanks for your first pull request! There seems to be an issue in the typescript, you can see that in the tests:

https://github.com/Felienne/hedy/runs/4073690029?check_suite_focus=true

The error is:

static/js/app.ts(222,29): error TS2339: Property 'start' does not exist on type 'void'.

The 222, 29 means that the error exists on line 222 column 29. Lines 230 and 232 have the same issue. Does that error message mean anything to you?I am not super well versed in typescript so I don't immediately know what the issue is either but if you don't I can dine into it a bit more.

hedy.py
```diff
@@ -205,8 +205,12 @@ def __init__(self, **arguments):
...
 
 class InvalidCommandException(HedyException):
-    def __init__(self, **arguments):
-        super().__init__('Invalid', **arguments)
+    def __init__(self, level, location, invalid_command, guessed_command):
+        super().__init__('Invalid')
+        self.level = level
+        self.location = location
+        self.invalid_command = invalid_command
+        self.guessed_command = guessed_command
```
Felienne, line 214: If you place the arguments in a dictionary you do not have to set individual fields because the ** will take care of that for you.

dolfein:
The typescript error seems to be resolved with my latest commit, though I am now getting another failed test:
image
It seems like all the runs for this test have failed so far, is the test itself faulty?

And I get ```FAILED tests/tests_level_04.py::TestsLevel4::test_pront_should_suggest_print```
This is probably because I commented ```error_message``` and passed it as a string, which makes it just print "error_messsage" instead of the usual invalid command error message.

Felienne:
> And I get `FAILED tests/tests_level_04.py::TestsLevel4::test_pront_should_suggest_print`
> This is probably because I commented `error_message` and passed it as a string, which makes it just print "error_messsage" instead of the usual invalid command error message.

That is not necessarily the behaviour we want of course, can you revert that change so the test passes again?

static/js/app.ts
```diff
@@ -6,6 +6,8 @@ import { auth } from './auth';
 
 export let theGlobalEditor: AceAjax.Editor;
 
+declare var range: any;
```
rix0rrr, line 9: There's not really a need to make this a global variable, since we only use it in the block below. Better to avoid making this variable unnecessarily visible (especially if it's mutable). That could only lead to mistakes later on.

Also typing it as `any` makes it possible to misuse it. We can have the compiler automatically derive the type for us if we assign the variable immediately when we declare it, that saves typing and headaches.

It would suggest making this a local, immutable value below.

static/js/app.ts
```diff
@@ -218,7 +220,33 @@ export function runit(level: string, lang: string, cb: () => void) {
       }
       if (response.Error) {
         error.show(ErrorMessages['Transpile_error'], response.Error);
-        if (response.Location && response.Location[0] != "?") {
+        if (response.Invalid_command)  {
+          range = editor.find(response.Invalid_command, {
```
rix0rrr, line 224: A replacement for the `var range` above:

```suggestion
          const range = editor.find(response.Invalid_command, {
```

hedy.py
```diff
@@ -1605,7 +1609,7 @@ def transpile_inner(input_string, level):
                                                             invalid_argument=''.join(invalid_info.arguments))
                 # clearly the error message here should be better or it should be a different one!
                 raise ParseException(level=level, location=["?", "?"], keyword_found=invalid_command)
-            raise InvalidCommandException(invalid_command=invalid_command, level=level, guessed_command=closest)
+            raise hedy.InvalidCommandException(invalid_command=invalid_command, level=level, guessed_command=closest, location=[line, "?"])
```
rix0rrr, line 1612: This is not necessary. This exception is already available in the current file, so you definitely don't need to namespace it.

I'm noticing other exceptions are doing it too... but honestly: we should remove the `import hedy` statement, and remove `hedy.` everywhere it's being used.

(But you can do that in a different PR as well, it doesn't really have bearing on the change you're making here)

hedy.py
rix0rrr: I'm not opposed to making the expected arguments explicit in the constructor signature. It will make it harder to misuse the class later on. For example, if you did the following:

```python
InvalidCommandException(invalid_commadn='xyz')
```

Then:

* The linter would see that you misspelled 'command'
* The linter would see that you are missing two arguments.

However, what happens to the arguments here is *different* than what used to happen:

## Old

```python
    def __init__(self, level, **arguments):
        super().__init__('Invalid', **arguments)
```

This passes all keyword arguments in `arguments` up to the **superclasss**.

##  New

```python
    def __init__(self, level, location, invalid_command, guessed_command):
        super().__init__('Invalid')
        self.level = level
        self.location = location
        self.invalid_command = invalid_command
        self.guessed_command = guessed_command
```

This doesn't pass any arguments up to the superclass anymore. Instead, it stores all arguments onto the current class.

## What's the difference

The mechanism where all exception parameters are passed up into the superclass (where they get stored into the `ex.arguments` dict) is used to inject these strings into formatted error messages.

So everything you don't pass to the superclass cannot be used in error message format strings! 

At the same time, `hedy_error_to_response()` will look at attributes that exist on the exception object itself (specifically `location`), to format them into the response object.

So we need to pass certain attributes up to the superclass (to be packed into `ex.arguments`), as well as store some onto the exception object directly.

## Recommendation

I would recommend writing it like this:

```python
    def __init__(self, *, level, location, invalid_command, guessed_command):
        super().__init__('Invalid', level=level, invalid_command=invalid_command, guessed_command=guessed_command)
        # Mark the error location for rendering in the frontend
        self.location = location
        self.location_length = invalid_command.length
```

This will have the following benefits:

* It makes the required arguments explicit, so the class cannot be accidentally misused
* The lone `*` in the argument list makes it so that all arguments MUST be passed as keyword arguments, not positionally. This makes it harder to accidentally mess up the order of the arguments.
* It still passes all the arguments that are likely to be used in the error format string up to the superclass
* It sets `location` and `location_length` to the values that are expected by `hedy_error_to_response`.

app.py
```diff
@@ -357,10 +357,10 @@ def parse_error_to_response(ex, translations):
 arguments_that_require_translation = ['allowed_types', 'invalid_type', 'required_type', 'character_found', 'concept']
 
 def hedy_error_to_response(ex, translations):
-    error_message = translate_error(ex.error_code, translations, ex.arguments)
+    #error_message = translate_error(ex.error_code, translations, ex.arguments)
     location = ex.location if hasattr(ex, "location") else None
-    return {"Error": error_message, "Location": location}
-
+    invalid_command = ex.invalid_command if hasattr(ex, "invalid_command") else None
+    return {"Error": "error_message", "Location": location, "Invalid_command": invalid_command}
```
rix0rrr, line 363: I think you are trying to add some fields into the JSON here if they exist on the error, but in the process you turned off error translation for all errors. That's probably not what you intended to do.

I would recommend leaving the existing `Error` field to be what it is: a human-readable string that will be used for various purposes. 

-------------

Something else: I'm seeing you are adding the `Invalid_command` field to the response... but you are only using it to later on get the **length** of the invalid command, so that you can use it to mark a subsection of a line.

But what if we want the same feature later on to indicate type errors? Or where kids still need to fill in a placeholder? Are we going to reuse `Invalid_command`? Or are we going to add fields `Invalid_type`, `Invalid_placeholder`, etc? When all we need is just the length?

To keep it nice and simple, I would call the field something like `LocationLength`, and if given it will indicate the number of characters starting at `Location` that need to be highlighted because they contain a mistake.

If you do that, the code could look something like this:

```python
def hedy_error_to_response(ex, translations):
    return {
        "Error": translate_error(ex.error, translations, ex.arguments),
        "Location": getattr(ex, "location", None),
        "LocationLength": getattr(ex, "location_length", None),
    }
```

static/js/app.ts
```diff
@@ -218,7 +220,33 @@ export function runit(level: string, lang: string, cb: () => void) {
       }
       if (response.Error) {
         error.show(ErrorMessages['Transpile_error'], response.Error);
-        if (response.Location && response.Location[0] != "?") {
+        if (response.Invalid_command)  {
```
rix0rrr, line 223: With the new generic `LocationLength` field:

```suggestion
        if (response.Location && response.Location[0] != '?' && response.LocationLength)  {
```

Probably, there is now a bunch of duplication between the "complete line" error marker and the "specific" error marker blocks.

So probably the logic should look something like this:

```ts
        if (response.Location && response.Location[0] != '?') {
            // Add (line) annotation

            if (response.LocationLength) {
                // Add (range) marker
            }
        }
```

Felienne:
Awesome, this seems to work @dolfein!!

image

What a lovely extension of our error messages. We can leave the exception refactoring for a separate PR and move on with this.

Felienne:
Hi @dolfein! I cam confirm this works:

image

But you are making a lot of changes in the Python code that are not really needed (as @rix0rrr commented on earlier), so I am going to make a new branch with just your ts changes and make a new PR for that. Because you worked from your main I cannot commit to this PR otherwise I would fix it here.

rix0rrr:
Thanks for all the work put into figuring this out @dolfein! 

I reused the principles of what you did, but did them slightly differently here (hopefully, I'd say a bit simpler): https://github.com/Felienne/hedy/pull/1184

PR #1155, opened 2021-11-03
fix: translate page JavaScript doesn't load

rix0rrr:
The JavaScript app bundle depends on `client_messages.js` having
been loaded beforehand in the same page, but this script had not
yet been added to the HTML template.

Add an error message to indicate what went wrong, and fix the error.


Felienne:
@HuibSprangers-leiden I think this fixes your issue, can you confirm this?

Felienne (approved):
Works for me!

teundemast:
Yes, I can confirm this fixes our issue! Thank you! @Felienne @rix0rrr

PR #1160, opened 2021-11-03
Users are always able to visit the for-teachers page

TiBiBa:

**Description**

Currently, users are always able to visit the for-teachers page, even when they are not a teacher, by directly navigating to the url. The shown class information is empty and the users are unable to create a class but still: this is undesirable behavior. In this PR we add an additional check in `app.py` to verify if the user is teacher when requesting the for-teachers page. If not, the user is redirected to `index.html`.

**Fix for**

**How to test**

Try to visit the for-teachers page directly using this link: [](http://0.0.0.0:5000/for-teachers). If you're currently not logged in as a teacher you should be re-directed to the homepage. Otherwise the for-teachers page should be shown as expected.

**Checklist**


Felienne:
Works but... shouldn't we throw a 403 or something? Supposed I have this page bookmarked as a teacher and I am logged out because of a time out, it is a bit weird to go to the home page?

TiBiBa:
Changed to return a 403 error! It might be nice to create a template to catch these similarly to the 404 page.

Felienne:
> Changed to return a 403 error! It might be nice to create a template to catch these similarly to the 404 page.

I can make a drawing with an burglar Hedy that tries to break in if you make a template :D

Felienne:
> > Changed to return a 403 error! It might be nice to create a template to catch these similarly to the 404 page.
> 
> I can make a drawing with an burglar Hedy that tries to break in if you make a template :D

We can also make an issue, could be a great good-first-issue!

Felienne (approved):

PR #1165, opened 2021-11-04
Fix: Pop-up to "save program" shown when not signed in

### LINKED ISSUE #1162: [UI idea] Don't show "save program" pop-up when user is not logged in
**Idea incl level**
Currently the user is always shown a warning pop-up when they attempt to switch pages and there are currently unsaved changes, being checked by the `window.State.unsaved_changes` boolean. However, when a user is not logged in, he/she will be unable to save the program so the pop-up is redundant. It would be nice to add an additional check to only show the pop-up if the user is logged in.

In the future it would be better to remove the pop-up all together or replace by a less intrusive message attending the user on the unsaved program. As the current way of a pop-up can be seen as forced and annoying by users. But, for now the current pop-up is a bug and should be fixed with priority.

TiBiBa:

**Description**

In this PR we fix the issue where the pop-up to save a changed program is shown every time a user tries to switch webpage. However, when the user is not logged in this is undesirable as they are unable to save the program anyway. Returning a rather annoying pop-up (every time!) a user switches Hedy levels. This is solved by adding an additional check to the `prompt_unsaved()` function to check if there is a `auth.profile` value. If so, we know that there is a user logged in, otherwise: don't return the pop-up.

**Fix for**

This PR fixed #1162

**How to test**

Create or alter a standard program in a Hedy level of choice when not being logged in. No pop-up should occur when switching Hedy levels. Verify the correctness by logging in and repeating the steps to find a pop-up showing up.

**Checklist**


Felienne:
Hi Timon!

The modal is gone, but now I get a browser popup:

image

I am not sure this is a thing we can control; maybe this is Chrome being smart?

TiBiBa:
That might be chrome issue, it didn't occur for me (I only tested with Safari). There must be a fix somewhere as I can imagine more web developers with this issue. Will take a look!

TiBiBa:
The last commit should contain a fix with the default browser pop-up. The mistake was setting the `window.State.no_unload_prompt` variable **after** a possible return. This should be fixed now and the pop-up should no longer occur

dvberkel:
I have tested this pull-request. I tested it with two browsers: Firefox and Chrome. Both browser display the intended behavior; no popup is shown when you are not signed in.

I did not experience the pop-up mentioned by @Felienne

dvberkel (dismissed):
This looks good to me: :shipit:

dvberkel (approved):
The files `static/js/appbundle.js` and `static/js/appbundle.js.map` can be safely removed because they are generated on the server

PR #1166, opened 2021-11-04
fix: signup fails if languages are selected

rix0rrr:
If any "previous experience" languages are selected, the signup button
doesn't do anything.

There was a programming mistake around using `this` in JavaScript:

```ts
$('input[name=languages]').filter(':checked').map(function() { return $(this).val() as string }).get()
```

Would have been correct, because `function()` doesn't bind `this` (and
so `this` would have referred to the checkbox.

However, we wrote this:

```ts
$('input[name=languages]').filter(':checked').map(() => $(this).val() as string).get()
```

And because a `() => { }` arrow function DOES bind `this` to whatever
it's currently bound to, we wouldn't be bound to the checkbox but to the
global object.

Instead, just make the actual argument explicit, and don't rely on
`this` at all. It's always tricky in JS!


rix0rrr:
Note that you will get an error when trying this in offline mode, because of https://github.com/Felienne/hedy/issues/1163

If you get an error, that's a sign it's working! (Otherwise you wouldn't see anything)

Felienne (approved):

PR #1175, opened 2021-11-05
Adding confetti and green notification for successful running code

fRedelaar:
**Description**
Whenever a user successfully runs their code, a green notification shows up saying "Great job" in their own language. (For now, I have only implemented Dutch and English). In the green notification, a confetti emoji is also present. If the user clicks on this, there will be confetti on the screen for 2 seconds. 

https://user-images.githubusercontent.com/33595265/140512193-4d5c0c84-25b8-4eb2-ba43-d3f2d384939f.mov

In the video, you can see that the green error will **only** appear if the user runs code successfully. The confetti will **only** appear if the user clicks on the emoji. 

The code that has been pushed to this branch will produce the video like above. This story is still a work in progress!


static/js/app.ts
```diff
@@ -230,6 +231,10 @@ export function runit(level: string, lang: string, cb: () => void) {
         }
         return;
       }
+      if (response.Code){
+        console.log("success!");
+        success.show(ErrorMessages['Transpile_success']);
+      }
```
bjorn3, line 237: I think this should happen after the python program runs. Currently running it may still give an error.

MarleenGilsing:
I love it, @fRedelaar ! One addition: Could you make the confetti emoji only once clickable, so that the students don't get distracted by clicking it over and over again?

templates/incl-editor-and-output.html
```diff
@@ -49,16 +50,10 @@
...
-                viewBox="0 0 20 20">
-                <path
-                  d="M2.93 17.07A10 10 0 1 1 17.07 2.93 10 10 0 0 1 2.93 17.07zm12.73-1.41A8 8 0 1 0 4.34 4.34a8 8 0 0 0 11.32 11.32zM9 11V9h2v6H9v-4zm0-6h2v2H9V5z" />
-              </svg>
-            </div>
-            <div>
+            <div class="flex" style="justify-content: space-between; width: 100%; align-items: center;">
               <p class="caption font-bold">Success</p>
-              <p class="details text-sm">Something went according to plan.</p>
+              <button onclick="hedyApp.confetti_cannon()">🎉</button>
```
bjorn3, line 55: I think it is a bit non-obvious that clicking it will give confetti. Maybe adding a border would work?

Felienne: Nah I think it is clear enough! It is also an extra so if kids don't find it that is not al too bad (and they will all tell each other when one kid finds it 🤣 )

fRedelaar:
> I love it, @fRedelaar ! One addition: Could you make the confetti emoji only once clickable, so that the students don't get distracted by clicking it over and over again?

Smart thinking! I've just sent a commit where the confetti popper emoji will be hidden/invisible so also unclickable for users after it has been clicked once.

Felienne:
Love this idea @fRedelaar and your clear communication on your plans! Merging it into main now, so you can test it on alpha soon (hey-alpha.herokuapp.com)

Felienne (dismissed):

rix0rrr (approved):

rix0rrr:
I made changes to include the JavaScript package directly into our source (instead of loading it from a CDN), and change the include tag to be compatible with old browser (schools might still be using old versions of Internet Explorer).

Also had to add TypeScript declarations to make the code compile: we have to tell TypeScript what the `JSConfetti` class looks like.

fRedelaar:
Thank you @Felienne for the kind words! I am planning to make (at least) one more commit because there is a small bug if the user switches to a different tab (adventure). I still need to hide the green notification. I can make progress on this later today or early tomorrow.

https://user-images.githubusercontent.com/33595265/140609613-a1db9ef4-6f00-42be-bdaf-5f5bc4b30b52.mov

PR #1177, opened 2021-11-05
Fix: Set maximum amount of "try it" attempts when logged in

### LINKED ISSUE #1176: Block the try buttons in the example codes after a while
I noticed in my lessons that my students love playing with the example codes, but therefore they barely do any coding themselves. So I'd like them to be able to use the example code 3 times (or so) and then the code should be removed from their input screen and the try button should be blocked. So they have to type in the code themselves if they want to keep playing. 

Another option would be for the teacher to be able to block the try button in their students' adventures.

TiBiBa:

**Description**

Currently, users are able to try the example programs unlimited times. But, as pointed out by @MarleenGilsing, students no longer create their own programs. In this PR we implement a counter to restrict logged in users to only try out three different example codes per level, enforcing them to write their own code after looking at some examples.

**Fix for**

This PR fixed #1176.

**How to test**

Notice the different view on the code page when being logged in or not. When logged in a header should show with information on how many attempts of example code you have left. When reaching 0 the commands window on the left side will disappear. If, for some reason, the student still tries to copy example code a pop-up warning will appear. In the case of not being logged in the UI and functionality should be identical as before this PR. See screenshot below for clarification:

Schermafbeelding 2021-11-05 om 18 39 14

Schermafbeelding 2021-11-05 om 18 39 23

**Checklist**


Felienne:
Quickly implemented @TiBiBa!! I was wondering though if we should now show the message that kids are out of tries as a modal? That is a bit more aligned with the rest of the site, and also this line takes a bit of space from the UI.

Felienne (approved):
Maybe make the feedback as modal?

Felienne (changes requested):
Feedback in a modal?

TiBiBa:
> Feedback in a modal (second try because previous one was stale)

Do you mean only return a modal when there are no attempts left? I think it won't be user-friendly if we return a modal each time an example program is used (as the user has to close the modal as well). But otherwise, the user has no indication on the amount of attempts he/she has with example programs. I think it is nice to return some way of feedback to shown them the attempts left. But I agree that the added line unnecessary takes some space of the UI.

TiBiBa:
I've changed the structure to only show a modal after x attempts. Removing the extra header shown in the first implementation.

coursedata/texts/en.yaml
```diff
@@ -153,6 +153,10 @@ Auth:
     other_text: "Another text language"
     teacher_invitation_require_login: "To set up your profile as a teacher we will need you to log in. If you don't have an account, please create one."
     teacher_welcome: "Welcome to Hedy! Your are now the proud owner of a teachers account which allows you to create classes and invite students."
+    example_left: "You can still try 1 example program"
```
Felienne, line 156: I think we can remove the texts for x left if we only use the modal?

coursedata/texts/nl.yaml
```diff
@@ -144,7 +144,10 @@ Auth:
     other_text: "Een andere teksttaal"
     teacher_invitation_require_login: "Om je leerkrachtenaccount te activeren moet je eerst inloggen. Als je nog geen account hebt, maak dan eerst een account aan."
     teacher_welcome: "Welkom bij Hedy! Jouw account is omgezet naar een leerkrachtenaccount. Je kan nu klassen maken en er leerlingen in uitnodigen."
-
+    example_left: "Je kunt nog 1 voorbeeld programma proberen"
+    examples_left: "Je kunt nog ... voorbeeld programma's proberen"
+    no_examples_left: "Je kunt geen voorbeeld programma's meer proberen"
+    examples_used: "Je hebt jouw maximaal aantal voorbeeld programma's voor dit level al gebruikt! Probeer zelf iets te programmeren."
```
Felienne, line 150:
```suggestion
    examples_used: "Je hebt jouw maximaal aantal voorbeeldprogramma's voor dit level al gebruikt! Probeer nu zelf iets te programmeren."
```

static/js/app.ts
```diff
@@ -42,6 +44,12 @@ export let theGlobalEditor: AceAjax.Editor;
     }
   }
 
+  function initializedExampleCounter() {
+    if (window.State.examples_left <= 3) {
```
Felienne, line 48: Shall we place the 3 somewhere so that it is easy to customize? Maybe teachers should be able to set it for their classes? And maybe it should also be dependent on the level (unlimited for 1, 10 for 2, 3 for 3 etc.)?

Felienne (changes requested):
Thanks for this @TiBiBa! I think it is a lot better than the text above, and technically everything seems to work well, but I am just not sure how I want this feature to work. The way we have it now might be a bit too restrictive, maybe we should allow teachers to set the number of tries in their class, and maybe also per level?

What do you think @MarleenGilsing?

coursedata/texts/en.yaml
TiBiBa: Yes, this these can be removed!

static/js/app.ts
TiBiBa: Agree! Not sure what would be the best place to store this value, will take a look!

TiBiBa: I just noticed this function is redundant as we don't give inline feedback to the user anymore. It is removed. To keep track of the amount of examples_left we use a `window.State` value called `examples_left`. Which is set to 3 if the user is logged in, otherwise it is set to max int

Felienne: ok!

Felienne:
Hi @TiBiBa!

I think we can merge this just so we do not have too many PR's open with potential merge conflicts, but let's set the number really high for now (20?) so kids do not hit it all too often. Ultimately what I think I want with this feature is that teachers can set it for a class and for a level in their teacher's interface (maybe you can make an issue for that?)

TiBiBa:
> Hi @TiBiBa!
> 
> 
> 
> I think we can merge this just so we do not have too many PR's open with potential merge conflicts, but let's set the number really high for now (20?) so kids do not hit it all too often. Ultimately what I think I want with this feature is that teachers can set it for a class and for a level in their teacher's interface (maybe you can make an issue for that?)

I will change the value! It indeed would be really nice to add this as a feature to the teacher's interface. If I recall there isn't any class customize page (yet)? I will create an issue and can change the class overview structure to support this change.

TiBiBa:
I just changed the threshold to 20 with the last commit! For the idea of customizing through class I still have some design issues we have to discuss. What if a student is in multiple classes? We should allow student to select an "active" class for which they receive the customizations. Next I think it would be best to create a new table in the database, being dependent on the class id. Allowing us to store all customizations (which in turn will have to be loaded as maybe sessions values so we have access to them from every page?).

Felienne:
> What if a student is in multiple classes? We should allow student to select an "active" class for which they receive the customizations. 

I love that you are thinking of edge cases but I suspect this is not likely to happen so I would not worry about it (we coud for example pick the last joined one as the active one by default)

> Next I think it would be best to create a new table in the database, being dependent on the class id. Allowing us to store all customizations (which in turn will have to be loaded as maybe sessions values so we have access to them from every page?).

This sounds like a good plan (and of course I know you are working on that already)

Felienne (approved):

Issue #1182, opened 2021-11-06
[BUG] Sign-up is broken (not because of email sending)


boryanagoncharenko:
**Describe the bug**
Sign-up does not work (not because the welcome email fails to send). The root cause seems to be that this commit: https://github.com/Felienne/hedy/commit/744bfbbf569a6149a8c27c662e44e9e7f3f1663c removes the "auth" class of the form which is used as in auth.ts

Currently, an attempt to sign up gives the error "Please enter a username" because no values of the form are passed to the front-end validation.

**Expected behavior**
The sign-up does pick up the username and all fields of the form and validates them correctly.

TiBiBa:
I think it is bad practice to make the javascript form handling dependent on a classname. Would be better to use an `id` for this as classes are supposed to be used by multiple elements. I removed the class in the commit because also some form styling was created in `styles.css` on the `auth` class.

boryanagoncharenko:
I agree and I wanted to make the minimal fix that will unblock me. Do you want to take this over and address it the way you believe it is best? I see you are making many cool changes to the front-end.

Felienne:
Fixed by #1183

PR #1183, opened 2021-11-06
Fixed broken sign-up #1182

boryanagoncharenko:
Addressed issue with sign-up.

Fix for #1182

To check it is working:
1. Run Hedy locally and try to sign-up by filling the sign-up form correctly
2. You should NOT see the error "Please enter a username". Instead you will see the "There was an error, please try again" error because Hedy cannot send emails locally.


rix0rrr (dismissed):

rix0rrr:
Approved, but I see in the issue discussion that @TiBiBa might want to do some more deep fixes.

Leaving up to you all to decide whether or not to proceed with this or not.

boryanagoncharenko:
Changed the form selection from class to id as @TiBiBa suggested. Shall we merge @TiBiBa or is there further refactoring that you would like to do?

rix0rrr (approved):

rix0rrr (approved):

Issue #1178, opened 2021-11-05
Exception parameters are used inconsistently


rix0rrr:
(This is a refactoring/cleanup task.)

----------------

The following occurs side-by-side, and may get confusing to contributors:

```py
class ParseException(HedyException):
    def __init__(self, level, location, keyword_found=None, character_found=None):
        super().__init__('Parse')
        self.level = level
        self.location = location
        self.keyword_found = keyword_found
        self.character_found = character_found

class UndefinedVarException(HedyException):
    def __init__(self, **arguments):
        super().__init__('Var Undefined', **arguments)
```

There are differences in two dimensions:

## Parameters

Are they enumerated (`level, location, keyword_found, character_found`), or do we just implicitly accept everything that's given (`**arguments`) ?

The former is probably better against accidental misuse, while the second is less typing 🤗.

It would *probably* be a good idea to use the style of the former to protect against accidents, while potentially even adding in the concept of *mandatory* keyword arguments:

```py
    def __init__(self, *, level, location, keyword_found=None, character_found=None):
```

The `*` in there makes it so that the arguments *can not* be passed positionally, they *must* be passed as keyword arguments. So:

```py
ParseException(5, [0, 1], 'xyz', 'abc')        # Not allowed

ParseException(level=5, location=[0,1], keyword_found='xyz')  # Must be this
```

The mandatory keyword arguments make it hard to accidentally swap two positional arguments around: it's very clear what you are doing.

## Storage

Are arguments passed up to the superclass, or stored on the exception object itself?

If they get passed to the superclass, then they all get stored in the `ex.arguments` dictionary. This dictionary is used in `hedy_error_to_response` to pass all the arguments into the error formatting strings found in `coursedata/texts/$LANG.yaml:HedyErrorMessages`.

That's very convenient, and the uniform mechanism removes the need for helper functions like `parse_error_to_response`.

At the same time, `hedy_error_to_response` also looks for an attribute called `location` on the exception and copies it into the response object if it exists, so that the client can use it to render error locations into the editor.

So there's no getting around either having:

```py
        self.location = location
```

In the exceptions for which `location` makes sense, OR treating the `location` argument specially in the `HedyException` class (and storing it in a location where `hedy_error_to_response` will be able to find it back).

I don't have a super strong opinion here. I think the following:

```py
class ParseException(HedyException):
    def __init__(self, *, level, location, keyword_found=None, character_found=None):
[...]

PR #1184, opened 2021-11-06
fix: higlight error location in line

rix0rrr:
This is a resubmission of #1147, which had done most of the heavy lifting
to figure out what needed to happen already but had some unnecessary
changes in there as well.

The `Location` field in the error response already contained both row
and column number, so good enough to achieve column-specific
highlighting.

To support `InvalidCommandException`, which *can* only have a row and
not a column (since we don't have that information available), we
introduce the convention that `Location` can be either `[row, col]` or
just `[row]`. The client will highlight the entire line or just a single
character.

(I added a helper future-proofed a bit by adding a parameter so that
once we want to highlight whole words, we can do that too).

There is a bit of a short-term hack in `InvalidCommandException`,
which should be tackled as part of addressing #1178.


rix0rrr:
Hi @dolfein, thanks for all the heavy lifting you did in your PR figuring out where all the pieces need to go! I agree with most of what you did, but I thought we could achieve the same thing a bit simpler, with fewer overall changes to the Python and the server<-->client protocol.

Right now this PR highlights errors in one of two ways:

We know the column:

image

We do not know the column:

image

Felienne (approved):

Felienne:
Hi @rix0rrr!

This works but t seems Ace is a bit confused when an error occurs, and gives us a weird square attached to the mouse cursor:

image

Not sure if that can be removed?

bjorn3:
The weird square is what I talked about in https://github.com/Felienne/hedy/pull/987#issuecomment-941173195. This is where the error message would be if one was given to ace.

Issue #1135, opened 2021-10-29
[BUG] left, right, number could be represented nicer


Felienne:
**Describe the bug**
The allowed types could be formatted prettier: left, right, number -> left, right or a number
This can be done with the new translate code I think:

https://github.com/Felienne/hedy/blob/158ede3e23776808c3c76ec41165c7407a957ecf/app.py#L363

We then also want to localize to -> left, right of een getal but that can be done in one go.

**Paste the Hedy code & level**
`turn right 90`

Level: 2

**Add a screenshot (optional)**
![image](https://user-images.githubusercontent.com/1003685/139476509-8f82a3e1-5904-4f24-9ae2-f545594851d6.png)

Felienne:
Closed by #1186

PR #1186, opened 2021-11-06
Improve error message when an list needs to be translated #1135

boryanagoncharenko:
Improve error message when an list needs to be translated #1135
- left, right, number -> left, right or a number
- left, right, number -> left, right of een getal

**Description**

**Fix for**

#1135 

**How to test**

Try the following commands on levels 1 and 2 and in English and Nederlands:
`turn lalala`
`forward lalala`
The expected response is "left, right or a number" instead of "left, right, number"


Felienne (approved):
Works!! Lovely nice error message!

image

Issue #1179, opened 2021-11-05
Make template globals in Flask consistent


rix0rrr:
(This is a cleanup/refactoring task)

---------------

We have a number of things we inject into the Flask templates (language, translations, some other things) in various ways.

Turns out that Flask has nice mechanisms to make certain functions and variables available to their templating mechanism, so we didn't need to do all the passing and copying that we ended up doing. 

There are some good opportunities for cleanup in this area.

Felienne:
Thanks for the suggestion @rix0rrr, do you have some pointers to these mechanisms, so we can help people get started?

rix0rrr:
Here are some recent issues I ran into trying to convert the Quiz functions to a [Flask Blueprint](https://flask.palletsprojects.com/en/2.0.x/tutorial/views/), which is currently really hard because the handlers rely on all kinds of global variables:

```py
        return utils.page_404 (TRANSLATIONS, render_main_menu('adventures'), current_user(request) ['username'], requested_lang (), 'Hedy quiz disabled!')
```

* `render_main_menu`
* `currrent_user`
* `TRANSLATIONS`

And the `g` object:

```py
        g.lang = lang = requested_lang()
        g.prefix = '/hedy'
```

And `is_teacher`:

```py
                                   is_teacher=is_teacher(request),
```

These things occur everywhere a template is rendered, but that is not necessary.


We can make all of these automatically available in templates, without constantly passing them, by using something like:

* [Flask Context Processors](https://flask.palletsprojects.com/en/2.0.x/templating/#context-processors)
* [@app.template_global](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.template_global)

dvberkel:
I will look into this

dvberkel:
The arguments that are always passed around, such as `current_user`, `TRANSLATIONS` etcetera as mentioned by @rix0rrr, will by made available to the template via [context processors](https://flask.palletsprojects.com/en/2.0.x/templating/#context-processors).

In short 
> a context processor injects variables automatically

dvberkel:
At the moment, when user information is passed to a template, the `current_user` function is used. Under the covers the database is used to retrieve the token associated with this session and the currently logged in user. Some predicates on the user, such as `is_admin` and `is_teacher` are calling the `current_user` function. This possible results in multiple calls to the database for each request.

[...]

PR #1188, opened 2021-11-07
User information no longer necessary in `render_template` calls

dvberkel:
This pull-request encompasses the following improvements:

* It places the user on the Flask session.
* It automatically injects user information in the context for templates.
* It refactors `is_teacher` and `is_admin` to accept a user instead of a request.

Below you can find a detailed motivation for these changes.

This pull-request uses a [_context processor_](https://flask.palletsprojects.com/en/2.0.x/templating/#context-processors) to inject user information, such as `username`, `is_teacher` and `is_admin` into the context automatically. This relieves developers from passing along this information with each call to `render_template`.

The context processor reads the user information from the user on the [Flask session](https://flask.palletsprojects.com/en/2.0.x/api/#flask.session). The user is placed on the session when they login and is removed when they user logs out or when they are destroyed.

The `is_teacher` and `is_admin` function accepted a request. This request was used to retrieve the current user from the database. This is an example of a function with side-effects, which should be avoided. Therefore the functions are changed to accept a user instead.

This pull-request is related to issue #1179. It removes a few of the globals that hamper progress on that front. 

## How to test
Since this is a refactoring, the application should work as before.


website/auth.py
```diff
@@ -113,6 +113,8 @@ def login():
         if not check_password(body['password'], user['password']):
             return 'invalid username/password', 403
 
+        session['user'] = user
```
rix0rrr, line 116: This should become a slice of the data.

I would prefer to automatically invalidate `is_teacher` (based on time).

app.py
```diff
@@ -231,6 +231,12 @@ def check_language():
 def before_request_begin_logging():
     querylog.begin_global_log_record(path=request.path, method=request.method)
 
+# A context processor injects variables in the context that are available to all templates.
+@app.context_processor
+def enrich_context_with_user_info():
+    user = session.get('user', {'username': '', 'email': ''})
```
rix0rrr, line 237: This object should be standardized somewhere, probably close to the `copy_data_from_database_user` function...

dvberkel: As discussed:

We will repurpose the `current_user` function to return session user.

website/auth.py
dvberkel: The following slice of the data is placed in the current session
* `username`
* `email`
* `is_teacher`

dvberkel: I have implemented a time-to-live for the current user

website/auth.py
```diff
@@ -35,26 +35,52 @@ def hash(password, salt):
...
+# The main purpose of the current user is to provide a convenient container for
+# * username
+# * email
+# * is_teacher
+#
+# Since the is_teacher can be changed during a session we also store a time-to-live. When retrieving the current user, we can check if we need to reload data from the database.
+#
+# The current user should be retrieved with `current_user` function since it will return a sane default.
+# You can remove the current user from the Flask session with the `forget_current_user`.
+def create_current_user(db_user):
```
rix0rrr, line 48: I would expect a function called `create_` to return something, but it has a side effect. Let's use a different verb?

```suggestion
def remember_current_user(db_user):
```

website/auth.py
```diff
@@ -35,26 +35,52 @@ def hash(password, salt):
...
+# If the current user is to old, as determined by the time-to-live, we repopulate from the database.
+def current_user():
+    now = times()
+    user = session.get('user', {'username': '', 'email': ''})
+    ttl = session.get('user-ttl', None)
+    if ttl == None or now >= ttl:
+        username = user['username']
+        if username:
+            db_user = DATABASE.user_by_username(username)
+            create_current_user(db_user)
```
rix0rrr, line 66:
```suggestion
            remember_current_user(db_user)
```

website/auth.py
```diff
@@ -113,6 +139,8 @@ def login():
         if not check_password(body['password'], user['password']):
             return 'invalid username/password', 403
 
+        create_current_user(user)
```
rix0rrr, line 142:
```suggestion
        remember_current_user(user)
```

rix0rrr (changes requested):

website/auth.py
dvberkel: It puts a tear in my eye :man_cook: :kiss:

rix0rrr (dismissed):

rix0rrr (approved):

PR #1189, opened 2021-11-07
2 Confetti bug fixes

fRedelaar:
**Description**

Fixed 2 bugs regarding the confetti and green error box. In my last draft PR #1175 I mentioned two bugs. The green error box wouldn't hide after changing tabs. Also, if the confetti emoji was clicked once, it wouldn't appear again. These bugs are fixed in this PR. Also made tiny adjustments so its in line with tailwind, only meant for consistency. 

**Video**

https://user-images.githubusercontent.com/33595265/140640874-27571fcd-b003-4e1a-afef-194b17db820f.mov

**Possible future fixes**
If there is a request, the confetti could be adjusted slightly to be smaller/somewhere else on the page for example.


Felienne:
> Fixed 2 bugs regarding the confetti and green error box. In my last draft PR #1175 I mentioned two bugs. The green error box wouldn't hide after changing tabs. 

Thanks for the fixes @fRedelaar!

> Also, if the confetti emoji was clicked once, it wouldn't appear again. These bugs are fixed in this PR. Also made tiny 
> adjustments so its in line with tailwind, only meant for consistency.

Haha I did not realize that was a bug, we in fact liked that because otherwise we suspect kids would continue to click the button. Once per level would be good I think (or maybe in a few weeks when the gradual feedback model is further along, after a series of failed programs). Could you revert that?

Felienne (approved):
Oops accidentally approved

Felienne (changes requested):
See above

fRedelaar:
> Haha I did not realize that was a bug, we in fact liked that because otherwise we suspect kids would continue to click the button. Once per level would be good I think (or maybe in a few weeks when the gradual feedback model is further along, after a series of failed programs). Could you revert that?

Oh haha! I misunderstood something because I thought the confetti emoji would be limited once per adventure in a level :D. I will revert that back to limit the confetti emoji to once per level.

Felienne:
> > Haha I did not realize that was a bug, we in fact liked that because otherwise we suspect kids would continue to click the button. Once per level would be good I think (or maybe in a few weeks when the gradual feedback model is further along, after a series of failed programs). Could you revert that?
> 
> Oh haha! I misunderstood something because I thought the confetti emoji would be limited once per adventure in a level :D. I will revert that back to limit the confetti emoji to once per level.

Ow that actually makes sense! Let's keep it at once per level.

Felienne (approved):

fRedelaar:
Cool! Thank you for the extremely fast replies and feedback @Felienne ! Please let me know if anything needs to be changed.

PR #1191, opened 2021-11-07
New level 14

Felienne:
Following the discussion on #903 

As explained there, I am not sure this is not too hard for kids and if we can come up with good content but let's see what @MarleenGilsing can imagine with these concepts!


coursedata/yaml_transformer.py
```diff
@@ -40,25 +51,25 @@ def transform_adventures(transformations):
...
-def transform_levels_in_all_YAMLs(transformations):
+def transform_levels_in_all_YAMLs(old_level, new_level=None, function=nop):
   # Set the current directory to the root Hedy folder
   os.chdir(os.path.join(os.getcwd(), __file__.replace(os.path.basename(__file__), '')))
-  transform_level_defaults(transformations)
-  # transform_adventures(transformations)
-
-
+  transform_level_defaults(old_level, new_level, function)
+  # WARNING: adventure do not properly use the transformer fucntion yet!
```
dvberkel, line 68:
```suggestion
  # WARNING: adventure do not properly use the transformer function yet!
```

dvberkel (changes requested):
There is a typo in a comment

dvberkel (approved):
You are on :fire:

PR #1192, opened 2021-11-07
fix: quiz is broken if user is not logged in

rix0rrr:
If the user is not logged in, the username is `''`.

First of all, DynamoDB rejects this value with an error.

Second of all, if we would record all anonymous users under the same
key, the data partition would become too big.

Instead, we'll record the responses under a username like
`'anonymous:e91f5ac2a6c946e893683c5c02781117'`.

We also store a timestamp in seconds in the database (instead of milliseconds). The resolution is good enough.


dvberkel (approved):
:lady_beetle:

PR #1194, opened 2021-11-07
refactor(exceptions): clean up exceptions

### LINKED ISSUE #1178: Exception parameters are used inconsistently

rix0rrr:
An effort to clean up the exception handling. Contains the following
changes:

* All exceptions now live in a separate file, `exceptions.py`, and in
  the code must be referenced like `[hedy.]exceptions.ParseException`.
* Document the purpose of `arguments`.
* Make the formatting arguments explicit for each type of exception.
  This makes it clear what values are expected for each.
* Remove the different way `ParseException` and `InvalidSpaceException`
  record their parameters, and made it the same as other exceptions.
* Remove the `parse_error_to_response` function: it can easily be
  rendered like any other exception.

`InvalidSpaceException` was the odd one out, in that it's not really an
error condition: it functions more of a quick escape from some deep
layers of recursion, but ultimately doesn't lead to an error response,
just to a warning.

To get rid of the duplication in `invalid_space_error_to_response`,
this PR introduces the new concept of an `FtfyException` (Fixed That For You
exception), which gets caught directly after calling `transpile_inner`,
and gets treated as a "success with a warning" result.

Fixes #1178.


Felienne (approved):
Thanks @rix0rrr!

tests/tests_level_07.py
```diff
@@ -127,7 +127,7 @@ def test_allow_space_after_else_line(self):
     a is 1
     if a is 1
       print a
-    else   
+    else
```
bjorn3, line 130: This test was testing that spaces after `else` work.

PR #1195, opened 2021-11-07
Splitting level1.lark file

teundemast:
The keywords, which were originally placed in level1.lark, can now be accessed from keywords-en.lark and keywords-nl. This can be used to add multiple languages for the keywords.


hedy.py
```diff
@@ -1367,7 +1367,8 @@ def merge_grammars(grammar_text_1, grammar_text_2):
 def create_grammar(level):
     # start with creating the grammar for level 1
     result = get_full_grammar_for_level(1)
-
+    keys = get_keywords_for_language("en")
+    result = merge_grammars(result, keys)
     # then keep merging new grammars in
     for i in range(2, level+1):
         grammar_text_i = get_additional_rules_for_level(i)
```
bjorn3, line 1374: The grammar saving below will need to differentiate between keyword languages I think.

bjorn3: Also I think the grammar cache needs to consider keyword language.

Felienne: Yes @bjorn3 is right, the cache now indexes with level number only, that will need to be updated to level + lang (for example: "en-1" and "nl-7")

teundemast: Thanks for your feedback! I have now made a commit ('Resolving comments on pull request') which hopefully resolves these issues. If not, please let me know.

teundemast:
> The `save_total_grammar_file` call still only uses the level and not the language. Other than that I think it looks fine.

Ah totally missed that! With the last commit, that should be fixed as well. But now I have no idea why it doesn't pass all tests... When I run the tests locally, all tests are passed. Do you have any idea @bjorn3 why that is the case?

bjorn3:
I think that is either a spurious error or one of our indirect dependencies published a version with a breaking change.

Felienne:
> I think that is either a spurious error or one of our indirect dependencies published a version with a breaking change.

Looking into the test output, I see this:

```
/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/botocore/httpsession.py:10: in <module>
    from urllib3.util.ssl_ import (
E   ImportError: cannot import name 'OP_NO_TICKET' from 'urllib3.util.ssl_' (/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/urllib3/util/ssl_.py)
```

You can also see this output if you go to Checks and then to build (https://github.com/Felienne/hedy/runs/4144964190?check_suite_focus=true)

My suspicion is that things fail because you use Python 3.9.7 instead of 3.9.5 as we specify in the runtime, but I am not sure, this error is new to me too.

Felienne:
> > I think that is either a spurious error or one of our indirect dependencies published a version with a breaking change.
> 
> Looking into the test output, I see this:
> 
> ```
> /opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/botocore/httpsession.py:10: in <module>
>     from urllib3.util.ssl_ import (
> E   ImportError: cannot import name 'OP_NO_TICKET' from 'urllib3.util.ssl_' (/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/urllib3/util/ssl_.py)
> ```
> 
> You can also see this output if you go to Checks and then to build (https://github.com/Felienne/hedy/runs/4144964190?check_suite_focus=true)
> 
> My suspicion is that things fail because you use Python 3.9.7 instead of 3.9.5 as we specify in the runtime, but I am not sure, this error is new to me too.

Hmmmm maybe it was a caching issue because another run of the tests fixed it! Will try to approve later today.

teundemast:
> > > I think that is either a spurious error or one of our indirect dependencies published a version with a breaking change.
> > 
> > 
> > Looking into the test output, I see this:
> > ```
> > /opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/botocore/httpsession.py:10: in <module>
> >     from urllib3.util.ssl_ import (
> > E   ImportError: cannot import name 'OP_NO_TICKET' from 'urllib3.util.ssl_' (/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/urllib3/util/ssl_.py)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can also see this output if you go to Checks and then to build (https://github.com/Felienne/hedy/runs/4144964190?check_suite_focus=true)
> > My suspicion is that things fail because you use Python 3.9.7 instead of 3.9.5 as we specify in the runtime, but I am not sure, this error is new to me too.
> 
> Hmmmm maybe it was a caching issue because another run of the tests fixed it! Will try to approve later today.

I think that was indeed the problem. Thank you!

bjorn3 (approved):

PR #1204, opened 2021-11-08
chore: re-introduce dropped space in unit test

rix0rrr:
In #1194, I accidentally dropped a space from the end of a line in
a unit test.

Significant trailing whitespace in a source file feels dangerous.  It is
invisible, many style checkers will complain about it, and personally I
have my editor set to always strip trailing whitespace at the end of
lines when I save my files (hence this change).

This change re-introduces the space in the code that is being tested,
without relying on literal trailing whitespace in the source code.


Felienne (approved):

Felienne:
Thanks @rix0rrr, this version indeed makes it a lot more clear what the test is there

Issue #1203, opened 2021-11-08
Check that we generate valid Python


Felienne:
In the current transpile and test setup, we never check that we generate Python that is valid. Of course this should not happen "by construction" but sometimes is does, leading to "bad input" on the client side. 

Things we could to do to make the chance of that lower:

~~1. When running tests, validate that the generated Python is valid by executing run_code. This can be integrated into the multi-level tester as a start?~~
~~2. When testing adventures and level defaults, validate that the generated Python is is valid by executing run_code~~
3. Maybe even add an exception in typescript that executes the Python and gives a nicer error message? This is expensive but might be worth it?

Felienne:
Points 1 and 2 are now taken care of (see PR's linked above)

I am doubting if we should implement 3 on the typescript side, for performance reasons... Do you have a feeling for that @boryanagoncharenko?

Felienne:
Let's close this, 1 and 2 are fixed and 3 will really be expensive (we already have performance issues as it is now)

PR #1206, opened 2021-11-08
Runs programs in validator to detect programs that return invalid Python

Felienne:
Addresses idea 2 of #1203


Felienne:
Note that #1205 needs to be merged before this one since it fixes a program that we now test (`if eten is 'pizza'`)

Issue #1209, opened 2021-11-08
[BUG] Unindented second block gives ugly error


Felienne:
Unindented second block gives ugly error:

![image](https://user-images.githubusercontent.com/1003685/140756659-669448a8-9806-4ba7-beba-324aff9b2422.png)

For example in level 10. Code:

```
for x in range 1 to 10
  for y in range 1 to 10
  print 'x*y'
```

PR #1211, opened 2021-11-08
Fix for 1209

Felienne:
We have end-blocks in the grammar, but not start-blocks and I have wanted to implement that for a while, but I never got around to it. Which was maybe a good idea, because now we have the Indentation Exception, it is actually nicer to throw that instead of relying on ugly parse errors :)

There is one nasty hack in here, will  make a separate issue for that


Felienne:
Fixes #1209, obviously

PR #1217, opened 2021-11-09
Fixes Issue1996: if and elses might be separated by newline

racertop:
This PR is a partial fix for issue #1996. This way we can accept this type of code in levels 4, 5 and 6

**Description**

I changed the `ifelse` rule in the level 4 grammar so it's similar to the `ifs` rule in  the same grammar.

**Fix for**
This fix is so we can accept this type of code.
```
if condition
command
else
command
```
Notice that there aren't any leading spaces in this code, just separated by newlines.

**Why a draft**
I'm having some issues making it work. It works fine enough when I'm debugging and the parser is not cached, but once it's cached it says invalid in some cases. The issue lies in this line of code:
When it falls in that if, something happens that my grammar is not valid anymore, as opposed to when is build from scratch. 
image

 When there's a mathematical operation, it doesn't work either, which is something that also happens with the current ifs rule that accepts newlines.
```
people is 1, 2, 3, 3
test is 1
if dishwasher is test
else
print 'luckily no dishes because' dishwasher + 1 'is already washing up'
```
All the help you can provide will be much appreciated.
**How to test**

You can use this code in levels 4, 5 or 6. Or anything that resembles it:
```
if dishwasher is Sophie
else
print 'luckily no dishes because' dishwasher 'is already washing up'

```

**Checklist**


Felienne:
Hi @racertop! 

I think I managed to fix your issue by disallowing else in the invalid clause. That way it will never match and we always match with the real actual else. In a near future, we might be able to parse ambiguously so this type if hacking is no longer needed :) 

Hope this helps!

racertop:
Hi @Felienne, thank you so much for your help! I think it's ready now, I edited the `invalid` rule a bit further disallowing the if else too, because the same problem happened with this type of programs:
```
if condition
statement
```

Felienne:
Awesome work @racertop!! This is really going to help kids a lot. Let me know when this is ready to be merged.

racertop:
Thank you, @Felienne! I think it's ready for merge right now, I added a few test too so you can check them out!

Felienne (approved):

Felienne:
Thanks again @racertop for your hard work on this!

PR #1248, opened 2021-11-12
Translate new levels and fix endless loop in 13

oscarkarnalim:
Translate level 8,9,and 11, also fix endless loop in 13


Felienne (approved):

Felienne:
Thanks @oscarkarnalim!!

Issue #1252, opened 2021-11-12
Recieving an error when parsing anything in languages other than English or Dutch


teundemast:
As a result of adding different languages for keywords, it is not possible to use keywords when the keywords for that language are not available yet. For now, it is probably the best solution to accept English keywords for that language.

Issue #915, opened 2021-10-08
Allowing one program to use keywords in different languages (f.e. mix als (nl) with repeat (en)


HuibSprangers-leiden:
Long term goal 3.

PR #1253, opened 2021-11-12
Fix for issue #1252

teundemast:
Retrieve english keywords when keywords for a language are not availabe (yet) 

Fixes issue #1252


teundemast:
Because of recent commits, the language was not passed as an argument into the transpiler. This is fixed with the most recent commit and will be fixed if this pull request is accepted.

Felienne (changes requested):
Hi all! In principle this works, but if we merge this, in the Duthc version, only Dutch programs can be parsed (f.e. only `drukaf hallo`) that is not ok for now because we do not have all other things in place (distance algo, explanations etc)

So can we put this behind a flag? Something like

```
if local_keywords_enabled:
.....
```

teundemast:
Ah, ofcourse! We are working on #915, so English keywords can also be used when the specified language is Dutch. But for now,  with the most recent commit (made a typo in the name), I have added a flag.

Felienne:
Hi @teundemast!

The PR is quite a bit behind because so many things changed in the meantime, sorry. Can you either fix the PR or start anew from a fresh branch of of main?

teundemast:
Hi @Felienne, no problem! I will create a new PR.

Issue #1220, opened 2021-11-09
Come up with 5 succes-messages for different languages


fRedelaar:
Stuff like "Well done", "Goob job!" etc.

maximedongen:
Hi

johannadekker:
French:
- Bien fait! (Goed gedaan)
- Bon travail!
- Bravo!
- Continuez comme ça! (Ga zo door)
- Beau boulot! (Good job)

johannadekker:
Chinese: 做得好! (Well done)
Russian: Ты хорошо постарался, молодец! (You did a good job, well done!)

Issue #1218, opened 2021-11-09
Show several random succes-messages


remva:
Right now, the same 'well done' message is showed every time. Several messages could be shown randomly like 'good job' and 'perfect!'.

PR #1259, opened 2021-11-12
Five random success messages will be shown #1218 #1220

Jasper1408:
@johannadekker came up with success messages in multiple languages. (Mandarin, Dutch, Spanish, German, French, English). 

@remva and I implemented the following. A random success message should be shown from a list of five different ones. This change was made in app.ts. Here a random number is generated by a random number generator. In all the language yaml files five different success messages were added. We were unable to push yaml files and the ts files separately because they were dependent on one another.


coursedata/texts/bg.yaml
```diff
@@ -27,6 +27,11 @@ ui:
     by: "от"
     start_quiz: "Започни теста"
 ClientErrorMessages:
+    Transpile_success1: "Good job!"
```
Felienne, line 30: We do not have to add these to the yaml, because we default to English anyway if there is no text in the yaml for a certain field

coursedata/texts/en.yaml
```diff
@@ -40,7 +40,11 @@ ui:
     invalid_teacher_invitation_code: "The teacher invitation code is invalid. To become a teacher, reach out to hedy@felienne.com."
     page_not_found: "We could not find that page!"
 ClientErrorMessages:
-    Transpile_success: "Great job!"
+    Transpile_success1: "Good job!"
```
Felienne, line 43: Could we make this into a list instead:

Transpile_success = Good job!, Amazing, etc...

In yaml you can make a list like this:
Transpile_success:
      - Good job!
      - Amazing
      - ...

Felienne: That was different languages can also have different lengths of messages, so you'd need to change the code a bit too

static/js/app.ts
```diff
@@ -228,7 +228,9 @@ export function runit(level: string, lang: string, cb: () => void) {
       }
       if (response.Code){
         console.log("success!");
-        success.show(ErrorMessages['Transpile_success']);
+        var errormessageslist: string[] = ['Transpile_success1','Transpile_success2', 'Transpile_success3','Transpile_success4','Transpile_success5'];
```
Felienne, line 231: If we grab a list from the yaml, this can be a bit cleaner too!

static/js/app.ts
```diff
@@ -228,7 +228,9 @@ export function runit(level: string, lang: string, cb: () => void) {
...
+        var randomnum: number = Math.floor(Math.random() * 5);
```
Felienne, line 232: then of course the 5 here needs to be changed to

Felienne (changes requested):

Felienne:
Nice idea!! I made some small remarks

coursedata/texts/el.yaml
```diff
@@ -28,7 +28,7 @@ ui:
     commands_title: Εντολές
     by: "by"
 ClientErrorMessages:
-    Transpile_success: "Great job!"
+    Transpile_success: 'Great job'
```
Felienne, line 31: This entire line may be removed, and Hedy will then by default pick the English text

Felienne (changes requested):

Felienne (dismissed):
Well done!!

Issue #930, opened 2021-10-09
Forgotten indent could get a better error message


Felienne:
e.g. at level 7:

```
print 'kassabon'
prijs is 0
repeat 7 times
ingredient is ask 'wat wil je kopen?'
if ingredient is appel
prijs is prijs + 1
print 'Dat is in totaal ' prijs ' euro.'
```

![image](https://user-images.githubusercontent.com/1003685/136656458-223032d6-7241-4d85-98de-2cabcd609ede.png)

remva:
Whenever you start a block with 3 spaces (for example), every block should start with 3 spaces to be correct. But the explanation at the top of level 7 says: "... That means putting four spaces at the beginning of the line.". 
This means that a block can only start with 4 spaces. 

In this case, should the text at the top be changed to "start every block with the same number of spaces"? Or should the error message be changed to "Oops! You used 3 spaces instead of 4".

PR #1260, opened 2021-11-12
#930 Block should start with 4 spaces instead of None

remva:
**Fix:** Whenever you start a block with 0 spaces, part the error message that occurs in level 7 is: "You used too few spaces in line 2. You used 0 spaces, which is not enough. Start every new block with None spaces more than the line before". 'None'  is changed to '4' to make the sentence grammatically correct. This is also the indent size you learn in level 7. 
 
**Small discussion:** It is still possible to start a block with more than 0 spaces, but different than 4. For example: whenever you start a block with 3 spaces, every block should start with exactly 3 spaces. There are two changes possible:
1. The text at the top of level 7 should be changed to "start every block with the same number of spaces"
2. The error message should be changed to "Oops! You used 3 spaces instead of 4".


Felienne:
Nice catch! I changes the default behavior a bit but otherwise this is all ok!

Felienne (approved):

PR #1266, opened 2021-11-14
chore: remove unnecessary `lang` parameters

rix0rrr:
The "language" parameter should just be a global, automatically
available in every request and template. No need to set and pass
it in every Python handler and every call to `render_template()`
anymore.

Relates to #1179.


Felienne (approved):

dvberkel:
I am late to the party :partying_face:, but this looks good to me :rocket:

PR #1270, opened 2021-11-15
A new module for program repair

dolfein:
Nothing special (yet), just a new module for program repair, as requested by @Felienne in #1225.
The existing repair (removing leading spaces) is moved to this new module.


program_repair.py
```diff
@@ -0,0 +1,3 @@
+def repair(input_string):
```
Felienne, line 1: I think it is better if we rename this function into remove_leading_spaces so it is clearer what it does and that we will have other repair methods too.

Felienne (changes requested):
Good refactoring @dolfein, I just have a small suggestion for naming the repair function.

program_repair.py
dolfein: Done! I was planning on changing the name once I made another one.

Felienne (approved):

PR #1280, opened 2021-11-15
Radio button size changes in Q6 quiz level 7

LterHaar:
The radio button is moved under the option symbol. This makes more space for the option text and eliminates an problem where the radiobutton resizes

**Description**

The radio button  has been swapped with the option symbol. The tailwind CSS is adjusted with flex-col and justify-center, items-center tags.

**Fix for**

Radio button size changes in Q6 quiz level 7 #1075

**How to test**
image
 
 Is changed to this, with radiobuttons moved to underneath the option symbol.
image

### DIFF
--- a/templates/quiz_question.html
+++ b/templates/quiz_question.html
@@ -47,16 +47,19 @@
                     {% for column in row %}
                         <div class="flex flex-col">
                             {% if question.correct_answer == chosen_option %}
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Top, ziet er goed uit zo Lotus! Mooie voorbereiding ook op straks de symbooltjes ipv de letters

Issue #1109, opened 2021-10-27
Adding .lark files to the translation page


brucemartens:

teundemast:
I think the most convenient way to create a .lark file from translated keywords on the translation page, is to first create a yaml from these keywords (because there is already a structure for creating yaml files). And then, using a function which transforms a .yaml file into a .lark file. 
So my current approach is as follows.
First add a keywords folder to coursedata. Then give a translator the option to fill in the translation of keywords and download a yaml file from this. The translation page will then look as follows: 
<img width="771" alt="Capture" src="https://user-images.githubusercontent.com/36051227/141691541-987317d3-064f-4c5c-b811-534f7007e05b.PNG">
The yaml file will be structured as follows:
<img width="217" alt="Capture2" src="https://user-images.githubusercontent.com/36051227/141691738-f9a4e049-1b1b-4236-af12-37d0320ddea9.PNG">
This file will be the input for the following function (I am not sure yet when this function should be called):
<img width="465" alt="Capture1" src="https://user-images.githubusercontent.com/36051227/141691772-41667ca0-b224-4e4a-bdf1-49322900c50b.PNG">
The output will be a keywords-{lang}.lark file in the grammars folder. 

@Felienne do you think this approach is the best way to attack this? Do you foresee any roadblocks? 
I would love to have your feedback on this before implementing and possibly wasting a lot of time :)

Felienne:
This to me seems a fine approach!

Felienne:
> The translation page will then look as follows: <img alt="Capture" width="771" src="https://user-images.githubusercontent.com/36051227/141691541-987317d3-064f-4c5c-b811-534f7007e05b.PNG"> 

It might be slightly nicer if we remove the _ from the token names and name the lowercase, so print instead of _PRINT but that is not super important right now.

teundemast:
Thanks for your reaction! That will not be much hustle I think, so that's going to be all right!

Felienne:
Closed by #1310

PR #1281, opened 2021-11-15
Add Keywords to the translate page

teundemast:
**Description** 
With this pull request, it will be possible to give a translation for all the keyword tokens on the translation page. Then, these translations can be downloaded in a yaml file and can be added to coursedata/keywords/. I have added a function in the file yaml_transformer.py, which can create lark files in the grammars directory for yaml files in the coursedata/keywords/ directory. There is an option to overwrite all these lark files or just add new lark files without overwriting the other ones. The approach in the function I have added in yaml_transformer.py, differs a bit from the existing transforming functions. This is because I could not get these function to work on my device (I kept getting import errors while all packages in requirements are installed). I don't know if this happened because of my device or because of some updates in modules. Anyway, I hope that the way the 'transform_yaml_to_lark' function works is also fine. 

I have updated TRANSLATING.md accordingly. 

Fixes issue #1109 

**How to test**

- Go to http://localhost:5000/translate/en/new
- Fill in the translations for the keywords
- Download keywords/new.yaml 
- Add it to the directory coursedata/keywords
- Run yaml_transformer.py (at least the function transform_yaml_to_lark)
- If all goes well, a keywords-new.lark file should be added to the directory grammars


Felienne (changes requested):
Great PR with clear changes, and awesome that you also updated the MD files.

Just one (small) issue, if you do not translate all keywords, you get a weird yaml file:

image

Can we default to English if a keyword was not translated? Because your yaml transformer does not properly consume this version (you may also change the transformer)

teundemast:
Thanks for your feedback! I will look into that! For now, I will convert this PR to a draft and open it again when PR #1293 is accepted (this will prevent conflicts) and I have fixed that issue.

Felienne:
> Thanks for your feedback! I will look into that! For now, I will convert this PR to a draft and open it again when PR #1293 is accepted (this will prevent conflicts) and I have fixed that issue.

ok!

teundemast:
Hi @Felienne, I have changed the transformer so it will retrieve the English keyword when the translation of a keyword is not yet filled in. Due to recent commits in the main branch, changes I made with PR #1293 were erased. I don't think that was the intention, so I have put these changes in this PR.

Felienne:
Closing because of merge issues, see #1309

PR #1282, opened 2021-11-15
chore: end-to-end tests now written in unittest format

### LINKED ISSUE #1058: Rewrite e2e tests into 'unittest' format
(This is a refactoring/cleanup task.)

---------

The end to end tests currently send off a set of HTTP requests in a particular order, and each test depends on the previous one succeeding.

This has a couple of issues:

### Tests aren't very readable

It's not super obvious what a line like this does:

```py
        ['retrieve programs after deleting saved program', 'get', '/programs_list', {}, {}, 200, retrieveProgramsBefore],
```

For example, what do the `{}` in there mean? What does `retrieveProgramsBefore` do? Why is it there?

### Tests aren't self-contained

For example, the following 2 tests:

```py
        ['valid signup', 'post', '/auth/signup', {}, {'username': username, 'password': 'foobar', 'email': username + '@e2e-testing.com'}, 200, successfulSignup],
        # ...
        # 10 unrelated lines follow
        # ...
        ['valid login', 'post', '/auth/login', {}, {'username': username, 'password': 'foobar'}, 200, setSentCookies],
```

Depend on each other. The first one creates a user, and the second one tests that a login with the user that was created before. The fact that this relationship exists is not clear, however. The following would be a more readable of the same test:

```py
def test_login_works():
  # GIVEN
  user_exists(username, 'foobar')

  # THEN
  login(username, 'foobar').assert_successful()
```

(Or something similar, just making up an API here)

Because tests aren't self-contained, we also can't choose to run a single individual test--we must always run the entire suite.

### Progress and error reporting isn't standardized

If tests were in a standardized format, we could use standardized test runners like `py.test` to run individual tests, and have nicer error and progress reporting.

### The purpose of supporting code will become more clear

The  [`e2e_tests.py` file](https://github.com/Felienne/hedy/blob/main/e2e_tests.py) currently has `def getProfile1()` through `def getProfile5()`, with what looks like complicated logic inside. It's not clear what their purpose is, because the logic is far away from the actual test case that it supports.

fpereiro:
- Closes #1058
- Opening a PR for feedback, please don't merge yet.

**How to test**:
- Start a development server.
- Run `pytest -q tests_e2e.py`


fpereiro:
Hi @rix0rrr & @Felienne ! Here's a draft fot the refactoring of the E2E tests. Any and all feedback is welcome! A few notes on what's done:

- Covers most auth tests and a few programs tests.
- Includes auth helpers.
- Classes contain a group of tests (Auth, Programs).
- There's either one or two functions per endpoint. Endpoint that receive payloads have two test functions, one valid and one invalid.
- I have maintained the given-when-then structure and added comments, but let me know if I should make it more structured or explicit.

tests_e2e.py
```diff
@@ -0,0 +1,366 @@
...
+        for invalid_payload in invalid_payloads:
+            response = request ('post', 'programs', {'cookie': user ['cookie']}, invalid_payload)
+            self.assertEqual (response ['code'], 400)
+
+    def test_CreateProgram (self):
+        # We create a new user to ensure that the user has no programs
+        user = AuthHelper.assertUserIsLogged (AuthHelper.makeUsername ())
+
+        program = {'code': 'hello world', 'name': 'program 1', 'level': 1}
+        response = request ('post', 'programs', {'cookie': user ['cookie']}, program)
```
rix0rrr, line 355: We now have to repeat the cookie in every single call to `request('post')`.

If instead we put the state of the current cookies on the current object (on `self`), then we don't need to remember to do this all the time. We could be writing something like this:

```py
   def test_CreateProgram (self):

        # We can hide makeUsername() behind a descriptive name
        # This writes the user and the user's cookies to 'self.user' and 'self.cookie',
        # or some other member variable
        self.given_fresh_user_is_logged_in()

        program = {'code': 'hello world', 'name': 'program 1', 'level': 1}

        # Because `post_data` lives on `self`, it can transparently access self.cookie
        # while doing the POST request. User doesn't need to think about this anymore,
        # cannot get it wrong!
        #
        # Also proposing that `post_data` ALWAYS expects a call to return 200, that saves us
        # an assertEquals check.
        response = self.post_json('/programs', program)

        # If we want to have a response code besides 200, we can have an optional keyword
        # argument for that:
        response = self.post_json('/programs', program, expect_http_code=403)

        # Like 'self.post_json', automaticaly includes the cookie, checks the status code,
        # and reads the body as JSON.
        saved_programs = self.get_json('/programs_list')['programs']

        # Compare two lists of dictionaries
        self.assertEquals(saved_programs, [program])
```

tests_e2e.py
```diff
@@ -0,0 +1,366 @@
...
+
+# *** HEDY RESOURCES ***
+import utils
+from config import config as CONFIG
+
+# *** GLOBAL VARIABLES ***
+
+HOST = 'http://localhost:' + str (CONFIG ['port']) + '/'
+# This dict has global scope and holds all created users and their still current sessions (as cookies), for convenient reuse wherever needed
+USERS = {}
```
rix0rrr, line 21: Do we clean up these users at the end of the test? For the local database I don't care, but otherwise this would accumulate endless users on the `alpha` database and that concerns me a little.

rix0rrr (changes requested):
Hey, thanks for doing this! I like where this is going. 

I have one concern (about leaving testing crap in our database) and one opportunity, around reducing duplication as much as possible.

As a nit, PEP8 (the official style guide for Python code) [recommends `snake_case` for variable and method names](https://www.python.org/dev/peps/pep-0008/#toc-entry-32) and [no spaces before the parentheses of a function call](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements) and I see you're deviating from common style quite thoroughly. Can I ask that you change the style to match what is commonly accepted in the Python community?

rix0rrr:
Oh can you also in the same PR remove the old e2e test file and update the `validate-e2e` script to run the new tests, so that we can see the tests pass in the PR check?

Felienne:
I fixed the spaces because otherwise I am unable to read the code properly.

Felienne:
It is n ok start Federico but it missed the #GIVEN, #WHEN, #THEN style. Maybe it was not clear but the idea is to put those words in the tests literally. I have rewritten one tests in that style and I hope you see that this is not just fluff but helps to draft the tests. 

For example, in the one I wrote the final line is not an assert, so why it is there? What is the #THEN there? Is this setup for another test. Using the right structure helps design the tests.

Also of course many tests still need to be added.

Felienne:
Also, I am still getting issues while running this:

image

But I am not sure that is still because it is WiP

fpereiro:
Hi @rix0rrr and @Felienne ! Thank you for the detailed review. I agree on all counts. Going point by point:

@rix0rrr 
- I will iterate the users at the end of the test and remove them, so they don't build up on the database. This is very important indeed when we run against Dynamo.
- I will follow the PEP8 code style to keep things as readable and standard as possible.
- I will take your suggestion of using the current cookie through `self` so we don't have to send them explicitly every time we make a request. Great point.
- I'll make it so that this test is run together with the other ones. If you don't mind, I'd like to keep the old file until the end to have it as a reference, since I'm rewriting the tests but keeping an eye on the old file for some patterns.

@Felienne 
- Thank you for the clarification regarding #GIVEN, #WHEN and #THEN. I will be sure to add that to the rest of the tests, taking your examle.
- I just added the remaining tests for auth, but we're still missing some tests for programs, and all the tests for classes. If needed, I can even go further and add tests for endpoints that we haven't tested yet (like admin or quizzes).
- The tests were actually passing on my end. There was a tiny issue with the change password function that tripped up the rest of the following tests. I fixed it in the latest merge commit. All tests should be passing now. As a rule, I won't push commits to this branch that have broken tests, I'd rather leave them out so that the suite presented here always passes :)

I'll be working first on adding all the above changes, and then continue with the programs & classes E2E tests.

Thanks again for the feedback!

fpereiro:
Hi there! I just added the following changes:

- Add GIVEN/WHEN/THEN comments.
- Conform to PEP8 style.
- Clean up users after all tests.
- Run tests as part of the validate-e2e script and remove old E2E script.

@rix0rrr I still have to implement the following changes:
- put cookie in `self`, also user
- Define and use `self.post_json` and `self.get_json`
- Assume 200, pass `expect_http_code` as optional

After we're all 100% agreed on the code style & approach, I can proceed with the remaining tests.

Felienne:
Looks a lot better this way, thanks! Feel free to finish those final changes.

tests_e2e.py
```diff
@@ -0,0 +1,603 @@
...
+        # WHEN signing up a new user
+        response = request('post', 'auth/signup', {}, body)
+
+        # THEN receive an OK response code from the server
+        self.assertEqual(response['code'], 200)
+        # THEN receive a body containing a token
+        self.assertIsInstance(response['body'], dict)
+        self.assertIsInstance(response['body']['token'], str)
+
+        # THEN Store the user and its token for upcoming tests
```
Felienne, line 175: I think technically this is not a #THEN because it does not concern the testing. Mayvbe we can use #FINALLY here (or is there another standard @rix0rrr?)

fpereiro:
HI @rix0rrr & @Felienne !

I just implemented the last requested changes. We now use `self.user`, `self.cookie`, `self.get_data` and `self.post_data`, as specified by @rix0rrr 

Putting the assertion for the return codes in `get_data` and `post_data` definitely saves some code and looks nicer - in some cases, it compacts the `WHEN` and `THEN` statements, but I don't think that is a problem.

I also added the `FINALLY` statements as well.

build-tools/github/validate-e2e
```diff
@@ -37,7 +37,7 @@ if ! $up; then
 fi
 
 # Finally!
-if ! python e2e_tests.py --endpoint http://localhost:$PORT; then
+if ! pytest -q tests_e2e.py; then
```
rix0rrr, line 40: How do we pass in the endpoint now? It would still be beneficial to be able to run these tests against alpha, for example.

If nothing else, we can use an environment variable.

tests_e2e.py
```diff
@@ -0,0 +1,780 @@
...
+    def test_get_admin_page(self):
+        # WHEN attempting to get the admin page
+        # THEN receive an OK response code from the server
+        # (Note: this only happens in a dev environment)
+        self.get_data('/admin')
+
+class TestSessionVariables(AuthHelper):
+    def test_get_session_variables(self):
+        # WHEN getting session variables from the main environment
+        body = self.get_data('/session_main')
```
rix0rrr, line 196: Lovely 👍

tests_e2e.py
```diff
@@ -0,0 +1,780 @@
...
+            self.cookie = self.user['cookie']
+        else:
+            self.cookie = None
+
+    def post_data(self, path, body, expect_http_code=200, no_cookie=False, return_headers=False):
+        headers = {}
+        if no_cookie == False and hasattr(self, 'cookie') and self.cookie:
+            headers['cookie'] = self.cookie
+
+        response = request('post', path, headers, body)
```
rix0rrr, line 160: Should we not retrieve the cookie from the response as well and set it on `self.cookie` every time if applicable?

You are now assuming that cookies are ONLY set by the http calls you expect to set cookies (login calls etc). If we change the server in the future, those assumptions might not be true anymore, and the tests would not behave the same as a real browser would anymore (and so they wouldn't be testing the right thing).

In fact, this already happens today: the quiz pages liberally change the session variables, which get stored in a cookie. So if we don't constantly update the cookies anytime the server says it wants to, we wouldn't be able to test the quiz properly.

`requests` has cookie jar helpers that make this trivial to do, see the previous iteration of the e2e tests where I made the same change already.

tests_e2e.py
```diff
@@ -0,0 +1,780 @@
...
+        # FINALLY remove user since it has already been deleted in the server
+        USERS.pop(self.username)
+
+# *** CLEANUP ***
+
+# We delete all the test users we created during the tests.
+# For this purpose, we use a pytest fixture. This requires us to use the `request` variable, without any possible renaming.
+# For this reason, we must rename our `request` function to `Request` so it will be referenceable from within the fixture.
+Request = request
+@pytest.fixture(scope='session', autouse=True)
```
rix0rrr, line 773: This is pytest-specific. If that is necessary, then I guess it will just have to do but for flexibility I was hoping to keep the code restricted to `unittest` if possible.

Does [tearDownModule](https://docs.python.org/2/library/unittest.html#setupmodule-and-teardownmodule) not work?

fpereiro:
Hi @rix0rrr -- thank you for the suggestions!

I added a cookiejar and removed the manual cookie setting. Great suggestion - I was actually having trouble with session variables because of this, and it was gone thanks to the cookie jar. I also used the `TearDown` method as well to clean up the users at the end of the test.

I added further tests for classes. I am having an issue on the last test because of the TTL of the session - basically, a user is made a teacher and the session is not updated for five minutes. I'll submit a patch in the next commit that reduces the TTL to 0 when receiving a test request.

I will also add back the parameters for endpoint and concurrent tests - I'm saving that for last.

Thanks!

rix0rrr:
>  basically, a user is made a teacher and the session is not updated for five minutes.

Ah yeah, true 😀. Well a workaround for that is logging out and in again. I'll make it do that.

rix0rrr (approved):
Thanks for the effort Federico! 

Made some final tweaks and I'm merging it now!

PR #1286, opened 2021-11-16
To remove the indenting of the code blocks, span is removed around th…

LterHaar:
…e <code> tag

**Description**

The span tag is unnecessary and creates a indenting for the code, so is removed.

**Fix for**

Code blocks always start intended, which can be confusing for childeren. 

**How to test**

Before 
image

After

image


Felienne (approved):
Looks good @LotusterHaar!!

Issue #1198, opened 2021-11-08
A bulb icon should occur when a program ran but is fixable by Hedy itself


fRedelaar:
This has been completed in branch:
https://github.com/Felienne/hedy/tree/hint-icon

fRedelaar:
Video that shows the bulb icon
https://user-images.githubusercontent.com/33595265/140723524-fae38901-ce54-4794-aad3-f4b60c73feed.mov

PR #1287, opened 2021-11-16
A bulb icon appears with a modal that contains an animation of the fixed code

fRedelaar:
**Description**

This PR is meant for #1198 . Whenever users make mistakes in their code, Hedy often gives a good error notification to help users. Errors such as "Did you mean `print`?" when a user accidentally wrote `pint` for example. This PR contains a bulb icon that appears whenever a user makes mistakes such as keyword spelling mistakes or using too many spaces. After clicking on the **bulb icon**, a **modal** appears (with a read-only ACE editor) that contains an **animation** of the users **repaired code**.


hedy.py
```diff
@@ -1621,7 +1621,12 @@ def transpile_inner(input_string, level):
             raise exceptions.UnsupportedFloatException(value=''.join(invalid_info.arguments))
         else:
             invalid_command = invalid_info.command
-            closest = closest_command(invalid_command, commands_per_level[level])
+            closestmatch = closest_command(invalid_command, commands_per_level[level])
+            closest = style_closest_command(closestmatch)
+            if closestmatch:
+                fixed_code = input_string.replace(invalid_command, closestmatch)
+                if fixed_code != input_string: #only if we have made a successful fix
+                    result = transpile_inner(fixed_code, level)
```
bjorn3, line 1629: If the "repaired" code has a problem too, the problem for the "repaired" code will be thrown as exception by this `transpile_inner`. You will need to catch it so you can still report the original problem.

hedy.py
```diff
@@ -1622,6 +1627,17 @@ def transpile_inner(input_string, level):
         else:
             invalid_command = invalid_info.command
             closest = closest_command(invalid_command, commands_per_level[level])
+            fixed_code = None
+            result = None
```
bjorn3, line 1631: @fRedelaar I fixed all tests. I think you will need to handle the case where fixed_code or result is none in the frontend/translations.

fRedelaar: Cool, great work! I've just sent a commit that made the bulb disappear whenever the code is fixed by the user. Or did you mean something else?

fRedelaar:
The PR seems to be working well and without bugs for levels 1 and 2. I couldn't find bug myself. After level 2, the Hedy code becomes a bit more complex which makes this PR a bit more vulnerable to bugs. To prevent this PR from getting even bigger, we've only implemented the code repair "Hint Button" for levels 1 and 2. We can improve the code repair for more levels in new PRs. 

Could someone please review this PR? :) A working video of the code can be seen below: 

https://user-images.githubusercontent.com/33595265/142623979-434b7051-c4fb-43d4-8b72-ca1ed1f3c21c.mov

Felienne:
I will take a loot later today @fRedelaar, thanks a lot!!

Felienne:
Supercool work team!! Just a little suggestion, can we add an explanation to the modal?

Something like:

"This is the correct code, can you fix it?" (localized of course)

This can be done in a separate PR, we can merge this now!

hedy.py
Felienne: I think Bjorn means the case happening when no fixed_code is given (but that should be ok for now)

hedy.py
Felienne: yes you would want to check not only if fixed_code is not the same as the original code, but also if it is valid code (but that is a program repair issue while this PR is mainly a UX PR, so we can fix that separately)

Felienne (approved):

fRedelaar:
> Supercool work team!! Just a little suggestion, can we add an explanation to the modal?
> 
> Something like:
> 
> "This is the correct code, can you fix it?" (localized of course)
> 
> This can be done in a separate PR, we can merge this now!

Thank you for the fast review @Felienne  !! My apologies for the big PR. The explanation of the modal is a good idea! Our team will create an issue and work on it.

Felienne:
> My apologies for the big PR. 

No problem, that sometimes happens even though we try to avoid it

> The explanation of the modal is a good idea! Our team will create an issue and work on it.

Awesome!

Issue #1263, opened 2021-11-14
Add sleep in level 2, make all list content new level 3


Felienne:
### Discussed in https://github.com/Felienne/hedy/discussions/1207

1. As it is now
2. Introduce only simple variables (like name), also add sleep to make it all more fun
3. Introduce lists with assign as we have it now (dieren is hand, kat, cavia) and also add and remove 
4. What is now 3, etc.

Felienne:
Closed by #1284 and #1288

PR #1288, opened 2021-11-16
Split level 3

Felienne:
First stab at #1263 

Works and may be merged but is not 100@ complete yet, the transpiler code still needs to be split and add and remove need to be added to the new level 3.


Felienne:
Hi @boryanagoncharenko!

You can merge this and proceed with your work.

It is not 100% complete, because level 2 and 3 still use the same transpiler, that needs to be split but I can do that when you are ready!

boryanagoncharenko (approved):

Issue #1163, opened 2021-11-04
[BUG] Signup is broken in offline mode


rix0rrr:
**Describe the bug**

When I used to run `python3 app.py`, I would be in offline mode (with an offline database) so I could do a full test of Hedy without having to use actual AWS credentials.

Signup is now broken when I run like that. It manifests like this:

![image](https://user-images.githubusercontent.com/524162/140293431-9d9be077-1dfd-4388-b49f-6cdef74e97d1.png)

But the actual cause is that the Python code is throwing an error, most likely because it's trying to send a confirmation email (irrelevant stack frames left out):

```
Traceback (most recent call last):
...
  File "/Users/huijbers/Dev/hedy/website/auth.py", line 238, in signup
    send_email_template('welcome_verify', email, os.getenv('BASE_URL', 'http://localhost') + '/auth/verify?username=' + urllib.parse.quote_plus(username) + '&token=' + urllib.parse.quote_plus(hashed_token))
  File "/Users/huijbers/Dev/hedy/website/auth.py", line 562, in send_email_template
    send_email(email, subject, body_plain, body_html)
  File "/Users/huijbers/Dev/hedy/website/querylog.py", line 170, in wrapped
    return fn(*args, **kwargs)
  File "/Users/huijbers/Dev/hedy/website/auth.py", line 527, in send_email
    result = email_client.send_email(
  File "/Users/huijbers/Dev/hedy/.env/lib/python3.9/site-packages/botocore/client.py", line 386, in _api_call
    return self._make_api_call(operation_name, kwargs)
...
  File "/Users/huijbers/Dev/hedy/.env/lib/python3.9/site-packages/botocore/auth.py", line 373, in add_auth
    raise NoCredentialsError()
botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

**Expected behavior**

Expected behavior would be to skip trying to send an email if:

* There are no credentials; and
* We're running in development mode.

An alternative could be to catch the error and ignore it if we're in development mode.

boryanagoncharenko:
This is related to https://github.com/Felienne/hedy/issues/1115

boryanagoncharenko:
tested locally - signing up does not fail
tested on alpha - received a welcome email successfully

PR #1289, opened 2021-11-16
Suppress error when sending email fails because of no aws credentials during local development #1163

boryanagoncharenko:
**Description**

Contains the minimal changes to avoid local development failures caused by no credentials for sending emails.

**Fix for**

#1163

**How to test**

1. Run hedy locally and sign up as a new user. The sign-up call should be successful since now the email part is skipped.
2. Sign up on alpha as a new user. The sign-up call should be successful and an email should be sent.


Felienne (approved):
Ah this will make local development so much easier!!

PR #1292, opened 2021-11-16
Created function to get translated keyword suggestions

brucemartens:

**Description**

Deleted commands_per_level and added the function get_translated_suggestions(). The function make a dictionary with the levelnumbers as a key and for the item it makes a list with the suggestions. The suggestions are stored in grammars/suggestions-"language".yaml. For example the dutch translated keywords suggestions can be found in grammars/suggestions-nl.yaml. If a translated keyword suggestion yamlfile can't be found, then it uses English by default.

**How to test**

- Make a typo in Hedy to get a suggestion for the keyword you were trying to type.
- It then will give a suggestion
- To test in Dutch change the 4 occurrences of lang="en" to  lang="nl" in hedy.app
- Try typing dukaf instead drukaf and it will give drukaf as a suggestion.

Fixed issue #1169


Felienne:
Can you look into merge issues and failing tests for this PR @brucemartens? We'd like to have PR's open only for a short time to prevent conflicts.

brucemartens:
All should be good right now @Felienne

grammars/suggestions-en.yaml
```diff
@@ -0,0 +1,199 @@
+1:
+ - print
+ - ask
+ - echo
+ - turn
+ - forward
```
bjorn3, line 6: Would it make sense to write it like

```suggestion
1: [print, ask, echo, turn, forward]
```

? That makes it easier to compare the list of keywords between levels, which may reduce the risk that you forget a keyword.

Felienne: I agree that is a bit more readable but I don't think it is a blocker at this point

Felienne: Made a separate issue for this: #1350

Felienne:
Added a test for the Dutch keyword suggestion, otherwise all good!

Felienne (approved):

Issue #942, opened 2021-10-09
Set up translation structure for keywords


HuibSprangers-leiden:
Currently, even if new keywords were added, the program has no structure in place to translate them. (app.py currently has the functions for other translations)

Felienne:
Good point! This means we will need to make a central location for translations and use that in both app.py and hedy.py.

Felienne:
(alternatively you can also make the issue part of a milestone, that also gives me an overview in the issue list)

emil-meyknecht:
We are currently working on this issue and would like to share our approach.
We are currently working in Hedy,py and are making Transformers to transform a Parse tree for each level, just as is done to convert the Hedy code to python, only this time we translate the Tree to the Hedy code with translated keywords. Then we write tests to see if each level is translated correctly.
We were wondering if this a a good approach and if we should continue with it.

PR #1306, opened 2021-11-17
Added Transformer to hedy.py. Added test File tests_translation_level_01

emil-meyknecht:
Review for the translation of keywords from one language to another

This draft request is mainly meant to show what the approach to translating keywords is.

Adds "ConvertToLang()" Transformers to the hedy.py file.
Adds "test_translation_level*" test files to test the translation

Is for Issue #942

Test are available on the branch and can be run if necessary.

The code can be run by running the tests.

It is a bit too much code for one pull, but it will be separated in smaller bites.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -12,14 +12,16 @@ import re
 HEDY_MAX_LEVEL = 22
 MAX_LINES = 100
 
-#dictionary to store transpilers
+# dictionary to store transpilers
 TRANSPILER_LOOKUP = {}
 
 # Python keywords need hashing when used as var names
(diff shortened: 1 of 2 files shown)


emil-meyknecht:
@Felienne The changes are all at the bottom of hedy.py and the new test file.

We are currently using the Lark transformers to translate the keywords from the parse tree. We believe this is the best way to make sure the code only translated the commands and not words in strings or vars while keeping the code the same way it was before. The test functions currently only check from English to Dutch but can easily be expanded to test from and to several languages with a keyword dictionary.

Felienne:
> @Felienne The changes are all at the bottom of hedy.py and the new test file.
> 
> We are currently using the Lark transformers to translate the keywords from the parse tree. We believe this is the best way to make sure the code only translated the commands and not words in strings or vars while keeping the code the same way it was before. The test functions currently only check from English to Dutch but can easily be expanded to test from and to several languages with a keyword dictionary.

This sounds like a fine approach to me (as discussed in the meeting today)

teundemast:
We will split this in multiple pull requests :)

Issue #882, opened 2021-10-05
Show quiz score as part of total


OnnoEbbens:
In the quiz of level 6 the maximum score is lower than in previous quizzes. Students think they did a bad job because they get a lower score. Can we show the score as part of the total points they can get, e.g. score: 50 / 70

Felienne:
Good idea! This can go on @LotusterHaar's next sprint. An easier solution that is quicker to reach (potentially even by you!) is to add some questions to level 6 to even out the points.

LterHaar:
My solution is, showing a percentage of the number of questions answered correctly

PR #1307, opened 2021-11-17
The end score will be shown as a percentage of the total score/max score * 100

LterHaar:
**Description**

The end score is shown as a percetage of the max score so that with different total scores the ratio is still in balance.

**Fix for**

In the quiz of level 6 the maximum score is lower than in previous quizzes. Students think they did a bad job because they get a lower score. Can we show the score as part of the total points they can get, e.g. score: 50 / 70

**How to test**

score 50/70 -> 71 %

### DIFF
--- a/app.py
+++ b/app.py
@@ -570,7 +570,7 @@ def quiz_finished(level):
     g.prefix = '/hedy'
 
     return render_template('endquiz.html', correct=session.get('correct_answer', 0),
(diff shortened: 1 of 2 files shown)


Felienne:
Addresses #882

Felienne:
Works!

image

Felienne (approved):

Issue #1312, opened 2021-11-18
Update to lark 1.0


bjorn3:
https://github.com/lark-parser/lark/releases/tag/1.0.0

The new grammar composition feature of lark should make it possible to replace the current grammar merging of hedy.

Felienne:
Yeah I saw that, it looks really promising! 

It will be a big refactor to update this (both fixing things that the update might break, and implementing the composition) 

Are you interested in picking that up @bjorn3?

Felienne:
Closed by #1314

PR #1314, opened 2021-11-18
Update lark to latest 1.1.1

bjorn3:
This updates lark to the latest version. None of the breaking changes affect hedy AFAICT. All tests pass. I think you need to do `pip uninstall lark-parser; pip install -r requirements.txt` when updating locally due to the package having been renamed.

Part of #1312


Felienne:
Thanks for this @bjorn3, but we are now in a bit too much churn with too many people working on Hedy daily for this to be smart now. Will do this over Christmas.

Felienne:
Hi @boryanagoncharenko!

I think since we are in a state to start improving things, maybe we can follow through on this one now? Once we do this (quite easy, just update this and see all tests still work) we could improve/simplify the code A LOT by using grammar compositions instead of our own grammar merging (https://github.com/Felienne/hedy/issues/1312)

Felienne:
Ah it seems they changed things wrt meta so we will need to make at least some changes to make this work (if absolutely needed we could drop meta since it is not actively used now)

boryanagoncharenko:
> Ah it seems they changed things wrt meta so we will need to make at least some changes to make this work (if absolutely needed we could drop meta since it is not actively used now)

Merged and resolved the latest changes without removing the meta information for now.

Felienne:
> This updates lark to the latest version. None of the breaking changes affect hedy AFAICT. All tests pass. I think you need to do `pip uninstall lark-parser; pip install -r requirements.txt` when updating locally due to the package having been renamed.

For me, `pip uninstall lark-parser; pip install -r requirements.txt` did not work, but `pip uninstall lark; pip install -r requirements.txt` did the trick. Maybe I had both installed?

Felienne (approved):

PR #1316, opened 2021-11-18
Add/remove in level 3

Felienne:
Fixes #1207


boryanagoncharenko (approved):
With this change I think we should reconsider the decision to disallow printing a list. Now users add or remove items from a list, but they cannot 'see' the change because they can only use "at random" or list access. Maybe we can print the list in the same way we create it, without quotes and brackets?

Felienne:
> With this change I think we should reconsider the decision to disallow printing a list. Now users add or remove items from a list, but they cannot 'see' the change because they can only use "at random" or list access. Maybe we can print the list in the same way we create it, without quotes and brackets?

I had the same idea yesterday!! (and this should really be a blog post haha) but I feel printing as a means of debugging is an old-fashioned way of doing it, which also mixes concerns. Printing is for showing to the user, and what we want here is to show a value for debugging/understandability purposes.

What we really want is to show the value of a variable, which is why I opened this issue: https://github.com/Felienne/hedy/issues/1329 Hope that makes sense!

PR #1318, opened 2021-11-18
semantics description in natural language additions

juliabolt:
**Description**

additions for semantics description in natural language
so far added level 2 and 3, not yet with types. 
Not yet complete.

### DIFF
--- a/SEMANTICS.md
+++ b/SEMANTICS.md
@@ -71,7 +71,7 @@ Level 1 supports:
     <td>Incomplete Exception</td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


Felienne:
Nice work @juliabolt!! This will be super helpful to people. Not that some things have already changed haha, level 2 is now split into two and the new level 3 has new features (see #1303 and (upcoming) #1316) but for now this is already better than what we have!

Felienne (approved):

juliabolt:
Oh haha I will check the alpha version! :) Thanks!

PR #1339, opened 2021-11-19
String manipulation for program repair

dolfein:
With these string manipulations, insert, delete and replace, I can do the necessary changes to repair the program, according to the error type.

Locations in [hedy.py](https://github.com/Felienne/hedy/blob/main/hedy.py) are indicated by a line and column integer (starting from 1), `calc_index` will return the corresponding index.


program_repair.py
```diff
@@ -1,3 +1,39 @@
+def calc_index(input_string, line, column):
```
Felienne, line 1: It is still not 100% clear to me what this function does. Maybe you can put an example run in the comments? (an example of what goes in and out)

Felienne (changes requested):

program_repair.py
dolfein: I hope the latest commit makes it more clear.

The example is a `ParseException` at `location = e.line, e.column`:
https://github.com/Felienne/hedy/blob/3640a6fd4336b66d6fda893b3021b8705e82608f/hedy.py#L1742-L1749

Felienne (approved):
Ok I still don't understand the function at all, but I think I ave to see it in context and I will rewrite the name when I understand it myself.

Issue #1338, opened 2021-11-19
 Setup translation structure for level 1


emil-meyknecht:
This is the start of the keyword translation setup for Hedy. The setup for level 1 will add all the basic functionalities and a structure to translate the keywords present in level 1. This issue will not change anything to the current Hedy site.

This include a function to make a dictionary out of Lark files, a first transformer for level 1 and tests.

PR #1342, opened 2021-11-19
Translation structure level 1 setup

emil-meyknecht:
Added structure for translation of level 1 to Hedy.py, issue #1338 .

Adds a way to translate code from one language to another with a transformer.
Works if the keyword lark file is available for the chosen language and if the "local_keywords_enabled" is set to True.
Adds a "keywords_to_dict" helper function to make a dictionary out of a Lark file.
Does not currently do anything on the Hedy site, only when tests are run.
Adds a test file for the translation of level 1.
Removed one empty line from "keywords-en.lark".

Tests can later be added to the "test_level_01" file, but have but separated initially for ease of use.
Same for the Transformers, which could be put in a new python file.

It is maybe too much for a single pull request bit it seemed better to me if the main functionalities were present with a testing structure

**How to test**

Tests are available in the new test_translation_level_01 file, otherwise the "translate_keywords" function can be run.

### DIFF
--- a/grammars/keywords-en.lark
+++ b/grammars/keywords-en.lark
@@ -18,7 +18,6 @@ _FROM: "from"
 _AT: "at"
 random : "random" //random needs to appear in the tree for further processing so does not start with _ or is uppercase
 
-
 //level 4
 _IN: "in"
 _IF: "if"
--- a/hedy.py
+++ b/hedy.py
@@ -20,7 +20,7 @@ MAX_LINES = 100
 LEVEL_STARTING_INDENTATION = 8
 
 # Boolean variables to allow code which is under construction to not be executed
-local_keywords_enabled = True # If this is True, only the keywords in the specified language can be used for now
+local_keywords_enabled = False # If this is True, only the keywords in the specified language can be used for now
 
 #dictionary to store transpilers
 TRANSPILER_LOOKUP = {}
--- a/hedy.py
+++ b/hedy.py
@@ -20,7 +20,7 @@ MAX_LINES = 100
 LEVEL_STARTING_INDENTATION = 8
 
 # Boolean variables to allow code which is under construction to not be executed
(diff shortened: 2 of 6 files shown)


Felienne:
It seems quite a few tests fail, you can check the output here: https://github.com/Felienne/hedy/runs/4268753086?check_suite_focus=true (or maybe run the validations locally, they are located in buildtools/github

emil-meyknecht:
Yes indeed, I overlooked something. I am working on a fix. I will just close this pull request. 
However is it normal that the "test_z_adventures.py" and "test_z_default" don't run correctly, because they givce me errors, with or without my modifications.

Felienne:
Nice work, smart move with the decorators!

I see the adventures and level_defaults are also working now. They should work (you may always assume that all tests run in main, since we cannot merge otherwise.

I rewrote your tests a little but because I think the dictionary lookup makes it a bit hard to see what is going on (I initially thought the translation worked the other way around) Can you rewrite them all in the more explicit way? It is a bit less generic but what we want to test here is the specific translation of a few Dutch to En programs and the other way around, not the very generic case. It is fine for me if you also do a few ones in your "dictionary style" but I also want a few explicit ones.

Felienne:
Als as you can see from my latest commit, enabling your boolean does seems to break a number of tests, I guess that is something you also want to look into, maybe in a separate PR?

emil-meyknecht:
@Felienne I changed the tests so they are more readable. The team is working on a fix for the tests that do not pass when the local_language_bool is True. Once it is done I will add it to this branch

The translation should work the other way around now.

Felienne:
Hi @emil-meyknecht! 

The issue was that drukak in the the keywords file was changed to print (as discussed) but some tests still used drukaf. All should be ok that this final change.

Felienne (approved):

PR #1343, opened 2021-11-19
Set level to 1 by default when run from command line

### LINKED ISSUE #1341: [BUG] Level is 0 by default when running from command line
If you do a clean checkout and try to run  a simple Hedy program from the command line, you get an error because the level hasn’t been set properly.

Steps to reproduce:

1. Do a clean checkout
2. Run

```
$ python3 -m venv .env
$ source .env/bin/activate
(.env)$ pip install -r requirements.txt
```

3. Create a file, `hello.hedy`, with the following contents:

```
print Hello World!
```

Then run the program from the command line according to [this](https://github.com/Felienne/hedy/blob/main/CONTRIBUTING.md#running-hedy-files-from-the-command-line) description:

```
(.env)$ python3 runhedy.py hello.hedy
An error occurred: Invalid Argument Type
Error attributes:
command=print
invalid_type=string
allowed_types=[]
invalid_argument=
(.env)$
```

I expected the program to print `Hello World!`.

As far as I can tell, the error happens because the Hedy source file doesn’t declare a level. And indeed, if you explicitly provide a level, the program works:

```
(.env)$ python3 runhedy.py hello.hedy —level 1
Hello World!
(.env)$
```

 According to the docs [here](https://github.com/Felienne/hedy/blob/main/CONTRIBUTING.md#running-hedy-files-from-the-command-line), indicating a level is only necessary if you want to use a level besides level 1. Thus, I think the example provided above should have worked.

As far as I can tell, there are two options for fixing this:

1. Change the documentation in `CONTRIBUTING.md` to say that you _always_ need to indicate a level, or
2. Fix the source code around [here](https://github.com/Felienne/hedy/blob/2eebc8930c9f3c15dbc5593f6a5f77c7e9567766/runhedy.py#L38) in `runhedy.py` so that `level` is set to `1` if it hasn’t been set by a command line flag or a declaration in the source file (AFAICT, it is left as `0` if not otherwise specified).

Despite never having written a single line of Python, I am happy to take a stab at fixing either the documentation or the Python code :-) 

PS. Thank you for the chat “at” Build Stuff @Felienne

ibsenrune:
The purpose of this PR is to fix #1341.

The current code base first tries to read the level from the command line. If no level was provided on the command line, we try to find a `#LEVEL <level>` in the source file. If none of these approaches render a level, the level will be left at `0`. 

This PR ensures that in that situation, the level will be set to `1`.

As part of this work, I also took the liberty of ensuring that you cannot explicitly set a non-positive level.

### DIFF
--- a/runhedy.py
+++ b/runhedy.py
@@ -26,8 +26,11 @@ def main():
                 try:
                     level = int(args[2])
                     if level > 8:
-                        print("Level has been set to 8, because the value specified was to high")
+                        print("Level has been set to 8, because the value specified was too high")
                         level = 8
+                    elif level <= 0:
+                        print("Level has been set to 1, because the value specified was too low")
+                        level = 1
                 except ValueError:
                     print("Level argument is not an integer, skipping argument.")
 
@@ -41,8 +44,14 @@ def main():
             if len(parts) >= 2:
                 level = int(parts[1])
                 if level > 8:
-                    print("Level has been set to 8, because the value specified was to high")
+                    print("Level has been set to 8, because the value specified was too high")
                     level = 8
+                elif level <= 0:
+                    print("Level has been set to 1, because the value specified was too low")
+                    level = 1
+
+    if level == 0:
+        level = 1 # Set level to 1 if not provided through other means
 
     program = '\n'.join([line
         for line in lines


runhedy.py
```diff
@@ -26,8 +26,11 @@ def main():
                 try:
                     level = int(args[2])
                     if level > 8:
```
Felienne, line 28: This 8 is now hardcoded (from way back when we had 7 levels) but hedy has a MAX_LEVEL constant that can be used here.

runhedy.py
```diff
@@ -41,8 +44,14 @@ def main():
...
                 if level > 8:
-                    print("Level has been set to 8, because the value specified was to high")
+                    print("Level has been set to 8, because the value specified was too high")
                     level = 8
+                elif level <= 0:
+                    print("Level has been set to 1, because the value specified was too low")
+                    level = 1
+
+    if level == 0:
+        level = 1 # Set level to 1 if not provided through other means
```
Felienne, line 54: Some duplication from above, should this be a function? get_correct_hedy_level or something?

Felienne (changes requested):
Thanks for your first Hedy contribution @ibsenrune!!!

Felienne:
Closing this for inactivity (and also the bug is not our priority)

PR #1344, opened 2021-11-20
Implemented a 403 error page

### LINKED ISSUE #1161: [UI idea] Implement a 403 error page
**Idea incl level**
I would like to have a 403 page when a user tries to visit a page he/she isn't allowed to visit; returning an unauthorized error. Currently we simply return a string "unauthorized" and the corresponding 403 error code. It would be nice to have a specific page for this error code, similarly to the already implement 404 and 500 error code pages. Which can be found in #992.

### LINKED ISSUE #1323: [BUG] repeat (and other constructs) do not check var existance
**Describe the bug**
print checks whether a variable it uses actually exists, f.e. here:

https://github.com/Felienne/hedy/blob/3f4dbab96e00e2c6d97f20bd810b2372fd3a0b5e/hedy.py#L1019

But other constructs like loops and ifs can also access vars and we do not check that, generating syntactically valid but not working Python code.

Level 9 (current alpha)

```
repeat x times
    eten is ask 'Wat wil je bestellen?'
    if eten is pizza
        print 'lekker'
    else
        print 'pizza is lekkerder'
```

image

**Expected behavior**
An undefinedvar exception should be thrown (which will happen if you call check_var_usage)

eylu4090:

**Description**

Fixes #1161. Adds a dedicated 403 page and image for the html page (as a note, the current image used was something I put together based on the current image used and general 403 images. Let me know if another image was in mind)

**Fix for**

For for #1161

**How to test**

1. Set up testing environment and login (some links only show the 403 page when the user is logged in, otherwise the user is redirected to the login page)
2. Visit the following links and observe the page:

Only teachers can retrieve classes
http://localhost:5000/classes
http://localhost:5000/class/foo
http://localhost:5000/customize-class/foo

Looks like you are not a teacher!
http://localhost:5000/programs?user=foo
http://localhost:5000/for-teachers


Felienne:
Wow thanks for this @eylu4090 and for your superclear PR description with links. Love the image also :D 

image

At one point I want to make a "burglar Hedy" but this will certainly do for now.

Felienne:
Just one little thing, I think you continued on your local branch from your prior commit: 
image

It is not a superbig deal but it is easier for us if you create a fresh local branch so you do not push all old commits every time (If you become a regular, that list can get really long :) )

Felienne (approved):

Issue #1298, opened 2021-11-16
Improve error messages from the type system 


boryanagoncharenko:
Iterate through all unit tests that raise an invalid type exception and check whether the customer-facing message can be improved. 

- In smaller levels should we use 'integer' or 'number'. The turn command seems to return both indifferent cases.
- Consider the following cases:
`turn tralala`
and
`var is tralala 
 turn var`
Currently both of the above return the same error: "You cannot use turn with text. Try using turn with left, right or a number". While this error makes sense for the first case, it might be confusing for the second. A better message would say "Try changing the variable var to a number".

boryanagoncharenko:
Closed by #1348

PR #1348, opened 2021-11-20
Review user-facing type error messages; Refactor comparisons #1298

boryanagoncharenko:
**Description**

- Revise all user-facing type error messages
- Refactor shamelessly comparisons' grammar and tests
- Add type validation for comparisons, add to list, remove from list commands, in list and equality checks

**Fix for**

#1298

**How to test**

1. Check whether the new user-facing messages are ok:
- In level 1 run the command `forward ten`. This is the general exception telling the user that they are using a type that is does not work for the command at hand. Is the error message clear?
- In level 2 run the command `a is ten\n forward a`. While this is the same error, it has a slightly different twist because now it is not referring to a literal value (as in the previous example), but to a variable. Is the error message clear?
- In level 12 run `x is 'test1'\n y is x + 1`. This is the error indicating that both types are allows with the command but they are not compatible with each other. Is it clear?
- In level 2 run `turn right 90`. This is a variation of the ParseError: `right 90` is something wrong and it does not really have a type. However, we suggest that it is changed to something that is supported. Is it clear?

2. Check that the grammar change of level 14 is correct. The change is solely  done based on assumptions made from the available tests.

**Checklist**


Felienne:
> * In level 1 run the command `forward ten`. This is the general exception telling the user that they are using a type that is does not work for the command at hand. Is the error message clear?

Yes, although I rewrote it a little bit. 

> * In level 2 run the command `a is ten\n forward a`. While this is the same error, it has a slightly different twist because now it is not referring to a literal value (as in the previous example), but to a variable. Is the error message clear?

Hmmm, 'a' is a text so it is a bit confusing. Maybe 'a contains text'?

Also the 'a number' part is not localized yet, can we fix that (this is true for the above error too) The Dutch word is "een getal" and that already is present in the texts yaml (can be a separate PR) 
image

> * In level 12 run `x is 'test1'\n y is x + 1`. This is the error indicating that both types are allows with the command but they are not compatible with each other. Is it clear?

Perfect!

> * In level 2 run `turn right 90`. This is a variation of the ParseError: `right 90` is something wrong and it does not really have a type. However, we suggest that it is changed to something that is supported. Is it clear?

Yes!

Felienne:
> 2\. Check that the grammar change of level 14 is correct. The change is solely  done based on assumptions made from the available tests.

Yes, that was exactly what I had in mind but had no time for Friday, thanks!

Felienne:
I fixed a merg conflict, hope I did not mess anything up :D

Felienne (approved):

Issue #1351, opened 2021-11-21
[Language idea] Make the ask command similar to the print command


boryanagoncharenko:
**Language idea incl level**
Applies to all levels. Let's make the ask and print commands symmetrical, so that they act similarly in all levels.

This will resolve the following issues:
1) (Should not be addressed, see comment) At level 2, print takes variables, while ask does not:
```
number is 10
answer is ask hello, number
print hello, number
```
The text of the input is ```hello, number``` while the print outputs ```hello, 10```

2) At level 12, ask gives a python error:
```
a is 10
answer is ask 'is your number ' a
```
Results in: ```cannot concatenate 'str' and 'int' objects```

boryanagoncharenko:
The first issue cannot be addressed. The reason is that before text is wrapped in quotes, the following command can cause cyclic reference:
```
color is ask What is your favourite color?
```

Issue #1329, opened 2021-11-19
[UI Idea] Show variable value in view


Felienne:
### Discussed in https://github.com/Felienne/hedy/discussions/807

<div type='discussions-op-text'>

**- Showing a variable value in the screen**
It would be nice if we could add a value counter to the output screen that would be constantly updated. For example if you make the multiplication game in level 6/7 that it would keep track of your score live. Or if you'd make the story adventure that you could see the items your carrying. So you could see: 'bag: rope, sword'. It would have to be a seperate bar at the top or bottom of the screen that could show a variable's value. 
This doesn't only contribute to the fun of playing a game, but it might also help in the understanding of variables. When  students can see live what's 'saved in the variable' they might understand it better. 
I think the command could be 'show'. This could be incorporated in level 2. 

</div>

I think this is an excellent idea and will help students indeed. Scratch also supports something like this, where you tick a variable and its value is shown in view:

![image](https://user-images.githubusercontent.com/1003685/142588313-d9c0b582-6c39-4213-8ab9-3b1fbfb549e8.png)

Maybe this would be a great BSc project, it is very clear in scope a sort of separate of the rest of the project

fRedelaar:
Uploading some mock-ups here to give more visual context for the staff of bachelorklas (thesis).

**Mock up 1**
![mock1_level6](https://user-images.githubusercontent.com/33595265/145963795-05a45c93-de7f-425d-ada4-12821e6a9f1b.png)

**Mock up 2**
![Mock1](https://user-images.githubusercontent.com/33595265/145963821-90fad749-859f-4e63-abe5-54695057598c.png)

**Mock up 3**
![kleurtjesPOC](https://user-images.githubusercontent.com/33595265/145963837-d0500862-3d6c-4738-917b-a0cf4d3a0161.png)

fRedelaar:
In the previous comment made in December, I've shown some mock-ups. These mock-ups were made to show the possible end-goal of this issue. I've redesigned the mock-ups to make Hedy as user-friendly, shape/color/font consistent as possible. 

### The latest designed mock-ups:
In the latest design, I've focused on consistency and "calmness". This is to prevent Hedy from getting an interface that is getting too full, like discussion #1835 mentioned. Since TiBiBa created the NL/EN dropdown, I've used this in the design for the variable view for consistency.  There are two mock-ups. One without and one without color-coding. Color-coding is a functionality brought up by @Felienne where string, integer, and list types will be made distinctive of each other using colors.


[...]

PR #1352, opened 2021-11-21
Address code review comments #1298

boryanagoncharenko:
**Description**

- Fix translation of 'a number' in Dutch error messages
- Remove single quotes around in type error messages and changed them for code snippets
- Make the ask command similar to the print command in terms of allowed types

**Fix for**

#1298 

**How to test**

- At any level try `forward ten` and check that the suggested type is translated to Dutch properly. With the same example check that there are no single quotes and, instead, all code references are code snippets
- At level 12, check that the ask command accepts variables which are of type int or float: `a is 10\n answer is ask 'is your number ' number` should not yield a type error. NOTE that python will fail because changes are needed in the transpiler. To be addressed with #1351
- At level 16, check that the ask command accepts list variables: `colors is ['orange', 'blue', 'green']]\n favorite is ask 'Is your fav color' colors` should not yield a type error. Again python will fail because changes are needed in the transpiler.

**Checklist**


Felienne:
> **How to test**
> 
> * At any level try `forward ten` and check that the suggested type is translated to Dutch properly. With the same example check that there are no single quotes and, instead, all code references are code snippets

I think we would like forward to also be highlighted in the error but I already fixed that (code was so clear it was a supereasy fix!!):

image

I am not sure we want ten to be pink, maybe white would be better because text is white but that is ok for now. This is better than 'ten' I think

> * At level 12, check that the ask command accepts variables which are of type int or float: `a is 10\n answer is ask 'is your number ' number` should not yield a type error. NOTE that python will fail because changes are needed in the transpiler. To be addressed with [[Language idea] Make the ask command similar to the print command #1351](https://github.com/Felienne/hedy/issues/1351)

Check!

> * At level 16, check that the ask command accepts list variables: `colors is ['orange', 'blue', 'green']]\n favorite is ask 'Is your fav color' colors` should not yield a type error. Again python will fail because changes are needed in the transpiler.

Check, but I am not sure I want to allow printing of lists at any level, because I think it is always a debugging thing that I want to do differently (because of #1329) But in any case ask and print should be the same so it is easy to disallow it in one place :)

Felienne (approved):

PR #1356, opened 2021-11-21
refactor: remove 'menu' parameter

rix0rrr:
All calls to `render_template()` used to include a `menu=` parameter,
to render the entries of the main menu (in every language).

It was an annoying requirement to pass in the menu everywhere, and
unnecessary because the menu could just be a global variable in the
template.

Add a template global `main_menu_entries()` which returns the same
as what `render_main_menu()` used to do. Instead, `menu` is now
a boolean that controls whether or not the menu should be rendered
at all, and the default is `True` so in most cases you don't need
to pass anything: you only need to pass `menu=False` in cases where
you don't want the menu rendered, which is only for embedded frames
like for the quiz.

Relates to #1179.


Felienne (dismissed):

templates/incl-menubar.html
```diff
@@ -6,19 +6,19 @@
...
       {% if item.short_name != 'for-teachers' or is_teacher %}
         <a class="menubar-btn {% if item.selected %}border-{{item.accent_color}}{% else %}border-transparent{% endif %} " href="{{localize_link (item.href)}}">
           {{ item.caption }}
         </a>
       {% endif %}
     {% endfor %}
     {% if username %}
-      <a class="menubar-btn border-{% if current_page == 'my-profile' %}white{% else %}transparent{% endif %}" href="{{localize_link ('/my-profile')}}">{{auth.profile}}</a>
-      <a class="menubar-btn border-{% if current_page == 'programs' %}white{% else %}transparent{% endif %}" href="{{localize_link ('/programs')}}">{{auth.program_header}}</a>
+      <a class="menubar-btn {% if current_page == 'my-profile' %}border-white{% else %}border-transparent{% endif %}" href="{{localize_link ('/my-profile')}}">{{auth.profile}}</a>
```
dvberkel, line 17: Thank you

dvberkel (approved):
Very nice!

PR #1363, opened 2021-11-23
Improve coding and cleanup class customizations 

### LINKED ISSUE #1300: [UI idea] After creating class, re-direct to customization page
**Idea incl level**
Currently when we create a new class the page is reloaded and the class can be found in our class overview. Directly linked with #1299 it would be nice to re-direct the teacher to the customization page with instructions on the possibilities. This idea in turn is related to #775.

**Additional context**

### LINKED ISSUE #1326: [CLEAN-UP] Variable naming for class customization unclear
**Describe the bug**
Currently we use the words `preferences` and `customizations` side-by-side when talking about the class/level customizations made by a teacher. It would be nice to align the variable and function naming to be better readable and make more sense. This also applies to the database table and related database functions.

### LINKED ISSUE #1360: [BUG] Class customization "example programs" toggle incorrect on programmers mode switch
**Describe the bug**
Currently teachers are able to hide example programs within the class customizations. However, when we hide the example programs for a specific level and toggle the programmers mode twice we are still able to show the example programs. This is clearly undesired behavior and should be prevented.

**Expected behavior**
Example programs will stay hidden, even when toggling the programmers mode

TiBiBa:

**Description**
In this PR we cleanup the code related to the class customizations. In the original implementation a lot of redundant database calls were made. By restructuring the code the amount of queries is reduced. We also fix a small bug where the programmers mode toggle was always able to show the example programs, even when being disabled in the class settings. We also implement a re-direct when creating a new class: in this PR the teacher is automatically re-directed to the customization page.

**Fix for**
This PR fixed #1300, fixed #1326 and fixed #1360.

**How to test**
No really visual changes. One can verify that the programmers mode correctly toggles() even when class customizations are active. All other changes are on the back-end and don't influence the user.

**Checklist**


static/js/teachers.ts
```diff
@@ -16,7 +16,7 @@ export function create_class() {
       contentType: 'application/json',
       dataType: 'json'
     }).done(function(_response) {
-      location.reload ();
+      window.location.pathname = '/customize-class/' + _response.id ;
```
rix0rrr, line 19: Not specifically a comment on this PR, but this makes me notice the URL does not start with `for-teachers`.

If it did, we would automatically properly detect and render the main menu appropriately (with the "for teachers" tab selected).

So maybe worth changing?

static/js/teachers.ts
```diff
@@ -119,25 +119,33 @@ export function save_level_settings(id: string, level: number) {
...
-     let toggle_example_programs = false;
-     let toggle_level = false;
+     let hide_level = false;
+     let example_programs = false;
+     let hide_prev_level = false;
+     let hide_next_level = false;
+
      if ($('#hide_level' + level).prop('checked')) {
-       toggle_level = true;
+       hide_level = true;
```
rix0rrr, line 128: Perhaps this can be rewritten as

```ts
const hide_level = !!$(`#hide_level${level}`).prop('checked');
const hide_next_level = !!$(`#hide_level${level - 1}`).prop('checked');
const example_programs = !!$(`#example_programs${level}`).prop('checked');
const hide_prev_level = !!$(`#hide_level${level - 1}`).prop('checked');
```

website/database.py
```diff
@@ -227,30 +227,62 @@ def delete_class(self, Class):
         for student_id in Class.get ('students', []):
             Database.remove_student_from_class (self, Class ['id'], student_id)
 
-        CLASSES.delete({'id': Class ['id']})
-        PREFERENCES.delete({'id': Class['id']})
+        for customization in CUSTOMIZATIONS.get_many({'id': Class['id']}):
```
rix0rrr, line 230: Isn't there a `delete_many`?

website/database.py
```diff
@@ -227,30 +227,62 @@ def delete_class(self, Class):
...
+    def get_customizations_class(self, class_id):
+        customizations = {}
+        for customization in CUSTOMIZATIONS.get_many({'id': class_id}):
+            customizations[customization['level']] = customization
+        return customizations
+
+    def get_level_customizations_class(self, class_id, level):
+        try:
+            return CUSTOMIZATIONS.get({'id': class_id, 'level': level})
+        except:
```
rix0rrr, line 260: Does this actually throw an exception? It shouldn't, if so that's a bug I feel.

website/database.py
```diff
@@ -227,30 +227,62 @@ def delete_class(self, Class):
...
+            return None
+
+    def get_student_restrictions(self, adventures, user, level):
+        restrictions = {}
+        found_restrictions = False
+        if user:
+            student_classes = self.get_student_classes(user)
```
rix0rrr, line 267: Don't we know yet what class a student is in?

website/database.py
```diff
@@ -227,30 +227,62 @@ def delete_class(self, Class):
...
+            customizations[customization['level']] = customization
+        return customizations
+
+    def get_level_customizations_class(self, class_id, level):
+        try:
+            return CUSTOMIZATIONS.get({'id': class_id, 'level': level})
+        except:
+            return None
+
+    def get_student_restrictions(self, adventures, user, level):
```
rix0rrr, line 263: Does `adventures` here represent `all_adventures`, and are we filtering that list down?

website/database.py
```diff
@@ -227,30 +227,62 @@ def delete_class(self, Class):
...
+        if user:
+            student_classes = self.get_student_classes(user)
+            if student_classes:
+                level_preferences = self.get_level_customizations_class(student_classes[0]['id'], level)
+                if level_preferences:
+                    found_restrictions = True
+                    display_adventures = []
+                    for adventure in adventures:
+                        if adventure['short_name'] in level_preferences['adventures']:
+                            display_adventures.append(adventure)
```
rix0rrr, line 275:
```suggestion
                    display_adventures = [a for a in adventures if a['short_name'] in level_preferences['adventures']]
```

website/database.py
```diff
@@ -8,7 +8,7 @@
 TOKENS = dynamo.Table(storage, 'tokens', 'id')
 PROGRAMS = dynamo.Table(storage, 'programs', 'id', indexed_fields=['username'])
 CLASSES = dynamo.Table(storage, 'classes', 'id', indexed_fields=['teacher', 'link'])
-PREFERENCES = dynamo.Table(storage, 'preferences', 'id')
+CUSTOMIZATIONS = dynamo.Table(storage, 'class_customizations', partition_key='id', sort_key='level')
```
rix0rrr, line 11: It's probably helpful if you describe the schema here in a couple of comments. What are the keys, how do we use them, what do the records contain, etc?

rix0rrr:
Don't mistake me picking on your code. I like it, thanks for the refactorings :)

TiBiBa:
> Don't mistake me picking on your code. I like it, thanks for the refactorings :)

Thanks for the feedback! Will take a look to implement some changes.

static/js/teachers.ts
TiBiBa: Not sure how we would change this, do you mean `for-teachers/customize-class` ?

static/js/teachers.ts
TiBiBa: Agree, bit less readable but I assume someone else never has to alter this later on

website/database.py
TiBiBa: Didn't know that! Fixed!

website/database.py
TiBiBa: Agree! Changed the notation to `return customizations if customizations else None`

website/database.py
TiBiBa: In `current_user()` we only store the email, is_teacher and username values

website/database.py
TiBiBa: all adventures. Changed variable naming with last commit to make it more clear

website/database.py
TiBiBa: Agree! Changed with commit.

static/js/teachers.ts
rix0rrr: That's what I meant. You can also say "neh", it was just an idea.

You will start to see the difference after combining with this commit: https://github.com/Felienne/hedy/commit/42ab2641d2c26a101a870371e7c16f0de8729439

website/database.py
rix0rrr: Probably not then.

rix0rrr (approved):
❤️

static/js/teachers.ts
TiBiBa: I agree that this would be better! Other than the menu bar issue we might also implement more pages such as `customize-student` that would belong to the `for-teachers` section. Makes more sense to group them there. Is it okay if I pick this up in a different issue?

Felienne: Yeah it has the benefit of the menu bar being correct "for free" But sure, let's do that separately!

Felienne (approved):

Issue #1365, opened 2021-11-23
[CHORE] Run all tests with Hedy testrunner


Felienne:
Since #1359 the multilevel tester runs the generated Python but of course it'd be good if all tests do that, so let's make a "single level" runner too so we can share that code.

Felienne:
Closed by #1461

PR #1367, opened 2021-11-23
Adds single_level_tester

Felienne:
Start of #1365

Introduces a single_test test runner to make test code briefer and more systematic (allowing for the checking of all generated Python more easily)

@boryanagoncharenko I am making this into a draft PR so you know I am working on this and you can take it into account when doing the "spaces refactor"


Felienne:
Let's merge this now or we will have more issues later!

Felienne:
@boryanagoncharenko, If this does not mess up your work too much, can you merge it?

PR #1369, opened 2021-11-24
[Fix] Turtle not showing when used in combination with ask keyword

### LINKED ISSUE #1361: [BUG] When an ask is present in a program, the turtle arrow disappears
**Describe the bug**
When an `ask` is present in a program, the turtle arrow disappears but otherwise, turtle programs work fine.This seems to happen at all levels and it does not matter if the ask has anything to do with the turtle code (like entering a value that is used in the turtle, that is *not* needed for this bug to happen)

I have no clue what can cause this!

For example level 1:

```
ask Wat is je lievelingskleur?
turn right
forward 50
turn left
forward 50
```

image

**Expected behavior**
Turtle should be shown.

TiBiBa:

**Description**
Currently when using the ask keyword in combination with the turtle drawing (in any possible format) the arrow of the turtle won't show. This occurred due to the turtle canvas being empties when a user input is read, but with this operation the arrow is removed as well due to being initialized before the input function is called. By replacing `empty()` with `hide()` and write an additional check to verify if we have a turtle combined with `show()` the error should be solved.

**Fix for**
This PR fixed #1361.

**How to test**
Run the code as described in the issue and verify that the arrow is showing in this case. Code can be found below as well:

```
ask Wat is je lievelingskleur?
turn right
forward 50
turn left
forward 50
```

**Checklist**


Felienne:
Nice, I would never have found this myself I think!

Felienne (approved):

PR #1372, opened 2021-11-24
chore: remove dependency on `BASE_URL`

rix0rrr:
This environment variable is only rarely necessary, and even then can be
automatically derived from the current request.

This PR makes it so you don't get the following error:

```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

If you forget to set `BASE_URL`.


Felienne:
Thanks @rix0rrr but this causes some of the e2e tests to fail (not sure that is expected?)

rix0rrr:
Well the test is testing we definitely do include the `BASE_URL`, even when it's not necessary.

So that's why it's failing. No biggie but test needs to be fixed.

Felienne:
> Well the test is testing we definitely do include the `BASE_URL`, even when it's not necessary.
> 
> So that's why it's failing. No biggie but test needs to be fixed.

Ah got it! And I guess it makes no sense to fix that now, but rather do that when #1282 has been merged?

Felienne (approved):

PR #1375, opened 2021-11-24
Include error line in invalid exception

bjorn3:
Previously in certain cases `?` could be reported as line and column for parse errors. This PR will now cause the line number to always be available. `IsValid()` doesn't report the column of the error and I don't know how to change it to report it, so for now the column will remain as `?`.

Part of https://github.com/Felienne/hedy/issues/845


Felienne:
Thanks @bjorn3, can you give some example programs that have changed behavior to help me understand the change better?

bjorn3:
An example is this test from level 4:

```
is Foobar
print welcome
```

Previously this would say:

```
The code you entered is not valid Hedy code. There is a mistake on line ?, at position ?. You typed is, but that is not allowed.
```

With this PR it will say:

```
The code you entered is not valid Hedy code. There is a mistake on line 1, at position ?. You typed is, but that is not allowed.
```

Felienne (approved):

Felienne:
> An example is this test from level 4:
> 
> ```
> is Foobar
> print welcome
> ```
> 
> Previously this would say:
> 
> ```
> The code you entered is not valid Hedy code. There is a mistake on line ?, at position ?. You typed is, but that is not allowed.
> ```
> 
> With this PR it will say:
> 
> ```
> The code you entered is not valid Hedy code. There is a mistake on line 1, at position ?. You typed is, but that is not allowed.
> ```

Ah got it! I was a bit misled by the "always" in the PR title (updated that)

PR #1376, opened 2021-11-24
Adding a check if there are any duplicate translations of keywords

teundemast:
**Description**
Since having duplicate translations of keywords will cause errors, we need to make sure there are none in the lark files. For example, 'anders' can't be a Dutch translation for 'elif' and 'else'. 
Having duplicate 'translations' of commands in English won't cause any trouble, so the English language is excluded in the check. 

I think the best time to check if there are any duplicate translations, is when filling in the translation of keywords on the translation page. But since this requires a lot of programming for just a small problem, I have decided to not go down that path. So, I have added a warning in the title of the keywords section on the translation page:
Capture
When transforming the yaml files of the keywords to lark files, I have added a check for any duplicate translations. If this check fails a Value Error is raised.  

The function transform_yaml_to_lark might be a bit long now, but I think splitting this function up in multiple functions won't do any good for the readability. 

**Fix for**

This fixes issue #1345

**How to test**

To test if the check works, you could put a duplicate translation to a keywords yaml file. When running yaml_transformer.py, it should not transform this yaml file.


Felienne (approved):

Felienne:
> I think the best time to check if there are any duplicate translations, is when filling in the translation of keywords on the translation page. But since this requires a lot of programming for just a small problem, I have decided to not go down that path. 

Fair enough, but can you open an issue for the actual check in the translation page? Maybe someone else wants to build it!

Issue #1046, opened 2021-10-19
Quiz could be a bit wider?


Felienne:
There seems to be some whitespace on the right hand side where the quiz could be larger. Esp when folding the palette in this would give us more room. Not sure to fix this in the quiz itself (@LterHaar) or in the iframe (@fpereiro or @TiBiBa)?

![image](https://user-images.githubusercontent.com/1003685/137896473-6c22980e-55e3-489f-9407-001b535f5171.png)

fRedelaar:
Closing issue since MR #1381 has been merged

PR #1381, opened 2021-11-25
Possible fix for #1046: Made quiz wider

fRedelaar:
**Description**

Could be a fix for issue #1046. In the PR the scale has been deleted, so the quiz will be wider. Feel free to check out the branch and improve the UI or even discard this MR if there is a better solution, I just looked into it a bit. :) 

**Previously**
image

**Result of PR**
image


Felienne:
Nice this is a lot better!

Felienne (approved):

Issue #1275, opened 2021-11-15
[UI idea] Do not show confetti when running demo code


TiBiBa:
**Idea incl level**
Currently, the user is given a positive message when successfully running a program, combined with the option to show confetti with the click on an icon. I think it would be nice to only show this to the user when the created program is unique program and not when simply running a start or demo program. This should be easy because we already keep track if a program is either a start or demo program

**Additional context**
Add any other context about the idea here.

Felienne:
Good idea!! At one point the confetti will be more limited anyway with a few victory conditions but this is a good idea to start with!

Felienne:
Maybe a fun little issue for you @jpelay?

jpelay:
Yes! 

> Maybe a fun little issue for you @jpelay?

jpelay:
@TiBiBa I'm working on this one, but I can't see what you by "we already keep track if a program is either a start or demo program" I can't see that in the /parse endpoint, or anywhere, can you help me with that? Do you mean that we can know if a program is a `start_code` or `demo_code` by checking if it's in the `load_adventures_per_level` list?

Felienne:
> @TiBiBa I'm working on this one, but I can't see what you by "we already keep track if a program is either a start or demo program" I can't see that in the /parse endpoint, or anywhere, can you help me with that? Do you mean that we can know if a program is a `start_code` or `demo_code` by checking if it's in the `load_adventures_per_level` list?

Yes I am pretty sure that it what @TiBiBa meant!

What we could alternatively do (and will be helpful going forward anyway...) is to track if a button is clicked from an example or form a cheatsheet. If we start tracking that we can gather more finegrained data about how kids use the site (atm we do not know how often examples and the cheatsheets are used apart from our own observations)

PR #1397, opened 2021-11-26
Fix #1283 Delaying confetti to end of execution

### LINKED ISSUE #1283: Delay confetti to end of execution
**Describe the bug**
When you create a program with ask, you immediately get the well done/confetti bar

For example in level 2:
image

**Expected behavior**
Can we make the confetti appear only when the program is done?

fRedelaar:
**Description**

Fix for issue #1283. The confetti emoji won't be shown until the program is over, so ask programs have to be filled in by the user.

https://user-images.githubusercontent.com/33595265/143560470-2d7c85f0-088f-4358-947a-1b0561623d94.mov


fRedelaar:
I just noticed a bug with warnings (because it can run, it shows a green and yellow notification), so I converted the MR to a draft. Will fix it probably later today.

TiBiBa:
A bit of topic: I just noticed that we are also getting compliments and the option to show confetti when viewing a shared program of another user. Is that expected behavior? Otherwise we should implement some check if we are currently in the code editor or just executing someone else's their code.

Felienne:
> **Edit** In extend to this suggestion. What do we want to do if a user runs an example program? They didn't do any work so should they receive a compliment?

I think this is a good point, shall we make a separate issue for that?

fRedelaar:
> A bit of topic: I just noticed that we are also getting compliments and the option to show confetti when viewing a shared program of another user. Is that expected behavior? Otherwise we should implement some check if we are currently in the code editor or just executing someone else's their code.
> 
> **Edit** In extend to this suggestion. What do we want to do if a user runs an example program? They didn't do any work so should they receive a compliment?

Haha I wasn't aware of the feature to share a program and view the program until this comment. Really cool feature and I think the confetti is fun, but I can imagine if the confetti should be removed. I don't feel very empowered to make this decision. 

About the example/demo programs, I think you are right! I saw your issue #1275 and totally agree. Maybe our team (team 13 Rocket scientists for the course Software Engineering at LIACS) could pick up this issue next week?

fRedelaar:
> > **Edit** In extend to this suggestion. What do we want to do if a user runs an example program? They didn't do any work so should they receive a compliment?
> 
> I think this is a good point, shall we make a separate issue for that?

There already is I believe, #1275

Felienne:
Lovely, that is a lot better!

Felienne (approved):

TiBiBa:
> > A bit of topic: I just noticed that we are also getting compliments and the option to show confetti when viewing a shared program of another user. Is that expected behavior? Otherwise we should implement some check if we are currently in the code editor or just executing someone else's their code.
> > **Edit** In extend to this suggestion. What do we want to do if a user runs an example program? They didn't do any work so should they receive a compliment?
> 
> Haha I wasn't aware of the feature to share a program and view the program until this comment. Really cool feature and I think the confetti is fun, but I can imagine if the confetti should be removed. I don't feel very empowered to make this decision.
> 
> About the example/demo programs, I think you are right! I saw your issue #1275 and totally agree. Maybe our team (team 13 Rocket scientists for the course Software Engineering at LIACS) could pick up this issue next week?

Completely forgot I already made an issue about this. My bad! Sounds great if your group could pick this up next week.

Felienne:
> > > **Edit** In extend to this suggestion. What do we want to do if a user runs an example program? They didn't do any work so should they receive a compliment?
> > 
> > 
> > I think this is a good point, shall we make a separate issue for that?
> 
> There already is I believe, #1275

Ah my bad, should have checked!

PR #1404, opened 2021-11-26
Move coding=utf-8 to the top of app.py

bjorn3:
It must be in either the first or the second line to be effective according to [PEP 263](https://www.python.org/dev/peps/pep-0263/).


bjorn3:
While python 3 defaults to utf-8 for source files, python 2 doesn't. If someone forgets to `source .env/bin/activate` running `python app.py` will likely default to python 2. As `app.py` contains a non-ascii character, it will fail to load in python 2 before this PR. It won't even reach the version check at the top of the file, thus causing the following potentially confusing error:

```
SyntaxError: Non-ASCII character '\xc3' in file app.py, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

After this PR trying to run with python2 will result in a message of the form

```
Hedy requires Python 3.6 or newer to run. However, your version of Python is 2.7.0
```

Felienne (approved):

Issue #1302, opened 2021-11-17
Translate characters should return a string depending on the language used


teundemast:
The function translate characters only returns English words

teundemast:
<img width="527" alt="Capture" src="https://user-images.githubusercontent.com/36051227/142722889-acfb2ecd-7f70-4831-98a5-d7e3c98c753e.PNG">
What is returned should be depending on the language of the user

PR #1409, opened 2021-11-26
Localizing translate characters

teundemast:
**Description**
With this PR the function 'translate_characters' is localized, which means that instead of the following error message: 
translate_characters

The programmer will recieve this error message: 
translate_characters2

Fixes issue #1302

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -12,6 +12,7 @@ import hashlib
 import re
 from dataclasses import dataclass, field
 import exceptions
+import hedy_translation
 
 import yaml
 import program_repair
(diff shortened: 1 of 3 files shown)


Felienne:
Going to tag @boryanagoncharenko here, because I think we should have already supported this before (we already have the translations) 

I think you might have found a bug rather than a missing feature. 

Can you share some code snippets where you got this error originally so I can look into it?

Felienne:
Update, the function `def translate_error(code, translations, arguments)` in app.py should handle this

teundemast:
Ah totally missed that, I just assumed it was not translated because the function just returned an English word. Sorry about that!

Felienne:
Teah I gues my comment there also did not help! 

WOuld still love to see the code though attached to your screenshot because that should still be fixed!

teundemast:
It was some code I wrote in my local environment on the branch from this pull request, but with local keywords disabled. I just assumed I would get the same error when I would run the code on the main branch 😅. So no worries, all good!

Issue #1408, opened 2021-11-26
Improve undefined error messages


boryanagoncharenko:
1. Add error messages for cyclic references: a = a + 1
2. [Timebox to 15-30 mins] Try moving the undefined variable errors to the TypeValidator. We could provide better messages from there since we have more information from the AST.

PR #1417, opened 2021-11-28
Add cyclic var definition error message #1408

boryanagoncharenko:
**Description**

- Add error for cyclic var definitions: b is b + 1
- Remove duplications in lookup table. Now if a variable is present, it is hashed if it needs to be hashed or in plain text if it does not need to be hashed.
- Remove dead code from the transpiler: get_value function

**Fix for**

#1408 

**How to test**

- Run Hedy locally and go to any level >= 6. Type `b is b + 1` and see the error in English. Is the error text ok?
- Please translate the error in Dutch. You can add it as a comment or commit it directly.
- Have a thorough look at the adventure yaml files that are updated (because of the new cyclic error). Maybe there is something else with these that needs to change?

**Checklist**


hedy.py
```diff
@@ -587,19 +589,20 @@ def save_type_to_lookup(self, name, inferred_type):
...
-            if not matches[0].inferred:
-                if matches[0].inferring:
-                    # TODO we know at this point there is a cyclic var reference, e.g. b = b + 1, for now do not raise
-                    matches[0].inferring = False
-                    matches[0].type_ = HedyType.any
-                    matches[0].inferred = True
+            # TODO: we should not just take the first match, take into account var reassignments
+            match = matches[0]
+            if not match.type_:
+                if match.inferring:  # there is a cyclic var reference, e.g. b = b + 1
```
Felienne, line 597: The comments combined with the name are really clear, but I am not sure `inferring` i a good name for the flag (without looking at the comment). I think it means something like "is this the var we are currently inferring"? So maybe `currently_inferring` or `is_currently_inferred` is better?

boryanagoncharenko: The inferred flag is gone and the inferring is changed to currently_inferring

hedy.py
```diff
@@ -259,8 +259,7 @@ class LookupEntry:
     tree: Tree
     skip_hashing: bool
     type_: str = None
-    inferred: bool = False  # types of lookup entries are inferred on demand
-    inferring: bool = False  # used to detect cyclic type inference
+    currently_inferring: bool = False  # used to detect cyclic type inference
```
Felienne, line 262: Thanks, this is just a bit more clear!

Felienne (approved):

Issue #531, opened 2021-06-24
Point out missing translations in UI texts


fpereiro:
Currently we have a script that checks for missing entries for UI texts on each supported language. However, we have many of those texts that are not translated. It would be good to have a way to see which ones, at a glance, to work on the remaining translations.

My initial idea was to check for this on the build tools and point them out. However, @Felienne suggested this might be incorporated into the translation tool, perhaps by either showing only the untranslated segments (or providing a button to show only those).

Felienne:
Hi @boryanagoncharenko!

Maybe this could also be an issue for you to pick up? I am imagining a tick mark in the translation UI (https://hedycode.com/translate/en/nl) which shows only the untranslated fields so it gets easier to keep translations up to date.

Felienne:
Closed by #1418

PR #1418, opened 2021-11-28
Add option to show/hide the fields that are already translated #531

boryanagoncharenko:
**Description**

Add option to show only the fields that need translation or show all fields. Note that the fields that need translation are fields that 1) have empty translation or 2) have translation equivalent to the original. The second condition covers for cases in which the English properties are copied to the yaml file of another language.

**Fix for**

#531 

**How to test**

If you test this locally, please make sure to run the generate-typescript.
- Run Hedy and open the translation tool: 0.0.0.0:5000/translate/en/bg and check the button behaviour.
- Try translating a field and download the files (with both translated fields hidden and shown). Check that the files contain ALL fields.

**Checklist**


Felienne:
> **Description**
> 
> or 2) have translation equivalent to the original. The second condition covers for cases in which the English properties are copied to the yaml file of another language.

Oh so smart you thought of that!! First I was like: "hmmmm why are all those En translations in there?" but this makes total sense! Love that you keep delivering things that I did not ask for but that I should have asked for :D

Felienne (approved):

PR #1419, opened 2021-11-29
[FIX] Replace file structure of for-teachers page from .md to .yaml files

### LINKED ISSUE #1389: [UI idea] Frequently made mistakes page needs an update
Not only the content (which I will take care of myself) but also the lay-out of this page needs an update. Right now, somehow most mistakes are mentioned twice and the indentation with the bulletpoints looks very messy. 

Maybe @TiBiBa can take a look at it and maybe add some buttons, so that the teachers can select the level of which they want to see the frequently made mistakes? Because now it's just a long messy list. 
Maybe you or @Felienne have some other ideas as well about how we can make it more structured and look more clear?

TiBiBa:
**Description**
Currently we depend on .md files for the content of several pages. For correctly retrieving and displaying sections several pre-process functions are written. However, the .md file structure gives us less freedom with front-end customization than we would like. By replacing the structure to depend on .yaml files (similar to how adventures are stored) we are better able to customize our UI dependent on the content. In this PR we replace the .md file structure by a .yaml file structure for content translations. A new function is written to retrieve the correct text and the front-end template page is adjusted to correctly display this content.

**Fix for**
This PR fixed #1389 (partly).

**How to test**
Navigate to for-teachers page and notice so difference on the front-end. On code base, notice the addition of a new folder for storing the .yaml files and verify the correct retrieval of translations.

**Checklist**


coursedata/pages/for-teachers/en.yaml
```diff
@@ -0,0 +1,91 @@
+title: "Hedy documentation"
+sections:
+    introduction:
+        title: "Introduction"
+        text:
+            whats_hedy:
```
Felienne, line 6: Do we need these names (like `whats_hedy`) hard coded into the yaml? We can also simply make a list of key, text pairs? That gives us a bit more flexibility and less duplication. I see in the HTML template we do not really use these names anyway.

Otherwise, really good idea this way!!

TiBiBa: I changed the keys to numeric values, simply counting up with each (sub)sections!

coursedata/pages/for-teachers/en.yaml
```diff
@@ -0,0 +1,91 @@
+title: "Hedy documentation"
+sections:
+    1:
+        title: "Introduction"
+        subsections:
+            1:
```
rix0rrr, line 6: Why give them numbers in the first place if you're never referring to them anywhere? Could this not just be a list?

I think you want the `section`s to have string keys (since they seem to represent different "pages", so you could potentially link them in the URL bar), but probably not the subsections (as they seem to be rendered onto a single page):

```
title: "Hedy documentation"
sections:
    introduction:
        title: "Introduction"
        subsections:
            - title: What's hedy
              text: <TODO>
            - title: Target audience
              text: <TODO>
```

Also, since you're ultimately going to render this to:

```
<h2>What's Hedy</h2>

(Text for What's hedy here)

<h2>Target audience</h2>

(Text here)
```

Instead of a list of sections, couldn't you store it as MarkDown inside the YAML:

```
## What's Hedy

(Text here)

## Target audience

(Text here)
```

What's the advantage of splitting every section into a separate Python object?

(Or is it just so that user's don't have control over the heading level?)

app.py
```diff
@@ -920,6 +920,21 @@ def main_page(page):
         else:
             return utils.page_403(TRANSLATIONS, current_user()['username'], g.lang, TRANSLATIONS.get_translations(g.lang, 'ui').get('not_user'))
 
+    user = current_user()
+
+    if page == 'for-teachers':
+        for_teacher_translations = hedyweb.PageTranslations(page).get_page_translations(g.lang);
+        if is_teacher(user):
+            welcome_teacher = session.get('welcome-teacher') or False
+            session['welcome-teacher'] = False
```
rix0rrr, line 929:
```suggestion
            session.pop('welcome-teacher', None)
```

Remove the value from the session so we don't need to store it. Sessions get transmitted over the wire (in cookies)

coursedata/pages/for-teachers/en.yaml
TiBiBa: Agree that the numbers don't make any sense! Will take a look to improve the structure (not that familiar with .yaml files (yet)).. The main reason for splitting is indeed to don't give the translators any control on the front-end level. I noticed that currently .md files have mixed styling due to mistakes being made on the translation side. Translating headers but forgetting or miscounting the amount of # characters needed to be consistent.

app.py
TiBiBa: Didn't know that! Code was already there and it seemed important, will change right away.

Felienne (approved):

PR #1420, opened 2021-11-29
Changing structure keyword yaml files

teundemast:
With this PR the structure of the keyword yaml files is changed. Visually this only affects the the translate page;
From: 
translatepage1
To: 
translatepage2

I think this looks a bit cleaner and with this structure, and it is much easier to work with. 
I have changed the yaml transformer accordingly. 

With this change, it was possible to simplify the keywords_to_dict function. This is also included in this PR. @emil-meyknecht  @Livavdv do you think this works as good as the original function? With the original function there was also flaw where 'to' : 'tot' was an item of the dictionary but it needs to be possible to translate  'to' to 'toe aan'. In the new function, 'to_list' : 'toe aan' is now also part of the dictionary. We probably need a way to detect whether 'to' is used as 'tot' or as 'toe aan'.


emil-meyknecht:
The new function to create dictionaries indeed gets rid of some flaws the other one had, and makes it a lot simpler.  I am not sure how the different "to" will impact the translation yet, but I believe the parser will be able to differentiate the 2 "to" and then it can be translated accordingly.

Felienne:
Ah this is a lot clearer!

> The new function to create dictionaries indeed gets rid of some flaws the other one had, and makes it a lot simpler. I am not sure how the different "to" will impact the translation yet, but I believe the parser will be able to differentiate the 2 "to" and then it can be translated accordingly.

Do you mean the fact that we have two keywords that use the word "TO"? That should be ok I think indeed the parses knows which one to expect (this is of course also covered by the unit tests)

Felienne (approved):

Issue #1405, opened 2021-11-26
Clear output of Ace console after unsuccessful execution of code


fRedelaar:
Currently whenever executing a program, the output appears in the console. It would be nice however if the console clears whenever code was unsuccessful to execute. 

![image](https://user-images.githubusercontent.com/33595265/143588179-88d0ce5a-de14-43e8-8bca-6737f76116ce.png)

The expected behavior is that Ace clears the right panel.

fRedelaar:
Closing since MR #1421 has been merged

PR #1421, opened 2021-11-29
Output is being cleared if a user made mistake

fRedelaar:
**Description**
Right now, when a child runs a program successfully there is output on the right screen which is really good. But if the child edits the code (that will run unsuccessfully) the output hasn't cleared on the output screen. This might come across to the child as if the code still works. This work is for issue #1405 

**Video of current situation**

https://user-images.githubusercontent.com/33595265/143907758-6ca55547-d165-477c-9288-397b3ba6e032.mov

**Video of merge requests result**

https://user-images.githubusercontent.com/33595265/143907751-c82bf145-ac31-4617-b58f-004e3f19b50d.mov


TiBiBa:
Hi! Great QOL suggestion. I noticed that the turtle canvas still remains visible (even after making a mistake). This is due to the `#output` only containing the ace editor output. You can fix this by adding something like `$('#turtlecanvas').empty();`

fRedelaar:
> Hi! Great QOL suggestion. I noticed that the turtle canvas still remains visible (even after making a mistake). This is due to the `#output` only containing the ace editor output. You can fix this by adding something like `$('#turtlecanvas').empty();`

Thanks @TiBiBa ! That's a really good suggestion that I totally missed. I've just sent a commit where I've added in your suggestion!

Felienne (approved):
really good idea team!

Issue #1350, opened 2021-11-21
Refactor keyword yamls


Felienne:
It would make sense to write it like

```suggestion
1: [print, ask, echo, turn, forward]
```

? That makes it easier to compare the list of keywords between levels, which may reduce the risk that you forget a keyword.

_Originally posted by @bjorn3 in https://github.com/Felienne/hedy/pull/1292#discussion_r753679478_

PR #1422, opened 2021-11-29
Refactored the yamls files

brucemartens:

**Description**

Refactored the yaml files as stated in issue #1350 .

### DIFF
--- a/grammars/suggestions-en.yaml
+++ b/grammars/suggestions-en.yaml
@@ -1,199 +1,23 @@
-1:
- - print
- - ask
- - echo
- - turn
- - forward
-2:
- - print
- - ask
- - is
- - turn
- - forward
-3:
- - print
- - ask
- - is
- - turn
- - forward
-4:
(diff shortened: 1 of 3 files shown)


grammars/suggestions-en.yaml
```diff
@@ -1,199 +1,23 @@
...
- - ask
- - is
- - if
- - for
- - elif
- - while
- - turn
- - forward
\ No newline at end of file
+1: [print, ask, echo, turn, foward]
```
Felienne, line 1: this says `foward` instead of `forward`

Felienne (changes requested):

grammars/suggestions-en.yaml
Felienne: (and lines 2 and 3 do so too)

teundemast:
This PR has become unnecessary when PR #1424 is accepted.

PR #1423, opened 2021-11-30
Translation structure for levels 3 to 7

Livavdv:
Added structure for translation of levels 3 to 7 to hedy_translation.py, issue #942.

Works the same way as the structure for levels 1 and 2.
Works if the keyword lark file is available for the chosen language and if the "local_keywords_enabled" is set to True.
Does not currently do anything on the Hedy site, only when tests are run.
Adds test files for the translation of levels 3 to 7.
Tests are available in the new test_translation_level_03 (and 04, 05, etc.) file.


teundemast:
I think something went wrong

PR #1424, opened 2021-11-30
Made suggestions more feasible

teundemast:
Changed the structure of getting keywords to be more flexible. I think the current way is very inconvenient when considering a lot of languages. 

Better fixes issue #1169


Felienne:
I agree this is a more feasible solution!

Felienne:
Ther are some issues now in the distance algorithm, can you fix these (see the failing tests)

teundemast:
Hi @Felienne, I think all should be good right now. The tests are also passed when local keywords are enabled.

teundemast:
I foresee one problem when for example 'to' is added to the commands per level. Namely, when this is the case, 'toe aan' of/en  'tot' should be added to the Dutch suggestions. With the current structure it is not possible to specify if 'toe aan' or 'tot' should be added. This is the case , because I considered the items in the dictionary 'commands_per_level' as 'English translations' of commands.

This could be fixed by considering the items in 'commands_per_level' not as English translations but as a command. So, when 'to' (for lists) or 'add' is added to  'commands_per_level', we should add 'to_list' and 'add_list'. If we do so, the 'translate_list_keywords' needs to be structured differently. 
Currently, commands_per_level is only used for suggestions, so changing this won't affect anything else. 

@Felienne, do you think it is better to consider commands_per_level as English translations or commands? I hope you understand what I mean, because it's a quite confusing 😅. 
Is there by the way a reason why keywords like 'sleep' or 'add' can't be suggested right now? 
Or is this the case because 'commands_per_level' is outdated?

Felienne:
> I foresee one problem when for example 'to' is added to the commands per level. Namely, when this is the case, 'toe aan' of/en 'tot' should be added to the Dutch suggestions. With the current structure it is not possible to specify if 'toe aan' or 'tot' should be added. This is the case , because I considered the items in the dictionary 'commands_per_level' as 'English translations' of commands.

I am not sure I understand, it occurs double in English so why can't we have two different suggestions in other langs?

> This could be fixed by considering the items in 'commands_per_level' not as English translations but as a command. So, when 'to' (for lists) or 'add' is added to 'commands_per_level', we should add 'to_list' and 'add_list'.

Yeah that is what I meant above!

> If we do so, the 'translate_list_keywords' needs to be structured differently. 

How does it need to be adapted? I am afraid I don't follow!

> Currently, commands_per_level is only used for suggestions, so changing this won't affect anything else.
 
> @Felienne, do you think it is better to consider commands_per_level as English translations or commands? I hope you understand what I mean, because it's a quite confusing 😅. 

For me too hahaha!

> Is there by the way a reason why keywords like 'sleep' or 'add' can't be suggested right now? Or is this the case because 'commands_per_level' is outdated?

I am quite sure that is because `commands_per_level` is outdated because I forgot to add these two new keywords

teundemast:
Capture
Sorry 😂, so basically the translate function I designed considers inputs to be the orange words (see picture above). But at least for the suggestions, I think a translate function should use the blue words as input. So the words in commands_per_level should be the blue words instead of the orange words. (I don't know how to name them, but I hope this clarifies it 😅). 
Do you agree @Felienne? 
Then, I will change the current translate function or add a new one.

teundemast:
I have updated the translate function used for suggestions as described above. I have also updated the dictionary commands_per_level (not fully complete yet). Should all the commands I added in this dictionary be suggested @Felienne, or is there some reason to not suggest some commands? 
I also think there needs to be a way to automatically create the dictionary commands_per_level. Or is it okay to hardcode this dictionary? Shall I make an issue about this?

Felienne:
> Capture
> 
> Sorry 😂, so basically the translate function I designed considers inputs to be the orange words (see picture above). But at least for the suggestions, I think a translate function should use the blue words as input. So the words in commands_per_level should be the blue words instead of the orange words. (I don't know how to name them, but I hope this clarifies it 😅). Do you agree @Felienne? Then, I will change the current translate function or add a new one.

Yes I see now! We indeed ned to take the blue words (keys) and not the En translations as input. Wil you change that in this PR?

teundemast:
> > Capture
> > Sorry 😂, so basically the translate function I designed considers inputs to be the orange words (see picture above). But at least for the suggestions, I think a translate function should use the blue words as input. So the words in commands_per_level should be the blue words instead of the orange words. (I don't know how to name them, but I hope this clarifies it 😅). Do you agree @Felienne? Then, I will change the current translate function or add a new one.
> 
> Yes I see now! We indeed ned to take the blue words (keys) and not the En translations as input. Will you change that in this PR?

As you can see in my most recent comment, it's changed :)

Felienne:
hmmm if I run this locally I get a path issue:

image

Ligt dat aan mij?

teundemast:
> hmmm if I run this locally I get a path issue:
> 
> image
> 
> Ligt dat aan mij?

Ik heb hetzelfde probleem. Wanneer ik '../' weghaal uit de pathing (zoals in mn vorige commit), werkt het locaal maar dan falen de tests op github. 
Wat dit veroorzaakt, is voor mij een raadsel.

teundemast:
@Felienne I have no idea how to fix this, as you can see in my recent commits. If the pathing is specified as ' ../coursedata' it does work in the test environment, but it does not work locally. When the pathing is specified as './coursedata' or 'coursedata' it does work locally, but it does not work in the testing environment. 
If it works in the testing environment, does that mean it will work if this is pushed to the server?

Felienne:
> @Felienne I have no idea how to fix this, as you can see in my recent commits. If the pathing is specified as ' ../coursedata' it does work in the test environment, but it does not work locally. When the pathing is specified as './coursedata' or 'coursedata' it does work locally, but it does not work in the testing environment. If it works in the testing environment, does that mean it will work if this is pushed to the server?

Yeah I think it is a configuration issue in your IDE. Mine (Pycharm) allows me to set this:

image

Can you do something like that as well? If not, you may leave this as is (github can also run it they way it is)

Felienne (approved):
Let's merge like this, path issues are local it seems

Felienne:
ok I think I fixes the path issue in a way that works!

Felienne (approved):

PR #1429, opened 2021-11-30
[FIX] Turtle hidden after program containing 'ask'

### LINKED ISSUE #1427: Running the forward example after the echo example in level 1 [BUG]
**Describe the bug**
In level 1, if I first run the `echo` example and then the `forward` example the output screen remains black and no line is shown. With the examples I mean the command examples from the pane on the left. This also happens when I use the `turn` example instead of the `forward` example. I think it has something to do with the `ask` command but I am not sure.

first I use this:
```
ask What is your favorite color?
echo so your favorite color is 
```
after running I replace the code with this:
`forward 100`

The feedback is positive but no line is drawn (see screenshot)

image

**Expected behavior**
I want to see a line when I run `forward 100`

TiBiBa:
**Description**
Currently when we run a program retrieving user input (by using 'ask') the program runs correctly. However, if the next program contains the turtle the turtle stays hidden. This is due to a `hide()` function on the turtle canvas to prevent overlay of the turtle on the user input when both are used in the same program. In this PR we solve this issue of a hidden turtle by checking if the program has a turtle and make sure the canvas is shown (even if already visible, just to be sure). 

**Fix for**
This PR fixes #1427 

**How to test**
First run a program containing user input but not a turtle (!). Next run a program only containing a turtle. The turtle should be visible and execute correctly.

**Checklist**


Felienne (approved):
Wow thanks for this incredibly quick fix!!

PR #1435, opened 2021-11-30
More information to InvalidInfo

dolfein:
**Description**
Adds line and column to the dataclass `InvalidInfo` for invalid nodes. They are not being used yet in this PR, but they will be crucial for [program_repair.py](https://github.com/Felienne/hedy/blob/main/program_repair.py).
Also removes the unnecessary conversion from (line, column) to an index in [program_repair.py](https://github.com/Felienne/hedy/blob/main/program_repair.py) (from #1339).

**Fix for**
Progress for #1225.

**How to test**
`invalid_info` in `is_program_valid` has attributes `line` and `column`, you can view it in debug mode or print it to the terminal (Lark starts counting at 1).

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -648,11 +648,9 @@ class Filter(Transformer):
         if all(bool_arguments):
             return [True] #all complete
         else:
-            command_num = 1
             for a in args:
                 if not a[0]:
-                    return False, a[1], command_num
-                command_num += 1
+                    return False, a[1]
 
     #leafs are treated differently, they are True + their arguments flattened
     def var(self, args):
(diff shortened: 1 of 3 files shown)


Felienne:
Nice start @dolfein! I am using your approach in a few more places, will merge when ready.

Felienne:
Finished! Well done again @dolfein, supercool Lark supports this out of the box, I would have found that myself!

I made a note in the code because we might move the v_args to Filter so everyone who inherits from it can use it, simplifying the code a bit, maybe you can take that into account in a future version?

Felienne (approved):

Felienne:
ow and @dolfein can you make a local branch next time instead of working on your own main, because this way it gives me an error since I can't run actions on your main.

dolfein:
Thank you!

> I made a note in the code because we might move the v_args to Filter so everyone who inherits from it can use it, simplifying the code a bit, maybe you can take that into account in a future version?

Sure thing, I tried it on `Filter` early on when I was using a different approach but then other problems arose, I will take another look at this.



> ow and @dolfein can you make a local branch next time instead of working on your own main, because this way it gives me an error since I can't run Github actions on your main, causing my mailbox right now to show this 🤣
> 
> image

Sorry😅 I will!

PR #1436, opened 2021-12-01
[Improvement] Return warning of user overwrites saved program

### LINKED ISSUE #1433: [BUG] Users can overwrite saved program by saving with identical program name
**Describe the bug**
Currently when saving a program it is stored with a chosen name. However, when storing a program with an identical name the stored program is overwritten. Even if this is written on a different level! This is, of course, undesirable behavior. Programs should only be overwritten if we open a previously stored program and then save it again. But even in that case we want to return the user with a warning such as "Saving this program will over-write an older version, are you sure?". In any other case of saving a program if the name already exists: return an error "you already have a program saved with this name".

**Edit:** This bug has some priority because of this bug students are able to overwrite submitted programs!

**Expected behavior**
When saving a program with the same name as an already saved program we want to return an "name already exists error". In the case that the user alters an already stored program (which is currently handled identical) we want to return a "are you sure you want to overwrite the older version" error. The second case should only occur iff the program is opened from the programs overview, otherwise we should view it as a new program.

TiBiBa:
**Description**
Currently users can store programs which are identified by their name. When saving a program with an identical name to a previously saved program the older version is overwritten without a warning. This is undesirable as this might accidentally overwrite a submitted program (automatically losing the submit status!). In this PR we implement a fix to return the user an additional confirmation pop-up if they are about to overwrite an already existing program.

**Fix for**
This PR fixed #1433.

**How to test**
Save a program with and without an already existing saved program name. Notice the additional pop-up to confirm saving the program with occur.

**Checklist**


TiBiBa:
I think this is ready for review! One issue I still have is that after this implementation we check if there is a duplicate program twice, creating unnecessary db calls. I tried working around this by storing the program id if we already found a duplicate program, but didn't work out as expected.

Felienne:
Nice idea @TiBiBa!

I think this fits with a bigger question about saving programs which currently is a bit of an undesigned nightmare, it more or less "happened" in this form.

Felienne:
@rix0rrr do you want to take a peek at the double db calls or is this ok with you?

rix0rrr:
> One issue we know create is that after this implementation we check if there is a duplicate program twice, creating unnecessary db calls.

Thanks for the diligence paying attention to this! 

Another subtlety I didn't discuss last time around we talked about this, is the following: how often is the code path invoked?

* If the DB call(s) happen(s) on every page load (let's say: 10 times per second), nearly every user will trigger the condition, and every millisecond counts! (*)
* If the DB calls happen only occasionally, let's say once per minute: we don't care as much, as their relative impact won't be as large.

So, *assuming* that saves don't happen as often, we are okay caring less about these database accesses than those on the "hot path". The question will always be: "do our assumptions indeed hold", but in this case I would guess it's okay to make that assumption.

Nevertheless: 

>  I tried working around this by storing the program id if we already found a duplicate program, but didn't work out as expected.

This interests me as well. This feels like a natural solution, hold on to the information so you don't need to look it up twice, so I'm curious to hear what "didn't work out as expected" means in practice 😅

> (*) For two reasons. 1️⃣: the $$$ cost of the DB calls themselves. And even if we don't care too much about those, 2️⃣: every millisecond waiting for a database response is a millisecond our server is not ready to service someone else's request. Add enough of those milliseconds together, and we're going to need to add new servers to service all the users, which itself costs $$$.

rix0rrr (approved):
After reading the source, I'm not even sure where the duplicate calls are.

I'm not sure you could be doing much better than this. Seems good to me! 👍

Felienne (approved):

PR #1447, opened 2021-12-02
Restructure .md files to .yaml structure

### LINKED ISSUE #1014: Replace .md structure with .html and *.yaml structure
Currently we create a separate .md-file for each language in the case of the learn-more, start and for-teachers pages. However, for other pages such as account and programs we use a .html template and add the language through the use of *.yaml-files. Which is a way more efficient way to add and alter languages. We should replace the old .md structure to align with the (easier) and more maintainable structure of .html templates and .yaml files.

TiBiBa:
**Description**
Currently some pages are structured using a template html page and rendering a translated .md file for the content. Other pages are similarly using a template page but fill their language-dependent content based on a .yaml file. This inconsistency in how we render pages is undesirable and leads to inconsistencies in UI between different languages. This inconsistency is due to the use of .md files with which the translator has influence on the front-end of the website. For example, a translator can (accidentally) use `##` instead of the `###` used in the original file when translating a page. Making it a `<h2>` instead of `<h3>` header. When using keywords in .yaml files to store translations mistakes like this can no longer occur.

In this PR we re-structure the rendering of pages to make all pages dependent on .yaml files. Enabling us to have control over the UI of the pages as well as enabling us to easier style pages for all languages without having to alter x amount of .md files. For all related pages a dedicated template is made (for example, `learn-more.html`) for which the .yaml file is retrieved of the language of choice. If it doesn't exist we automatically render the English one. As was the same case with the .md files.

Translation pages are structured as follows in the .yaml files:

```
<at the top possible page unique keywords we want to place at specific places in UI>
title: ""
sections:
    - title: ""
    text: |
    - title: ""
```

In the case of subsections the same structure can be continued such as is implemented in the for-teachers page. Notice that this structure can be repeated multiple times in the case we want to create sub subsections. The structure will look at follows:

```
<at the top possible page unique keywords we want to place at specific places in UI>
title: ""
sections:
    - title: ""
    subsections:
        - title: ""
        text: |
```

There are quite some inconsistencies between the translated content. For some languages several sections did not exist in the .md files. In these cases the English sections are kept in the translated files as well to make clear that these still need translating. It would be nice that, when updating the dutch or English page, we keep track of other languages being "out-of-date". Notifying possible translations through Github if they are willing to update a language file.

**Fix for**
This PR fixed #1014.

**How to test**
On the front-end side of the website nothing should have changed. Navigate to either the homepage, learn-more page or for-teachers page and visit all languages to verify the translations for each language are retrieved correctly and shown.

**Checklist**


Felienne (approved):
All seems to work fine!

PR #1457, opened 2021-12-02
fix: automatically log users in from long-lived cookies

### LINKED ISSUE #1425: Not showing For teachers, My profile and My programs tabs [BUG]
**Describe the bug**
When I go to hedycode.com I only see only 4 tabs: Home, Hedy, Learn More and Log in and not the three tabs (For teachers, My profile and My programs) that I need. When I go to Log in I am already logged in. Only when I logout and log back in again the 3 tabs (For teachers, My profile and My programs) appear. Every time I reopen my browser I have to do this again. I tried deleting my cookies but it did not help.

NA

**Expected behavior**
It would be nice if I can stay logged in and see all relevant tabs every time I go to hedycode.com 

I use Windows 10 with Firefox. I got the same problem when using Edge and Chrome.

rix0rrr:
The Hedy code base was operating in a split-brain scenario:

* Some code would inspect the session to find the currently logged
  in user. Sessions get cleared when the browser is closed.
* Some code would look at the long-lived login cookie and perform
  database lookups on every request to figure out what the currently
  logged-in user is.

The user was purposely stored in the session to save database calls
on every request, but if you close and reopen the browser the two
code paths would disagree:

- According to the session, you are now logged out.
- According to the login cookie, you are still logged in.

This PR unifies the behavior. The new behavior is as follows:

- All code will look at the session to determine whether you are logged
  in or not.
- The session would be cleared when you close the browser. In order to
  support persistent login sessions, if we detect you *don't* have
  a session user but you *do* have a login cookie, we immediately log
  you in from the cookie.

This restores the old behavior, with the advantage that we still save a
lot of database calls by keeping login information in the session.

Also contains a fix for development mode: `SECRET_KEY` will be set
to a consistent value automatically, so that if you restart your local
dev server your session doesn't automatically invalidate.

Fixes #1425.

### DIFF
--- a/app.py
+++ b/app.py
@@ -19,7 +19,7 @@ from ruamel import yaml
 from flask_commonmark import Commonmark
 from werkzeug.urls import url_encode
 from config import config
-from website.auth import auth_templates, current_user, requires_login, is_admin, is_teacher, update_is_teacher
+from website.auth import auth_templates, current_user, login_user_from_token_cookie, requires_login, is_admin, is_teacher, update_is_teacher
 from utils import timems, load_yaml_rt, dump_yaml_rt, version, is_debug_mode
 import utils
 import textwrap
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -206,7 +231,9 @@ def before_request_https():
     app.config['SECRET_KEY'] = os.getenv('SECRET_KEY')
 
 else:
-    app.config['SECRET_KEY'] = os.getenv('SECRET_KEY', uuid.uuid4().hex)
+    # The value doesn't matter for dev environments, but it needs to be a constant
+    # so that our cookies don't get invalidated every time we restart the server.
+    app.config['SECRET_KEY'] = os.getenv('SECRET_KEY', 'WeAreDeveloping')
```
dvberkel, line 236: I propose to change the key `WeAreDeveloping` to `WeAreInventing`, honoring the brilliance of Hedy Lamarr

dvberkel (approved):
I approve these commits with a proposal for a single change

TiBiBa (approved):

Issue #1448, opened 2021-12-02
[UI idea] Keep the scrollbar in the output screen down with long codes.


MarleenGilsing:
When you're running longer codes, everytime you answer a question the output screen scrolls up after you answer the question. So you have to scroll down to see the rest of the code. It would make more sense if you didn't jump to the beginning of the code everytime you answer a question. 

For example: play the calculator game in level 9 (till the end) and you'll know what I mean.

Felienne:
Maybe this is a task for you @fRedelaar?

fRedelaar:
> Maybe this is a task for you @fRedelaar?

I’ll take a look at it :)

PR #1458, opened 2021-12-02
Editor output will scroll automatically down with new input

fRedelaar:
**Description**
Possible fix for #1448. 

**Fix for**
Users had to scroll down in the output if there were too many print statements. 

**Currently**

https://user-images.githubusercontent.com/33595265/144500608-b9447e80-2cbf-43c1-a966-4ad90604017b.mov

**Result of this Pull Request**

https://user-images.githubusercontent.com/33595265/144500629-ba37db58-a433-4c61-8c55-c1eec0830676.mov


TiBiBa (approved):
Love this QOL fix! Awesome!

templates/incl-editor-and-output.html
```diff
@@ -117,4 +117,11 @@
     </div>
     <div class="ml-3">{{ developers_mode }}</div>
   </label>
-</div>
\ No newline at end of file
+</div>
+<script>
+function ScrollOutputToBottom()
```
Felienne, line 122: Nice fix @fRedelaar, but maybe we want to place this code in app.ts rather than directly in the HTML? Or is there a reason that that is not possible?

fRedelaar: Nope no reason for this! I will create a new PR to place the code in app.ts. Thanks for the review, my bad!

fRedelaar: Moved the code at PR #1460

PR #1460, opened 2021-12-03
Auto scroll output, moved js code from .html to app.ts

fRedelaar:
**Description**

Follow-up PR for #1458 . Moved js script from the HTML file to the app.ts file. 

> Nice fix @fRedelaar, but maybe we want to place this code in app.ts rather than directly in the HTML? Or is there a reason that that is not possible?

_Originally posted by @Felienne in https://github.com/Felienne/hedy/pull/1458#discussion_r761951776_


Felienne (approved):
Thanks for the fix @fRedelaar, it is a bit cleaner to have all scripts in one place!

PR #1465, opened 2021-12-04
Translation structure level 8-9 + tests

emil-meyknecht:
Added level 8 and 9  into the translation structure.

These levels do not add any new keywords but correctly format the code if indentation is present.
For this the code is first passed to the "hedy.preprocess_blocks()" function.
The "Indent()" function add the right amount of white space in front of the code lines.
For now only a indentation of 4 white spaces works correctly

-Adds support for indentation after the repeat, if and else command.
-Adds the "indent()" function to hedy_translation.py.
-Adds tests for levels 8 and 9.

Tests available with test_translation_level_08.py and test_translation_level_09.py


hedy_translation.py
```diff
@@ -51,10 +52,11 @@ def translate_keywords(input_string, from_lang="en", to_lang="nl", level=1):
 
     punctuation_symbols = ['!', '?', '.']
 
+    keywordDict = keywords_to_dict(to_lang)
```
Felienne, line 55: deze var naam is niet zo "Pythonic". Dit is beter:
```suggestion
    keyword_dict = keywords_to_dict(to_lang)
```

Felienne (changes requested):

emil-meyknecht:
@Felienne The variable name was indeed not very "Pythonic". I Believe it has been changed everywhere.

PR #1466, opened 2021-12-04
Move profile button, rename, add icon

rix0rrr:
In the current website, when you log in two new buttons appear:

"My profile" and "My programs".

This PR makes those buttons stand out more from the rest of the menu bar, to mimic the way other sites behave:

- The "you" buttons align visually to the right instead.
- Rename the "my profile" button, showing the user's username instead.
- Add icons to the buttons (using [Font Awesome](https://fontawesome.com/v5.15/icons?d=gallery&p=2&s=solid&m=free)) to visually distinguish them a bit more from the rest.


rix0rrr:
Old:

image

New:

image

templates/incl-menubar.html
```diff
@@ -14,8 +14,9 @@
       {% endif %}
     {% endfor %}
     {% if username %}
-      <a class="menubar-btn {% if current_page == 'my-profile' %}border-white{% else %}border-transparent{% endif %}" href="{{localize_link ('/my-profile')}}">{{auth.profile}}</a>
-      <a class="menubar-btn {% if current_page == 'programs' %}border-white{% else %}border-transparent{% endif %}" href="{{localize_link ('/programs')}}">{{auth.program_header}}</a>
+      <div class="flex-grow"></div>
+      <a class="menubar-btn {% if current_page == 'my-profile' %}border-white{% else %}border-transparent{% endif %}" href="{{localize_link ('/my-profile')}}"><span class="fas fa-fw fa-user"></span> {{username}}</a>
```
TiBiBa, line 18: Can't we just use `ml-auto` on this element to make sure they float to the right?

TiBiBa (approved):

TiBiBa:
Really love this UI change on the menubar! I was thinking; might be nice to hide the language selection as well when a user is logged in. I assume a logged in user has a native language and doesn't switch between languages. We can maybe enable the user to select a preferred language for Hedy in the user settings. (Which we will always use when the user is logged in, if not available we use the defined fallback language).

Felienne:
> Really love this UI change on the menubar! I was thinking; might be nice to hide the language selection as well when a user is logged in. I assume a logged in user has a native language and doesn't switch between languages. We can maybe enable the user to select a preferred language for Hedy in the user settings. (Which we will always use when the user is logged in, if not available we use the defined fallback language).

Love this idea! Save space in the menu bar. Shall we implement that by showing the switcher in view when kids log in first and then, when they select a language, show a popup telling them it is now set and can be changed in their settings?

PR #1471, opened 2021-12-05
refactor: remove need for passing TRANSLATIONS object everywhere

rix0rrr:
This is another in the line of cleanup PRs: it reduces the unnecessary
duplication of invoking and passing `TRANSLATIONS` everywhere.

- We automatically inject the `ui`, `auth` and `texts` objects
  into the template, so they don't need to be passed all the time.
  - this needed updating some templates since they were expecting to
    see strings from the `ui` object directly (referencing them as
    `{{ run_button }}`, which turned into `{{ ui.run_button }}`.
- We make the `page_404`, `page_403` and `page_500` handlers
  look up their own keys in the translations so we don't get so much
  repetition (do this by sticking the UI strings for the
  current language in `g.ui_texts` so they are available everywhere)

Does not introduce any change to the website itself, everything
looks and works the way it did before.

Relates to #1179.


Felienne:
> This is another in the line of cleanup PRs: it reduces the unnecessary
> duplication of invoking and passing TRANSLATIONS everywhere.

Thanks for the further cleanup @rix0rrr!! 

> * We can make the `page_404`, `page_403` and `page_500` handlers
>   look up their own keys in the translations so we don't get so much
>   repetition (do this by sticking the UI strings for the
>   current language in `g.ui_texts` so they are available everywhere)

I am not exactly sure what this means, it looks like there is still a bit of work to do for the error pages? Should we make an issue for that?

rix0rrr:
> I am not exactly sure what this means, it looks like there is still a bit of work to do for the error pages? Should we make an issue for that?

Sorry, wrote that comment in the wrong tense. It is already done.

Felienne:
> > I am not exactly sure what this means, it looks like there is still a bit of work to do for the error pages? Should we make an issue for that?
> 
> Sorry, wrote that comment in the wrong tense. It is already done.

Ah got it, thanks!

Felienne (approved):

PR #1479, opened 2021-12-06
translation structure for levels 10-12 + tests

emil-meyknecht:
First PR of 3 to finalize the translation structure up to level 17.
Issue #1451 
Added level 10, 11 and 12 into the translation structure.

-Adds the translation for the "for_loop" command
-Adds the translation for the "while_loop" command
-Adds the "quoted_text" function to put quotes around values used for the 'ifelse' and 'is' commands

Tests available with test_translation_level_10.py, test_translation_level_11.py and test_translation_level_11.py


emil-meyknecht:
@Felienne we have worked on the translation structure quite a bit this weekend and have finished up to level 17. I do not know the best way to put this in a pull request. Would you prefer seeing everything together in one PR or should I split them in smaller ones ?

tests/test_translation_level_10.py
```diff
@@ -0,0 +1,54 @@
...
+    level = 10
+
+    @check_local_lang_bool
+    def test_for_english_dutch(self):
+        code = "dieren is hond, kat, vis\n" \
+               "for dier in dieren\n" \
+               "    print 'hallo' dier"
+
+        result = hedy_translation.translate_keywords(code, from_lang="en", to_lang="nl", level=self.level)
+        expected = "dieren is hond, kat, vis\n" \
```
Felienne, line 23: Shall we use `textwrap.dedent` here (like the regular tests do)? That reads a bit nicer without the \

Felienne (changes requested):
Can we use textwrap.dedent?

tests/test_translation_level_10.py
emil-meyknecht: @Felienne Yes, we forgot the regular tests use that. We can update tests to use textwrap.dedent

Felienne:
> @Felienne we have worked on the translation structure quite a bit this weekend and have finished up to level 17. I do not know the best way to put this in a pull request. Would you prefer seeing everything together in one PR or should I split them in smaller ones ?

I am ok with one big one, it seems the code is all quite similar!

tests/test_translation_level_10.py
Felienne: Thanks, please do (for future levels also, of course)

hedy.py
```diff
@@ -157,17 +157,48 @@ def promote_types(types, rules):
 
 character_skulpt_cannot_parse = re.compile('[^a-zA-Z0-9_]')
 
+
+def get_list_keywords(commands, to_lang):
```
emil-meyknecht, line 161: @Felienne This function was causing a circular import so I moved it to Hedy.py, with permission of @teundemast  
It does bring up issue #1482 since not everyone knows that the boolean has to be set to True for the tests to run

Felienne: I see! Can we merge this now (it seems the tests are passing?) or do we have to wait until #1482 is fixed?

tests/test_translation_level_11.py
```diff
@@ -0,0 +1,44 @@
...
+                               "    print counter")
+
+        result = hedy_translation.translate_keywords(code, from_lang="en", to_lang="nl", level=self.level)
+        expected = textwrap.dedent("voor counter in bereik 1 tot 5\n"
+                                   "    print counter")
+
+        self.assertEqual(result, expected)
+
+    @check_local_lang_bool
+    def test_for_in_dutch_english(self):
```
Felienne, line 32: Almost right :) 

with textwrap we don't need the \n's (that was sort of the point haha), so we can write it like this:

```suggestion
    def test_for_in_dutch_english(self):
        code = textwrap.dedent("""\  # this is to escape the newline that follows
        nummer is vraag 'hoe oud ben je'"
        voor counter in bereik 1 tot 5"
            voor count in bereik nummer tot 0"
                print 'hoi' counter""")
```

Felienne (dismissed):
See textwrap example

hedy.py
emil-meyknecht: I can add it in this branch, since i need to update the textwrap dedent anyway :)

emil-meyknecht: I mean add Issue #1482

Felienne: Yeah I got that, let's do it here than so all is in order at once.

Thanks!

emil-meyknecht:
@Felienne I believe the proposed fix for #1482 might not work in the test environment used for the check on Github. I am not yet sure how to fix this but I'll will look around a bit

emil-meyknecht:
After a lot of trying the problem has been found. First of all a check has been added to local_lang_bool has been added for the github environment #1482 . 
There seemed to be another problem when trying to reach the ./coursedata/keywords/nl.yaml file when in the Github environment. A fix for this has been added in hedy_translation.
Hopefully everything is working correctly now

Felienne:
> @Felienne I believe the proposed fix for #1482 might not work in the test environment used for the check on Github. I am not yet sure how to fix this but I'll will look around a bit

You can also run the github tests locally (with buildtools/validate) but what exactly seems to be the issue? All the tests pass on GitHub so I am not sure I fully understand...

emil-meyknecht:
@Felienne The Issues have all been solved. I do not know exactly why it didn't work, but when running the github tests locally everything worked just fine, only not when it was ran on github after a git push was made. But it all works fine now.

> After a lot of trying the problem has been found. First of all a check has been added to local_lang_bool has been added for the github environment #1482 . There seemed to be another problem when trying to reach the ./coursedata/keywords/nl.yaml file when in the Github environment. A fix for this has been added in hedy_translation. Hopefully everything is working correctly now

tests/test_translation_level_08.py
```diff
@@ -15,68 +15,72 @@ class TestsTranslationLevel8(HedyTester):
...
         self.assertEqual(result, expected)
 
-    @check_local_lang_bool
     def test_repeat_multiple_indent_english_dutch(self):
-        code = "repeat 3 times\n" \
-               "    print 'Hedy is fun!'\n" \
-               "    print 'print 3 keer'\n" \
-               "print 'print 1 keer'"
+        code = textwrap.dedent("""\
+                                repeat 3 times
```
Felienne, line 32: this is ok, but I would align it with the assignment like this, that is a bit more readable:
```suggestion
        code = textwrap.dedent("""\
        repeat 3 times
        etc...
```

emil-meyknecht:
@Felienne It does indeed look better without all the extra whitespace

emil-meyknecht:
@Felliene That fix will probably also work. In this branch I used
image

This also solved the problem.
But the solution you proposed also seems to work so I can implement that in this branch

Felienne:
> @Felliene That fix will probably also work. In this branch I used image
> 
> This also solved the problem. But the solution you proposed also seems to work so I can implement that in this branch

This will not work from the tests I fear, but let's see!

emil-meyknecht:
@Felienne I added `dir = path.abspath(path.dirname(__file__))` which seems to work fine as far as I can see

Felienne (approved):

Issue #1414, opened 2021-11-27
Put confetti emojis in Yaml for easier customization


fRedelaar:
> Just to summarize my question from yesterday: I would prefer that these emoji live in coursedata/adventures/ so they are more easy to customize.
> 
> There is logic to grab data from the yamls and use it in the frontend already so that can be repurposed here.
> 
> _Originally posted by @Felienne in https://github.com/Felienne/hedy/issues/1411#issuecomment-980622042_

PR #1481, opened 2021-12-06
Moved adventure confetti emojis to yaml 

fRedelaar:
**Description**

For issue #1414 the confetti emojis have been moved from app.ts to the English yaml file. This is so content developers can adjust emojis with more ease in the future.


coursedata/texts/en.yaml
```diff
@@ -52,6 +52,13 @@ ui:
     not_enrolled: "Looks like you are not in this class!"
     retrieve_class: "Only teachers can retrieve classes"
 ClientErrorMessages:
+    Emoji_turtle: "🐢"
```
Felienne, line 55: Great you are working on this @fRedelaar! However, maybe we can make a dictionary here, that would look a bit nicer? Something like:

Client
```suggestion
    ClientErrorMessages:
        turtle: "🐢",
        rock: "✂️, 📜, 🗻"
```

Then in ts we can look up a list with the name of the adventure, simplifying the code too? (see second suggestion)

static/js/app.ts
```diff
@@ -874,15 +874,26 @@ function getConfettiForAdventure(adventure: string){
 
   switch (adventure) {
     case 'turtle':
-      return [['🐢']];
+      let turtleEmoji = Array.from(ErrorMessages['Emoji_turtle'])
```
Felienne, line 877: With the dict in yaml, we can do something like `ErrorMessages[adventure]` here, reducing the neem for the switch.

Felienne (changes requested):

static/js/app.ts
fRedelaar: Great suggestion! I've pushed a new commit.

Felienne (approved):
Awesome, thanks!!

Issue #1219, opened 2021-11-09
Show a special "well done"-message when the newest command learned is mentioned in the code


remva:
Right now, you get a "well done" message everytime you submit a code that works. So, you can submit the code "print 'hi'" in every level and get a "well done" message. The message could be improved.

PR #1483, opened 2021-12-06
Specific success message (#1219)

remva:
With this pull request, it should be possible to see more specific success messages for levels 1 to 7.

Right now, you get a success message whenever you type a code that works. But with these code additions, you'll sometimes see the text "Your code works, but you forgot to use the newly learned commands". This happens when you submit "print hello world" in level 2 for example. Then you'll receive this message because you didn't use "ask" or "is", which were the new commands learned.

image

There were 3 things added to the existing code.
- There is a function `newly_learned_subjects(level: any, code: string)` added in file "app.ts". This boolean function checks if a newly learned command is used in a kids code. True gives a regular success message, like "well done". False gives the new message. If you check this function for level 8 or higher, the code will automatically return "true" en show the normal success message. This is done with a try-catch statement. 

```
export function newly_learned_subjects(level: any, code: string) { 
  try {
    var necessary_commands = ErrorMessages['New_subjects'][level];
    for ( var i = 0; i < necessary_commands.length; i++ ) {
      if ( code.includes(necessary_commands[i]) ) {
        return true
      }
    }
    return false
  }
  catch(err) {
    return true
  }
}
```

- There is an addition to app.ts within the function "export function runit(level: string, lang: string, cb: () => void)". Under "if (response.Code)" the function "newly_learned_subjects" is now called. If it returns True, a normal success message will be shown. If it returns False, the code is incomplete. The new message will be shown. 
A small part of the code says `if ( typeof allsuccessmessages === 'object' )`. This checks if there are several success message in the yaml-file, so "allsuccessmessages" is a list. The else statement works if there's only one succes message.

- There is a list added in the en.yaml file called `New_subjects`. This lists the new commands you learn in level 1 to 7.

This is a fix for issue #1219 it belongs to branch Specific_success_message


Felienne:
Thanks @remva! There is an issue in app.ts that I do not know how to resolve, could you have a look at that?

coursedata/texts/en.yaml
```diff
@@ -47,6 +47,7 @@ ClientErrorMessages:
       - "Well done!"
       - "Excellent!"
       - "You did great!"
+    Transpile_incomplete: "Your code works, but you forgot to use the newly learned subjects!"
```
Felienne, line 50: Can we replace subjects by 'commands' here? That is the word we commonly use. Also... can you add an nl translation too? In principle, the Hedy code team delivers all features in nl and en.

coursedata/texts/en.yaml
```diff
@@ -56,6 +57,32 @@ ClientErrorMessages:
     CheckInternet: "Have a look if your Internet connection is working properly?"
     ServerError: "You wrote a program we weren't expecting. If you want to help, send us an email with the level and your program at hedy@felienne.com. In the mean time, try something a little different and take another look at the examples. Thanks!"
     Program_too_long: "Your program takes too long to run."
+    New_subjects:
```
Felienne, line 60: Here too can we use new_commands, and... I know it is more complex but can we extract this from hedy.py?

We have a list of commands per level there too and it would be nicer if we do not have to do double bookkeeping

static/js/app.ts
```diff
@@ -185,6 +185,21 @@ function clearErrors(editor: AceAjax.Editor) {
   }
 }
 
+export function newly_learned_subjects(level: any, code: string) {
+  try {
+    var necessary_commands = ErrorMessages['New_subjects'][level];
+    for ( var i = 0; i < necessary_commands.length; i++ ) {
```
Felienne, line 191: ok I know it is a bit nitpicky but... if we just match strings it is not very precise. For example if I write `print random` it would count as a use of random. I think it should be possible to extract a list of commands in `hedy.py` and use that to match here?

(but we can also leave that for your BSc thesis of course and merge this now as it is :D )

Felienne (changes requested):

coursedata/texts/en.yaml
Felienne: If you convert the lists into sets you can do a set subtraction to get the difference I think, something like `set(list_2) - set(list_1)`

Jasper1408: We thought it would be best to have a different set of commands per level. Because in later levels there are no new commands learned, only the way the commands work has changed.  In level 4 the ' character is introduced and these are not a command.

static/js/app.ts
remva: Wow, we didn't think of that, but `print random` is very clever haha. @Jasper1408 and I haven't found the solution for this yet. Since the official due date for our project is today, I'd like to take the opportunity to finish this for my BSc thesis! 

We pushed solutions to the first 2 comments you posted. We'd like to merge this now as it is.

coursedata/texts/en.yaml
remva: Solved with commit: '#1219 Dutch translation transpile_incomplete'

remva:
> Thanks @remva! There is an issue in app.ts that I do not know how to resolve, could you have a look at that?

Solved with commit '#1219 issue surrounding 'never' solved'

static/js/app.ts
Felienne: Hahaha yeah kids come up with the cleverest programs :D Totally fine if you pick this up for your thesis, will you make an issue for it an assign yourself?

Felienne:
> > Thanks @remva! There is an issue in app.ts that I do not know how to resolve, could you have a look at that?
> 
> Solved with commit '#1219 issue surrounding 'never' solved'

Ah it is already solved, cool! Will check out this PR in a bit

Felienne:
ow we got some non-trivial merge conflicts now, could you try to fix those? I am afraid to destroy something!

coursedata/texts/en.yaml
Felienne: Ok, I can see how we prefer to keep this list, but can we rename it to `new_commands`? Subjects is a term that we do not use in other places so that will be confusing.

static/js/app.ts
```diff
@@ -188,6 +188,21 @@ function clearErrors(editor: AceAjax.Editor) {
   }
 }
 
+export function newly_learned_subjects(level: any, code: string) {
```
Felienne, line 191: Can we change this too:

```suggestion
export function newly_learned_commands(level: any, code: string) {
```

(and of course make the same change in other relevant places)

Felienne (changes requested):

Felienne:
Ah on a second thought, this feature might be too similar to the unlocking of levels, so maybe it is best to integrate it there in a more generic way? So let's close this for now but the code can be used in the unlocking config!

PR #1484, opened 2021-12-07
Repair indentations, missing quote and unexpected char

dolfein:
**Description**
Repairs for 3 (UnexpectedCharacters), 4 (UnquotedTextException) and 6 ((No)IndentationException) from #1225.

The repair for 4 only puts a quote at the begging or end, this will be expanded to take numbers and variables into consideration, but it at least removes the error for now.

The repairs are stored in `fixed_code`.

**Fix for**
#1225 

**How to test**
Raise one of the above exceptions and see the content of `fixed_code`.

**Important**
I moved `meta` to `Filter` and it gets accessed in `IsValid` again.

For some reason `meta.column` was always 1 when `meta` was in `IsValid`. So for example if Hedy code on level 4 or higher is:
`print Hello'`
The invalid node would be "Hello'", which is correct since it misses a quote, but `meta.column` would be 1 instead of 7.

I tried to debug it and find out why, but I have no logical explanation why this is the case.
Moving `meta` to `Filter` seems to fix this, but I am not sure if it covers every invalid node this way, since `meta` is now only available in the leaves.


program_repair.py
```diff
@@ -23,5 +23,37 @@ def replace(input_string, line, column, length, new_string):
...
-    # the only repair we can do now is remove leading spaces, more can be added!
     return '\n'.join([x.lstrip() for x in input_string.split('\n')])
+
+
+def remove_unexpected_char(input_string, line, column):
+    return delete(input_string, line, column, 1)
+
+
+# TODO: check if the user tried to print a variable
+def add_missing_quote(input_string, line, column, length):
```
Felienne, line 34: This is called add while the other function is called fix, can we make that the same?

Felienne:
Hai Dolfein!

Thanks for this. 

A few notes:

* The `help wanted` label is not needed for PR's we will always help you get things merged or closed. You might have noticed no other PR's use this label.
* What would greatly help if you want better help is to expand **how to test**. For example, add a program and a level giving the exception and give me the program you expect to the the repaired program, so I can run the code and see if I get the same result. Can you add that?
* Also, the tests fail, can you fix these first? Without passing tests we can't merge and it seems your questions are not related to the passing tests, although I am not sure because I have no example runs.
* On your remark: I wanted to move meta to Filter anyway so that to me seems to be a good move, I don't know why that would mean we can only access it in leaves, but I would have to dive deeper into the code. 
* Finally, these are three exceptions, maybe you can split it into one exception for one PR. The smaller, the easier it is for me to debug and to help you. This already is quite large because I need to test many things.

hedy.py
```diff
@@ -665,16 +665,25 @@ def program(self, args):
...
-        return True, ''.join([str(c) for c in args])
-    def random(self, args):
-        return True, 'random'
-    def punctuation(self, args):
-        return True, ''.join([c for c in args])
-    def number(self, args):
-        return True, ''.join([c for c in args])
-    def text(self, args):
-        return all(args), ''.join([c for c in args])
+    @v_args(meta=True)
```
Felienne, line 668: Ah I see why this means we only have access to the leafs. Could we add a default rule here that passes `meta`?

Felienne: (and then calls the method on the inheriting method)

Felienne (changes requested):

hedy.py
```diff
@@ -1748,6 +1760,7 @@ def preprocess_blocks(code, level):
                 raise hedy.exceptions.LockedLanguageFeatureException(concept="nested blocks")
 
         if next_line_needs_indentation and current_number_of_indents <= previous_number_of_indents:
+            fixed_code = program_repair.fix_indent(code, line_number - 1, leading_spaces, indent_size)
```
Felienne, line 1763: If we always subtract - from the line number, maybe we can do that in the method?

PR #1487, opened 2021-12-07
[UI Improvement] Implement and add page titles to each render_template

### LINKED ISSUE #1449: [BUG] Page titles are inconsistent
**Describe the bug**
Page titles are currently inconsistent. The pages that rely on a .md file had a correct page_title as this is given in the file itself. However, all other pages such as for-teachers and login just so the url. This is not likely and should be solved.

**Expected behavior**
Each page should so a clear page title

TiBiBa:
**Description**
Currently the handling of page titles is a bit of a mess. With the older .md file structure some titles where retrieved from these files whereas the titles of other pages, such as `my-profile` and `for-teachers` were hard-coded or didn't exist. With this PR we align all page title translations into one .json file, enabling us to easier add, translate and retrieve these titles. A function is added to enable each `render_template` to retrieve the correct page title.

**Fix for**
This PR fixed #1449

**How to test**
Navigate through the website and notice that a correct page title is shown for each page. Currently they are only translated in dutch and English. The automatic fallback language is English. If a key in the dict doesn't exist we automatically return the title of the English homepage ("Hedy - A gradual program language").

**Checklist**


templates/layout.html
```diff
@@ -1,7 +1,7 @@
 <!doctype html>
 <html>
   <head>
-    <title>{{page_title}}</title>
+    <title>{% if page_title %}{{page_title}}{% else %}Hedy{% endif %}</title>
```
Felienne, line 4: In the R description you write: "If a key in the dict doesn't exist we automatically return the title of the English homepage ("Hedy - A gradual program language")"

but this seems like it will be 'Hedy' when there is no title? (Not that it matters a lot to me either way :))

TiBiBa: We return the start title for "en" if for some reason we can't find the key or lang in the dict. But if for some reason we add no `page_title` at all to the `render_template` we will still show "Hedy" as title (and prevent the page from crashing).

Felienne: Ah got it!

Felienne:
Ah nice this fixes something that was bugging me for a year!!!!

image

Felienne (approved):

PR #1494, opened 2021-12-07
Move meta from IsValid to Filter

dolfein:
**Description**

Moves `meta` to `Filter` and it gets accessed in `IsValid` again.

For some reason `meta.column` was always 1 when `meta` was in `IsValid`. So for example if Hedy code on level 4 or higher is:
`print Hello'`
The invalid node would be "Hello'", which is correct since it misses a quote, but `meta.column` would be 1 instead of 7.

I tried to debug it and find out why, but I have no logical explanation why this is the case.
Moving `meta` to `Filter` seems to fix this, but I am not sure if it covers every invalid node this way, since `meta` is now only available in the leaves.

**Fix for**

Progress for #1225 

**How to test**

Example: in level 4 or higher run the following Hedy code:
`print 'Hello'`
`print 'World`
`line` and `column` from `InvalidInfo` now contains the correct positions for the invalid node 'World (2 and 7, respectively).


dolfein:
This PR is still a draft since @Felienne suggested some changes in #1484.

Felienne:
I am not sure how to handle this PR, because I don't know what the plan is. Can you clarify? Are you planning to fix the leaves issue in this PR going forward or do more debugging about the column number?

Felienne:
Hi @dolfein!

I am looking into this issue and I have made a branch for us to explore: https://github.com/Felienne/hedy/tree/meta-issue

The first thing I have done is make a test so we have traceability. It was a bit hard to make the test because your example is not a true issue since we will throw an unquoted exception and do not expose the error location to the user in that case.

So I am also wondering whether you have an example in mind where we show the wrong error location to the user? That would make it easier to test!

Felienne:
The test though is interesting because in the current main, the v_args are defined on IsValid, *not* on Filter and yet we get line 2, column 1 and not column 7 as a result!!

dolfein:
That is also what I observed:
> For some reason `meta.column` was always 1 when `meta` was in `IsValid`.

Before, when it was in `Filter`, the column would be correct.

> So I am also wondering whether you have an example in mind where we show the wrong error location to the user? That would make it easier to test!

I have not come across an error message that shows the wrong column to the user. The only error message that even shows the column is the parse error and that uses the `UnexpectedCharacters` exception from Lark itself (which is probably the reason why column does work here):
https://github.com/Felienne/hedy/blob/0278ff4db527f4974392384aca9be720d2c7947c/hedy.py#L2002-L2010

The issue was not that a wrong column was shown to the user, but that some repairs require that information.

Felienne:
> That is also what I observed:
> 
> > For some reason `meta.column` was always 1 when `meta` was in `IsValid`.
> 
> Before, when it was in `Filter`, the column would be correct.

Ok, but then could we move it to Filter? I have tried to make a few steps towards that but I ran into other issues. Did you find the same? What don't we move it back?

> The issue was not that a wrong column was shown to the user, but that some repairs require that information.

Ok clear! But it would still be helpful to have a set of examples that fail so we have something that guides us in the change. Can you list a few, do you have some examples?

boryanagoncharenko:
@dolfein I did some amendments to your PR, so now the test succeeds. However, there are a couple of issues that you need to take into account: [see comment ](https://github.com/Felienne/hedy/pull/1494/files#diff-2f9217cafc0f8d8df0537257ce1f91b8c54434877c960dbd55bb3dd17941e9a4R901-R911) 

I am not sure whether the current solution will allow you to provide the specific column of the error in following code on level 5:
```
a is hello
print a 'world
```
Let me know if you need more help changing the transformer.

Felienne:
Superceded by #1685

Felienne (approved):

Felienne:
Sorry missed that #1685 was not into main!

Felienne:
@dolfein This seems to fix the problem of moving v_args, but there are some open issues that might not work straight away, see @boryanagoncharenko's comments in the code. If you are happy with the change, can you approve it?

The PR is set to auto merge, so as soon as you give the ok it will be merged into main

tests/test_level_04.py
```diff
@@ -407,3 +407,46 @@ def test_chained_assignments(self):
...
+    is_valid = instance.transform(program_root)
+    _, invalid_info = is_valid
+
+    invalid_info = invalid_info[0]
+
+    line = invalid_info.line
+    column = invalid_info.column
+
+    # Boryana Jan 22
+    # Proabably we want to change the column so that it points to
```
dolfein, line 429: I was thinking of doing this check in the repair itself, since we know what and where the error is.

hedy.py
```diff
@@ -876,36 +881,47 @@ def all_print_arguments(input_string, level, lang='en'):
...
+
+        # Boryana Jan 22
+        # The meta variable passed as a parameter here contains the metadata of the current node.
+        # The args parameter is an array of the triples, of which the last part holds the metadata of the
+        # IMMEDIATE children. Two important points:
+        # - If an error rule has more than 2 levels of nested children, we lose the metadata of the
+        # deeper ones and get only the data of the immediate children. This does work for now, because the parse trees
+        # of all error grammar rules are only 1 level deep. For now.
+        # - At level 5, the error_print_nq can have multiple arguments and we cannot assume that it is the first one
+        # that has a missing quote. Further analysis is required to pinpoint where the quote should be added.
```
dolfein, line 910: @boryanagoncharenko First of all, thank you of course!

I get the problem, so if there are error rules with deeper levels, we only have meta available from the children and not the children's children. So we want to have the meta from the deeper nodes also available (in `args` again?).

But like you said, our error rules are currently only 2 levels. So I am not sure if we can implement this already, without have those deeper level error rules.

Felienne: Well there are 2 issues here:

* error grammar rules can be deeper -> yes, you are right we could leave that for now, although I would prefer this to be fixed fo the future
* error_print_nq can have multiple arguments -> further analysis is needed to pinpoint the right location of the missing quote

Can youat least look into point 2 @dolfein? This is also visible from the tests around this that do not always show the right place.

boryanagoncharenko: Let's consider the following program at level 5:
```
name is 'B'
L is 1, 2, 3
print 'Person ' name ' said ' L at random things'
```
The print node in the AST produced by the above code will have 5 arguments: 
- (text, "Person '")
- (var, ['name'])
- (text, "' said '")
- (list_access, [...])
- (text, "things'")

We want to skip the children of types var and list access and examine only the ones of type 'text'. So when we reach the error_print_nq method in the IsValid transformer, we need the type, value and meta information for each children. However, IsValid is a transformer which by default is transforming bottom up. This means that by the time we reach the error_print_nq, its children are transformed, so we lose the information we need.

The easiest way to address the problem at hand is to use a visitor instead. Consider the following code:
```
class IsValidVisitor(Visitor):
    def error_print_nq(self, tree):
        pass
...
def is_program_valid(program_root, input_string, level, lang):
    visitor = IsValidVisitor()
    visitor.visit(program_root)
...
```
[
NOTE Please have in mind that there is a tiny improvement/fix in the grammar that needs to be done in the grammar of level 5. The line 
```error_print_no_quotes: _PRINT _SPACE (textwithoutspaces | list_access | var) (_SPACE (textwithoutspaces | list_access | var))*  -> error_print_nq``` needs to change to ```error_print_no_quotes: _PRINT _SPACE (var | textwithoutspaces | list_access) (_SPACE (var | textwithoutspaces | list_access))*  -> error_print_nq``` instead. The reason is that var is more restrictive than textwithoutspaces and, therefore, needs to appear first.
]

In the above suggestion when we enter the error_print_nq we have the whole tree, including the type, values and meta information of the children. Figuring out which child has the missing quote is simpler now.
 
However, I am not sure whether changing the current implementation of the IsValid/Filter classes is something we want. When using a visitor we will fail at the first occurrence of an error while at the moment the transformer allows us to accumulate multiple errors. @Felienne maybe you could confirm this was the original idea behind the implementation? If so, we could make a hybrid transformer which returns the tree and accumulates the collected errors but this might increase the complexity.

Also before making any changes to the structure of the Filter, we need to ensure every code path is covered with tests.

Comment (dolfein) [reply hedy.py:910]: > The print node in the AST produced by the above code will have 5 arguments:
> 
> - (text, "'Person '")
> - (var, ["name"])
> - (text, "' said '")
> - (list_access, [...])
> - (text, "things'")

I am getting these arguments, am I doing something wrong? This is before IsValid.
image

I also came across these 2 issues: https://github.com/lark-parser/lark/issues/252 and https://github.com/lark-parser/lark/issues/556. Can we solve this with an Interpreter?

Comment (dolfein) [reply hedy.py:910]: >1. From this PR I assume we want to use the meta information to provide the concrete place of the error, e.g. a missing quote.

Yes, meta provides us with the line and (first) column of a **node**. If we know in which node the error is, we can look within that node for the exact column and store that in InvalidInfo.column.

>2. If 1) is correct, please consider the following program. Should we add or remove a quote? Both options yield perfectly valid programs.
>```
>name is Hedy
>print 'name
>```

I would say, in this case, remove the quote. I think this one is more interesting:
```
name is Hedy
print 'my name is name
```
>3. If 1) is correct, please consider the following program. Should we add a quote and where? Or should we remove a quote and which one?
>```
>people is mom, dad, Emma, Sophie
>print 'today the dishes are done by ' people at random'
>```

Removing the last quote would be optimal.

>4. Do you need meta information for other IsValid violations, e.g. 'invalid command' or 'invalid space'. If so, can you provide examples of program inputs (with their Hedy level) and the expected meta information?

Invalid space is, according to the comment, always a leading space, so meta.column is unnecessary here.
Invalid command can be useful in the future. I tried this example but a wrong error is raised. But the input should hopefully still give you an idea, we want to know the column of prind.
image

But I remember that @Felienne also said that meta could be useful for other stuff as well.

Felienne: > > 2. If 1) is correct, please consider the following program. Should we add or remove a quote? Both options yield perfectly valid programs.
> > 
> > ```
> > name is Hedy
> > print 'name
> > ```
> 
> I would say, in this case, remove the quote. 

Reading this, I think maybe we are all confused about what the others are thinking. So let me clarify where I want to go with program repair. It is *not* about finding the exact right location of an error. We only want to have the location to have a general sense of where to start trying to remove, add or swap out characters.

Absent of the right meta information, we can still very much do program repair, but then we just would do totally random things and see where they get us. The meta info helps us be more effective.

So there is not the situation that "remove the quote" is the right answer here. With program repair, we will generate many solutions, maybe one for adding *and* one for removing the quote, and we somehow evaluate the results. Maybe we present all solutions that parse to the user and heave them pick, but that is of later concern. The sky is the limit in terms of tries, but the bottom line is that we do not need a precise or perfect location for meta, just a good enough one.

> Invalid space is, according to the comment, always a leading space, so meta.column is unnecessary here. Invalid command can be useful in the future. I tried this example but a wrong error is raised. But the input should hopefully still give you an idea, we want to know the column of prind. image

Please @dolfein, for now focus on only the quotes and maybe indeed the leading spaces, that is tricky enough for your project. Invalid command could first use some other improvements (from me)

> But I remember that @Felienne also said that meta could be useful for other stuff as well.

Yes, but let's not consider that now, we need to scope your project down zo it becomes doable.

I propose the following:
* Based on the meta information there is now, you start to work on code that does random mutation (with the code you already wrote) and evaluates mutants (mutants = programs changed by code)
* In the mean time, @boryanagoncharenko will see if she can improve the meta information

boryanagoncharenko: @Felienne thank you for the clarification. This sounds like a nice project and I am looking forward to its results.

@dolfein I highly recommend keeping the IsValid transformer as it is for now. Please create your own traverser and plug it right before the IsValid.transform() invocation and handle the error cases that are interesting to you. Make sure that you are raising the needed exception immediately in the visitor (as opposed to the IsValid transformer which returns information about the error and an exception is raised later). Once you are ready with your project, I will clean up the IsValid transformer so that it does not repeat what you have already covered.

You can use the following PR as a starting point: https://github.com/Felienne/hedy/pull/1815. It defines a new Visitor and plugs it right before the IsValid.transform() method. Once you are in the Visitor, you will get a tree object which has a meta property and all its children will have a meta property too. This solves the problem of getting the meta of the immediate children, since you now get the full unprocessed tree annotated with meta for every node. If you need the original program code, you can use self.code. So this is all the information you need. Note that the UnquotedTextException is thrown in the visitor. 

Let me know if you have any questions about the meta information.

PR #1495, opened 2021-12-07
Repair for UnexpectedCharacters exceptions

dolfein:
**Description**

Repairs `UnexpectedCharacters` exceptions by removing that character.

**Fix for**

Progress for #1225, point 3.

**How to test**

For example, on level 4 or higher run the following Hedy code:
`print ,'Hello'`

`fixed_code` will have the `,` removed,  remaining `print 'Hello'`. It is still possible that other errors arise after the repair.

### DIFF
--- a/exceptions.py
+++ b/exceptions.py
@@ -53,13 +53,16 @@ class InvalidSpaceException(FtfyException):
             fixed_result=fixed_result)
 
 class ParseException(HedyException):
-    def __init__(self, level, location, found):
+    def __init__(self, level, location, found, fixed_code):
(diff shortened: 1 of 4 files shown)


Felienne:
Thanks @dolfein!

I added a test so we know what the code does, and we are protected against people that accidentally break this feature. I have also looped the `fixed_code` into the exception so we can use it in the front end and we can test more easily. Can you do the same for future fixes?

Felienne (approved):

Felienne (approved):

Issue #1478, opened 2021-12-06
[UI idea] When logged in, set language automatically to user preferred language


TiBiBa:
**Idea incl level**
Currently the UI is set to a user selected language. When logging in the language will stay the same. However, when we built upon #1477 the language selection will disappear from the menubar when logged in. In this case we would like the user to selected a preferred language in the profile settings to use every time the user is logged in. This should be aligned with #1467 to make sure users are able to change their preferred language on the profile page.

Felienne:
Lovely idea! 

I guess we can ask preferred language on signup for new accounts, and ask existing users to set it once in their profile?

PR #1497, opened 2021-12-08
[Cleanup] Removed language dependency on url argument

### LINKED ISSUE #1496: [UI idea] Store current language in session and prevent url argument dependency
**Idea incl level**
Currently we depend on the `?lang=` addition to the url to retrieve the user selected language. Because of this we use a parsing function to localize each url to make sure the user gets the next page in the same language when switching pages. This is cumbersome, especially with the goal of #1467 and #1478 to enable logged in user to set a fixed language in their profile settings. I propose that we alter this structure and store the current language in the user sessions, cleaning up the url and making the implementation of the other mentioned issues easier.

TiBiBa:
**Description**
Currently we depend on an `lang` argument added to the url to tell us the language we have to render the current page in. This is cumbersome as we have to pre-process each url to make sure the user keeps the same language when switching pages. In this PR we implement the chosen language in a session variable, living on the server. Removing the need from sending the argument with the url and in the same process cleaning this url as well. This PR will also make it significant easier to implement #1478.

**Fix for**
This PR fixes #1496.

**How to test**
Navigate through different pages and notice that the `lang` attribute is no longer passed in the url. Try different languages to verify the correct updating of the stored session variable.

**Checklist**


Felienne:
It seems that the lang is still used if people click the advance button, is that as expected?

image

Felienne:
> It seems that the lang is still used if people click the advance button, is that as expected?
> 
> image

I think this is located in incl-editor-and-output.html which indeed this PR does not seem to update?

TiBiBa:
> > It seems that the lang is still used if people click the advance button, is that as expected?
> > image
> 
> I think this is located in incl-editor-and-output.html which indeed this PR does not seem to update?

Forgot that one, will fix with a new commit right away!

Felienne:
Ah looks better now the langs are clickable again!

Felienne (approved):

PR #1499, opened 2021-12-08
Added translation for levels 12 to 17 + tests

emil-meyknecht:
Push to finalize the translation structure up to level 17.
Issues #1453 , #1454 

Added levels 12 to 17 into the translation structure.

-Adds the translation for 'and', 'or' conditions.
-Adds the support for mathematical comparison signs.
-Adds the translation for the while loop
-Adds brackets around arrays after level 16
-Adds a semicolon after indentation in level 17

Tests available with test_translation_level_10.py to test_translation_level_17.py


hedy_translation.py
```diff
@@ -28,8 +28,9 @@ def translate_keywords(input_string, from_lang="en", to_lang="nl", level=1):
     punctuation_symbols = ['!', '?', '.']
 
     keyword_dict = keywords_to_dict(to_lang)
-    if level > 7:
```
Felienne, line 31: Why are we removing the level check here? processing blocks only needs to be done for levels over 7

Felienne (changes requested):

hedy_translation.py
emil-meyknecht: I added the check that was already present in hedy.py, the preprocess_input() function. That one also only uses preprocess_blocks when level > 7

hedy_translation.py
```diff
@@ -28,8 +28,9 @@ def translate_keywords(input_string, from_lang="en", to_lang="nl", level=1):
...
-        input_string = hedy.preprocess_blocks(input_string, level)
+
+    input_string = hedy.process_input_string(input_string, level)
```
emil-meyknecht, line 32: This function

> I added the check that was already present in hedy.py, the preprocess_input() function. That one also only uses preprocess_blocks when level > 7

Felienne: Ah of course!

Felienne (approved):

PR #1501, opened 2021-12-08
Multiple keyword languages

teundemast:
This PR makes sure, English keywords are always allowed when using the hedy website in another language. So for example, when the language of the website is in Dutch, users can use English en Dutch keywords. 
To achieve this, I have updated the yaml to lark transformer in a way that English keywords are added to the lark files. 
I have also updated the keyword suggestions to suggest both Dutch keywords and English keywords. I'm not sure if this is desired though. 

Fixes issue #915 

How to test? 
Make sure local_keywords_enabled is True. 
When running hedy locally and changing the language to 'Nederlands', you should be able to use 'vraag' and 'ask'.


teundemast:
Had some trouble with running the tests locally, so that's why could not use the test feedback immediately. But I will do so now :)

Felienne:
Ah awesome that you took at stab at this team! It is a complex piece of work so I will try to check it tomorrow but it might be Friday too

Felienne (dismissed):
Wow team this is truly awesome!! I did not think you would get this far!!!

I just pushed a few tests but otherwise, we are very close to finishing this (I saw a PR for syntax highlighting also!!)

Felienne (approved):

Issue #1475, opened 2021-12-06
is, = and ==?


Felienne:
### Discussed in https://github.com/Felienne/hedy/discussions/1274

1. In current level 6 when we add calculations, change `is` in assignments to `=` 
2. [NEW!] And change `is` in _conditions_ to `=` too 
3. In current level 12 where we introduce < and >, we also switch `is` in conditions to `==`

I think in all situations we want to allow both, or give a warning but not a hard error for the use of `is`.

hapiel:
`=` is an action, `is` is a conclusion. Maybe it is an option to replace `is` as `=` with `becomes`
`count becomes 1` == `count = 1`
`count is 1` == `count == `

Just a random idea, can't tell if it's worth considering or if this overcomplicates stuff.

Felienne:
> `=` is an action, `is` is a conclusion. Maybe it is an option to replace `is` as `=` with `becomes`: `count becomes 1` == `count = 1` `count is 1` == `count == 1`
> 
> Just a random idea, can't tell if it's worth considering or if this overcomplicates stuff.

Yeah initially we use `is` for both assignment and conditions and we slowly want to move to Python, so the switch is part of that.

I could see how `becomes` would be a great word initially for assignment, better conceptually that `is` but that does create two different words, which we deemed too complex

Felienne:
Closed by #1504

Felienne:
Closed entirely by #1580

PR #1504, opened 2021-12-08
`is` to `=`

Felienne:
Addressing #1475


grammars/level16-Additions.lark
```diff
@@ -3,5 +3,7 @@
 ?atom: NUMBER | var | list_access | unsupported_number
 list_access: var _LEFT_SQUARE_BRACKET (INT | random | var) _RIGHT_SQUARE_BRACKET
 change_list_item: var _LEFT_SQUARE_BRACKET (INT | var) _RIGHT_SQUARE_BRACKET _SPACE _EQUALS _SPACE (var | textwithoutspaces)
-list_access_var: var _SPACE _EQUALS _SPACE var _LEFT_SQUARE_BRACKET (INT | random | var) _RIGHT_SQUARE_BRACKET
-assign_list: var _SPACE _IS _SPACE _LEFT_SQUARE_BRACKET (quoted_text | NUMBER) (_COMMA _SPACE? (quoted_text | NUMBER))+ _RIGHT_SQUARE_BRACKET
+list_access_var: var _SPACE (_EQUALS|_IS) _SPACE var _LEFT_SQUARE_BRACKET (INT | random | var) _RIGHT_SQUARE_BRACKET
+assign_list: var _SPACE (_EQUALS|_IS) _SPACE _LEFT_SQUARE_BRACKET (quoted_text | NUMBER) (_COMMA _SPACE? (quoted_text | NUMBER))+ _RIGHT_SQUARE_BRACKET
+
+# the two rules above now overlap :(
```
bjorn3, line 9: They shouldn't be. `var` shouldn't match `[`. Just some wild guesses, but maybe `var: NAME -> var` or the complexity of the `ID_START` regex confuses lark about which rules overlap and which don't or cause it to hit some edge case in the lexer generator?

Felienne: yeah this was really a note to self so I can continue tomorrow, no help needed at this point

PR #1509, opened 2021-12-08
Ace now highlights keywords from other languages

### LINKED ISSUE #909: Ace can highlight keywords in other languages
Currently, if you change the keyword in the parser, it will use it but not highlight it

HuibSprangers-leiden:
**Description**

- highlights the keywords correctly in the online ace instance, should the option be turned on.
- contains syntaxModesRules-en and -nl, templates for the creation of modes for the highlighter in ace/modes/level1en.js as opposed to just ace/modes/level1.js
- syntaxRules.ts now contains exported rules shared by each individual highlighter, to prevent duplicate code errors.
- The original syntaxModesRules.ts has been changed to syntaxModesRules-en.ts, which serves the same function as the original
- slight changes in app.ts and index.ts to allow for these different highlighting rules to actually be used (importing the syntaxModesRules-'lang'.ts files and calling the correct ace/mode file), also includes an enable to turn this on or off. It is currently off by default.

**Fix for**

Closes #909 

**How to test**

Set **local_keywords_highlights_enabled** in app.ts to true and rebuild the appbundle, afterwards run the website. You should now see that the keywords highlighted for the current language (currently just dutch and english) are highlighted appropriately according to the language's keywords as given in 'lang'.yaml
So print and echo should both be highlighted for level 1, forward should only highlight for english, vooruit should only highlight in dutch, etc.


static/js/syntaxModesRules-en.ts
```diff
@@ -0,0 +1,355 @@
...
+      syntaxRules.rule_printParen(_PRINT),
+      syntaxRules.rule_isInputParen(_IS,_INPUT),
+      syntaxRules.rule_is(_IS),
+      syntaxRules.rule_ifElse(_IF,_ELSE,_IS,_IN),
+      syntaxRules.rule_expressions(_AT,_RANDOM),
+      syntaxRules.rule_arithmetic(),
+      syntaxRules.rule_forRangeParen(_FOR, _IN,_RANGE),
+    ),
+  },
+// ----------------------------------------------------------------
```
Felienne, line 175: Are we ready to merge this? :) I guess the lang changes are finished further than 12?

static/js/syntaxModesRules-nl.ts
```diff
@@ -0,0 +1,354 @@
...
+  {
+    // Adds if/else
+    name: 'level4',
+    rules: syntaxRules.pipe(syntaxRules.baseRules(_AT,_RANDOM),
+      syntaxRules.rule_printSpace(_PRINT),
+      syntaxRules.rule_isAsk(_IS,_ASK),
+      syntaxRules.rule_is(_IS),
+      syntaxRules.rule_ifElseOneLine(_IF,_ELSE,_IS,_IN),
+      syntaxRules.rule_expressions(_AT,_RANDOM),
+    ),
```
Felienne, line 82: This (and all other levels) are basically a copy-paste of en, could we somehow centralize this (if now, we can leave it as is)

static/js/syntaxModesRules-nl.ts
```diff
@@ -0,0 +1,354 @@
...
+var _IN='in';
+var _IF='als';
+var _ELSE='anders';
+var _REPEAT='herhaal';
+var _TIMES='keer';
+var _FOR='voor';
+var _RANGE='bereik';
+var _INPUT='invoer';
+/* These keywords currently don't have highlighing rules, but are also listed in the .lark files
+var _OR='of';
```
Felienne, line 20: Yeah these should have rules (in nl and en of course)

Felienne (changes requested):

Felienne:
Cool, this seems to work!!

image

image

It is just not up to date with multilang programs yet as added in #1501, is it planned to fix that in the highlighting too (and shall we do that in this PR or in a new one?)

HuibSprangers-leiden:
> Cool, this seems to work!!
> 
> image
> 
> image
> 
> It is just not up to date with multilang programs yet as added in #1501, is it planned to fix that in the highlighting too (and shall we do that in this PR or in a new one?)

I think that depends on how easily I can make it combine modes into one with Ace/use both, I might know how to do that easily, but it could end up more complicated if that doesn't work.
I'll see if I can add the functionality quickly, otherwise I'll convert this to draft.

HuibSprangers-leiden:
So after looking it up, it seems that Ace does not currently support multiple syntax modes:
https://github.com/ajaxorg/ace/issues/3435
It should still be possible to implement multiple language highlights by hardcoding the english rules into the rules for other languages, I'll try doing that. Turing this into a draft request in the meantime.

Felienne:
> So after looking it up, it seems that Ace does not currently support multiple syntax modes: [ajaxorg/ace#3435](https://github.com/ajaxorg/ace/issues/3435) It should still be possible to implement multiple language highlights by hardcoding the english rules into the rules for other languages, I'll try doing that. Turing this into a draft request in the meantime.

Ah that is a pity! But... we can do something like this to recognize both forms:

`var _ASK = '(?:vraag|ask)';`   

 // the ?: indicates a non-capturing group

We can place that in a header and dynamically switch the header from the en version with 1 option, to the nl version with both. That also removes the need to have the copy-pasted code.
 
Does that make sense?

If you still have the time, you could write code to generate the header from the yamls, but if not we can do that manually too (the keywords will not change often) or I can write that script.

HuibSprangers-leiden:
> > So after looking it up, it seems that Ace does not currently support multiple syntax modes: [ajaxorg/ace#3435](https://github.com/ajaxorg/ace/issues/3435) It should still be possible to implement multiple language highlights by hardcoding the english rules into the rules for other languages, I'll try doing that. Turing this into a draft request in the meantime.
> 
> Ah that is a pity! But... we can do something like this to recognize both forms:
> 
> `var _ASK = '(?:vraag|ask)';`
> 
> the ?: indicates a non-capturing group, and the | of course indicates two options
> 
> We can place that in a header and dynamically switch the header from the en version with 1 option, to the nl version with both. That also removes the need to have the copy-pasted code.
> 
> Does that make sense?
> 
> If you still have the time, you could write code to generate the header from the yamls, but if not we can do that manually too (the keywords will not change often) or I can write that script.

I didn't know you could declare variables that way, I'll try to implement it like that. Also wasn't aware of headers in typescript, I'll see if I can figure that out. It makes sense now that I've read up on it a bit. Like that you would only need to have syntaxModesRules and I guess a bunch of files containing those headers (Or just manually putting them in, but a seperate file might be easier for generating them through a script), correct? Turning the variables into more like a dictionary and reading that dictionary in syntaxModesRules.

Something like:
`const LANG-NL { _FORWARD: '(?:forward|vooruit)'}`
with all the keywords contained within that LANG-NL, from what I see.

I'll see if I can generate them automatically, I had a script to catch the keywords from the yaml files already (it got depreciated and kinda useless towards the end so I didn't include it on the merged branch), though parsing the correct ones from both languages might be difficult.

I'll try and see how much I can get done tomorrow. The alternate declarations shouldn't be an issue but editing that script and using headers might take a bit. I'll see what I can do.

Felienne:
> I didn't know you could declare variables that way, I'll try to implement it like that. 

Great!!

> Also wasn't aware of headers in typescript, I'll see if I can figure that out. It makes sense now that I've read up on it a bit. Like that you would only need to have syntaxModesRules and I guess a bunch of files containing those headers (Or just manually putting them in, but a seperate file might be easier for generating them through a script), correct? Turning the variables into more like a dictionary and reading that dictionary in syntaxModesRules.
> 
> Something like: `const LANG-NL { _FORWARD: '(?:forward|vooruit)'}` with all the keywords contained within that LANG-NL, from what I see.

Yeah exactly! I did not mean an actual header file (not sure ts supports that) but just a file with the nk declarations and one for en etc. Precisely as you describe above with ` LANG-NL`.

We could also consider creating one big dictionary instead of separate files, and then fetch the rule based on the lang.

> I'll see if I can generate them automatically

Cool! But as I said, if we can't make them automatically, that is no biggie, manual is also an option since it will not have to be done often.

HuibSprangers-leiden:
> Yeah exactly! I did not mean an actual header file (not sure ts supports that) but just a file with the nk declarations and one for en etc. Precisely as you describe above with ` LANG-NL`.
> 
> We could also consider creating one big dictionary instead of separate files, and then fetch the rule based on the lang.

I actually thought you might have meant these kinds of headers at the time I made that post:
https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers
To be fair, I was quite tired at that hour. I implemented it like you said, just as a file that gets imported by syntaxModesRules.ts

With the latest commit, it should work as specified, or functional at the very least:
god I hate typescript
I'll give it another look over to see if I missed anything and add some commentary, then turn this back into a pull request, unless you or I find another issue that really needs to be looked at.

HuibSprangers-leiden:
And the script was already pushed earlier, it currently sits in build-tools under the name generate-syntax-keywords

Felienne:
> I actually thought you might have meant these kinds of headers at the time I made that post: https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers To be fair, I was quite tired at that hour. 

Ah sorry my use of the word headers was a bit confusing!

> I implemented it like you said, just as a file that gets imported by syntaxModesRules.ts

> With the latest commit, it should work as specified, or functional at the very least: god I hate typescript I'll give it another look over to see if I missed anything and add some commentary, then turn this back into a pull request, unless you or I find another issue that really needs to be looked at.

COOOL! Will test this now.

static/js/syntaxLang-en.ts
```diff
@@ -0,0 +1,31 @@
+//Generated through generate-syntax-keywords
```
Felienne, line 1: keurig commentaar zodat mensen weten dat dit een gegenereerde file is!

build-tools/heroku/generate-syntax-keywords
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
Felienne, line 1: wow impressive in a bash script (py would have been fine too)

Not sure I will ever be able to decipher this though haha

static/js/syntaxModesRules.ts
```diff
@@ -63,266 +84,265 @@ const LEVELS = [
...
+    rules: pipe(baseRules(currentLang._AT,currentLang._RANDOM),
 
-      rule_printSpace('expression_eol'),
-      rule_isAsk('gobble'),
-      rule_is('gobble'),
-      rule_turtle(),
-      rule_sleep(),
+      rule_printSpace(currentLang._PRINT,'expression_eol'),
+      rule_isAsk(currentLang._IS,currentLang._ASK,'gobble'),
+      rule_is(currentLang._IS,'gobble'),
```
Felienne, line 91: I see that this works, but I am not sure it is very pretty with all the extra arguments. Could we add the whole currentLang to the global state and access it in the functions?

static/js/syntaxModesRules.ts
```diff
@@ -437,48 +457,45 @@ function rule_is(next?: string) {
...
-
-
 /**
  * Add an 'is input' rule with brackets
  */
-function rule_isInputParen() {
+function rule_isInputParen(_IS: string, _INPUT: string) {
   return recognize('start', {
-    regex: '(\\w+)( is input)(\\()',
+    regex: '(\\w+)( ' + _IS + ' ' + _INPUT + ')(\\()',
```
Felienne, line 498: so here instead of taking the `_IS` and `_INPUT` as arguments, can we remove it and do something like this:

```
function rule_isInputParen() {
  return recognize('start', {
    regex: '(\\w+)( ' + currentLang._IS + ' ' + currentLang._INPUT + ')(\\()',
```

I know using global state is not typically advised but I think here it makes sense

Felienne (changes requested):
This is fantastic, team and it works like a charm!!

I just have a tiny idea to improve readability a bit more bit but otherwise it looks great!

static/js/syntaxModesRules.ts
HuibSprangers-leiden: Yeah, that should be possible, the main reason I turned them into proper methods in the first place was because I seperated them into syntaxRules, which is now gone. I'll get on that.

build-tools/heroku/generate-syntax-keywords
HuibSprangers-leiden: The regex took a while to figure out, so I'm glad to hear you find it good.
I'll maybe add some per line comments so it's easier to figure out exactly what that script is doing.

HuibSprangers-leiden:
Added the change asked for, currentLang is now used within the functions instead of in their arguments. The arguments were thus removed from the functions.
In exchange currentlang has to be initialized in it's typing(s) somewhat due to implicit typing errors
Schermafbeelding 2021-12-10 144423
.

static/js/syntaxModesRules.ts
Felienne: New version looks a lot better!!

Felienne:
> Added the change asked for, currentLang is now used within the functions instead of in their arguments. The arguments were thus removed from the functions. In exchange currentlang has to be initialized in it's typing(s) somewhat due to implicit typing errors Schermafbeelding 2021-12-10 144423 .

Yeah that's ok!

Felienne:
Cool, the code looks great now and also it still works:

image

Felienne (approved):

JoshMcCullough:
Does this fix https://github.com/ajaxorg/ace/issues/3435 ? @Felienne / @HuibSprangers-leiden ?

PR #1518, opened 2021-12-09
Clean up adventures

MarleenGilsing:
Mainly changing 'in level x' to 'in this level', so the texts are accurate.


Felienne (approved):
Smart, that will save us rewrites if we ever update the levels again!!

PR #1519, opened 2021-12-09
Use a proper regex string in the index route

bjorn3:
Previously python would first try to interpret the \d as an escape sequence and throw "DeprecationWarning: invalid escape sequence \d" upon noticing that \d is not a valid escape sequence. By escaping the \, the regex engine will get a proper \d literal without python throwing a warning.


bjorn3:
Merging is blocked on an explicit approval.

Felienne:
> Merging is blocked on an explicit approval.

yeah working on some other things rn, will merge when ready

Felienne (approved):

Issue #1256, opened 2021-11-12
Keywords in the sidebalk should also be translated, just as the demo code


teundemast:

FlorisHessels:
Since this is done in the Yaml files, I would like to know if you, @Felienne, agree with these Dutch translations for the keywords:

![teun](https://user-images.githubusercontent.com/91470479/142026433-49f55269-ccf0-43db-8032-99e6afb20858.jpeg)

Then, I can change all the Dutch Yaml files accordingly.

Felienne:
Hi all!

willekeurigewijze I guess is a bit weird because we use it for list access, like `print dieren at random` so I guess `willekeurig` is better. And it can always be changed right, because I assume we will use some sort of converter to grab the existing programs and translate those?

FlorisHessels:
Okay that's fine, I will use `willekeurig` instead of `willekeurigewijze`. Furthermore, we are in the process of making such a converter for programs, but we cannot use this converter for the Yaml files. This is because for example not every 'if' in the text should be converted to 'als'. The picture down below depicts an instance where the 'if' should not be translated to 'als'.

![happy](https://user-images.githubusercontent.com/91470479/142030121-bb627580-4eba-4529-8510-3e44f8b4336d.PNG)

Do you think that this is just a rare occasion or is this a roadblock for a converter for the Yaml files, @Felienne?

Felienne:
This is not a roadblock perse, it just means you will have to use the
parser to parse the code, then transform the tree and pretty print the tree.

FlorisHessels:
Okay that's clear. I will resolve this issue whenver issue #942 is done.

teundemast:
So for this issue we are adding a function to yaml_transform.py. But we need to import the file hedy_translation, which is in the root folder. But since the root folder is of course not a package, we think this is the only way to do so: 
<img width="174" alt="Capture" src="https://user-images.githubusercontent.com/36051227/144456977-d63147f1-7dfd-4fde-b10b-7f87e9400967.PNG">
Hardcoding the path is of course not feasible for the hedy site, so we are thinking of creating a new transformer file in the root folder. @Felienne do you think this is the right solution? Or are we missing a solution?

Felienne:
I don't think I understand, can't you just do this?

![image](https://user-images.githubusercontent.com/1003685/144478684-39db86c7-7e1f-46f2-a531-2f8bd945bfd1.png)

hedy can be imported too from the main dir into yaml_transform.py?

[...]

Issue #1506, opened 2021-12-08
[BUG] The example codes in what's next adventures don't always work


teundemast:
If the language is Dutch, running the demo code in 'Vooruitblik' (or Fortune Teller in English) gives an error. Is this the intention, because it is in 'Vooruitblik', @Felienne ? 
<img width="503" alt="Capture" src="https://user-images.githubusercontent.com/36051227/145233083-a58b2b8c-1e45-48df-8900-4f78e1974b83.PNG">

MarleenGilsing:
The fortune teller adventures do work. The what's next (vooruitblik) adventures sometimes don't work yet.

Felienne:
Yeah that is the case because it is level 8 code shown at level 7, @MarleenGilsing will change the examples.

Felienne:
Hi @MarleenGilsing! 

Thanks for working on this issue! In principle we don't close issues until the PR's are tested and merged, so is m reopening this one until I merge your fixes :)

Felienne:
Closed by #1557

Issue #1498, opened 2021-12-08
Calling translate_keywords


Livavdv:
We are working on the translate_keywords function in hedy_translation.py. We want to call on this function so it will translate the keywords in the written code when you change the language on the website. However, we can't find a way to do this without parsing the code. Do you have any suggestions on where and how we could implement this? @Felienne @brucemartens

Felienne:
I don't see how this is possible without parsing the code! (which also seems to me to be a fine solution, why would we want to avoid that?)

Felienne:
Closing this, we will not translate all language snippets but instead add a switcher in the UI (see #1577)

PR #1524, opened 2021-12-09
Translating keywords in yaml files (adventures and level-defaults) 

teundemast:
With this PR, @FlorisHessels and I added 2 new transform functions in the yaml_transformer.py. These functions make it possible to translate the keywords in the yaml files to local keywords. Fixes issue #1256 

When we were working on this, we discovered some errors in the translations and in the translation structure. Those will also be fixed with this PR.

We will convert this PR to a draft request since the transform functions have some imperfections. Namely,

- If there is a word like 'to' in for example an intro text (in code this works fine), it is impossible to know if it should be translated to 'tot' or 'toe aan' in Dutch. So we think this should be translated manually. 
- If there is a word like 'ask' in a text but it is contained with ` `` `, than it won't be translated. 
- If there are multiple words in a text inside the ` `` ` the transformer function won't translate these words since most of the times this is invalid code. Thus, we can not use the current translation structure for this because that only accepts valid code.
- There are some pieces of code in 'Vooruitblik' which are invalid code as mentioned in issue #1506. So we cannot translate this for the same reason above. That's why, if the transform functions are used right now with current Dutch yaml, they will break. 
- If we execute the transform functions right now, the transformed yaml file won't have the same structure as the original yaml file. Namely, new lines will be replaced with \n in the transformed yaml file. We are not sure if there is a way to fix this. However, this has no effect on the website itself. 

For now, we have only tested the transform functions for Dutch.


Felienne:
Thanks for this awesome piece of work team!!

> There are some pieces of code in 'Vooruitblik' which are invalid code as mentioned in issue [BUG] The example codes in what's next adventures don't always work #1506. So we cannot translate this for the same reason above. That's why, if the transform functions are used right now with current Dutch yaml, they will break.

Yeah that is ok! We can fix those examples and then run this code

> * If we execute the transform functions right now, the transformed yaml file won't have the same structure as the original yaml file. Namely, new lines will be replaced with \n in the transformed yaml file. We are not sure if there is a way to fix this. However, this has no effect on the website itself.

Yeah I know this issue, it happens when you just dump yamls. You can use the ruamel version to write used in other places of the code bae as well to avoid that issue.

teundemast:
Hi @Felienne, today is the deadline for delivering our product. We have decided to not continue with this issue, because we don't have time to do so. Some of our team members are going to work on their issues: issue #909 and issue #1498. 

Shall we leave this as a draft request or transform it into a Pull Request?

Felienne:
> Hi @Felienne, today is the deadline for delivering our product. We have decided to not continue with this issue, because we don't have time to do so. Some of our team members are going to work on their issues: issue #909 and issue #1498.
> 
> Shall we leave this as a draft request or transform it into a Pull Request?

You might leave it as is, I can finish it myself.

Is the only thing left to make sure that the yaml output looks readable or were there other things still open?

Felienne:
Thanks for the clarification! We will pick this up further in the Core Team!

PR #1525, opened 2021-12-09
Fixes #1311: adding new operator to merge grammars

### LINKED ISSUE #1311: Make expansion of lists non-terminals possible in the merger
Several levels now redefine the list of commands with a lot of repetition, for example:

level 3:
`command: print | ask | turtle | assign | assign_list | add | remove | sleep | invalid_space | ask_dep_2 | echo_dep_2 | comment | invalid`

level 4:
`command: print | ask | turtle | assign_list | add | remove | sleep | ask_no_quotes| assign | print_no_quotes | comment | invalid
`

Would be nicer if we would define something like += and generate proper Lark in the merger

racertop:
**Description**

As stated in issue #1311 there is a lot of redefinition of the `command` rule in the grammars, and that makes understanding which rules are added or removed in each level hard to follow. That's why we're adding 2 new operators to the grammar +=,  and -= They work as follows:
If we have this rule:
`command: rule1 | rule2 | rule3`

And we want to add a rule4, we simply do this in the next level:
`command:+= rule4`

The result is
`command: rule1 | rule2 | rule3 | rule4`
 
If we want to add a new rule, and remove another we do this:
`command:+= rule4 -= rule1 | rule3`

The result is:
`command: rule3 | rule4`

Note that at the moment, if we want to remove a rule, it always comes after adding a new one, since I figured that we never remove a rule and not add another one.

Another caveat is that sometimes the commands need to be in a certain order, in those cases we might do this:
`command: add | remove | invalid -= invalid`

This way invalid will be in the end of the rule.

**Fix for**

Issue #1311

**How to test**

Since I basically changed all the grammars, the best way is to simply run all the unit tests.

**Checklist**


Felienne:
> As stated in issue #1311 there is a lot of redefinition of the `command` rule in the grammars, and that makes understanding which rules are added or removed in each level hard to follow. That's why we're adding 2 new operators to the grammar +=, and -= They work as follows: If we have this rule: `command: rule1 | rule2 | rule3`
> 
> And we want to add a rule4, we simply do this in the next level: `command:+= rule4`
> 
> The result is `command: rule1 | rule2 | rule3 | rule4`
> 
> If we want to add a new rule, and remove another we do this: `command:+= rule4 -= rule1 | rule3`
> 
> The result is: `command: rule3 | rule4`

Wow thanks, this is fantastic work and makes the grammars a lot more readable!!!

> Another caveat is that sometimes the commands need to be in a certain order, in those cases we might do this: `command: add | remove | invalid -= invalid`
> 
> This way invalid will be in the end of the rule.

This I don't really understand, sorry! Because -= means removal, I would think that

`command: add | remove | invalid -= invalid`

leads to:

`command: add | remove`

But from your writing I understand that when we name ánd remove a rule in one line, it will be placed at the end?

Maybe we would like to have a different operator there? Maybe something like this:

`command: add | remove > invalid` indicating that whatever is added, invalid should be last?

racertop:
> This I don't really understand, sorry! Because -= means removal, I would think that
> 
> `command: add | remove | invalid -= invalid`
>
> leads to:
> 
> `command: add | remove`

Yes, I admit it's a bit unclear. The way the `comand:+= add -= remove` works is by removing the `remove`rules and then merging the remaining rules and the new rules like this:
`command: remaining | add`, so if `invalid` needs to be in the end of the list, just remove it and place it in the end of the `add` list.

But yours is a much better idea hahaha

> But from your writing I understand that when we name ánd remove a rule in one line, it will be placed at the end?
> 
> Maybe we would like to have a different operator there? Maybe something like this:
> 
> `command: add | remove > invalid` indicating that whatever is added, invalid should be last?

I'll implement this! It's much clearer.

racertop:
@Felienne I added the `>` operator and now the grammars are even clearer :) for example in level 5 the rule needs that assign is placed at the end of the command rule, so it ends up looking like this:

`command:+= ifelse | ifs  |  list_access_var -= invalid | ask_no_quotes > assign`

It adds `ifelse | ifs  |  list_access_var` removes `invalid | ask_no_quotes` and places  `assign` at the end.

I hope it makes things much clearer!

grammars/level8-Additions.lark
```diff
@@ -1,4 +1,5 @@
-command: print | repeat | ifs elses? | turtle | ask | assign_list | add | remove | sleep | list_access_var | assign | comment  //placing assign at the end means print is will print 'is' and print is Felienne will print 'is Felienne'
+command:+= ifs elses? -= ifelse| ifs
```
Felienne, line 1: owwww so much clearer and nicer!!

Felienne (approved):

Issue #1412, opened 2021-11-26
Add small animations to adventure images


johannadekker:
We now have certain custom Hedy images in the adventures. For some, it could help in clarifying the purpose of the level if some small-scale animations could be added (and it would look nice). 

For example, for the turtle code, it could be nice to see the turtle moving along the line so children can immediately see what their code is going to do as a hint.

PR #1526, opened 2021-12-10
#1412 replaced turtle.png with a gif

johannadekker:
I created a gif for the turtle adventure that shows the turtle moving along the line. I thought it would be visually clearer what turtle code exactly does, and also fun. The way I implemented it so that it doesn't distract is by having the gif loop twice before reverting to the first frame (turtle.png). This is implemented directly in the GIF file.

I have attached both the gif that loops twice (turtle.gif) and the gif that loops infinitely (turtle_loop.gif).

turtle
turtle_loop


fRedelaar:
Looks great Johanna! Good to see that the gif stop looping by itself is a solution.

Felienne (approved):
Lovely!

Issue #777, opened 2021-09-16
Localize quizzes


Felienne:
Hi @fpereiro!

#757 made me realize quizzes can't really be localized at this point (I assumed it was just about content). Can we make that happen?

Felienne:
Hi @LterHaar,

What I propose we do there is make the quiz yaml files as similar as possible to the other files, which means I would like to have all content (all quizzes) for one language in one file, something like:

1: 
   1:
    question_text: "What's this programming language called?"
    code: "..."
    mp_choice_options:
    - "option_text": "Hedy"
      "feedback": "Good job!"
    - "option_text": "Heddy"
      "feedback": "Not this one!"
    - "option_text": "Haydie"
      "feedback": "Not this one!"
    - "option_text": "Henk"
      "feedback": "Not this one!"
    correct_answer: "A"
    hint: "It's named after Hedy Lamarr."
    question_score: 10
 2:
    question_text: "Which command makes text appear?"
    code: "..."
    mp_choice_options:
    - "code": "echo"
      "feedback": "Echo repeats a given answer."
    - "code": "print"
      "feedback": "Correct!"
    - "code": "hello"
      "feedback": "Hello isn't a command."
    - "code": "ask"
      "feedback": "With ask you can ask a question"
    correct_answer: "B"
    hint: "(?) Hello world! "
    question_score: 10

Where the first number is the level and the next ones are the question numbers.

The whole English file then will be called en.yaml and the dutch one called nl.yaml. Can you make this change and make sure the quizzes can be localized (so the right one is selected when the language changes)

This should be 2 PR's, 1 for the changed files and 1 for the lang switcher.

Felienne:
Closed by #1415

Issue #1531, opened 2021-12-10
[CONTENT] Typos in English level 1 questions 7 and 9


johannadekker:
I have attached screenshots from the English-language level 1 quiz below. You can see that in both questions, option B has some mistakes. 7b) states "In line 1 **moet** print needs to be replaced with echo". 9b) states "print in line should be **aks**".

![Screenshot 2021-12-09 at 13 08 21](https://user-images.githubusercontent.com/90705403/145598983-29b47cf9-983c-4abb-805a-0e6eb5ff1d00.png)
![Screenshot 2021-12-09 at 13 08 41](https://user-images.githubusercontent.com/90705403/145598994-e2fa297a-ebf1-473b-92fe-0ea6e1f3a607.png)

Felienne:
Closed by #1532

PR #1532, opened 2021-12-10
#1531 fixed typos and improved some of the grammar in quiz level 1

johannadekker:
I fixed both typos in questions 7 and 9. I also noticed some of the feedback had some very small errors that I fixed. For example:

- In level 1 question 5, one of the options was echo misspelled as ego. However, this only works as a spelling mistake in Dutch. I changed it to 'ekko' so that it would remain similar to the original Dutch version.
- In level 1 question 4, feedback for option B first said: ""Im is spelled wrong on purpose**,** Hedy doesn't understand the apostrophes yet." However, that is incorrect grammar as these could be two separate sentences. I changed the comma to a semi-column ( ; ).
- One of the options for question 1, what language is this was the name 'Henk'. For an anglophone child, that is an unfamiliar name. I changed it to Heidi, which should be more familiar.
- The feedback for the 'no' option in Q10, asking if they want to progress to the next level was changed from 'Too bad' to 'What a shame'. I believe this is a more accurate translation of 'Jammer'. I think too bad could be experienced as sarcastic.
- Somewhere, in the feedback for a wrong answer, it said correc instead of correct. I fixed that too.

tl;dr : Fixed small language errors


Felienne:
Thanks for the fixes @johannadekker but this PR is a bit far behind main to be merged (since we now closed #777) I will make a new PR with these changes in it.

PR #1534, opened 2021-12-10
Animate program repair #1430

Jasper1408:
This pull request changes the way the modal containing the repaired code shows. First, the broken code input code is shown, then the cursor moves to the part of the code that doesn't work. The broken code is overwritten by the repaired code so that the changes can be seen by the user.


fRedelaar:
The result of this PR should look like the video below. It isn't perfectly finished yet, since the animation for the correct code needs to be implemented. Maybe @remva  or I could pick this up and upgrade it for our thesis :) 

https://user-images.githubusercontent.com/33595265/145613989-356887ea-9d98-4da3-929f-3986d290b703.mov

Felienne:
> The result of this PR should look like the video below. It isn't perfectly finished yet, since the animation for the correct code needs to be implemented. Maybe @remva or I could pick this up and upgrade it for our thesis :)
> 
>  2021-12-10.18-10-35.mov

Ah thanks for the explanation @fRedelaar! 

I think as it is, it is not better than what we have now where the right code is shown immediately. My proposal is to close this PR and indeed you or @remva can make the animation for your thesis! Of course the branch remains in this repo, so you can start with this code when you pick it up later.

PR #1535, opened 2021-12-10
Fix en level 2 quiz

bjorn3:
The extra indentation caused all questions to be considered fields of question 1

Fixes https://github.com/Felienne/hedy/issues/1533

image


Felienne:
Ah nice catch @bjorn3!

Felienne (approved):

Issue #665, opened 2021-08-23
User statistics


fpereiro:
Opening an issue to brainstorm possible user statistics that can help teachers better understand how a class is going. Also perhaps useful for the project itself.

Ideas:
- Programs run in the last 24 hours/week/month
- Distribution of levels per program run
- Perhaps the two above but for saved programs instead of run programs

@Felienne let's discuss ideas here and we can start implementing a couple of them. Feedback from you and from teachers using Hedy would be very helpful so we can create metrics that actually matter :)

Felienne:
Let's brainstorm on this Thursday!

fpereiro:
To implement:
- Runs per level (granularity: month, all time)
- Highest_level_reached per user (maximum)
- Error rate per user per run programs (granularity: week)
- Exceptions per level and per type (granularity: month, all time)

PR #1540, opened 2021-12-13
Add metrics about program runs #665

boryanagoncharenko:
Add metrics about program runs

**Description**

- Add support for database query on index with sort key
- Add /stats age available for admins

**Fix for**

#665

**How to test**

1. Try opening the /stats page as an anonymous user. You should get a 403. Test opening the /stats page as a logged in user that is not an admin. Again, you should get a 403.
2. As an anonymous user, go to level 1 and run a program. As an admin open the /stats page. Ensure they page loads correctly and there are no metrics.
3. As a logged in user go to levels multiple levels and run programs. Make sure that you run programs that yield success and failure. Go to /stats page and ensure that data is present. Note that the data will be identical for past month and past week. If you want to see diff data, open the dev_database.json file and change any date in the program-stats to be more than a week ago.

Now that there is data, examine the charts. Since this is the admin API, the data is for all users (as opposed to a teacher looking at a class only).
1. The Hedy levels are on the X axis and the values per level on the Y axis. For now we will stay away from showing the same data with time on the X axis. This feature will be reevaluated with teachers.
2. The bug in which levels are ordered 1, 12, 2 should be gone.
3. The logs drill-down will be prototyped using Athena and integrated when further feedback is collected :)

**Checklist**


Felienne:
Ordering issue seems fixed!

image

Felienne:
This does not yet have a drill down yet right? I assume we will do that in a new PR? Otherwise it looks perfect!

boryanagoncharenko:
> This does not yet have a drill down yet right? I assume we will do that in a new PR? Otherwise it looks perfect!

Yes, I will make the drill-down in a separate PR since I need to discuss with you pricing first

Felienne:
Perfect than we can merge this now!

Felienne (approved):

PR #1543, opened 2021-12-14
Implemented user achievements

### LINKED ISSUE #1516: [UI idea] Implement "unlocking levels through achievements"
**Idea incl level**
Currently users are able to navigate through all levels and from the moment of entering the Hedy environment execute code on all these levels. It would be nice to have a bit more of an "enforced" level-structure. We already have a "forbidden" level structure implemented through the level customization dashboard so we are already able to hide specific levels for users. We can built upon this concept and the idea is as follows:

- We define achievements per next level (keywords used, programs created etc.)
- Students can see these achievements and receive badges when meeting the requirements
- When a certain threshold of achievements is achieved (for example, 60%; the next level is unlocked)

We should include this into the teacher customization to allow the teacher the following additional settings:

- Define relevant achievements
- Customize amount for these achievements (x amount of keywords used etc.)
- Customize threshold

Several other nice to have features would be:

- Global ranking of badges achieved
- Public profile with user overview of general statistics
- Profile overview of all achieved and to-be achieved achievements
- Implement some "global" achievements that only one user can have (most programs, most classes etc.)
- Implement "hidden" achievement which are a surprise until you get it!

**Food for thoughts**

- Should we implement the achievements as a dict on the user table or create a new table?
- For achievements (total amount of programs created for example) we have to keep track of a lot more of user data
- Might be useful to create an additional POST at every run containing relevant program info

TiBiBa:
**Description**
In this PR we implement user achievements. Small goals user can work towards and when reached unlock the achievement and receive a badge. Due to the complex structure of this project it is already stored here as a PR draft to keep track of WiP and possible errors. For a complete overview of the achievements that will be implemented as well as the current state and thoughts on the implementation, notice the discussion [here](https://github.com/Felienne/hedy/discussions/1541).

**Fix for**
This PR fixes #1516.

**How to test**
Notice the new profile option on the 'My-profile' hover that directs to the achievements page. This page contains an overview of the already received (and still to get) achievements.

**Checklist**


TiBiBa:
The front-end UI of the overview is almost done. Containing :hover elements to give the user an explanation on how to get the achievement and corresponding badge:

**New achievements UI**
Schermafbeelding 2021-12-14 om 14 02 22
Schermafbeelding 2021-12-14 om 14 02 04

TiBiBa:
**UI concept**
Uploading Schermopname 2021-12-14 om 15.11.01.mov…

TiBiBa:
As of right now a total of 39 achievements are written and implemented!🎉

app.py
```diff
@@ -1151,16 +1188,24 @@ def list_programs(user):
     return {'programs': DATABASE.programs_for_user(user['username'])}
 
 
-# Not very restful to use a GET to delete something, but indeed convenient; we can do it with a single link and avoiding AJAX.
-@app.route('/programs/delete/<program_id>', methods=['GET'])
+@app.route('/programs/delete/', methods=['DELETE'])
```
rix0rrr, line 1191: I agree with you that the `GET` here was wrong, and this needs changing, I don't think `DELETE` goes with this URL structure. `DELETE` should not have a request body: the URL alone should identify the resource to be deleted.

Any of the following would make sense:

HTTP verb | Parameters | Interpretation
-----------|------------|----------------
DELETE /programs/1234 | |  `/programs/1234` is a resource which we're asking to DELETE
POST /programs |  { command: 'delete', id: 1324 } |  send a 'delete' command to an endpoint that manages programs
POST /programs/delete | { id: 1234 } | send an 'id' to an endpoint that only does deletes
POST /programs/1234/delete | | invoke the 'delete' command of program 1234
POST /programs/delete/1234 | | alternative URL structure for the previous one
```

Of those, `DELETE /programs/1234` is the most HTTP-pure. However, one might ask: what do we buy for that? The API is not for other people to consume, only for us, and we might as well do what is convenient.

I myself used to have strong opinions on how REST APIs are "better" than RPC APIs (just like I had very strong opinions on brace styles)... but then I worked on some projects and found it didn't actually matter all that much and I didn't care anymore about style or purity, only about what was (a) correct and (b) convenient.

Long story short: I agree that `GET` is wrong and should be changed. Either POST or DELETE would be correct. The simplest change that will make our broken system correct is by changing GET to POST, and leaving everything else the same, in my opinion.

Of course, free advice is worth exactly what you're paying for it. Feel free to do with this opinion what you want.


(Another two cents: seems like this PR is growing quite large, and this change in particular doesn't seem like it has anything to do with the title of the PR. Are you trying to do too many things at once? You might want to make different PRs out of different logical changes to make it easier to review, discuss and approve them in isolation).

Comment (TiBiBa) [reply app.py:1191]: Thanks for the advice! I will re-write it to be conform one of your options. This PR is indeed getting a bit too big for its purpose. This is mainly due to having to re-write some stuff to enable us to return a response with possible achievements. (As re-writing the GET into a DELETE). So all is still related to the achievements, but got a big out of hand while implementing them.

**Comment by rix0rrr** [comment]
I'll go through the e2e failures one by one:

```
____________________ TestProgram.test_valid_delete_program _____________________

self = <tests_e2e.TestProgram testMethod=test_valid_delete_program>

    def test_valid_delete_program(self):
        # GIVEN a logged in user with at least one program
        self.given_user_is_logged_in()
        program = {'code': 'hello world', 'name': 'program 1', 'level': 1}
        program_id = self.post_data('programs', program)['id']
    
        # WHEN deleting a program
        # THEN receive an OK response code from the server
>       headers = self.get_data('programs/delete/' + program_id, expect_http_code=302, return_headers=True)

tests_e2e.py:761: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests_e2e.py:182: in get_data
    self.assertEqual(response['code'], expect_http_code, f'While reading {path} (user: {self.username})')
E   AssertionError: 404 != 302 : While reading programs/delete/ef9163a874c44f938deddd3ca00c2bde (user: user98490)
```

This is from this code: https://github.com/Felienne/hedy/blob/2ed4621472a47628c866f7abf53aec54e2a1a6a0/tests_e2e.py#L757-L759

It used to be sending a `GET programs/delete/1234`. Since you changed the server and `GET programs/delete/1234` is no longer the right command to send to delete a program, you need to update the test as well to send thew new command.

There is no `self.delete_resource` yet, so you will have to make that. Or you can use `self.post_data` if you are going to use a POST.

N.B: Don't forget to change this as well. This test now succeeds *for the wrong reason*:

https://github.com/Felienne/hedy/blob/2ed4621472a47628c866f7abf53aec54e2a1a6a0/tests_e2e.py#L747-L749

**Comment by rix0rrr** [comment]
Next failure:

```
    def test_join_class(self):
        # GIVEN a teacher
        self.given_teacher_is_logged_in()
    
        # GIVEN a class
        self.post_data('class', {'name': 'class1'})
        Class = self.get_data('classes') [0]
    
        # GIVEN a student (user without teacher permissions)
        self.given_fresh_user_is_logged_in()
        student = self.user
    
        # WHEN attempting to join a class without being logged in
        # THEN receive a forbidden status code from the server
>       self.get_data('class/' + Class['id'] + '/join/' + Class['link'], no_cookie=True, expect_http_code=403)
```

I imagine this is the same reason? Joining a class is a stateful operation, so I'm guessing you also changed that to a POST? So this needs to change into a `self.post_data()` call.

**Comment by rix0rrr** [comment]
That function is defined here btw: https://github.com/Felienne/hedy/blob/2ed4621472a47628c866f7abf53aec54e2a1a6a0/tests_e2e.py#L169

**Comment by rix0rrr** [comment]
Here is the same failure cause again, but in a different test:

```
    def test_see_students_in_class(self):
        # GIVEN a teacher (no classes yet)
        self.given_fresh_teacher_is_logged_in()
        teacher = self.user
        # GIVEN a class
        self.post_data('class', {'name': 'class1'})
        Class = self.get_data('classes') [0]
    
        # GIVEN a student (user without teacher permissions) that has joined the class
        self.given_fresh_user_is_logged_in()
        student = self.user
>       self.get_data('class/' + Class['id'] + '/join/' + Class['link'], expect_http_code=302)
```

Again, it's because joining a class uses a different URL now.

There is yet one more test that uses the same code. At this point, I wonder if it's not helpful to define this in one place:

```py
def given_user_has_joined_class(self, Class):
     self.post_data('class/' + Class['id'] + '/join/' + Class['link'], expect_http_code=302)
```

And use that everywhere. That will reduce duplication in the tests and if you ever decide to change the URL structure again, it will only require one change.

app.py
rix0rrr: > This is mainly due to having to re-write some stuff to enable us to return a response with possible achievements.

I understand. But assuming those other changes are useful on their own (which it looks like they are), it is not going to be wasted effort to extract them, make individual PRs out of them, merge those first, and then continue with this big work.

Felienne:
Down to two failing tests :) We are getting there!!

image

TiBiBa:
> Down to two failing tests :) We are getting there!!
> 
> 
> 
> image
> 
> 

It all seems to work! Will take a look later today it we need some additional testing written for the achievements.

TiBiBa:
I don't think we need additional e2e tests for the achievements. One possible issue can be a user making a POST and storing an achievement themselves. However, we implemented a check to verify if an achievement actually exists before sending it to the db and the users don't know the exact keys (unless they check out the code).

Felienne:
Hi @TiBiBa!!!

Just tested the branch extensively and it looks AWESOME!!! I updated the image into something more Hedy style but I think we might be good to go to push this to alpha??

One little layout thing is that when I save a program, the achievement popup ends up underneath the achievements one:

image

That is not super pretty, can we solve that? Maybe wait with the achievement popup until the modal is gone?

Felienne:
Ow and I want to stress once more "on the record" haha that you really came up with LOVELY badges with lovely names!!

TiBiBa:
> Hi @TiBiBa!!!
> 
> 
> 
> Just tested the branch extensively and it looks AWESOME!!! I updated the image into something more Hedy style but I think we might be good to go to push this to alpha??
> 
> 
> 
> One little layout thing is that when I save a program, the achievement popup ends up underneath the achievements one:
> 
> 
> 
> image
> 
> 
> 
> That is not super pretty, can we solve that? Maybe wait with the achievement popup until the modal is gone?

I think we can fix it if we have some sort of listener on the modal to check if it's displayed. Will take a look! Not sure if we'll be able to solve this quickly.

TiBiBa:
@Felienne just found a bug that causes us to store each achievements twice in the db, resulting in duplicate entries. Not really an issue (even when we sent it to Alpha). Will take a look what causes this and commit a fix, but this might be after Christmas.

Felienne:
> @Felienne just found a bug that causes us to store each achievements twice in the db, resulting in duplicate entries. Not really an issue (even when we sent it to Alpha). Will take a look what causes this and commit a fix, but this might be after Christmas.

Ah good to know I did not check the database. Of course totally fine if you fix this after Christmas, there is no rush! Enjoy the holidays!

TiBiBa:
> > @Felienne just found a bug that causes us to store each achievements twice in the db, resulting in duplicate entries. Not really an issue (even when we sent it to Alpha). Will take a look what causes this and commit a fix, but this might be after Christmas.
> 
> Ah good to know I did not check the database. Of course totally fine if you fix this after Christmas, there is no rush! Enjoy the holidays!

I was unable to find the cause of this bug, didn't remember it being there earlier. It might have something to do we some re-writing I did on the code to reduce server load. I just wrote a (very ugly) bug in which we parse the achievements to a set and then back to a list to efficiently remove duplicates. Not really nice, but it works. 

For the modal I would suggest to keep the current implementation but change the pop-up modal to be less intrusive to the user. Only showing a full-screen pop-up when we require some user feedback (yes/no, accept, remove etc.). In the case of an informative pop-up ("Your program is saved") I would like to implement another modal that only shows a minimalistic bar at the top of the screen. This has two advantages:

- Does not interfere with our achievement layout and thereby not overlap
- Less intrusive to the user as they get a lot of modal pop-up at the moment

TiBiBa:
> > I was unable to find the cause of this bug, didn't remember it being there earlier. It might have something to do with some re-writing I did on the code to reduce db load. I just wrote a (very ugly) bug in which we parse the achievements to a set and then back to a list to remove duplicates. Not really nice, but it works.
> 
> Ok, shall we merge this now then, so we can test it in alpha well before the schools open again?
> 
> Do we need an issue for the weird bug so we don't forget about it (with a link to this PR so we can dive into the issue)?
> 
> > For the modal I would suggest to keep the current implementation but change the pop-up modal to be less intrusive to the user. Only showing a full-screen pop-up when we require some user feedback (yes/no, accept, remove etc.). In the case of an informative pop-up ("Your program is saved") I would like to implement another modal that only shows a minimalistic bar at the top of the screen. This has two advantages:
> > 
> > * Does not interfere with our achievement layout and thereby not overlap
> > * Less intrusive to the user as they get a lot of modal pop-up at the moment
> 
> Sounds good! I see we already have an issue and PR there!

I think we are ready to merge then! Will create an issue for the duplicate value storing in the db. Weird thing is: When storing data and directly retrieving again the duplication is not there. Then when retrieving after a new code run some achievements are stored duplicate! I have absolutely no clue what can cause this weird issue. Maybe @rix0rrr has an idea?

Felienne (approved):

PR #1547, opened 2021-12-14
[BUG FIX] Fixes issue where the modal won't show due to error in ModesRules

TiBiBa:
**Description**
In this PR we temporary disable the localize keywords because the `Window.State.lang` value in this function breaks other js code for some reason.

**How to test**
Verify that the modal is currently not working in specific places when running locally or on alpha. For example, when trying to share or delete a saved program. After this PR the modal should work everywhere again.

**Checklist**


Felienne:
Superseded by #1548

PR #1507, opened 2021-12-08
[UI] Improved my-profile UI and functionality

TiBiBa:
**Description**
In this PR we clean-up the my-profile page and improve functionality. Currently the my-profile page only has functionality to change personal settings, password and/or delete the account. In this PR we introduce user statistics, class overview and a section-structure creating a more clean page. Shortly the followed functionality is introduced:

- Collapsable tabs, containing sections of relevant information
- User statistics (currently only amount of programs saved, can be expanded in future work)
- Self-removal of classes (student where till this day unable to leave a class)
- Setting preferred language (removing language selector from menubar when logged in)

**Old UI**
Schermafbeelding 2021-12-08 om 17 29 41

**New UI**
Schermafbeelding 2021-12-08 om 18 08 42

**Fix for**
This PR fixed #1467 and fixed #1478.

**How to test**

Describe how to run your code to see it is action. See this https://github.com/Felienne/hedy/pull/880#issue-1016304308 for an example.

**Checklist**

If you're unsure about any of these, don't hesitate to ask. We're here to help!
- [ ] Links to an existing issue or discussion (if not, create an issue first)
- [ ] Describes changes clear in the format above (present tense, no subject)
- [ ] Has a "how to test" section
- [ ] Only one thing is done in this pull request (specifically please try to refrain from mixing textual changes to the yamls from code changes)

PR #1548, opened 2021-12-14
[BUG FIX] Re-introduce auth.profile setting, incorrectly removed in #1507

TiBiBa:
**Description**
In the PR #1507 we remove a function from `auth.ts` because we receive the user data from the back-end instead of through a AJAX call in the front-end. However, a lot of code deepens on the existence of `auth.profile`. A value that is set in the removed function. In this PR we restore this function to make sure all functionality is working as it should be. 

**Fix for**
This PR fixes an issue introduces in #1507 where we no longer assign a value to auth.profile.

**How to test**
Store a program and notice that you are required to login (even when already logged in). After this PR this issue should no longer occur.

**Checklist**


Felienne:
Hi @TiBiBa!

I think my latest commit fixes the sharing without disabling the localized highlighting, can you check if that is the case?

PS this is my first real typescript contribution :D

Felienne (approved):

PR #1554, opened 2021-12-15
[BUG FIX] Fixed issue where users couldn't update profile with empty optional fields

### LINKED ISSUE #1553: [BUG] birth year, gender and country are mandatory when updating profile
**Describe the bug**
Currently users are able to create an account without setting an age, gender or country. These are all optional values and should remain optional. However, when updating your profile these are mandatory to contain a value. In addition to #1507 a lot of users will have to update their profile to select a preferred language. Therefore this bug has some priority. Also when trying to delete an already set birth year (something that should be possible) we receive an error as well.

We also perform some duplicate value checking. In the front-end as well in the back-end some value verification is performed. It would be nice to make sure everything is fine on the front-end (and return some useful feedback). And only check necessary values in the back-end (unique username, email, password). A lot of verifying can be performed within html by using input attributes.

It is super helpful if we can copy-paste the Hedy code to test, so please paste the code here, and don't forget to tell us what level you were in.

**Expected behavior**
Users are able to clear optional fields on their profile information as well as update their profile while keeping these optional fields empty.

TiBiBa:
**Description**
Currently we differentiate between and `signup` and a `profile` function. One for creating the function and one for updating the profile. In the first one (signup) we correctly set the values of all optional fields to `undefined` before sending the data to the back-end if the values didn't exist. However, this extra step was not implemented in the `profile` function, making it impossible for users with empty optional fields to update their profile. This is especially an issue as of #1507 users are obligatory to update their profile with a preferred language.

**Fix for**
This PR fixed #1553.

**How to test**
Create a new account and leave the optional fields empty. Then try to update this account in your profile settings and notice that you are now able to do so without setting values for each input. With this PR the bug of being forced to set values is solved and optional fields can stay empty if desired.

**Checklist**


Felienne (changes requested):
I just ran the code and this works! 

I just have a tiny nitpicking remark, sorry! We now have a menu section "mijn profiel" but a subheader of that is also called "mijn profiel" 

image

Shall we rename one of the two? Maybe we can make the overarching one "my account"?

TiBiBa:
> I just ran the code and this works!
> 
> I just have a tiny nitpicking remark, sorry! We now have a menu section "mijn profiel" but a subheader of that is also called "mijn profiel"
> 
> image
> 
> Shall we rename one of the two? Maybe we can make the overarching one "my account"?

Fixed!

Felienne (approved):

PR #1561, opened 2021-12-17
Further work for #1475

Felienne:
Allows = instead of `is` in comparisions.


Felienne:
@MarleenGilsing With this PR we allow the use of `=` in comparisons, for example (level 6)

```
naam is Hedy
    print 'ik heet' naam
    if naam = Hedy print 'leuk' else print 'minder leuk'
```

I have however left all examples as they were, I think it is better to stress in our examples that the is in if is different. But at least now from level 6 it is not forbidden anymore!

PR #1562, opened 2021-12-17
Add different error message for server errors on all auth functions

### LINKED ISSUE #1164: [BUG] Signup form will guess at the root cause of an error
**Describe the bug**

This is a sister ticket to #1163.

When a 500 server error occurs during form submission, the signup form will confidently state that the reason is because "username already exists".

This is misleading: the actual reason is there was a programming mistake on the server, and picking a different username is not going to help the user.

The form needs to make a distinction between:

* The form was submitted successfully, but the server said something was wrong with the contents of it
* The form was not submitted successfully

**Expected behavior**

An error message that accurately represents the situation (something went wrong on the server, you may try again but if the problem persists you should probably let us know)

fpereiro:
**Description**

- Adds a specific error message for all auth operations in the client (signup, login, change password, edit profile, recover/reset password.

**Fix for**

- Closes #1164 

**How to test**

- Add a generic error at the top of (say) the signup method (https://github.com/Felienne/hedy/blob/main/website/auth.py#L196). Something like `raise Exception('foo')` would work. This will trigger a 500 error when attempting to sign up.
- Attempt to sign up and notice that the error will contain a text referring to a *server* error.

Note: this PR adds both a code change and changes to the YAML texts because one part needs the other in order to be useful.


Felienne:
Thanks for this @fpereiro!!

Would be great if @TiBiBa could review this since he will be working with this code going forward

TiBiBa (approved):

PR #1563, opened 2021-12-19
Hashes variable inside turn command

Felienne:
**Description**

Hashes variables used in the `turn` command

**Fix for**

This:

**How to test**

This code now works at level 2:

```
ángulo is 90
girar ángulo
```

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -881,7 +881,6 @@ class ConvertToPython_1(Transformer):
             return "t.right(90)"  # no arguments defaults to a right turn
 
         arg = args[0]
-
         if is_variable(arg, self.lookup) or arg.isnumeric():
             return f"t.right({arg})"
         elif arg == 'left':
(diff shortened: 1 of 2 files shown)


hedy.py
```diff
@@ -881,6 +881,7 @@ def turn(self, args):
             return "t.right(90)"  # no arguments defaults to a right turn
 
         arg = args[0]
+
         if is_variable(arg, self.lookup) or arg.isnumeric():
             return f"t.right({arg})"
```
bjorn3, line 886: Does this need to be updated too?

hedy.py
```diff
@@ -881,6 +881,7 @@ def turn(self, args):
             return "t.right(90)"  # no arguments defaults to a right turn
 
         arg = args[0]
+
```
Felienne, line 884:
```suggestion
```

hedy.py
bjorn3: Never mind. Level 1 doesn't allow setting variables yet.

bjorn3 (approved):

PR #1576, opened 2021-12-22
Always check valid var usage

Felienne:
Fixes https://github.com/Felienne/hedy/issues/1323 (work by @dolfein)


dolfein:
About the failing test: it fails because the english variant of the variable is used (`people`) instead of the dutch one (`mensen`).

In the current state this does not raise an error until midway through the program, but with these changes this program will not run at all because a `UndefinedVarException` is raised.

Felienne:
> About the failing test: it fails because the english variant of the variable is used (`people`) instead of the dutch one (`mensen`).
> 
> In the current state this does not raise an error until midway through the program, but with these changes this program will not run at all because a `UndefinedVarException` is raised.

Ok, so I guess that is a matter of fixing the test?

coursedata/texts/nl.yaml
```diff
@@ -80,7 +80,7 @@ HedyErrorMessages:
     Parse: "De code die jij intypte is geen geldige Hedy code. Er zit een foutje op regel {location[0]}, op positie {location[1]}. Jij typte {character_found}, maar dat mag op die plek niet."
     Unquoted Text: "Let op! Bij print en ask moet voor én achter de tekst een aanhalingsteken komen. Jij bent er ergens eentje vergeten."
     Unquoted Assignment: "Let op. Vanaf dit level moet je tekst rechts van de `is` tussen aanhalingstekens zetten. Jij bent dat vergeten voor de tekst {text}."
-    Var Undefined: "Je probeert de variabele {name} te printen, maar die heb je niet ingesteld. Het kan ook zijn dat je het woord {name} wilde printen en aanhalingstekens vergeten bent."
+    Var Undefined: "Je probeert de variabele {name} te gebruiken, maar die heb je niet ingesteld. Het kan ook zijn dat je het woord {name} wilde gebruiken en aanhalingstekens vergeten bent."
```
Felienne, line 83: Really nice eye for detail also changing the error message @dolfein !

hedy.py
```diff
@@ -1516,12 +1520,17 @@ def assign(self, args):
         else:
             # we no longer escape quotes here because they are now needed
             return parameter + " = " + value + ""
-    
+
```
Felienne, line 1523: I am not sure why we need to add this? We already have such a var() in level 2: https://github.com/Felienne/hedy/blob/305f949991a6c344e42c2231224eeb994e412317/hedy.py#L1102

Felienne:
> > About the failing test: it fails because the english variant of the variable is used (`people`) instead of the dutch one (`mensen`).
> > In the current state this does not raise an error until midway through the program, but with these changes this program will not run at all because a `UndefinedVarException` is raised.
> 
> Ok, so I guess that is a matter of fixing the test?

Yes, I see now, your new code found a test in adventure code: 

image

I will be pushing a fix for that in the next commit. In this case it would have been totally ok for you to actually change the adventure code but I get that is a bit scary.

hedy.py
```diff
@@ -1496,6 +1498,8 @@ def assign_list_equals(self, args):
 
     def assign(self, args):
         right_hand_side = args[1]
+        left_hand_side = args[0]
+        self.check_var_usage([left_hand_side])
```
Felienne, line 1502: I don't understand why we are checking the left-hand side here, shouldn't that be the right-hand side? That should exist so we can assign it?

hedy.py
```diff
@@ -1249,6 +1250,7 @@ def equality_check(self, args):
     def in_list_check(self, args):
         arg0 = process_variable(args[0], self.lookup)
         arg1 = process_variable(args[1], self.lookup)
+        self.check_var_usage([arg1])
```
Felienne, line 1253: For this case, it would have been really nice to see a test added that catches an error that we did not catch before. Can you add one? Something like:

```if dier in dieren:
...```

Used to work, but now will now give a (desired!) exception.

Also shouldn't we tests both sides here?

Felienne:
Note to self: this does not yet check all constructs, f.e. loops are still missing.

Felienne:
Hi @dolfein,

I just pushed a bunch of tests, these are nice to have, because it makes it a lot more clear what is changed and it helps protect again future changes.

hedy.py
Felienne: Looked into it and this one is not needed, type checker can handle this

hedy.py
Felienne: This is not needed either, assignments will be checked in type checker

Felienne:
Think this is all ready now!

hedy.py
Felienne: Ah, I see, the answer seems to be: "because if we don't add this, the tests stop working"

But it should work I am diving into this.

PR #1580, opened 2021-12-23
Allows use of == from level 14

Felienne:
Addressing 3 of #1475


Felienne:
Note to self: Grammar works, but this still needs an update for the adventures and palette codes and some updates to the intro text

PR #1585, opened 2021-12-24
Workaround for 'copy share link to clipboard' in Safari after sharing program

### LINKED ISSUE #548: Copy to clipboard does not work on Safari
The reason is that "copy to clipboard" may only be done in response to a direct user click (and not in response to a user click which leads to an API call which THEN leads to the copy-to-clipboard).

The way around this is to implement a clipboard button. Instead of copying an invisible link to the clipboard, we can show a UI element with the link in it, and a copy-to-clipboard button right next to it, like for example PyPI does:

image

It is a pity that this leads to an extra click for kids but there does not seem to be another way.

fpereiro:
**Description**

Currently when making a program public from either the main programming page  or the programs page, we copy a share link to the clipboard. This, however, doesn't work in Safari because Safari doesn't allow to copy things to a clipboard after an API call if there hasn't been a click.

To solve this, we stop copying the share link altogether when making a program public, and instead add a 'copy share link' button on the modal that tells the user that the program has been saved successfully.

**Fix for**

- Closes #548 

**How to test**

- Share a program from the main programming view and make sure that there's a link to copy the program to the clipboard. Once you click on it, verify that a share link has indeed been copied to the clipboard.
- Repeat the above operation, but from the Programs page

**Implementation notes**

- it was also necessary to add extra logic to the Modal class so that if a modal B was shown before a modal A was dismissed, the timeout for hiding modal A would not hide modal B.
- We repurposed a text from the Programs section and moved it to the Auth section within the language yamls.


Felienne:
Thanks @fpereiro! 

Tested it on Chrome and Safari and it works well now! 

Just a few cosmic issues with the modal: Can we make the share button green too, and put the buttons next to each other? That is the way things are in other modal instances too. In the current form I worry kids will not see it it a button:

image

Other modal:
image

And maybe the 'ok' button should say 'close popup'? Ok is a bit meaningless here? I know it is the default in many situations but I feel for kids it might not be so clear.

coursedata/texts/el.yaml
```diff
@@ -113,8 +113,9 @@ Auth:
     unsaved_changes: 'Έχεις ένα μη αποθηκευμένο πρόγραμμα. Θέλεις να αποχωρήσεις χωρίς να το αποθηκεύσεις;'
     save_success: 'Επιτυχία'
     save_success_detail: 'Το πρόγραμμα αποθηκεύτηκε επιτυχώς'
-    share_success_detail: "Program shared successfully and share link copied to clipboard"
+    share_success_detail: "Program shared successfully"
```
Felienne, line 116: Do we need an En translation here? We would just default to En if it is missing right? (also doen not hurt of course...)

Felienne (changes requested):

coursedata/texts/el.yaml
fpereiro: Right now it is indeed superflous, since a missing text in a non-English defaults to English. However, this entry was already there; all I did was editing it so that the text would be correct, in keeping with the spirit of making the PR as focused as possible. This is the case with a few other languages as well, which have English texts for the `share_success_detail` key.

fpereiro:
Hi @Felienne ! I added a bit of JS trickery to place the copy button next to the OK button. I also made it green, to increase its visibility.

Regarding the "OK" button, I checked and we only use this text on all the alert-type modals throughout the app. Is this something that should be changed for them all, or just for this one?

coursedata/texts/el.yaml
fpereiro: If you want me to remove the untranslated `share_success_detail` entries from the yaml files, let me know and I can do it within this PR.

Felienne: yeah true, they were already there, we can also leave them (someone will work on a better translation UX anyway after Christmas)

Felienne:
> Hi @Felienne ! I added a bit of JS trickery to place the copy button next to the OK button. I also made it green, to increase its visibility.

Cool, will check it out now!

> Regarding the "OK" button, I checked and we only use this text on all the alert-type modals throughout the app. Is this something that should be changed for them all, or just for this one?

@TiBiBa plans some work on the modals anyway, I will add this to that open issue, so you can leave it as is.

Felienne:
> > Hi @Felienne ! I added a bit of JS trickery to place the copy button next to the OK button. I also made it green, to increase its visibility.
> 
> Cool, will check it out now!
> 
> > Regarding the "OK" button, I checked and we only use this text on all the alert-type modals throughout the app. Is this something that should be changed for them all, or just for this one?
> 
> @TiBiBa plans some work on the modals anyway, I will add this to that open issue, so you can leave it as is.

Nah looking at the modal now, I think we should remove the 'ok' in this case, because now it looks weird:

http://localhost:8080/hedy/60796e6780de42e084a65d143f97ed3d/view

Let's just do the 'copy' button alone, and no 'ok' here. If a kid clicks share, the modal is replaced, if a kid does not, it disappears after a while on its own.

fpereiro:
Hi @Felienne ! Indeed there was a bug that produced multiple buttons when the button was not clicked, sorry for that. I fixed it by adding an extra timeout. However, I'm not sure about how solid this solution will be -- for now, I wasn't able to break it. A better solution would add the show/hide logic as a callback passed to the alert method.

I also hid the OK button when showing the share button, and then re-showed it at the same time I removed the share button (either through a click or through a timeout).

@TiBiBa if you prefer, go ahead and merge your PR and then I can adapt this PR to your work - from what I see, it shouldn't be too complicated.

Felienne:
Cool, duplicate problem indeed seems fixed!

PR #1587, opened 2021-12-25
Split assigment rules from level 6 on

### LINKED ISSUE #1579: [BUG] Translation code does not preserve is/equal symbol
The translation code works well but has one small caveat since we allow `is` and `=` for assignment from level 6 on (since https://github.com/Felienne/hedy/pull/1504), code that uses `is` (which is allowed) get translated to `=` because in the grammar (and thus in the AST) we do not distinguish between the two.

There is a test for this in the code since https://github.com/Felienne/hedy/pull/1578, describing the current (undesirable) state.

https://github.com/Felienne/hedy/blob/c0effe1c66c2115848dda65b3e4c8b0373e6aeda/tests/test_translation_level_06.py#L41

What we want is code remains as is when translating, so code with lines with a `=` keep that and lines using a `is` keep the `is`.

This is not a big deal of course because the code will still work in the new form but I think it is not pretty. And in conditions (allowed since #1561) the example code uses `is` but then when you would translate it, it would become `=`.

And there will a similar issue with `is` versus `=` versus `==` when we address 3. of #1475 too!

The solution is not hard perse, but is a bit intricate and would involve:
1. making separate rules for an assignment with `is` and one for assignment with `=`
2. rewriting the parser so both rules map to the existing code and output the same Python
3. updating the translation code for both rules differently.

racertop:
**Description**

As stated on issue #1579 translation code does not preserve = and `is` converting everything to =. To solve this I split the following rules into 2 rules, one for is, and one for = from level 6 on:
* `ask` 
* `list_access` 
*  `assign` 
* `assign_list`

**Fix for**

Fixes #1579 

**How to test**

The transpiler to Python should behave the same, but translating to English should handle correctly `is` and =. I added a few tests on `tests\test_translation_level_06.py` to test this behavior. 

**Checklist**

### DIFF
--- a/grammars/level12-Additions.lark
+++ b/grammars/level12-Additions.lark
@@ -1,7 +1,12 @@
 //assigning a string now must be done with quotes
 
-assign: var _SPACE (_IS|_EQUALS) _SPACE sum | var _SPACE (_IS|_EQUALS) _SPACE text_in_quotes
-assign_list: var _SPACE (_IS|_EQUALS) _SPACE (text_in_quotes|NUMBER) ((_COMMA _SPACE|_COMMA) (text_in_quotes|NUMBER))+
+//assign: var _SPACE (_IS|_EQUALS) _SPACE sum | var _SPACE (_IS|_EQUALS) _SPACE text_in_quotes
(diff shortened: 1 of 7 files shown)


Felienne (approved):
Great work @racertop!

PR #1591, opened 2021-12-26
Complete German translation of UI

nichtich:
Some entries of German user interface had not been translated from the English original, so I tried to complete it. This subsumes <https://github.com/Felienne/hedy/pull/1586> (because it would otherwise result in a merge conflict) so better first discuss the rename before merging the rest.


nichtich:
I'll have another look at this before merge request.

Felienne:
ok!

PR #1592, opened 2021-12-27
Adding warn message for duplicated rules when merging grammars

### LINKED ISSUE #1559: Check for duplicate grammar rules
In https://github.com/Felienne/hedy/pull/1556 I removed a duplicate grammar rule. Would be nice if we could check, while merging, whether an exact rule is there already and warn for that.

racertop:
**Description**

Now the merger warns for duplicated rules: it tells the rule and the level where it's duplicated. 

**Fix for**

Fixes #1559 

**How to test**

There are no tests for checking the grammar, but you can try duplicating the print rule in, let's say, level 11, and it should output this message:

```
 hedy/hedy.py:1645: UserWarning: The rule print is duplicated on level 11. Please check!
```


Felienne (approved):
Ah great! Happy also to see no duplicate rules were found by this new warning :)

PR #1596, opened 2021-12-28
Implemented less intrusive modal alert

### LINKED ISSUE #1594: [UI idea] Make pop-up modal less intrusive on confirmation
**Idea incl level**
Currently we confront the user with quite a lot of modal pop-ups. When storing, sharing or submitting a program. As well as deleting one or joining or leaving a class. When we require user feedback it is relevant that the pop-up is intrusive and the user's attention is drawn to the question at hand. However, when returning some confirmation as feedback for which we **don't** require user interaction it would be nice to make the information less intrusive. I propose to create a separate modal that only shows a minimalistic information bar in this case. Concrete, this happens when:

- A user saves a program
- A user shares/unshares a program
- A user deletes a progrma
- A user joins a class
- A user leaves a class

This improvement is in line with our long-term goal to change saving of programs into a more intuitive way of working for children. Note that there is already a `timeoutMS` attribute on which we can apply a value after which the `alert()` function is automatically hidden. However, I don't like this implementation and the feedback to users is unclear if the modal just disappears.

TiBiBa:
**Description**
In this PR we implement a less intrusive modal alert when returning feedback to the user **when we don't require any user information**. Basically this pop-up is only shown to give the user some additional information if something is stored, saved, shared, submitted, deleted correctly. We re-wrote the `modal_alert` class in `layout.html` and added some additional `fade` to make it look nice.

**Fix for**
This PR fixed #1594.

**How to test**
Either delete, share, unshare or submit a program. In all these situations the new pop-up will show at the top of the screen.

**Screenshots**
Schermafbeelding 2021-12-28 om 10 05 35 (2)
Schermafbeelding 2021-12-28 om 10 06 09 (2)

**Checklist**


Felienne:
This looks lovely!!

But do not that the share program modal need to be "intrusive", because soon it will have a button to copy (see #1585) This is needed to work in Safari.

TiBiBa:
> This looks lovely!!
> 
> But do note that the share program modal need to be "intrusive", because soon it will have a button to copy (see #1585) This is needed to work in Safari.

Will take a look at how to best show the copy button! If we first merge #1585 I will alter this PR to work afterwards.

Felienne:
> > This looks lovely!!
> > But do note that the share program modal need to be "intrusive", because soon it will have a button to copy (see #1585) This is needed to work in Safari.
> 
> Will take a look at how to best show the copy button! If we first merge #1585 I will alter this PR to work afterwards.

Yes that sounds like a plan. There still is a weird bug in that PR but I am sure @fpereiro will be able to fix that.

TiBiBa:
Converted to draft as it conflicts with the latest update to the modal to fix the copy link button on shared programs. We have to do some re-structuring to still enable this copying. I would suggest to **not** use the `alert` modal when sharing a program, but instead the `confirm` modal. Making it easiest to implement this change.

Felienne:
> Converted to draft as it conflicts with the latest update to the modal to fix the copy link button on shared programs. We have to do some re-structuring to still enable this copying. 

Yeah sorry about the interference!

> I would suggest to **not** use the `alert` modal when sharing a program, but instead the `confirm` modal. Making it easiest to implement this change.

Sounds good to me it that is easiest for now. Will the confirm modal then include button with the link for pasting?

TiBiBa:
> Sounds good to me it that is easiest for now. Will the confirm modal then include button with the link for pasting?

Yes! But on second thoughts it might be best if we just create a new modal for the copying issue. We already have one for input, confirm and alert. Adding one for copying should be no issue. I would also prefer this method as with the latest fix we use the `alert` modal while it isn't actual an alert: it's a modal with a copy button. I suggest to split this issue into two parts:

- Re-write copy option into new `copy` alert
- Re-implement this PR to make `alert` modal less intrusive

TiBiBa:
After merging #1617 we should be able to do some clean-up and fix this PR as well!

Felienne:
Works perfectly for saving and sharing, but for deleting and submitting I do not seem to get this modal (or any modal) 

I guess that's ok because the state change is clear (program disappears or becomes grey) but the How to test states it should show so I am checking just to be sure.

Felienne:
> > 
> 
> My bad! There is currently no modal implemented when removing or submitting a program. It would be best to give the user some feedback if this is executed successfully (other than the program being removed / turning grey by a reload). But this requires quite some re-work to perform dynamically re-loading of the page. However, I just had an idea how we can also perform this visual change on the front-end only. This also decreases db calls as we call the programs again at a reload.
> 
> But for now this works as expected, on the long term I would like to improve the general process of showing a modal.

No problem, I thought that was the case but just wanted to check. I think it is clear enough as it is, so we can keep it like this until you have implemented your front-end magic!

Felienne (approved):

PR #1603, opened 2021-12-28
[BUG FIX] Users can self-delete an account

### LINKED ISSUE #1602: [BUG] Unable to delete an account
**Describe the bug**
Due to the delete account `button` being within the update profile form we create an issue. The button does two actions simultaneously:

- Update the profile
- Call the `auth.destroy()` function

This results in the page being reloaded while there is a modal pop-up with the question if the is sure to remove their account. This, of course, is not the desired behavior. We should move the delete button outside the form while keeping the same structure.  

**Expected behavior**
When we press the delete account button a modal pop-up should occur and wait for an answer. Only after the answer an action should be taken.

TiBiBa:
**Description**
In this PR we fix the issue where users where unable to delete their account due to a structure mistake. Due to this error the delete button was in the same `form` as the update profile button, creating a misunderstanding on the browser-side as to what action to perform onclick. By placing the button outside the form and style a surrounding div instead of the form itself we are able to solve the issue.

**Fix for**
This PR fixed #1602.

**How to test**
Verify that you can now correctly delete your account in profile settings.

**Checklist**


Felienne (approved):
Good catch @TiBiBa!

PR #1604, opened 2022-01-02
Extend German translation

nichtich:
I've extended and slightly polished the German user interface and examples.


coursedata/texts/de.yaml
```diff
@@ -1,21 +1,21 @@
 ui:
-    level_title: "Stufe"
+    level_title: "Level"
```
Felienne, line 2: This seems to be quite a big change, is `stufe` not right here? (I don't know a lot of German, so cc @dpprdan)

Felienne:
Nice work on the umlauts!! I think those examples stem from the days before we had unicode variable names!

I am not sure about stufe/level though, it seems a big change to me (but in Dutch we also use the English `level` and not the more Dutch sounding `niveau`, so I totally believe it makes sense!)

dpprdan:
"Level" is better, indeed. It's the same as in Dutch, "Level" is a borrowed word/loanword ("Lehnwort" in German), but is more appropriate in this context, especially given the target group. Also adds to the gamification. 😉 

I just skimmed over the other translations, but they look good to me as well. Nice work @nichtich!

Felienne:
> "Level" is better, indeed. It's the same as in Dutch, "Level" is a borrowed word/loanword ("Lehnwort" in German), but is more appropriate in this context, especially given the target group. Also adds to the gamification. 😉
> 
> I just skimmed over the other translations, but they look good to me as well. Nice work @nichtich!

Cool, thanks for letting us know!!

Felienne (approved):

Issue #167, opened 2021-02-19
R-side of check in if should be allowed to contain spaces


Felienne:
This code does not work:

kleur is ask Wat is je lievelingskleur?
if kleur is groene kleur print 'mooi!' else print 'niet zo mooi'

Because a space is not allowed in the right hand side of the if.

Felienne:
@boryanagoncharenko This has been bugging me forever so I am going to take a stab at it today!

Felienne:
Ok I have been playing with this a bit and the issue is that we allow things like this:

`if kleur is groen antwoord is ok else antwoord is stom`

This means:

```
if kleur is groen
antwoord is ok 
else antwoord is stom
```

So if we allow a space in rhs's the above will be parsed as something like "if kleur is groen antwoord" and then after that we fail because the is does not fit anything. 

I see two paths forward here:
* We start exploring ambiguous parsing (good to dive into anyway?) I know Lark should support it but I have never dove into it in detail
* We disallow assignment in a flat if in level 4 (and only allow this starting in level 7) 
* We only allow this type of if's when a newline (as layouted above) 

I like these two options because I am not a fan of "if kleur is groen antwoord is ok else antwoord is stom" When reading it I always get confused myself, because it misses a natural pause (or "then") between groen en antwoord

Felienne:
ok this turned out to be not as hard as I thought!! Do I get a trophy for closing the oldest issue? :D

PR #1607, opened 2022-01-03
Change grammar to allow spaces in rhs of if

Felienne:
Addresses #167


Felienne:
Ready for level 5 but needs a few fixes for higher levels

PR #1609, opened 2022-01-04
Achievement parsing

### LINKED ISSUE #1608: [BUG] Sample code on level 13 fails
**Describe the bug**
Locally or on alpha, navigate to level 13 (lang en) and execute the sample code. The error "Unexpected end-of-input. Expected one of: * _SPACE * _SPACE" is returned, which is not expected.

Level 13
Code:
```
name = ask 'what is your name?'
leeftijd = ask 'what is your age?'
if name is 'Hedy' and age is 2
    print 'You are the real Hedy!'
```

Screenshot 2022-01-03 at 21 09 38

**Expected behavior**
Sample code should not fail. It should work as it does on production.

Felienne:
Fixes #1608 

@TiBiBa, you created a small issue in the achievements checking. The first mistake was that you parse an individual line, which works fine for lower levels but once we get indented lines that no longer works. Secondly, you did not wrap the achievements checking iin a try block, allowing the error to escape and block further processing, resulting in a very weird state, a correct program with a parse error. Took me a bit of debugging to find that one!

This PR fixes both issues, but can you check I implemented the right idea for hedy-ious? I am now checking all arguments of prints and checking if there are 10 similar ones, but I am not explicitly checking whether a loop is used, so in principle it is not 100% correct but I guess it will do?

### DIFF
--- a/website/achievements.py
+++ b/website/achievements.py
@@ -190,7 +190,7 @@ class Achievements:
         if 'hedy-ious' not in self.achieved:
             all_print_arguments = hedy.all_print_arguments(code, level, self.lang)
             for argument in all_print_arguments:
-                if all_print_arguments.count(argument) > 10:
+                if all_print_arguments.count(argument) >= 10:
                     self.new_achieved.append("hedy-ious")
                     break
 
--- a/website/achievements.py
+++ b/website/achievements.py
@@ -174,8 +174,9 @@ class Achievements:
             self.new_achieved.append("deadline_daredevil_III")
 
     def check_code_achievements(self, code, level):
+        commands_in_code = hedy.all_commands(code, level, self.lang)
         if 'trying_is_key' not in self.achieved:
-            for command in set(hedy.all_commands(code, level, self.lang)):
+            for command in set(commands_in_code):
                 if command not in self.commands:
                     self.new_commands.append(command)
         if set(self.commands) == set(hedy.commands_per_level.get(hedy.HEDY_MAX_LEVEL)):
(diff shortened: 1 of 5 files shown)


TiBiBa:
> This PR fixes both issues, but can you check I implemented the right idea for hedy-ious? I am now checking all arguments of prints and checking if there are 10 similar ones, but I am not explicitly checking whether a loop is used, so in principle it is not 100% correct but I guess it will do?

Oops, my bad! I completely forgot that the `command` checker also parses the code. Will take a look right now! I think this solution for the Hedy-ious achievement is fine.

website/achievements.py
```diff
@@ -187,12 +188,12 @@ def check_code_achievements(self, code, level):
...
         if 'hedy-ious' not in self.achieved:
-            lines = code.splitlines()
-            for line in lines:
-                if "print" in hedy.all_commands(line, level, self.lang):
-                    if lines.count(line) >= 10:
-                        self.new_achieved.append("hedy-ious")
-                        return
+            all_print_arguments = hedy.all_print_arguments(code, level, self.lang)
+            for argument in all_print_arguments:
+                if all_print_arguments.count(argument) > 10:
```
TiBiBa, line 193: I think this should be `>=` instead of `>`, otherwise it all seems to work fine!

TiBiBa (changes requested):

website/achievements.py
Felienne: ah good catch, changed!

PR #1617, opened 2022-01-04
[CLEAN-UP] Implemented copy modal

### LINKED ISSUE #1612: [UI idea] Implement copy modal
**Idea incl level**
Due to an issue with the copying of shared programs an additional copy button is added to the `alert` modal when we want to share a program. However, in my opinion the `alert` modal should only be used to alert the user on something, without requiring interaction. We currently have a modal for input, confirm and alert. Adding one for copying should be no issue. This would also help solve the current merge issue of #1596 with which we implement a less intrusive alert modal.

TiBiBa:
**Description**
Currently we differentiate between three different types of modals: `confirm`, `alert` and `prompt`. Due to a bug with copying the share link in Safari an issue occurred: When wanting to return the user an alert to notify them that the link is copied we had to add an additional button to manually copy the share link. This change was implemented within the `alert` modal, but in my vision this is incorrect as the `alert` to only be used to notify users, not require some interaction.

In this PR we implement an additional modal for `copy` to solve this issue. When a users wants to share a program the `copy` modal is used instead of the `alert` modal with an added button. As this is a clean-up task no visual changes should be noticeable. This PR has the additional advantage that it helps us fix the merge errors on #1596. 

**Fix for**
This PR fixes #1612.

**How to test**
As this is a clean-up task no visual changes should be found. One can verify the correctness by either sharing a program directly from the coding page or within the programs page and notice that the copy modal is working correctly and the link is copied correctly as well.

**Checklist**


Felienne (approved):
Share modal still works!

PR #1625, opened 2022-01-05
Add types to SEMANTICS description for first levels

juliabolt:
**Description**

Adds types per command in the semantics description for level 1 and 2 

**How to test**

Read SEMANTICS.md file. Unsure about the types where I put a questionmark behind it, so maybe check those, and see if lay-out is okay like this.


SEMANTICS.md
```diff
@@ -5,14 +5,41 @@ in case of an error, the resulting HedyException.
...
+    <td>forward</td>
+    <td>integer</td>
+  </tr>
+  <tr>
+    <td>turn</td>
+    <td>'right' + integer | 'left' + integer</td>
+  </tr>
+  <tr>
+    <td>echo</td>
+    <td>any?</td>
```
Felienne, line 38: This can just be int + string I think, as we are doing for print? Because it supports the same operations. We can also do any for print of course, but it would be nicer to use the same.

SEMANTICS.md
```diff
@@ -107,15 +134,44 @@ Level 1 supports:
...
+  <tr>
+    <td>forward</td>
+    <td>integer</td>
+  </tr>
+  <tr>
+    <td>turn</td>
+    <td>integer</td>
+  </tr>
+  <tr>
+    <td>is</td>
```
Felienne, line 166: type signature of is should have two types I think, left and right? Left of the is we can have only valid variable names (not integers for example or symbols) and right we can have anything.

SEMANTICS.md
```diff
@@ -107,15 +134,44 @@ Level 1 supports:
...
+    <td>any?</td>
+  </tr>
+    <tr>
+    <td>sleep</td>
+    <td> empty | integer?</td>
```
Felienne, line 171: Yes that is correct!

Felienne (changes requested):

Felienne:
Thanks for your contribution @juliabolt!! I answered to the question marks above

SEMANTICS.md
juliabolt: valid variable names would be strings, I think?

juliabolt:
Added types to level 3 and 4 now too! :)

juliabolt:
I don't see the new/extra feedback you mentioned @Felienne ? Am I looking in the wrong place?

juliabolt:
I can re-request review if that helps?

Felienne (changes requested):
Zie eerdere opm

Felienne (approved):

PR #1628, opened 2022-01-05
[UI IMPROVEMENT] Update programs view without re-loading

### LINKED ISSUE #1627: [UI idea] Dynamically update programs page on submit, share or delete
**Idea incl level**
Currently we reload the whole page when we either share, un-share, delete or submit a program. It would be nice if we could perform this process of updating the programs view without the need of re-loading the whole page (and doing another db call!).

TiBiBa:
**Description**
Currently we reload the page when we either share, un-share, delete or submit a program. Through this process we retrieve an up-to-date version of our programs from the db and display this to the user. However, this approach has two large disadvantages:

- We use unnecessary db calls
- The reload complicates the showing of achievements

In this PR we implement a front-end approach to use the correct programs and state of these programs to the user. When we either share, un-share, delete or submit a program we call a javascript function that updates the page **without** needing additional interaction with the server. Reducing server load, db calls and improve our achievements experience.

**Fix for**
This PR fixes #1627.

**How to test**

**Checklist**


static/js/app.ts
```diff
@@ -546,10 +546,23 @@ export function viewProgramLink(programId: string) {
   return window.location.origin + '/hedy/' + programId + '/view';
 }
 
+function change_shared (shared: boolean, index: number) {
```
Felienne, line 549: What is the meaning of index here? Is it a program id? Of a page element id? In what situations is it -1?

Can we add that as a comment because as it is I don't really understand the meaning.

TiBiBa: My bad! The index is a front-end element id so we are able to remove / hide this element without have to verify with the server again. Will add some additional line comments.

Felienne: Thanks! I don't look at the front-end code that often so I need a bit of help to stay in the know :)

TiBiBa:
There still seems to be an issue that we the `program_container` is either hidden or removed the margin stays. Resulting that when we remove the top 5 programs we will keep a significant margin to the header. Not sure how that happens, but will take a look!

TiBiBa:
> There still seems to be an issue that when the `program_container` is either hidden or removed the margin stays. Resulting that when we remove the top 5 programs we will keep a significant margin to the header. Not sure how that happens, but will take a look!

TiBiBa:
> There still seems to be an issue that when the `program_container` is either hidden or removed the margin stays. Resulting that when we remove the top 5 programs we will keep a significant margin to the header. Not sure how that happens, but will take a look!

Fixed! @Felienne I think we are ready to merge

Felienne (approved):

PR #1629, opened 2022-01-05
chore(dynamo): support pagination in DB engine

rix0rrr:
Add support for pagination to Dynamo engine.

Both the `get_many()` and `scan()` functions now have two new
parameters: `limit` and `pagination_token`. These indicate how many
records to return, and at what location to start retrieving those
records, respectively.

The return value of these functions still mostly behaves like a list,
but is actually a `ResultPage` object, which has two new attributes:
`next_page_token` and `has_next_page`.

Together, they are supposed to be used like this:

```python
page1 = db.get_many({ 'key': 'value' }, limit=10)
for record in page1:
    print(record)

page2 = db.get_many({ 'key': 'value' }, limit=10, pagination_token=page1.next_page_token)

if page2.has_next_page:
  page3 = db.get_many(..., pagination_token=page2.next_page_token)

  # etc.
```

`pagination_token` is a string that safe for rendering in a webpage or
hyperlink, but don't make any assumptions on its structure. It's
currently not safe against tampering, so be aware of that.

If you need the return value of `db.get_many()` to be an ACTUAL list (because you
want to do additional manipulation like call `.append()` on it), convert it to a list 
like so: `xs = list(db.get_many(...))` (or access `.records` directly).


TiBiBa:
Awesome! This will be really useful for the `explore` page. I just did some testing and one issues occurred, when using:

'''
page1 = PROGRAMS.get_many({'username': username}, limit=1, reverse=True)
        print(page1)
        page2 = PROGRAMS.get_many({'username': username}, limit=1, pagination_token=page1.next_page_token, reverse=True)
        print(page2)
'''

TiBiBa:
Awesome! This will be really useful for the `explore` page. I just did some testing and one issues occurred, when using:

```
page1 = PROGRAMS.get_many({'username': username}, limit=1, reverse=True)
print(page1)
page2 = PROGRAMS.get_many({'username': username}, limit=1, pagination_token=page1.next_page_token, reverse=True)
print(page2)
```
The second call results in an empty `ResultPage`. I noticed that this is due to the `reverse=True`. When setting the value to `False` it works as expect.  But because we don't have a `prev_page_token` we are unable to call a subset of the programs in the desired order. Is there a way to change the implementation that this is supported?

Felienne:
> Awesome! This will be really useful for the `explore` page. I just did some testing and one issues occurred, when using:
> 
> ```
> page1 = PROGRAMS.get_many({'username': username}, limit=1, reverse=True)
> print(page1)
> page2 = PROGRAMS.get_many({'username': username}, limit=1, pagination_token=page1.next_page_token, reverse=True)
> print(page2)
> ```
> 
> The second call results in an empty `ResultPage`. I noticed that this is due to the `reverse=True`. When setting the value to `False` it works as expect. But because we don't have a `prev_page_token` we are unable to call a subset of the programs in the desired order. Is there a way to change the implementation that this is supported?

@rix0rrr Thanks for this and for fixing the tests! I am not sure if your latest changes also address this comment? Or are you not planning to add this option (also fine of course, in that case we can merge this)

rix0rrr:
> The second call results in an empty `ResultPage`. I noticed that this is due to the `reverse=True`.

Hmm. I suppose this is in the in-memory engine, not the real engine, right?

That's odd because I did write a test for that:

https://github.com/Felienne/hedy/pull/1629/files#diff-c548907ac1208ec2ee89456cd5b673f9c85a78b487c2ff7cb678d397e75e5534R202

 Can you make a minimal reproducing example?

rix0rrr:
Ohhh the difference might be that `users` is a table without a sort key. Lemme check.

boryanagoncharenko:
The change works great when I connect to my AWS account. However, the in-memory implementation has an issue indeed. I got the same issue (as the one mentioned above) when querying PROGRAM_STATS by index with reverse.

Felienne (changes requested):
Setting this to Changes Requested to keep a clean overview of open PR's

rix0rrr:
Got it -- indexes have their own sort key which the in-memory database was not respecting.

It is now.

TiBiBa:
I'm still running into the same issue, but I might be doing something wrong. I would like to access the last x amount of programs, followed my the next last x amount of programs. However, when using `reverse=True` of the `get_many()` the `next_page_token` is empty. Which makes sense from one perspective because we look at the last program, but as we use reverse I would like the token to point at location `last program - x` when retrieving x programs. See the code snippet below that results in `program2` being `None`.

```
program1 = PROGRAMS.get_many({'username': username}, limit=1, reverse=True)
program2 = PROGRAMS.get_many({'username': username}, limit=1, pagination_token=program1.next_page_token, reverse=True)

print(program1)
print(program2)
```

Felienne:
@TiBiBa Could you approve this? I set it to automerge so if you are happy it merges straight away!

TiBiBa (approved):

Felienne (approved):

PR #1630, opened 2022-01-05
Add integration with Athena and log search to /stats page

boryanagoncharenko:
**Description**

- Add integration with Athena
- Add accordion to the /stats page that hosts charts with program runs and exceptions data
- Add interface for searching logs to /stats page

**Fix for**

#665

**How to test**

The charts are now organised in accordion sections per type. In this way the admin can explore which charts are better with real world data and collapse the charts that will not be used.

Unfortunately, log search can only be tested on Alpha. The Athena setup on Alpha is complete, so once it is deployed exercise the following scenarios:
- Make a search with no params -> all data should be returned. While on that screen when there is data, check:
    - The date should contain hours -> 2022-01-10 14:56:21 instead of only 2022-01-10.
    - The code column should display the code formatted on multiple lines
    - The copy-code button should appear in every cell containing code. NOTE that if you are using http://0.0.0.0:8080/stats then copying will fail and the button will display "Failed!" if you click it. To test the copy button, go to http://localhost:8080/stats and click it.
- Make a query that is narrowing down search and ensure the returned data matches the restriction.
    - Make a query for a concrete level: 15
    - Make a query with a wildcard, e.g. Parse% instead of ParseException
    - Make a query for a specific interval. Note that now the start and end dates are strings. This means that if you put both the start date and end date to be 2022-01-01 you will get no results even if there are logs on that date. You will need to specify the minutes, so that the string comparison works. Probably this is something to improve for the teachers' stats page? Or even now for the admin page too? 

**Checklist**


Felienne:
Merging to deploy on alpha. Tested the graphs locally and they look great!

Felienne (approved):

Felienne:
> * Make a query for a specific interval. Note that now the start and end dates are strings. This means that if you put both the start date and end date to be 2022-01-01 you will get no results even if there are logs on that date. You will need to specify the minutes, so that the string comparison works. Probably this is something to improve for the teachers' stats page? Or even now for the admin page too?

Yeah it would be a bit more user-friendly if from 2022-01-01 to 2022-01-01 would give all the programs on that date, but it is not a pressing issue. Maybe we can make an issue for it just so we don't forget?

boryanagoncharenko:
> 

Created #1670

PR #1637, opened 2022-01-06
Fix #1636

### LINKED ISSUE #1636: [BUG] chart.js load on page call error
**Describe the bug**
Currently there seems to be a mistake with the `chart.js` library being called on every page. Resulting in a JS console error as some decencies are not found on every page. See the screenshot below for clarification. I expect this issue to occur due to a function always being called in `auth.ts` regardless of the current page. I expect this can be solved by storing this function inside a dedicated call to the `program stats` page.

Schermafbeelding 2022-01-06 om 17 25 41

**Expected behavior**
The `chart.js` dependency error no longer occurs.

TiBiBa:
**Description**

Fix the issue where `chart.js` is called on every page, resulting in some undesirable decency errors. By storing some code inside a function that is only called when we call the dedicated page the error is fixed.

**Fix for**
This PR fixed #1636.

**How to test**
Load any page on alpha and on this branch and navigate to your browser console. Notice that the `chart.js` error is no longer there. (The other errors are for a later moment to solve).

**Checklist**


Felienne:
Awesome, error seem gone indeed!

Felienne (approved):

PR #1642, opened 2022-01-07
[HOTFIX] Temporary disable achievements

TiBiBa:
**Description**
Due to an unknown bug achievements are often not achieved or the wrong ones are returned to the user. Until this is fixed they should be invisible to the user. In this PR we disable all functional and visible achievements functionality, making sure that the user is not bothered by incorrect achievements.

**Checklist**

### DIFF
--- a/app.py
+++ b/app.py
@@ -460,9 +460,9 @@ def parse():
 
             try:
                 transpile_result = transpile_add_stats(code, level, lang)
-                if username:
-                    DATABASE.increase_user_run_count(username)
-                    ACHIEVEMENTS.increase_count("run")
+                #if username:
+                #    DATABASE.increase_user_run_count(username)
+                #    ACHIEVEMENTS.increase_count("run")
             except hedy.exceptions.InvalidSpaceException as ex:
                 response['Warning'] = translate_error(ex.error_code, hedy_errors, ex.arguments)
                 response['Location'] = ex.error_location
@@ -487,11 +487,11 @@ def parse():
         except:
             pass
 
-        try:
-            if username and ACHIEVEMENTS.verify_run_achievements(username, code, level, response):
-                response['achievements'] = ACHIEVEMENTS.get_earned_achievements()
-        except Exception as E:
-            print(f"error determining achievements for {code} with {E}")
+        #try:
+        #    if username and ACHIEVEMENTS.verify_run_achievements(username, code, level, response):
+        #        response['achievements'] = ACHIEVEMENTS.get_earned_achievements()
+        #except Exception as E:
+        #    print(f"error determining achievements for {code} with {E}")
 
     except hedy.exceptions.HedyException as ex:
         traceback.print_exc()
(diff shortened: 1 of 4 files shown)


rix0rrr:
Hi @TiBiBa, 

I had a look at the code and I have a good idea of why it might not be behaving as expected.

`ACHIEVEMENTS` is a global object, and it has a cache of achievements (to alleviate database pressure, which you know I 👍 100% approve of 😂). However, the achievements that are cached inside the object are for a single user.

For example, here we call `ACHIEVEMENTS.verify_run_achievements()`, on a global, shared instance of the Achievements class:

https://github.com/Felienne/hedy/blob/dc934574b9f46b3e609c53d3190b460e853432f4/app.py#L491

But inside the method, we treat `self.achievements` as the achievements of a single user. For example, we add the current user's new achievements to the array here:

https://github.com/Felienne/hedy/blob/dc934574b9f46b3e609c53d3190b460e853432f4/website/achievements.py#L99

Effectively, this means that we initialize `ACHIEVEMENTS.achieved` with the achievements of the *very first user* that happens to use the website after it restarts, and then we check all subsequent users against that list.

The reason why we only see this on beta and not on alpha, is because probably you are the only user on alpha and so you would not run into this situation.

--------

## Solution

The solution is to keep separate achievement caches per user. We can do that in a couple of ways.

### Strategy 1: turn the cache into a map

The most obvious way to do this, is to turn `self.achieved` on the `Achievements` class from a list into a map (and honestly -- more of the members of that class, like `self.commands` etc.)

Example:

```py
# Instead of
self.achieved = ['question_asker', 'hedy-ious', 'make_some_noise']

# Turn it into this
self.achieved = {
  'rico': ['question_asker', 'hedy-ious'],
  'felienne': ['make_some_noise'],
}
```

**HOWEVER**! Don't do this, it has the following disadvantages:

* It's easy to know when to add elements to the map (when we need data for a user and we see the user for the first time). But over time we will see an unbounded amount of users, and so we will use more and more memory keeping track of the achievements of each of them. This requires that we also come up with some garbage collection mechanism to periodically clean out the map, and all in all comes with a bunch of complications and overhead.
* This cache lives inside a single Python process. You have to remember that we are running 16 Python processes in total (there are 2 servers running 8 processes each), and so we would need to load and keep this data in the worst case 16 times (because on every request every user might hit a different process, and on that process the cache would need to be loaded as well).

### Strategy 2: cache achievements in the user session

Flask supports [sessions](https://flask.palletsprojects.com/en/2.0.x/quickstart/#sessions). Sessions are scoped to the user: information you store in there only applies to the current user, and you don't have to remember to garbage collect it! We get the properties we want for free!

You might ask: where does that data get stored? It gets sent in a Cookie to the user's browser, and the browser sends it back on every requests. Therefore there's also only ever 1 copy stored, it's always available in the current Python process (regardless of which Python process loaded it) and it's automatically garbage collected when the user closes their browser. The only thing you have to pay attention to is that there is a size limitation on the cookie data. I don't know what it is, you have to look that up, but do be careful to store as little information as possible to keep within the limits. For example, if achievements had numbers you could save a lot of space by storing numbers instead of strings, or even a [bitmask](https://en.wikipedia.org/wiki/Mask_(computing).

Bottom line: if you rename `Achievements` to `UserAchievements` (to make the scope of its data very clear) and store the source of truth of per-user achievements in the `session` global, you should be fine.

If anything is unclear, don't hesitate to reach out!

Felienne:
Fixed properly with #1645

PR #1644, opened 2022-01-08
Spliting equal comparison rules and addressing #1643

### LINKED ISSUE #1599: [BUG] Translation code does not preserve == symbol
PR #1580 has created the exact same issue as we had in #1579 but for `==`.

### LINKED ISSUE #1643: [BUG] Cannot use boolean operators with number comparisons
**Describe the bug**
Starting from level 14 number comparison operators are introduced, but they can't be used with boolean operators.

Level 14
This gives a parsing error
```
age = ask 'How old are you?'
if age >= 0 and age < 18
    print 'You are a minor!'
else
    print 'You are OLD!'
```

**Expected behavior**
Comparisons operators should work with boolean operators.

racertop:
**Description**

I edited the grammar rules in levels 6, 12 and 14 to split the production rules for comparison using equals, is and double equals. Now each one of these have their own rule.

I also edited the way the condition rule is defined to use the merging operator introduced a few PRs ago. Also,  because the rules `andcondition` and `orcondition` manually added the comparison rules they worked with in the beginning of the rule it caused issue #1643. Since this was a small fix and was related with the work I was doing I'm addressing that issue in this PR as well.

**Fix for**

Fixes #1599 and fixes #1643 

**How to test**

I added a test in level 14 for translation and the hedy transpiler.

### DIFF
--- a/grammars/level12-Additions.lark
+++ b/grammars/level12-Additions.lark
@@ -7,7 +7,9 @@ assign_list_equals: var _SPACE _EQUALS _SPACE (text_in_quotes|NUMBER) ((_COMMA _
 
 ?atom: NUMBER | var //unsupported numbers are gone, we can now allow floats with NUMBER
 
-equality_check: (var | text_in_quotes | NUMBER) _SPACE _IS_OR_EQUALS _SPACE (var | text_in_quotes | NUMBER)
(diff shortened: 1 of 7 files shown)


Felienne:
Ah great catch (and fix) on the level 14 error!

Felienne (approved):

PR #1645, opened 2022-01-09
[BUG FIX] Achievements related data is stored in session

TiBiBa:
**Description**
Currently we use a global variable `achievements` to store the user progress and achievements. However, when multiple users are logged in at the same time they share this variable as well as the values. Resulting in the undesirable behavior that users where unable to achieve achievements (because another user already got them) or they received random pop-ups of other users their achievements. As pointed out by @rix0rrr we can solve this by storing all user-related data in a session.

In this PR the achievements implementation is re-written to store all user-related data in a session and compare the achievements against this data. Resulting in users no longer interfering with each other.

**How to test**
Use Hedy with multiple users at the same time and notice that they no longer interfere with each others achievements

**Checklist**

### DIFF
--- a/app.py
+++ b/app.py
@@ -206,10 +206,12 @@ def initialize_session():
     session_id()
     login_user_from_token_cookie()
 
-@app.after_request
+
+@app.before_request
 def initialize_achievements():
     if current_user()['username'] and 'achieved' not in session:
         ACHIEVEMENTS.initialize_user_data(current_user()['username'])
+
 
 if os.getenv('IS_PRODUCTION'):
     @app.before_request
@@ -491,9 +493,7 @@ def parse():
 
         try:
             if username and ACHIEVEMENTS.verify_run_achievements(username, code, level, response):
-                print("Hier komen we nog")
                 response['achievements'] = ACHIEVEMENTS.get_earned_achievements()
-                print("Hier niet meer?")
         except Exception as E:
             print(f"error determining achievements for {code} with {E}")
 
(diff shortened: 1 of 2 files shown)


Felienne (approved):

app.py
```diff
@@ -207,6 +207,12 @@ def initialize_session():
     login_user_from_token_cookie()
 
 
+@app.before_request
+def initialize_achievements():
+    if current_user()['username'] and 'achieved' not in session:
+        ACHIEVEMENTS.initialize_user_data(current_user()['username'])
```
rix0rrr, line 213: I'm not opposed to it, but doing it here depends on ordering. If in the same request a user logs in *and* is supposed to get an achievement (or not), using an `@app.before_request` requires that the code which sets `current_user()['username']` is executed in the same request before this code runs.

Did you verify that is the case? 

Otherwise, a simpler way to go would be to keep the first load completely a concern of the `ACHIEVEMENTS` class. It can easily see whether or not its data has already been initialized, and do the load whenever it's called.

Every method would start with a call to:

```py
self.initialize_session_if_necessary()
```

TiBiBa: I agree that it would be better to make it a concern of the `ACHIEVEMENTS` class. For now it all seems to work fine with `@app.before_request` as there is no achievements that can be achieved on the same request as login. As the highest priority is to get it correctly working before tomorrow I would like to leave it this way for now. I will create an issue to optimize the class overall, such as the storing of `commands` and `achievements` that can be optimized (as you suggested) through using maps and/or numbers for reference.

PR #1650, opened 2022-01-10
[FIX] Removed incorrect argument on function calling preventing program submitting

### LINKED ISSUE #1649: [BUG] Program doesn't correctly change visually when submitted
**Describe the bug**
Currently when we submit a program the user receives no feedback if their submission was successful. This issue might occur due to #1628 where we re-write the programs page structure to prevent re-loading to update user information. The program is correctly submitted in the back-end but not on the front-end (yet). With the additional issue that now the submitting achievements are not correctly shown to the user.

**Expected behavior**
When submitting a program the program box on the programs page should change visually and show the submission achievement if received.

TiBiBa:
**Description**
Due to an incorrect argument on the achievements function call to verify submitted programs the user interface was not correctly shown when submitting a program as well as the submitting achievements could not be achieved. In this PR we fix this function call automatically fixing the issue.

**Fix for**
This PR fixes #1649.

**How to test**
Submit a program, notice that the view is updated without re-loading and an submitting achievement is shown if achieved.

### DIFF
--- a/website/achievements.py
+++ b/website/achievements.py
@@ -109,7 +109,7 @@ class Achievements:
     def verify_submit_achievements(self, username):
         if not session['achieved']:
             self.initialize_user_data(username)
-        self.check_programs_submitted(session['submitted_programs'])
+        self.check_programs_submitted()
 
         if len(session['new_achieved']) > 0:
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Yes, this works as expected (as before graying out the program without a refresh)

PR #1652, opened 2022-01-10
Implemented public profiles

### LINKED ISSUE #1595: [UI idea] Create public user page
In line with creating a more "community-like" environment we would like to enable user to create a public webpage. On this page we can show some relevant user information. The users are also able to display a short bio, their country, age and, of course, their achievements. It would also be nice (in the future) that users can select a profile avatar from one of x (say 10) Hedy designed avatars. We can also show some relevant user information such as:

- Highest Hedy level achieved
- Total amount of run programs
- Total amount of stored programs

We will have to add a boolean to the sign-up sheet to ask users if they want their profile private or public. In alignment with this we have to ask all already existing users to choose an option. We can use the same implementation with header message as used for the preferred language selection.

TiBiBa:
**Description**
In this PR we introduce public profiles! Users are able to create a public profile in their profile settings. They are able to select a profile image, set a personal text and select a favorite saved program. Other users are able to visit public profiles by navigating to `/user/<username>`. When no public profile is set a 404 page is given.

**Fix for**
In this PR we fix #1595.

**How to test**
Navigate to profile settings and notice a new header: public profile. In this section you are able to set your personal public profile. Navigate to the profile by going to `/user/<username` and notice that your set values are shown.

**This is currently a draft! Already a PR to solve merge conflicts along the way**


Felienne:
Nice work @TiBiBa! It really looks amazing!!

I think it would be nice if it is clearer (maybe with a tickmark?) that setting the information makes the profile public, something like:

[] By selecting this box I make my profile visible for everyone. Be careful not to share personal information like your name or home address, because everyone will be able to see it!

And maybe we should also allow making the profile private again if people want to?

And regarding my remark on selecting the programs, I do think it is nicer if we allow kids to "pin" them form the programs page too, the dropdown is a bit hard to use (I do not remember what the coolest program I made was by name)

Felienne:
Ow and I am getting a weird header on the old profile:


image

TiBiBa:
> Nice work @TiBiBa! It really looks amazing!!
> 
> I think it would be nice if it is clearer (maybe with a tickmark?) that setting the information makes the profile public, something like:
> 
> [] By selecting this box I make my profile visible for everyone. Be careful not to share personal information like your name or home address, because everyone will be able to see it!
> 
> And maybe we should also allow making the profile private again if people want to?
> 
> And regarding my remark on selecting the programs, I do think it is nicer if we allow kids to "pin" them from the programs page too, the dropdown is a bit hard to use (I do not remember what the coolest program I made was by name)

Good suggestions! Will implement a text box as well as a "delete public profile" button. We already return a 404 is no public profile exists for a user so this should be easy to add. Regarding the "pin", this requires some re-work on the programs page, but should be easy to fix!

TiBiBa:
> Ow and I am getting a weird header on the old profile:
> 
> image

My bad, think I messed up a keyword somewhere. Will fix it right away.

TiBiBa:
> Nice work @TiBiBa! It really looks amazing!!
> 
> I think it would be nice if it is clearer (maybe with a tickmark?) that setting the information makes the profile public, something like:
> 
> [] By selecting this box I make my profile visible for everyone. Be careful not to share personal information like your name or home address, because everyone will be able to see it!
> 
> And maybe we should also allow making the profile private again if people want to?
> 
> And regarding my remark on selecting the programs, I do think it is nicer if we allow kids to "pin" them from the programs page too, the dropdown is a bit hard to use (I do not remember what the coolest program I made was by name)

All should be implemented with the last commit! After some testing I think we are ready to merge.

TiBiBa:
Reminder to self before merging: Always show favorite program at top of the `programs` page.

Felienne:
Hi @TiBiBa,

@rix0rrr had a good thought about this idea (also a sad thought) "what about spammers...." If people can create a profile including links, they can also post links to increase Google rank. So can we make sure all links posted get this tag:

<meta name="referrer" content="no-referrer"> (see: https://stackoverflow.com/questions/53430334/how-to-prevent-capturing-referral-disable-referral-or-hide-referral)

That should keep the spammers away.

TiBiBa:
> Hi @TiBiBa,
> 
> @rix0rrr had a good thought about this idea (also a sad thought) "what about spammers...." If people can create a profile including links, they can also post links to increase Google rank. So can we make sure all links posted get this tag:
> 
> (see: https://stackoverflow.com/questions/53430334/how-to-prevent-capturing-referral-disable-referral-or-hide-referral)
> 
> That should keep the spammers away.

Good point! Will implement this and some other verification checks to make sure the personal text is only used for normal information.

TiBiBa:
> Hi @TiBiBa,
> 
> @rix0rrr had a good thought about this idea (also a sad thought) "what about spammers...." If people can create a profile including links, they can also post links to increase Google rank. So can we make sure all links posted get this tag:
> 
> (see: https://stackoverflow.com/questions/53430334/how-to-prevent-capturing-referral-disable-referral-or-hide-referral)
> 
> That should keep the spammers away.

I'm struggling what would be the best approach. We can call a javascript function `onload()` and add the  `rel="noreferrer"` to all `<a>` element within our personal text space. But otherwise we can also catch it on the input side like suggested here: https://stackoverflow.com/questions/19080957/how-to-remove-all-a-href-tags-from-text.

TiBiBa:
As we set the `personal text` value of the user as plain text they are unable to use re-directs or any other way of linking on their profile. Of course, they are still able to use swear words etc. but I think we are unable to prevent this. In the future it might be nice to implement some "report" option on the profile.

Felienne:
I think this can be removed now that we have the stars in the program list:

image

Ow for a new account, the dropdown is no longer there but the validation remains:

image

Otherwise all is good to release this!

Felienne:
another little thing (feel free to leave this for now and make issue)

usernames are case sensitive:

image

image

And the count of programs seems off, I have 1 program saved locally but this says 3. These are not blockers, so I set this to automerge but please make issues because I am too lazy for that now :)

TiBiBa:
> I think this can be removed now that we have the stars in the program list:
> 
> image
> 
> Ow if you do not have shared programs, the dropdown is no longer there but the validation remains:
> 
> image
> 
> Otherwise all is good to release this!

Weird! I think it would be nice to keep the option in both places, but maybe add a message "you can also choose a favorite program in your program page". Will take a look at the bug, that's weird because it should be an optional field.

TiBiBa:
> another little thing (feel free to leave this for now and make issue)
> 
> usernames are case sensitive:
> 
> image
> 
> image
> 
> And the count of programs seems off, I have 1 program saved locally but this says 3. These are not blockers, so I set this to automerge but please make issues because I am too lazy for that now :)

Good catch! Will add an `lowercase()` to the `@app.route` to make sure we catch these correctly as well. For the saved programs: we use the total amount of programs saved by the user, as stored in the achievements db. I think we also have a current programs saved count stored somewhere (on the users db?) we can use that one as well. I don't know which one we prefer, both should be easy to implement.

TiBiBa:
All should be fixed and ready to merge!

templates/profile.html
```diff
@@ -23,7 +72,7 @@ <h2 class="mb-4 pb-4 border-b-2 inline-block w-3/4">{{auth.student_classes}}</h2
    <h1 class="bg-blue-200 px-8 py-2 mb-4 rounded-lg text-3xl cursor-pointer" onclick="hedyApp.load_profile ('{{user_data['username']}}', '{{user_data['email']}}', '{{user_data['birth_year']}}', '{{user_data['gender']}}', '{{user_data['country']}}')">{{auth.profile}}</h1>
```
boryanagoncharenko, line 72: Note that the {{auth.profile}} is used here but it is not defined in the translations. As a result I see 'Unknown Exception' on the /my-profile page:
Screenshot 2022-01-24 at 12 11 30

templates/public-page.html
```diff
@@ -0,0 +1,88 @@
...
+                <img src="{{static('/images/profile_images/' + user_info.image + '.png')}}" class="w-64">
+            {% else %}
+                <img src="{{static('/images/Hedy-logo.png')}}" class="w-64">
+            {% endif %}
+        </div>
+        <div class="w-full h-full bg-blue-200 flex flex-col rounded-lg border border-black">
+            <div class="flex flex-row items-center h-12 bg-blue-600 rounded-t-lg px-4 font-medium">
+                <div class="text-white tracking-wide text-4xl">{{ user_info.username }}</div>
+                {% if user_achievements['achieved'] %}
+                    <div class="text-white text-xl ml-auto">{{ user_achievements['achieved']|length }} {{ auth.achievements }}</div>
```
boryanagoncharenko, line 16: Maybe we should be using auth.my_achievements here and not auth.achievements?
Screenshot 2022-01-24 at 12 20 31

boryanagoncharenko (dismissed):
I probably did not test all possible scenarios, but everything I tried worked great. I am happy with my public profile (even though it runs only locally :))

website/database.py
```diff
@@ -100,6 +101,10 @@ def programs_for_user(self, username):
         """
         return PROGRAMS.get_many({'username': username}, reverse=True)
 
+    def public_programs_for_user(self, username):
+        programs = PROGRAMS.get_many({'username': username}, reverse=True)
+        return [x for x in [i if i.get('public') == 1 else None for i in programs] if x is not None]
```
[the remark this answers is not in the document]
TiBiBa: Agree, fixed this one!

app.py
```diff
@@ -1455,10 +1461,17 @@ def share_unshare_program(user):
     if not isinstance(body.get('public'), bool):
         return 'public must be a boolean', 400
 
+    print(body)
```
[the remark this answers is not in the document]
TiBiBa: Good catch, removed it!

templates/profile.html
TiBiBa: I think this is a merge issue as I solve this issue earlier. But will fix it again!

templates/public-page.html
TiBiBa: I think also a merge error as this should be `auth.my_achievements` already, but fixed!

TiBiBa:
**Before merging:** Corresponding DB table has to be created on Alpha!

Felienne (approved):

PR #1656, opened 2022-01-11
[FEATURE] Implemented explore page

### LINKED ISSUE #1575: [UI idea] Make public overview of all shared programs
Also related to #1573, make a public version here kids can browse all shared programs, like Scratch has:

image

Guess we have to think about if we also want thumbnails or just links (guess just links is totally fine for now)

TiBiBa:
**Description**
In this PR we implement an explore page. A way for users, old and new a-like to find interesting programs! They can filter on level and/or adventure within all shared programs to find interesting one to try, copy and alter. We implementation a new button in the menubar called `explore`. All public shared programs are automatically shown.

**We should merge #1629 first as we need this one for the new limit() function for db calls**

**Fix for**
This PR fixes #1575.

**How to test**
Notice the new option on the menubar. Navigate to this `explore` and try filtering on level, adventure or not at all. Select any program to open it in the view page, enabling you to run it!


Felienne:
Awesome work again @TiBiBa! Love the title on the programs!! But, could we make the top right part a slightly different color (a bit lighter or darker blue) now it looks a bit like someone scratched my programs :) 

image

Felienne (changes requested):
Just a few cosmetic changes to the title squares and we are good to go.

TiBiBa:
> 

This was harder than I thought and the CSS solution is pretty "houtje-touwtje", but it works:

Schermafbeelding 2022-01-14 om 15 11 59


Not sure what we could do about the duplicate level naming as the level in the header is within the program name. As students are also able to store adventures by altering the name (and removing the level) we can't simply say: "Remove level when an adventure program is shown".

Felienne (approved):

PR #1657, opened 2022-01-11
Fix #1655

### LINKED ISSUE #1655: [BUG] wrong "go back to main page" redirect at quiz end
**Describe the bug**
At the end of a quiz, the "go back to main page" button redirects inside the iframe and creates a nested page view.
When the quiz is empty (in the French version) the user directly lands on this page.

URL: https://www.hedycode.com/hedy#end

Screenshot 2022-01-11 at 09 21 44
Screenshot 2022-01-11 at 09 21 49

**Expected behavior**
The main page should be reloaded instead of just the quiz subpage.

TiBiBa:
**Description**
Fixed an issue where the 404 page within the quiz `iframe` would result in a redirect to the main page within the iframe. We added an additional argument to the rendering of the 404 page which can be set to through to make sure the template knows to render the `parent` window to the homepage instead of itself.

**Fix for**
This PR fixed #1655.

**How to test**
Try to start a quiz on a non-supported language. When the error button appears: press it and notice that the main page reload correctly instead of the iframe itself.


Felienne:
This is surely a fix, better than the iframe, but indeed (as per your other issue) if would be better to go back to /hedy than to the main page, preferably even to the level the kid was in.

Felienne (approved):

PR #1661, opened 2022-01-12
[BUG FIX] Store used commands correctly (#1658)

### LINKED ISSUE #1658: [BUG] Commands not tracked in Achievements db table
**Describe the bug**
For an unknown reason the uses Hedy commands are not correctly stored in the db. We need these values to determine the "Gotta Use Them All" achievement, which can be received after using all unique commands. 

**Expected behavior**
When a logged in user uses a command this is correct added to their entry in the Achievements db table

TiBiBa:
**Description**
In this PR we fix an issue where the used commands of users are not correctly stored in the Achievements db. This is due to comparing the wrong session variable to verify if there are new commands used. This comparison would always result in `False`, never storing new data. We also fix an issue where the "Gotta Use Them All" achievement could be obtained by only using the commands of `hedy.HEDY_MAX_LEVEL` (in this case lvl 18), however it should include the commands available in all levels together.

**Fix for**
This PR fixes #1658

**How to test**
Locally run several programs and notice the values being stored in the `Achievements` database in `dev_database.json`. For the fanatic testers: use alle possible commands from level 1 till level 18 and notice that you achieve the "Gotta Use Them All" achievement. 

**Checklist**


Felienne:
@TiBiBa with the sheer volume of PR's can we give them a bit more descriptive names? Is not only nicer now but also later when we do a git blame for example it will be more clear what change a commit belongs to.

Felienne:
Not all commands are yet counted correctly but that is my bad, I am on it (will finalize tomorrow)

Felienne:
ok this was a lot of testing (that did reveal a few bugs!) but it works!!!!! Thanks for bringing out the completionist is me to try it haha.

image

Felienne (approved):

PR #1662, opened 2022-01-12
[CLEAN-UP] Reduced templates by generalizing error page

### LINKED ISSUE #1660: [CLEAN-UP][UI idea] Improve UI of 403, 404 and 500 pages
**Idea incl level**
The 404 page can use an UI upgrade to better align with the rest of our website as well as being more responsive to the message given. For example, doe we always want to return the user to the main page? Maybe we want to return the user to the previous page in some cases? We also currently use three different templates for the 403, 404 and 500 error pages. However: these are identical! The dependencies on the variables can be easily parsed on the back-end and we should merge them into one template.

TiBiBa:
**Description**
In this PR we reduce the amount of templates by creating a general `error` template. Currently we use three different templates (and functions) for the 403, 404 and 500 errors. Here we simplify the process to one function with one template, receiving the error code as parameter. If the error code doesn't exist we return the page with a 404 error.
We also implement the 403 template within the `requires_login`. Resulting in a nice error message with UI when a user tries to access something he/she is not allowed without loggin in instead of the current blank `unauthorized` page.

**Fix for**
This PR fixes #1660.

**How to test**
As this is a clean-up task there are no noticeable difference on the user-side. The error template itself is slightly improved in UI but otherwise nothing changed. Can all be tested by forcing different error codes such as 403, 404 and 500. The easiest way is to:

- Visit a page that doesn't exist (403)
- Try to access profile without being logged in (404)

**Checklist**


Felienne:
@TiBiBa oops this has conflicts now, can you take a look?

TiBiBa:
> @TiBiBa oops this has conflicts now, can you take a look?

GitHub returns a "These conflicts are too complex to solve in the web editor" so I will have to take a look if I can somehow some these conflicts within PyCharm (never done that before).

PR #1673, opened 2022-01-13
Adds types to level 7,8,9 in SEMANTICS description

juliabolt:
**Description**

Adds types for the commands in level 7,8,9. The only new command is repeat, the previous commands and their types did not change. For repeat, I used integer + 'times' as type.

**How to test**

Read SEMANTICS.md file, check level 7/8/9 under "commands and types".


SEMANTICS.md
```diff
@@ -1100,21 +1100,69 @@ Level 6 supports:
...
+    <td>if</td>
+    <td>boolean</td>
+  </tr>
+  <tr>
+    <td>if else</td>
+    <td>boolean</td>
+  </tr>
+  <tr>
+    <td>repeat</td>
+    <td>integer + 'times'</td>
```
Felienne, line 1161: I am not sure what this means, what is `integer + 'times'` as type?

Felienne (changes requested):

SEMANTICS.md
juliabolt: first an integer and then the word times, that's also how I put 'left' and 'right' in possible types used with the command turn in level 1 for example.
(repeat n times) with n an integer is how this command is used, right? 
I think repeat and then an integer does not work? we need the word times behind it?
we could also put times in the command column, but it should be somewhere, not sure which place is the most logical.

Felienne (approved):

PR #1681, opened 2022-01-13
Support for parametrized multilanguage testing

### LINKED ISSUE #1589: [Tests idea] Make Multilanguage tester
**Language idea incl level**
Right now the tests for the Hedy translator are hardcoded in each language. So, for example, if we want to test a feature there we'd call the test something like: `test_is_spanish_english` and it would include Spanish keywords.

I see 2 problems with this approach: only people who can speak the language can write the tests or understand it at first glance (I can understand Dutch tests, but I have to go to the grammar and see each definition),  and also, more importantly, we can only test one language at a time. 

Maybe we can do something like the Multilevel test, but for languages? We'd build each test so it take a generic placeholder like `lang['nl']['is]` and iterate through a language dictionary that contains all keywords for each language. Or perhaps this is not needed? Since the keywords for each language shouldn't affect the grammar and a test in a single language is all that is needed.

I'd love to read your input!

racertop:
**Description**

This PR adds a way to use parametrized tests for the Hedy translator. For this I added:

- A dictionary with the languages we support.
- A function to feed the keywords and languages to the tester.

**Fix for**

Fixes #1589 

**How to test**

**Checklist**


tests/test_translation_level_02.py
```diff
@@ -11,6 +13,7 @@
 
 class TestsTranslationLevel2(HedyTester):
     level = 2
+    all_kwords = hedy_translation.all_keywords_to_dict()
```
Felienne, line 16: Thanks for this work @racertop it is a great way to test more extensively!!

I would just like to ask, can we make this var name `all_keywords`? It is just 2 more letters and it is a lot more clear! In general we try not to use abbreviations unless they really save letters and are wellknown (like AST for abstract syntax tree)

Felienne (changes requested):

tests/test_translation_level_02.py
racertop: Ok! Do you want me to change it in the places where I use `print_kword` or `if_kword` too?

Felienne: Yeah that would be nicer!

racertop:
The test is failing because the translation Dutch translation differs in a keyword with the Spanish and English version. Particulary the command `add` in English and Spanish differs from `add_list` in Dutch. I assumed those dictionaries would all have the same key.

Felienne:
Ow good catch! I updated that recently in English but I forgot to update
the others. Can you make it all the same as en?

racertop:
> Ow good catch! I updated that recently in English but I forgot to update the others. Can you make it all the same as en?

Done! I changed all occurences of kword to keyword as well. If there's anything else to change, I will fix that! 

Thank you for your time!

Felienne (approved):

Issue #1679, opened 2022-01-13
[UI idea] Support right-to-left languages on front-end


TiBiBa:
**Idea incl level**
This might be a challenging project, but the goal is to support a right-to-left view for languages that write/read from right-to-left. As mentioned by @Felienne there is a tailwind package that might be able to help a lot: https://www.npmjs.com/package/tailwindcss-rtl.

PR #1683, opened 2022-01-14
Updated to tailwind v3

### LINKED ISSUE #1480: [UI idea] Update to Tailwind 3
**Idea incl level**
Currently we are using Tailwind 1.9.6. Increasingly I find that classes described in the tailwind documentation are not supported by this version. Creating the issue that adding these classes and generating a new .css file does not resulting in the class actually being added (as it doesn't exist for the current version). It would be nice if we update to the newest Tailwind version to prevent this from happening. As we have a lot of dependencies on tailwind classes (our complete front-end is built upon the styling) we might break stuff when updating and do so carefully.

TiBiBa:
**Description**
For are our front-end styling we use the Tailwind css library. A great way to quickly style the front-end and automatically generate the corresponding css files. Currently we are using version `1.9.6`. However, this version is outdated and currently version `3.0.13` is the most up-to-date. Sticking to the old version has some advantages and disadvantages:

**advantages**
- We don't break anything -> as it already works
- We use the old color scheme, colors that are not supported in versions v2 and v3
- Programmers (especially me) are already familiar with all the options of v1

**disadvantages**
- Lack off online support and documentation
- Can't use some specific new styling features
- Not supporting styling of LTR and RTL, as needed for #1679
- We have to switch someday, better today than tomorrow

In this PR we switch from our current Tailwind version to the newest one. Several configuration options are made on color schemes, line height and font-weight to make sure the new styling looks (almost) identical too the old version. Some extensive testing still has to be performed to make sure nothing is broken. For more information on the new Tailwind version take a look at: https://tailwindcss.com/blog/tailwindcss-v3

**Important sidenotes**
Some important notes that are useful for documentation and possibly future updating. Tailwind looks in `tailwind.config.js` for configurations that are different from the standard settings and overwrites these with the values in the file. In our case we overwrite the v3 color scheme with the v1 color scheme to make sure that the Hedy website stays the same color. The file `tailwind.full.js` doesn't do anything! It is a generated file of all configurations that Tailwind standard uses. It is only useful for look-up as Tailwind only looks for a file named `tailwind.config.js` for possible overwrite values. **Don't change tailwind.full.js as it doesn't do anything**.

**Fix for**
This PR fixes #1480

**How to test**
Ideally this should only be a clean-up task, but it is impossible to keep the front-end identical due to all the large changes. To test: run the branch locally and notice as little change as possible to the front-end styling.


Felienne:
@rix0rrr could you give this your blessing? I am not sure enough about how we froze the tailwind colors and how this ties into the build pipeline

rix0rrr (approved):
LGTM. I don't understand why `appbundle.js` is changed in here. It looks like nothing has changed in the TypeScript, so the `appbundle` should't have been affected either

rix0rrr:
@TiBiBa, one more detail on this:

* `generated.css` that is checked into the repository is generated from the FULL tailwind build

The reason for this is that developers, when working on the HTML, can use any Tailwind class to their heart's content. The downside of this is that the file is a couple of MB big, which is obviously not great for serving as website content.

At the moment we deploy to Heroku, we do a build of Tailwind where it will remove all unused classes from the `generated.css` (by checking out all the files in the `templates` directory, and seeing which classes are used there).

> We *used* to have the stripped build in the repository as well, but that meant every time someone wanted to use a new class, they would have to run the build script again to make sure the new class they were using was actually defined, and it turned out to be a bit of a hassle and we switched to doing it this way.

You might have known this already, but wanted to make sure to highlight it again (because the file just got screwed up on main).

rix0rrr:
I will add a Tailwind build script to regenerate the full development version of generated.css.

rix0rrr:
See #1698 and the new script that I added, it  would need to be updated as well to add the new `-i` flag for Tailwind 3.

Issue #1668, opened 2022-01-12
Add percentage of errors to stats


Felienne:
Now that we have stats, it is kind of hard to see what level is doing better/worse since they all have different bars:

![image](https://user-images.githubusercontent.com/1003685/149171648-e8d375ee-1d36-4cc4-9e16-0d78ebfd6b44.png)

Could we add a percentage in or above the bars, I think in the error ones makes the most sense?

boryanagoncharenko:
Do you think want to see the absolute number of program runs per level and the error rate? For example, in the following image the left graph shows program runs (so that you see if a level got more program runs) and the right - the error rate per level.

![Screenshot 2022-01-13 at 9 25 15](https://user-images.githubusercontent.com/3010723/149284803-0f187dc9-7097-4c89-8c4c-70a7627d2aa2.png)

Or maybe you only need the graph with the error rate?

Felienne:
I like the total numbers too, so this side by side would be great! I would also be happy with a textual representation by the way it does not have to be a graph, f.e. just 80% on top of/in the bar, like this:

![IMG_6268](https://user-images.githubusercontent.com/1003685/149300697-896f854e-1952-48f9-8914-fe499c7383f7.JPG)

boryanagoncharenko:
As easy as it may seem to add the percentage values as annotations, I am afraid it is not easy to achieve. The closes thing that I could try doing is merge the 2 charts, so that there are 2 y-axes. The program runs bars will use the left y-axis and the error rate lines will use the right y-axis. Please note that I am not sure this is supported by the chart component because the bars are stacked while the error line should not be stacked. I will investigate.

Could there be a different way altogether to represent this data, so that it is convenient for you?

boryanagoncharenko:
Just found that there is a plugin that supports adding data labels, so I will investigate briefly this option (timeboxed!)

PR #1690, opened 2022-01-14
Add error percentage to admin /stats page #1668

boryanagoncharenko:
Change per level charts of admin /stats page to hold absolute program runs and error rates per level #1668

**Description**

- Consolidate the successful and failed program runs per level into a single chart as stacked bars.
- Add error rate per level in a separate line chart.

**Fix for**

#1668 

**How to test**

- Run hedy locally, make sure you are logged as an admin, and generate some program runs.
- Open the /stats page and see inspect the change in the first couple of charts. Do they present the data better?


Felienne (approved):
Yes, a lot more clear like this!!

PR #1693, opened 2022-01-15
Extra test for meta column

dolfein:
An extra test for the functionality of `meta.column` from pull request #1494.
The error here is in the middle of a sentence, `nice'` misses a starting quote.


Felienne (approved):
Ah great, here the error does seem to be in the right place!

Issue #1631, opened 2022-01-05
[LANGUAGE] Extend type system to take into account variable reassignments of different types


boryanagoncharenko:
**Describe the bug**
The type validation of Hedy fails when a variable is reassigned with a value of a different type. All variables are stored as entries in a lookup table and are accessed during type checking and transpilation. When a variable is reassigned, two entries are created in the lookup table. However, when validating the type of the variable, the type checker only gets [the first entry match from the table](https://github.com/Felienne/hedy/blob/main/hedy.py#L673). This means that during type validation the variable reassignment will be ignored, which could lead to incorrect errors provided by Hedy.

**Paste the Hedy code & level**
For example, consider the following level 12 snippet:
```
a = 'test'
a = 25
print a * a
```

**Add a screenshot (optional)**
![Screenshot 2022-01-06 at 1 18 53](https://user-images.githubusercontent.com/3010723/148303518-7159f9cb-f8b2-4968-a1df-8ce73839f5bc.png)

**Expected behavior**
Hedy incorrectly returns an error that multiplication cannot be performed with strings. Instead, the code should run successfully.

**What machine and browser you were using (optional)**
If the issue concerns things in the website, let us know:
- What computer you are using (Windows, Mac, Linux?)
- What browser you were using (Chrome, Edge, Safari)

racertop:
Hi! Is anyone planning to take this one? Can I assing it to myself?

boryanagoncharenko:
Hey, please go ahead! The only thing to keep in mind is that this bug requires program slicing (i.e. knowing what part of the code belongs to what line/program block). @Felienne wanted to work on slicing, so you should sync on that.

@Felienne do you have any plans for slicing in the near future? If not, is it possible to work on this bug/missing feature?

Felienne:
Feel free to take it up @racertop! I indeed want to work on slicing bu I have not found the time yet, so feel free to go for it and see where you can get. Maybe I can then steal some work for the slicing plans.

racertop:
Thank you @boryanagoncharenko and @Felienne! I'll work on this and let you know where I get :)

arjank:
According to PR #1694 this issue has been addressed but not fixed. It seems to me that the issue described in the ticket has been fixed (code runs as expected) and other calculations are also performed correctly. If there's still work that needs to be done, we should describe that, and maybe even offload it to a new ticket.

jpelay:
[...]

PR #1694, opened 2022-01-16
Making the Hedy type system a little more dynamic

racertop:
**Description**
Hedy right now fails when a variable is reassigned a different type. This is not a comprehensive fix, because that would be kinda difficult, instead it just fixes the case when the variable is reassigned a new type sequentially, if there's a loop or a branch, it will still fail.

For example this case is now valid, as `a` now has the type integer.
```
a = 'test'
a = 25
print a * a
```

**Fix for**

This issue addresses #1631 but doesn't fully fix it

**How to test**

I added a couple of tests, but you can try changing the type of a variable in any level.


Felienne (approved):
Nice work @racertop!

PR #1696, opened 2022-01-16
Fixing input doesn't work with equals

### LINKED ISSUE #1689: [BUG] input does not work with =
**Describe the bug**
Input fails to parse when used with `=` (does work with `is`)

```
naam = input('Hedy')
print('ik heet', naam)
```

image

**Expected behavior**

### LINKED ISSUE #1695: [BUG] Incomplete inputs are not handled
**Describe the bug**
In level 18, incomplete inputs are not detected and are instead treated as assignments (it also yields an error but not the correct one). 
 Also, on a related bug, a program that contains input without text inside yields a parsing error:

Level 18
```
x is input
```
```
x is input()
```

image

image

**Expected behavior**
It should raise a `IncompleteCommandException`

racertop:
**Description**

Similarly to previous PRs I split the input rule into 2 rules: one for is and one for equals, so they can be handled properly in the translator.

Also I addressed #1695 by editing the input rule so inputs with only brackets and no text inside, and without arguments yield a `IncompleteCommandException`. On a minor fix I removed unnecessary return statements from functions in my previous PRs

**Fix for**

Fixes #1695  and fixes #1689 

**How to test**

I added a few tests in `tests/test_level_18.py` to test the incomplete exception and also the input using equals.


tests/test_level_18.py
```diff
@@ -229,4 +231,16 @@ def test_var_undefined_error_message(self):
 
       # deze extra check functie kan nu niet mee omdat die altijd op result werkt
       # evt toch splitsen in 2 (pos en neg?)
-      # self.assertEqual('name', context.exception.arguments['name'])
\ No newline at end of file
+      # self.assertEqual('name', context.exception.arguments['name'])
+    
+    def test_input_without_argument(self):
```
Felienne, line 236: So this one I am totally aboard with!

tests/test_level_18.py
```diff
@@ -229,4 +231,16 @@ def test_var_undefined_error_message(self):
...
+      self.multi_level_tester(
+        code="name is input",
+        exception=hedy.exceptions.IncompleteCommandException
+      )
+
+    def test_input_without_text_inside(self):
```
Felienne, line 242: But here I feel we should allow this, because Python allows it too. It is not really pretty and kind of confusing but yeah we are getting close to Python now so I am inclined to allow this. What do you think?

racertop: Yes! I had a lot of doubts with that one and actually wanted to ask you. It's a bit tricky since there is not a direct way of differenciate between an input with no brackets and one with brackets and nothing inside, maybe add a new error rule for the ,`x is input` in any case I'll solve it and update this :)

Also, the same thing happens with print(), gives a parse error, I assume you also want that to be valid maybe?

Felienne: Yeah I understand, it *is* a tough call! 

But indeed if we want to fix it, it would require two rules. Do you want to pick it up further in this PR? And yes yes, then do it for print also in one go, it is the same issue.

racertop: Yes! I want to pick it up further in this PR, give me one day or maybe two :)

Felienne: Sure no rush!! Super grateful for all your hard work!

racertop:
Making a draft until I can work on it again.

racertop:
> Thanks! Let me know if you need more help or support!
Thank you so much!!! I think it's done now! Let me know about any further changes!

Felienne (approved):

PR #1700, opened 2022-01-16
Correction and additions in SEMANTICS.md

juliabolt:
**Description**

For level 1 to 9, adds empty as type used with commands forward and turn, also for echo in level 1. Also deletes the possibility of an incomplete exception for the echo command in level 1.

**How to test**

See SEMANTICS.md file. Is it correct that all through level 1 to 9, turn and forward can be used without an integer behind it? For echo, we also discussed this through email :) the lonely echo exception is still there!


Felienne (approved):
Making good progress @juliabolt!

PR #1701, opened 2022-01-17
[FIX] limits shown programs on explore page

TiBiBa:
**Description**
Currently we return all public programs to the user when visiting the `explore` page. This is undesirable as this amount will increase resulting in a very slow webpage. In this PR we implement a quick fix to limit this amount to `50` programs. This is only a temporary fix, the goal is to implement `Public` as a partition key on the db to enable us to use the new `filter()` function to retrieve the desired amount of programs.

**Fix for**
This has no related issue, just a (very) quick fix to prevent webpages from crashing

**How to test**
Navigate to the explore page and notice that the amount of shown programs is limited to 50.

### DIFF
--- a/website/database.py
+++ b/website/database.py
@@ -220,7 +220,7 @@ class Database:
                 if 'adventure_name' in program and program['adventure_name'] == adventure:
                     adventure_programs.append(program)
             result = adventure_programs
-        return result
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Thanks for the quick fix!

PR #1703, opened 2022-01-17
Implements support for RTL languages

### LINKED ISSUE #1679: [UI idea] Support right-to-left languages on front-end

TiBiBa:
**Description**
Currently we only support a LTR (left-to-right) layout. However, multiple languages such as Arabic and Hebrew are using a  RTL (right-to-left) structure. In this PR we implement a dynamic front-end setting to support both layouts. We set a global value to either `ltr` or `rtl` on load, based on the selected language. Based on this value we set the `dir` attribute on the body of `layout.html` to the correct value. This value is stored in a `const dir` within `app.ts` to enable us to, for example, set the ace editor direction to the right one.

With the merging of #1683 we are using Tailwind v3, with built-in LTR and RTL support. However, their support makes the styling of our front-end a bit complexer. For example, instead of using `pl-6` to create a padding of `6` on the left we would have to use `ltr:pl-6 rtl:pr-6` to support both layouts. As a padding to the left on LTR would mean a padding to the right on RTL. Every time we use either `margin`, `padding` or just `right/left` in our styling we have to define this for both LTR and RTL.This is not an ideal situation and styling will cost a bit more time and debugging. But, this is the only way Tailwind currently support LTR and RTL layouts.

**Fix for**
This PR fixes #1679.

**How to test**
Set the `g.dir` value in `app.py` on line 280 fixed to `rtl` and notice that the complete website changes. Navigate through all pages and verify that all margins and paddings are still correct. As this is only a front-end task everything should still work fine.

**Known bugs**

- [x] ~~Shadow on commands window does not work with RTL (stays on rights side of window)~~
- [x] ~~Re-sizing main editor through developers toggle with RTL does not align lines correctly~~
- [x] ~~Programming experience checkboxes on edit-profile are aligned on the wrong side of the page~~


TiBiBa:
Schermafbeelding 2022-01-17 om 16 01 54

Schermafbeelding 2022-01-17 om 16 02 15

Felienne:
> **Known bugs**
> 
> * Shadow on commands window does not work with RTL (stays on rights side of window)

Do we need an issue for this or has it been fixed? I recall seeing a commit message about shadows?

Felienne:
Wow this is so cool @TiBiBa!!

Some tiny issues I ran into: 

The Lightbulb covers the run button on a mistake:

image

Programmer's mode toggle goes out of bounds:

image

Icons in the read aloud dropdown get clipped a little bit:

image

But of course that is all "small beer" compared to all the work already done!!

TiBiBa:
> Wow this is so cool @TiBiBa!!
> 
> Some tiny issues I ran into:
> 
> The Lightbulb covers the run button on a mistake:
> 
> image
> 
> Programmer's mode toggle goes out of bounds:
> 
> image
> 
> Icons in the read aloud dropdown get clipped a little bit:
> 
> image
> 
> But of course that is all "small beer" compared to all the work already done!!
> 
> I guess the first issue is the only thing we really need to fix before merging (and we can even delay that since we have no translations yet to use the rtl feature)

Will take a look at these! I just figured that I only updated the `templates` to support RTL but not the `styles.css` file, so the generated styles (such as buttons and some forms) will still have to be updated. Will fix these as well.

TiBiBa:
Fixed the error, warning / ok box close button!

Schermafbeelding 2022-01-18 om 22 07 51

TiBiBa:
Fixed developers mode toggle!

Schermafbeelding 2022-01-18 om 22 42 29

TiBiBa:
Fixed light bulb!

Schermafbeelding 2022-01-18 om 23 02 56

TiBiBa:
Fixed editor content resizing bug:

Schermafbeelding 2022-01-19 om 09 35 11

TiBiBa:
Last bug... shadow is fixed on the commands window:

Schermafbeelding 2022-01-19 om 10 20 12

TiBiBa:
The read aloud dropdown seem to be a Google Chrome bug. It works fine in other browsers. I tried to fix it with styling, but that results in this:

Schermafbeelding 2022-01-19 om 16 09 47

Felienne:
> The read aloud dropdown seem to be a Google Chrome bug. It works fine in other browsers. I tried to fix it with styling, but that results in this:
> 
> Schermafbeelding 2022-01-19 om 16 09 47

Ah pity, well, let's leave it as it is, I find this is not a showstopper. Can we figure out how to submit a bug with Chrome?

Felienne:
I think all little things here now are fixed?

Felienne (approved):

PR #1707, opened 2022-01-17
Adds types to level 10 and 11 in SEMANTICS

juliabolt:
**Description**

The new command is `for`. In level 10, this can only be used as `for i in list`, but in level 11 it can also be used as `for i in range j to k`. The other commands and types did not change, but in level 11, forward and turn disappear. 

**How to test**

See SEMANTICS.md file, commands and types under level 10 and 11.
The notation is consistent with the notation used already, but if it is not clear, maybe we should think of another notation.


juliabolt:
It might be clearer to use in the table:

for / for i in list / i:string
     / for i in range j to k / i:string; j,k:integer

instead of all the + signs. and then keep that notation consistent.

but if you think it is clear enough, we can keep it like this.

juliabolt:
(with the / I mean that a new column of the table starts )

Felienne:
> It might be clearer to use in the table:
> 
> for / for i in list / i:string / for i in range j to k / i:string; j,k:integer
> 
> instead of all the + signs. and then keep that notation consistent.
> 
> but if you think it is clear enough, we can keep it like this.

I think it is ok for now!

I am seeing that you based this branch on a previously merged branch/PR, it would be nicer if you start over with a fresh branch of of the current main (to prevent a really long list of commits)

PR #1714, opened 2022-01-17
[FIX] Adds cursor pointer to missing clickable objects

### LINKED ISSUE #1713: [UI idea] Change mouse in pointer where object is clickable
Several clickable objects, headers, buttons around the Hedy website don't show a mouse _pointer_ when they are clickable. It would be nice to implement the pointer as this is useful feedback for the user if they can interact with an element or not. This shouldn't be too hard as there is a `cursor-pointer` class built-in into Tailwind.

**Additional context**

TiBiBa:
**Description**
In this PR we add mouse pointers to clickable objects that didn't have a pointer on hover yet. This way users are receiving more feedback on which elements on the website they can and can't interact with.

**Fix for**
This PR fixes #1713

**How to test**
Navigate the site, all clickable elements should show a pointer on hover.


Felienne:
The PR seems to work, but it would be helpful in reviewing if you could give 1 or 2 examples of things that currently are not clickable and that are made clickable with this PR.

TiBiBa:
> The PR seems to work, but it would be helpful in reviewing if you could give 1 or 2 examples of things that currently are not clickable and that are made clickable with this PR.

My bad! For example the buttons for "save program" and "save and share program" on the code page didn't show a pointer. As well as all sections on the profile page.

Felienne:
> > The PR seems to work, but it would be helpful in reviewing if you could give 1 or 2 examples of things that currently are not clickable and that are made clickable with this PR.
> 
> My bad! For example the buttons for "save program" and "save and share program" on the code page didn't show a pointer. As well as all sections on the profile page.

Got it! I see the difference now!

Felienne (approved):

PR #1715, opened 2022-01-17
[FIX] Change font size dynamically based on username length

### LINKED ISSUE #1712: [BUG] Too long username messes up explore layout
**Describe the bug**
When a username is too long on the explore page it shifts down and in the process pushes the code example down as well. In the case of a program containing more 4 or more lines, they go out-of-bound. This can be fixed by a simply truncate on the username or font-size scaling dependent on username length.

Schermafbeelding 2022-01-17 om 15 47 04

**Expected behavior**
When a username is too long we either truncate the last x characters or reduce the font-size

TiBiBa:
**Description**
In this PR we fix an issue where a too long username would mess up the layout of the explore page. By changing the font-size dynamically based on the username length the problem is solved.

**Fix for**
This PR fixes #1712.

**How to test**
Navigate to the explore page and make sure that a program with a long username is shared. Notice that the font of the longer username is smaller and, when still too long, truncated.


Felienne:
Fanatic tester testing :D 

image

Felienne (approved):

PR #1717, opened 2022-01-17
Implements commands language-switcher

### LINKED ISSUE #1577: [UI idea] Add language switcher to the UI
### Discussed in https://github.com/Felienne/hedy/discussions/1564

image

There is a method already in `hedy_translation` that you can call to translate snippets back and forth:

https://github.com/Felienne/hedy/blob/2ed4621472a47628c866f7abf53aec54e2a1a6a0/hedy_translation.py#L24

So this is just a matter of adding it to the UI.

Flags can be a bit tricky though, since not all languages have a good 1:1 mapping to countries (f.e. do we do the Indian flag for both Hindi and Bengali? Does the Spanish flag also represent the Spanish speaking community we have in South-America)

Alternatively we could do little squares with language names/codes (Nl, En, Es etc.)

TiBiBa:
**Description**
In this PR we implement the functionality to switch between languages within the code blocks. Each code blocks is implemented with an additional dropdown many iff the current user language supports translated keywords. At this moment English, Dutch and Spanish are supported. When using either Dutch or Spanish a dropdown menu is shown on each code block to support switching between English and the user language.

This implementation is created my making a dedicated `selector_container` on each code block. On selecting a language we send the `ace editor id`, `old language` and `target language` to `app.ts` for processing. A POST is sent to the back-end with the code from the editor to retrieve a keyword-translated version. Afterwards the two languages in the dropdown menu are switched (as the other language is now selected).

The starting `keyword_language` and `other_keyword_language` are stored in the window.State. As the syntax highlighter files for non-english languages also support the English keywords we initialize the syntax highlighter for each ace editor on `other_language`.

**Fix for**
This PR fixes #1577.

**How to test**
Notice the new blue button at the right-top of each code block. When selecting another language the keywords are translated accordingly. Verify that the copy button does copy the correct translated code. Also verify that you can switch back and forth between languages within the code editor.

**Checklist**


TiBiBa:
Schermafbeelding 2022-01-17 om 17 35 09 (2)

TiBiBa:
This is actually more complex when thinking about it, as we have to translate all possible keywords found around the website. In this process we also have to think about some user preferences, settings and edge-cases:

- Translate all `<pre>` elements using `translate_keywords`
- Translate all `<code` elements using `translate_keywords`
- How do we translate the example commands?
- What is most efficient? Sending a POST for each element increase server-load significant
- Do we want to allow users to set a preferred keyword language in their profile?
- Is the keyword language always set to profile preference, English or automatic matching?

TiBiBa:
Suggestion to use the following flags (for supported languages):

- English  🇬🇧
- Spanish 🇪🇸
- Dutch    🇳🇱

As they are built-in emoticons in (almost) all devices, this should work everywhere without needed an external library.

TiBiBa:
Proof-of-Work of language switcher without re-loading the whole page:
https://user-images.githubusercontent.com/38155424/149937545-82bad6fc-87d2-4c2d-8f7f-4717bce84663.mov

Felienne:
Waaaa this is so cool @TiBiBa!!! Really a game changer! 

one thing I noticed is that if the code is replaced by new code (f.e. from the palette or from the adventures) the language switcher is not set back to En but is should because demo code is always En at te moment. This is a bit confusing since then I cannot translate directly (I'd have to select En and then Nl again):

image

TiBiBa:
Forgot about that one! Will make a fix that the copy code block has the language as attribute. So if we copy from an adventure in dutch the code editor will also be in dutch. Shouldn't be to hard and will automatically fix this issue!

TiBiBa:
> Waaaa this is so cool @TiBiBa!!! Really a game changer!
> 
> one thing I noticed is that if the code is replaced by new code (f.e. from the palette or from the adventures) the language switcher is not set back to En but is should because demo code is always En at te moment. This is a bit confusing since then I cannot translate directly (I'd have to select En and then Nl again):
> 
> image

This should be fixed with the latest commit!

Felienne:
Ok, aaaaaalmost ready! I found one really weird thing:

https://user-images.githubusercontent.com/1003685/150403523-744a41d7-a6ae-416a-9060-14956eead613.mov

I suspect that the addition of the space is a mistake in the translation code and not in yours (but I did not check, I could be wrong!) but the issue here is of course that we can't parse, so also can't translate code with an error.

So could we throw a nice error if a kids translates broken code? Like: "Oops, deze code heeft nog een foutje, dus kunnen we m niet omzetten naar een andere taal?" You can also of course add the error message after so it is more clear what the issue is.

Alternatively, we can parse the code and only show the switched if the code parses but that is expensive!

Felienne:
ow no, I found another mistake, sorry! The distance algorithm does not properly give a closest command.

Current alpha:
image

This PR:
image

It does work on En in this PR so I guess something goes wrong with the switching?

image

TiBiBa:
> Ok, aaaaaalmost ready! I found one really weird thing:
> 
>  translatee_error_code.mov 
> 
> I suspect that the addition of the space is a mistake in the translation code and not in yours (but I did not check, I could be wrong!) but the issue here is of course that we can't parse, so also can't translate code with an error.
> 
> So could we throw a nice error if a kids translates broken code? Like: "Oops, deze code heeft nog een foutje, dus kunnen we m niet omzetten naar een andere taal?" You can also of course add the error message after so it is more clear what the issue is.
> 
> Alternatively, we can parse the code and only show the switched if the code parses but that is expensive!

I think this issue is with the `ask` translation between English and dutch, but I'm not sure. Not sure how we would implement this as we than have to parse the code each time before translating if I understand correctly. That would be expensive if kids translate a lot.

Felienne:
Ow no, the same space issue occurs in level 7 even when the code is right:

image

TiBiBa:
> ow no, I found another mistake, sorry! The distance algorithm does not properly give a closest command.
> 
> Current alpha: image
> 
> This PR: image
> 
> It does work on En in this PR so I guess something goes wrong with the switching?
> 
> image

Weird! How does the distance algorithm works? Essentially all changes are on the front-end. We keep track of the current code block language as an attribute to the html container, not in any other spot. I'll take a look at the code if I did something weird, but at this point have no clue what could cause this issue.

Felienne:
> I think this issue is with the `ask` translation between English and dutch, but I'm not sure. Not sure how we would 
> implement this as we than have to parse the code each time before translating if I understand correctly. 

Yeah indeed, try to parse and only switch language if the parse succeeds

> That would be expensive if kids translate a lot.

True, but I guess kids don't use it so often, and... this weird space hopping is worse

Felienne:
> > ow no, I found another mistake, sorry! The distance algorithm does not properly give a closest command.
> > Current alpha: image
> > This PR: image
> > It does work on En in this PR so I guess something goes wrong with the switching?
> > image
> 
> Weird! How does the distance algorithm works? Essentially all changes are on the front-end. We keep track of the current code block language as an attribute to the html container, not in any other spot. I'll take a look at the code if I did something weird, but at this point have no clue what could cause this issue.

Ah sorry a hard refresh seems to have solved this for me! Can you double check you don't see it?

image

But the spaces thing is still there :(

Felienne (changes requested):
Setting this to changes required for the space issue

Felienne:
ok this will also be interesting combined with r2l support, but let's first do this and then see what is impacted!

Felienne:
Level 1 works well but it seems level 7 does not yet translate to nl? It *does* work for en->es so something weird is going on!

Felienne:
Hmm it seems there is a way (by clicking a lot) to get n an inconsistent state, because after a while, Spanish also stopped working. Definitely related to trying out the code blocks from the adventures.

TiBiBa:
> Hmm it seems there is a way (by clicking a lot) to get n an inconsistent state, because after a while, Spanish also stopped working. Definitely related to trying out the code blocks from the adventures.

That might have something to do with checking if the editor language is the same as the code block language. If not we toggle() it. But, if for some reason the view and the attribute mess up (by clicking a lot for example) we can get in a stuck state. Will take a look how we can fix this!

TiBiBa:
> Hmm it seems there is a way (by clicking a lot) to get n an inconsistent state, because after a while, Spanish also stopped working. Definitely related to trying out the code blocks from the adventures.

Will work on this today and will hopefully fix this issue!

TiBiBa:
> Level 1 works well but it seems level 7 does not yet translate to nl? It _does_ work for en->es so something weird is going on!

I'm able to re-produce this error but have no clue how this happens. The front-end parsing all seems fine, for some reason the back-end just doesn't return a translated version. I think there is a bug within `hedy_translation.py` but I'm not really familiar with that code. I also just noticed another weird bug:

Schermafbeelding 2022-01-31 om 12 04 16
Schermafbeelding 2022-01-31 om 12 04 20

The first 'ask' on line 2 is translated, but it shouldn't be! This is not a keyword at this point and should simply remain ask. This might also have something to-do with a bug somewhere in the back-end.

I was unable to re-produce the "stuck in a state" bug. However, I do remember having the same issue the last time I was working on the code, so will try to make it happen again and look for a fix.

Felienne:
> > Level 1 works well but it seems level 7 does not yet translate to nl? It _does_ work for en->es so something weird is going on!
> 
> I'm able to re-produce this error but have no clue how this happens. The front-end parsing all seems fine, for some reason the back-end just doesn't return a translated version. I think there is a bug within `hedy_translation.py` but I'm not really familiar with that code. I also just noticed another weird bug:
> 
> Schermafbeelding 2022-01-31 om 12 04 16 Schermafbeelding 2022-01-31 om 12 04 20
> 
> The first 'ask' on line 2 is translated, but it shouldn't be! This is not a keyword at this point and should simply remain ask. This might also have something to-do with a bug somewhere in the back-end.

Ah yeah this surely is a backend problem! Can you make an issue for that and assign it to me?
 
> I was unable to re-produce the "stuck in a state" bug. However, I do remember having the same issue the last time I was working on the code, so will try to make it happen again and look for a fix.

Thanks for looking into it!

TiBiBa:
> > I was unable to re-produce the "stuck in a state" bug. However, I do remember having the same issue the last time I was working on the code, so will try to make it happen again and look for a fix.
> 
> Thanks for looking into it!

This should be fixed! We always updated the view when selecting another language, but if we switch again and the server did not respond (yet) something could go wrong. From now on we only update the view if we get a success from the back-end so this state should no longer occur.

Felienne:
Thanks!! I will look at #1856 so we can finally merge this (and with that really expose the multi language programs to kids!)

Felienne (approved):

Felienne:
ok, I am still making some improvements to the translation engine itself but let's merge this and not deploy so we do not keep running into conflicts

Issue #1718, opened 2022-01-18
[BUG] Repair modal not completely visible 


fRedelaar:
**Describe the bug**
When making a mistake in the code, the Hedy Bulb pops up. Whenever the bulb is clicked, the repair modal appears. The modal however seems to be positioned too high in the browser. This causes the modal to not be fully readable. 

**Add a screenshot**

https://user-images.githubusercontent.com/33595265/149893890-59056b90-4035-4b32-b18d-3e47e2791262.mov

**Expected behavior**
Lowering the modals viewing height.

https://user-images.githubusercontent.com/33595265/149893917-d58a0287-ad2d-4508-a4c1-990bd269f6a5.mov

TiBiBa:
One option might be to scroll automatically to the top when showing the modal using `window.scrollTo(0, 0);`. However, that is not a really user friendly solution. What might be nicer is the less intrusive alert modal solution, for which we use `style="top: 10%; left: 50%; transform: translate(-50%, -50%);`. Changing this to `top: 50%;` should always display the modal in the middle of the user view!

Felienne:
Nice to see you are doing good work even before the thesis starts @fRedelaar!

PR #1719, opened 2022-01-18
Fix for #1718: Lowered viewing height of repair modal

fRedelaar:
**Description**
In #1718 the repair modals viewing height is too high. It's fixed by this PR. See the issue for screenshots.


Felienne (approved):

TiBiBa:
I just noticed that this interferes with all other `modal.alert()` calls. As we use this alert function for all user feedback without interaction it is shown quite often. Through this fix all pop-ups are shown in the middle of the screen, not only the repair modal. I suggest to create a dedicated modal for the program repair as it is not really an "alert" after all. I'm currently working re-writing the alert functionality to support red colored messaged so I can add this to my work. If that's okay with you @fRedelaar ?

PR #1723, opened 2022-01-19
[UI IMPROVEMENT] Reduces width of code editor gutter

### LINKED ISSUE #1123: [BUG] Line counter is wider then necessary when more then 9 lines
**Describe the bug**
The code editor has a line counter with a margin on the left side. However, when having more than 9 lines (so the number 10 and higher) appear, the margin stays the same, shifting the UI of the code editor. 

Schermafbeelding 2021-10-28 om 12 52 43

Schermafbeelding 2021-10-28 om 12 52 50

**Expected behavior**
Make sure that the line count sidebar does not change width dependent on the number. (However, it might be necessary to be wider with 100+ lines but I think this situation will not occur often).

TiBiBa:
**Description**
In this PR we decrease the width of the code editor gutter. This had unnecessary margin on the left and the right, especially with more than 10 lines. We set a fixed value of `32px` in contrast to the original `48px`. The resulting issue is that the gutter no longer resizes. As long as a program is shorter than 1000 lines this is no problem, when the program is longer the line counter will go out-of-bound. I expect this not to be an issue.

**Fix for**
This PR fixes #1123.

**How to test**
Navigate to any code page and verify that the gutter is smaller than before. Also notice that it no longer resizes on different line counts.

**Screenshots**
Schermafbeelding 2022-01-24 om 15 20 43
Old view

Schermafbeelding 2022-01-24 om 15 21 17
New view


Felienne:
I just tried this out locally, but I still seem to have the "hopping" problem over 10 lines :(

TiBiBa:
> I just tried this out locally, but I still seem to have the "hopping" problem over 10 lines :(

Should all work fine now! (Until 999 lines that is...)

Felienne (approved):
Yes works now!!! (hope it will also work when the gutter is left....?)

Felienne:
Hope I got the regeneration right, it seems the gutter now hops again 😭  Can you check @TiBiBa?

TiBiBa:
> Hope I got the regeneration right, it seems the gutter now hops again 😭 Can you check @TiBiBa?

Weird! Will take a look!

TiBiBa:
Just generated all css and js files, it all seems to work fine for me!

Felienne (approved):

PR #1727, opened 2022-01-20
[BUG FIX] Re-run within program execution no longer crashes

### LINKED ISSUE #1724: No or incorrect output [BUG]
Sometimes when my students run a program they don't get any or incorrect output. It is hard to pinpoint when exactly this happens. I think it has to do with pressing "Run Code" while a program is still running. Student do this a lot and I see this issue multiple times per lesson. Running the program keeps on giving incorrect results once this has happened. My workaround is to save the program, refresh the webpage in the browser and try again. This usually works.

One example of this issue arises when I run the code below and press the "Run Code" button multiple times during execution. At one point it will create two figures and raise an error (see screenshot). However the issue is certainly not limited to drawings, it also happens when using the `repeat` and `ask` commands.

```
hoek = 90
repeat 10 times
    turn hoek
    forward 50
forward 100
```

image

TiBiBa:
**Description**
This is probably the most complex bug I solved till today, but it works! In this PR we solve an issue where the code running would return inconsistent results when running a new program while the old one was still executing. The issue occurred due to the following problem:

- When a program starts we increase the `window.State.programsInExecution` by `1`
- If this value is, for some reason, 1 or higher -> we clear the output view
- However, programs are run async -> we can run multiple programs together
- The `window.State.programsInExecution` is way higher than `1`
- We have to wait for alle programs to finish to "free" the output window and see code again
- Therefore, this resulted most of the times in a "programs takes too long to run error"

In this PR we implement a `suspension handler`. When starting a new program we set the `StopExecution` variable to `true`. All other async programs automatically throw an `program_interrupt` error and stop running. This automatically solves this weird bug and has the additional advantage that a user can no longer run multiple programs at the same time, not even on accident.

**Fix for**
This PR fixes #1724.

**How to test**
As described in the discussion of the issue one specific piece of code was always able to re-produce the error. When running the following code at level 7 multiple times without waiting for it to finish the error should no longer occur:

`repeat 50 times 'Hedy is leuk!'`


Felienne:
@rix0rrr could you take a peak at the ts error here? The code does seem to work and @TiBiBa and me can't really figure out what is the matter :(

Felienne:
ok, I know what's wrong (with a little help from my friend haha):

```return Sk.misceval.asyncToPromise( () =>
    Sk.importMainWithBody("<stdin>", false, code, true), {
      "*": () => {
        if (StopExecution) {
          window.State.programsInExecution = 0;
          throw "program_interrupt"
        }
      }
    }
```

asyncToPromise wants one argument (a function that will be made into a promise) but you give 2.

Argument 1: `() => Sk.importMainWithBody("<stdin>", false, code, true)`

Argument 2: 
```{
      "*": () => {
        if (StopExecution) {
          window.State.programsInExecution = 0;
          throw "program_interrupt"
        }
      }
    }
```

Now I do not know what you wanted to put in the async but that is why you are getting this error 🤣

rix0rrr:
The reason for this is that Skulpt is written in pure JavaScript, and doesn't come with TypeScript type definitions. So, I've had to write those type definitions myself. They are here:

https://github.com/Felienne/hedy/blob/4f899077956002c2b126dc17d4cf264309cebab1/static/js/vendor.d.ts#L24

However, such an ignorant fool I was! I didn't know any better than that `Sk.importMainWithBody` took one argument, so I wrote it as such! And now the TypeScript compiler is complaining that, as far as it knows, `importMainWithBody` takes only one argument and not 2 😅.

You probably want to change this to something like:

```ts
    asyncToPromise<A>(fn: () => Suspension, whateverTheseThingsAreCalled?: Record<string, () => void>): Promise<A>;
```

Don't know if that's the most accurate typing, but this is what I've written means:

```ts
 whateverTheseThingsAreCalled?: 
//                           ^-- optional
                           
                                Record<string, >
//                               ^^^-- a mapping of strings to

                                               () => void
//                                              ^^^-- functions that take no arguments and return nothing
```

TiBiBa:
> ```ts
> whateverTheseThingsAreCalled?: Record<string, () => void>):
> ```

Thanks for the explanation! I just made a commit with changes, it should all compile correctly now.

Felienne (approved):

Felienne (approved):

PR #1729, opened 2022-01-20
Sleep add remove content

MarleenGilsing:
Adds content to multiple adventures (story, dishes, restaurant) in multiple levels (mainly 2 and 3) using the sleep, add and remove command in both EN and NL. 

Adds a very easy new adventure called parrot for level 1-5. 

Also adds the sleep, add and remove command to the level default explanations and the try buttons in level 2 and 3.


Felienne (approved):
Lovely new adventure Marleen!

Lovely new adventure Marleen!

Lovely new adventure Marleen!

Lovely new adventure Marleen!

PR #1733, opened 2022-01-20
Adds types to semantics level 12 and 13

juliabolt:
**Description**

In level 12, the =-sign is added as a command, but 'is' is also still possible. (I noticed that in the examples, 'is' is used for comparison and '=' is used for assignment, but I tested it and the other way around it also still works.) Floats are also possible now, so that has been added to certain commands. 

In level 13, and & or become possible commands, it was not in the list of commands first, but I think it belongs there, so I added those.

Apart from that, nothing changed in these levels.

**How to test**

Read SEMANTICS.md file, commands and types under level 12 and 13.


Felienne (approved):

Felienne:
Nice work @juliabolt!

PR #1739, opened 2022-01-21
[BUG FIX] Added BASE_URL to class join link to create correct url

### LINKED ISSUE #1737: [BUG] Teacher's link copied incompletely
**Describe the bug**
Link for teachers only copies second half of the link (f.e. /hedy/l/123456)

TiBiBa:
**Description**
In this PR we add the `os.getenv('BASE_URL')` to the created "join class" link to make sure it works correctly. The current link only works if you are already on the Hedy website, otherwise it results in an error as their is no host name.

**Fix for**
This PR fixes #1737 

**How to test**
Navigate to the `class overview` page and notice that the class url and the copy button of the class link both work correctly when navigating to them from an external site (directly in the url bar for example).


Felienne (approved):
Works!!

Not sure how this has ever worked properly but maybe stuff was changed afterwards that I forgot about

PR #1740, opened 2022-01-21
[UI IMPROVEMENT] Improves admin UI, add filtering and attribute selection

### LINKED ISSUE #1734: [UI IMPROVEMENT] Clean-up the admin page, add filtering and pagination
**Idea incl level**
Currently the admin page supports and overview of all users. However, the table is out-of-bounds and all users are retrieved. As the amount of users on hedycode.com are rapidly rising this is no longer a feasible solution. With the support of pagination built in #1629 we are able to retrieve a limited amount of users from the database. It would be nice to give the admin page an overall re-style, improve the UI and additionally add filtering.

**Old view**
Schermafbeelding 2022-01-21 om 10 09 11

**New view (proposed)**
Schermafbeelding 2022-01-21 om 10 09 53

TiBiBa:
**Description**
In this PR we give the admin page an overall UI update to better align with the rest of the Hedy website styling. We also add dynamical attribute selection and query filtering on `creation` and `last login` date. The admin can now select either of these options and a corresponding start and end date to filter the users. The "empty" query automatically returns the last 500 users, the filtering function is without a limit. Currently we only support filtering on these two date value, but more can be added in the future.

**Fix for**
This PR fixes #1734.

**How to test**
Make sure you are an admin by setting the `os.getenv('ADMIN_USER')` value to your own username. Log in and navigate to `/admin`. Notice the new UI and verify that all the dynamic attribute selection and filtering on date is working.


Felienne:
Fantastic @TiBiBa!! I will make a separate issue but it'd also be really nice if we can filter on email addresses (f.e. all email addresses of "school5" with a simple string matching pattern)

website/database.py
```diff
@@ -187,9 +187,12 @@ def forget_user(self, username):
...
             self.delete_class (Class)
 
-    def all_users(self):
+    def all_users(self, filtering=False):
         """Return all users."""
-        return USERS.scan()
+        #If we have some filtering -> return all possible users, otherwise return last 500
+        if filtering:
+            return USERS.scan()
+        return USERS.scan(limit=500)
```
Felienne, line 195: ok to leave this for now but maybe we want to have an option to really see all users? I know that can become expensive, but maybe behind a tickmark or something if we really need it?

Felienne (approved):

website/database.py
TiBiBa: This can also be achieved by filtering without any restrictions (settings a starting date at 2000 for example). But I agree that it would be nice to have some dedicated option to achieve this. Will take a look to efficiently re-write the filtering, as we also add some additional filtering options in the near future.

Felienne: Yeah I realized I can use that but it is not very clear (but maybe ok, I am the only user)

Issue #1242, opened 2021-11-11
Add statistics to the teacher's UI


boryanagoncharenko:
Add metrics to the teacher's page which illustrate how the class is progressing. The concrete metrics will be defined in https://github.com/Felienne/hedy/discussions/1241 and the metric collection will be implemented in https://github.com/Felienne/hedy/issues/665. The scope of this ticket is to display the stats to the teacher's page.

PR #1762, opened 2022-01-24
Refactor CSS for /stats accordion #1242

boryanagoncharenko:
**Description**
Removed unnecessary CSS code.

**Fix for**

#1242 

**How to test**

- The user should be able to open and close the tabs of the accordion on the /stats page.

**Checklist**


Felienne:
> Looks good! I do think we can use a lot of Tailwind classes instead of native CSS to simplify our CSS structure and keep the `styles.css` file as small as possible. But that might be a nice clean-up task for later.

Do we need to open a concrete issue for this or was it more of an offhand remark?

TiBiBa:
> > Looks good! I do think we can use a lot of Tailwind classes instead of native CSS to simplify our CSS structure and keep the `styles.css` file as small as possible. But that might be a nice clean-up task for later.
> 
> Do we need to open a concrete issue for this or was it more of an offhand remark?

More an offhand remark! It would be nice of we read through `styles.css` once in a while to look if we can minimize some native CSS styling through Tailwind. Will add this to my regular to-do to clean-up some CSS coding once every month.

PR #1767, opened 2022-01-24
[UI IMPROVEMENT] Makes color of modal alert dependent of message

### LINKED ISSUE #1763: [UI idea] Return red alert when bad news, green when good
**Idea incl level**
Currently we use a less intrusive alert as an upgrade of the earlier used modal. However, this alert is always green: as if something is executed correctly. We throw this `alert()` on several occasions: when we remove a student, delete a program or receive an error message. It would be nice if the alert would color red in the case of an error to visually attent the user that something went wrong.

TiBiBa:
**Description**
In this PR we improve the modal alert implementation to support both green and red messages. This way we can return a more visual feedback to the user if the alert is successful or not. Currently all modal alert messages are green, but this might be confusing when an error is returned. We add an additional boolean parameter to the alert to define if a message is an error or not. We also implement a dedicated modal for the repair program. This to better differentiate between the "intrusive" modals and the alert one.

**Fix for**
This PR fixes #1763.

**How to test**
We have to test three things to make sure it works:

- Perform a successful action (for example, save or delete a program) -> should be green
- Perform an error action (try to rename a class to an empty string) -> should be red
- Call the repair modal (create a faulty program and call the repair logo) -> should be the "intrusive" modal


fRedelaar (commented):
Looks good to me! I'm not sure whether I am the right person to approve this PR though.

Felienne:
> Looks good to me! I'm not sure whether I am the right person to approve this PR though.

Sure, if you checked that it works, you can totally review!!

Felienne (approved):

Felienne:
Hmmm something might have gone wrong in a merge, because I am still seeing the blue modals:

image


(Note I did not run this locally myself, I was relying on @fRedelaar's assessment, so maybe it never worked well?)

Felienne:
Or is this because it ia button modal?

PR #1769, opened 2022-01-24
[UI IMPROVEMENT] Improves UI of pre-join page to align with overall UI

### LINKED ISSUE #1766: [UI idea] Improve pre-join class UI
**Idea incl level**
Currently the pre-join page of a class is nothing more than plain-text with a link. It would be nice to improve the UI a bit to better align with the rest of the Hedy website. See the screenshot below for the current situation:

Schermafbeelding 2022-01-24 om 13 22 28

TiBiBa:
**Description**
In this PR we clean-up the pre-join page when joining a class. Currently we have two different pages, a `class-prejoin.html` and a `class-already-joined.html`. They are almost identical and it makes no sense to keep track of two templates. We deleted the `class-alread-joined.html` page and differentiate through an additional boolean value `joined`. The texts are aligned in the middel and instead of a plain-text `<a>` we use buttons.

**Fix for**
This PR fixes #1766.

**How to test**
Navigate to a class and use the join link when both in the class and not in the class. Notice that the same template is rendered but the content is different.

**Screenshots**
Schermafbeelding 2022-01-24 om 15 00 09
Schermafbeelding 2022-01-24 om 14 59 59


Felienne (approved):
Ah that does look nicer @TiBiBa!

Felienne:
If you fix the conflicts, we should be good to go!

Felienne (approved):

PR #1770, opened 2022-01-24
Quiz UI revision

LterHaar:
The UI of the quiz question page is completely revisioned. 

**Description**

The (jinja2) code is written more efficiently. A lot of repitition in the html code is removed. A progress bar is added. And a lot of question information is omitted, because it's visualized in the progress bar. 
Questions with 6 options are possible. Each option has its own symbol, color and own submit button. Incorrect questions have a dashed border. 

**Fix for**

The old UI was not written efficiently and was not space and information efficient. 

**How to test**
UI revision 2 a
UI revision 2 b
UI revision 2 c
UI revision 2 d


TiBiBa:
Hi @LterHaar! It looks awesome! I just fixed the merging conflicts and it all seems to work. I'll do some additional testing tonight to make sure everything else is still working as expected. Other than that this is ready to merge🎉!

TiBiBa:
Sorry, said that a little bit too fast! I'm running into a weird bug where the quiz crashes when I'm at the end. It might have something to do with `quiz_answers` only being initialized on the path quiz question but not at the finished quiz path. As `quiz_question.html` extends `quiz.html` the variable is still needed (or we first have to check if it exists) when we want to load the finished screen. Can you take a look if this is the cause of the error? @LterHaar I'm afraid to change any quiz code to break anything.

LterHaar:
I removed the progressbar from the endquiz page. I'm not sure if we need the progressbar over there?

TiBiBa:
> I removed the progressbar from the endquiz page. I'm not sure if we need the progressbar over there?

Sounds good! As it already shows the score I think it is fine to remove the progress bar. The bug I found yesterday also seems fixed🎉! Will do some testing and will let you know if it's ready to merge.

Felienne:
Ziet er goed uit @LterHaar!

Een klein vraagje nog, er staat hier "Good" en niet "Goed". Is dat localizeerbaar of is dat een plaatje en is dit iedere taal altijd "Good"?

image

Felienne:
Ah never mind ik heb m gevonden (en aangepast) in de yaml!

PR #1775, opened 2022-01-24
Testing Hedy in Arabic Levels 1-5 

alaaeddin-swidan:
Level 1-5 translations

**Description**

Adds the Arabic language to the list of all languages and to the languages menu. Includes the translation levels 1-5 in Arabic (Yaml files affected: level-defaults and keywords). 

**Fix for**

No Fix.

**How to test**

You should be able to see "العربية" as a language option in the drop down menu. When chosen the site should switch to RTL and the first five levels can be tested. 
Issue related: https://github.com/Felienne/hedy/pull/1703

**Checklist**

- [X] Links to an existing issue or discussion (if not, create an issue first)
- [X] Describes changes clear in the format above (present tense, no subject)
- [X] Has a "how to test" section
- [?] Only one thing is done in this pull request (specifically please try to refrain from mixing textual changes to the yamls from code changes)


Felienne:
Wow awesome work @alaaeddin-swidan, thank you so much!! 

I needed to fix a few things in the tests, f.e. the passing of the language, because all other languages have examples with en keywords. It will be interesting to decide what we do once we have the switcher #1717, do we want to for rtl languages?

Also syntax highlighting is not working but I suspect that is because js does not properly see that the text is rtl. Will look into that tomorrow.

Felienne:
Ow also error messages look terrible and have the keywords in reverse:

image

Maybe you can take a break in translating more levels and pick up error messages first?

Felienne:
ok a few programs don't work well yet, can you have a look at these @alaaeddin-swidan? f.e. this one (level 4):

image

Sorry for the bad var name error but is seems that you print without quotes, what is this supposed to be? Code with turn in level 2 have an issue too, you can not use them with Arabic words left and right yet (but you can replace them with numbers for now)

Felienne:
This is the turn one:

image

The others seem to work now!

Felienne:
Hi @TiBiBa!

Look at this, we almost have Arabic working thanks to your RTL support. A few things that you can maybe help out with (in this PR or in another one:)

- The gutter shows on the wrong (left) side of Ace
- Syntax highlighting does not work. I am not sure why, it does not not work at all because in level 4 (see above) strings do become purple, but print in level 1 for example (or other keywords) do not get highlighted. Maybe the regex with Arabic letters does not parse, I am not even sure ts/js properly support this? Or maybe the text is sent left to right to the highlighter and thus the first keyword is not picked up because it does not "come first"? A nice puzzle for you to sink your teeth in :D

Felienne:
Hi @alaaeddin-swidan!

I checked the Arabic texts and the issues in level 1 to 4 occur because in the yaml file the ! after hello world! seems to be in the wrong direction. I do not know how to fix that, but it is not a real issue because it only happens in the tests, once copied into the interface it does work, so we can ignore them for now (I will try to see why the tests don't accept it, but maybe you can also double-check the direction is correct?)

In level 5 there is an issue though, in:

الالوان_الجميلة هو أخضر ، أصفر
لوني_المفضل هو اسأل 'ما لونك المفضل؟'
 اذا لوني_المفضل في الالوان_الجميلة قول 'هذا جميل!' وإلا قول 'يا خسارة!!

But I don't know what the issue is there, maybe you can look into that?

Felienne:
> * The gutter shows on the wrong (left) side of Ace
> * Syntax highlighting does not work. I am not sure why, it does not not work at all because in level 4 (see above) strings do become purple, but print in level 1 for example (or other keywords) do not get highlighted. 

Both these are fixed now!!

Felienne:
Hi @rix0rrr,

The look behind approach seems to fail for level 4 and above, with this error message:

image

I am not sure why though, could you have a peek?

rix0rrr:
It seems to be that we're triggering some kind of bug in Ace. Ace is doing a transformation on the regex we give it, to the regex it finally executes. Whatever we're doing in that regex there is making the transformation spit out something that's not legal.

Top: our regex

Bottom: what Ace spits out

image

I'll see if I can identify what triggers the bug and if we really need it.

rix0rrr:
I removed some unnecessary parentheses and that seems to have fixed it

Felienne:
Ok I think we are good to go for levels 1 to 4. The 2 remaining errors are in the test framework, that does not pick up the explanation marks well, somehow. I tested manually and they do work.

Felienne:
Ah darn, if the tests don't pass, we can't deploy to alpha so I removed them after all (sneaking it into #1838)

PR #1780, opened 2022-01-25
[BUG FIX] Tabs no longer overflow page layout

### LINKED ISSUE #1778: [BUG] Adventure tabs out-of-page with new adventures
**Describe the bug**
Due to the increases amount of adventures our tabs are getting out-of-bound. 

Schermafbeelding 2022-01-25 om 10 13 56

**Expected behavior**
The page should automatically handle too much adventures and fix the layout.

TiBiBa:
**Description**
In this PR we fix an issue where the increasing number of adventures would make the tabs overflow the code page. In this process also making the code editor think they could be wider than the page. We fix this issue by adding an `overflow-x-scroll` attribute to make sure that the overflow is caught by a scrollbar. This is not the nicest solution, but should work for now. We also improve some general styling of the tabs; for example, making sure the adventure title is always centered.

**Fix for**
This PR fixes #1778.

**How to test**
Navigate to any Hedy level and notice the changed UI of the adventure tabs. Navigate to either level 3 or level 5 in Dutch and notice that the tabs no longer overflow the code page.

**Screenshots**
Schermafbeelding 2022-01-25 om 10 50 33


Felienne (approved):
Nice temporary fix @TiBiBa!

Felienne:
Do we want to open an issue to fix this more proper so we don't forget?

TiBiBa:
> Do we want to open an issue to fix this more proper so we don't forget?

I think we first need to discuss how we want to approach this problem. Reducing adventure, unlockable adventure or res-styling the UI. I'm sure we can fix this nicely by re-thinking the UI but if we want to change the approach on adventures it would be nice to have a clear vision for this first before creating an issue.

Felienne:
> > Do we want to open an issue to fix this more proper so we don't forget?
> 
> I think we first need to discuss how we want to approach this problem. Reducing adventure, unlockable adventure or res-styling the UI. I'm sure we can fix this nicely by re-thinking the UI but if we want to change the approach on adventures it would be nice to have a clear vision for this first before creating an issue.

I agree more thinking/discussing is needed but maybe we can do that in the issue (or in a discussion?) I worry that if we don't write it down somewhere we will forget...

PR #1792, opened 2022-01-25
New and adapted german translations

krulls:

**Description**

Adds several translations out of https://www.hedycode.com/translate/en/de

**Fix for**

**How to test**

**Checklist**


krulls:
guess I need some help with the failing unit test. I tested the code that is printed in the test output: `winkel is 90
turn winkel
forward 100` in level 4 (https://www.hedycode.com/hedy/4#story) and it works. so ... not sure while the assert fails there.

could remove the translation for that key to work around the issue but rather like to get it to work. so any hint welcome.

Felienne:
> guess I need some help with the failing unit test. I tested the code that is printed in the test output: `winkel is 90 turn winkel forward 100` in level 4 (https://www.hedycode.com/hedy/4#story) and it works. so ... not sure while the assert fails there.
> 
> could remove the translation for that key to work around the issue but rather like to get it to work. so any hint welcome.

I saw the failing test, I will look into a later. A very brief glance at the tests seemed to indicate the problem was at level 2, not 4? But I will look at it in more depth later this week.

Felienne:
Hi @krulls!

I found the issue.... a trailing space. So sorry that tripped you up. The point is the editor removes it so if you copy-pasted there you would not see the error. It would be better if the tests would also do that, but they simply grab all code from the yaml and run it. Because many editors also delete trailing spaces, this does not often lead to issues but for you it did.

Fixed now and ready to go, thanks so much for helping Hedy grow!

Felienne (approved):

Issue #1773, opened 2022-01-24
Add program language to the stats pages log search results


boryanagoncharenko:
The language of the program is already in Athena and is present in the query response. What's left is to add a column in the front-end.

PR #1818, opened 2022-01-27
Add language to search log results on statistics pages #1773

boryanagoncharenko:
**Description**

Add language to the search log results on both statistics pages.

**Fix for**

#1773 

**How to test**

Unfortunately, for now log search can only be tested on Alpha.
- Log in as admin and go to /admin/stats page. Use the log search form and make sure your query yields results. Note that there is a "Lang" column included in the results table and that its value is presented correctly (e.g. you see values 'en' or 'nl').
- Log in as a teacher and go to /stats/class/$class_id. Double-check the same things about the log search results.


TiBiBa (approved):
Looks good to me! As already stated I'm unable to verify the log search. 

Two small nitpicky issues (non-related to this PR but more to the overall statistics):

- Teacher who are not a beta_teacher are still able to visit statistics by using the direct url to their class. I don't think that's an issue but just to make sure that this is intended.
- The selector on the class_statics is not styled:

Schermafbeelding 2022-01-27 om 15 18 17

boryanagoncharenko:
> Looks good to me! As already stated I'm unable to verify the log search.
> 
> Two small nitpicky issues (non-related to this PR but more to the overall statistics):
> 
> * Teacher who are not a beta_teacher are still able to visit statistics by using the direct url to their class. I don't think that's an issue but just to make sure that this is intended.
> * The selector on the class_statics is not styled:
> 
> Schermafbeelding 2022-01-27 om 15 18 17

Thanks for the review and the comments:
- It is ok if teacher can access their class stats even when they are not beta teachers. The most critical part here is that they do not access the stats of a class that they do not own. Indeed, the beta teachers are only used to show/hide the button.
- I am not able to replicate this on my end and I was wondering whether you ran ```./build-tools/heroku/tailwind/generate-css```?

boryanagoncharenko:
> > * I am not able to replicate this on my end and I was wondering whether you ran `./build-tools/heroku/tailwind/generate-css`?
> 
> I did! It might be a Safari issue, what browser are you using? I did find I work-around for force-styling a `<select>` with Tailwind but forgot where I used it. If I remember it will copy it to this selector, it's not really a high priority I guess.

Oh! I will investigate. It is not high priority but it is very annoying and frankly ugly. Thanks for the report!

PR #1819, opened 2022-01-27
Added emojis to the adventure example codes and added a try button in level 1 for printing emojis

MarleenGilsing:


Felienne (approved):
What a lovely additions @MarleenGilsing

Issue #1776, opened 2022-01-24
[UI idea] is ask / at random combination can be difficult to understand


ericverschoor:
Given the program

    eten is ask wat wil je eten?
    vanavond is eten at random
    print wij eten vanavond 

This is of course a strange program to run. It first asks for input (is ask) and then choose a random value (at random) from the input value.
A random char is chosen from the array of chars.
This is, i think, hard to understand for beginners.

Perhaps a warning can be shown that when variable that is set via ‘is ask’, is used as input for ‘at random’

Felienne:
Great catch @ericverschoor! This should have been caught by the type system! Tagging @boryanagoncharenko so she can have a peek at what went wrong here.

Felienne:
Hi @ericverschoor, this is fixed now, thanks again for pointing it out (goodies are on the way to you!)

PR #1821, opened 2022-01-27
Add input hedy type and forbid any list operations with the input type #1776

boryanagoncharenko:
**Description**

Introduced new type called 'input' meant to encompass all possible types that could be returned by the ask and input commands. For now the 'input' could turn into 'string', 'integer' and 'float'. Note that we cannot determine the exact type of the user input during type checking because the user input is only known when the python code actually runs in the browser.

**Fix for**

#1776 

**How to test**

1) Read the changes in the DESIGN.md. Are they clear?

2) Run Hedy, set the language to English, go to level 3, and run the following program:
```
eten is ask wat wil je eten?
vanavond is eten at random
print wij eten vanavond 
```
Is the error message clear?


Felienne (approved):

Felienne:
Lovely change, thanks @boryanagoncharenko!

@juliabolt maybe you want to take this into account for the semantics too?

PR #1826, opened 2022-01-27
fix quoting in example code in level 5

### LINKED ISSUE #1825: [CONTENT] improperly quoted string "'what's your name?'" in Level 5 code examples
**Describe the issue**
In [level 5](https://www.hedycode.com/hedy/5) in **English**, on the first tab ("Level 5") in sections **"Example Hedy Code"** and **"Tip"** are code blocks beginning with
```python
name is ask 'what's your name?'
```
i.e., the single-quoted string contains a straight apostrophe (`'`), which is what we also use for string-quoting here. Thus the string isn't properly quoted and will probably be terminated already after `what`.

Indeed, when that code is inserted into the code window and run, it doesn't behave like described in the explanation: The program immediately prints `boo!` instead of asking for the name first.

(I'm unsure whether such code should cause a syntax error instead, which it currently doesn't do.)

image

**Expected text**
The regular Python means of escaping using  backslash doesn't seem available at this level, and hasn't been introduced to the reader at this point anyway. Same for double quoted strings (with `"` as delimiters).

Thus the only reasonable option seems to avoid the contraction and write out "is", as other sentences in strings of this level seem to do, too:
```python
name is ask 'what is your name?'
```

das-g:
fixes #1825

**Description**

Fixes Issue #1825 for English by replacing the contraction "what's" by spelled out "what is", so that the quoting character `'` doesn't occur in the string.

**Fix for**

#1825

**How to test**

In both affected sections of [level 5](https://www.hedycode.com/hedy/5), on the first tab ("Level 5"):
- Example Hedy Code
- Tip

respectively, do the following:

1. click the "Try" button (⇥) to copy the example code there into the code area
2. click the green "Run code" button below the code area
3. check whether the program has printed anything to the output area (it shouldn't have)
4. check whether the program asks `what is your name?` in the output area, with an input field for entering the name

**Checklist**

### DIFF
--- a/coursedata/level-defaults/en.yaml
+++ b/coursedata/level-defaults/en.yaml
@@ -167,7 +167,7 @@
 
          ## Example Hedy Code
          ```
-         name is ask 'what's your name?'
+         name is ask 'what is your name?'
          if name is Hedy print 'nice' else print 'boo!'
          ```
 
@@ -187,7 +187,7 @@
          ## Tip
          Sometimes code with an `if` gets really long and does not fit on the line well. <br> You may also divide the code over two lines, starting the second line at the `else` like this:
          ```
-         name is ask 'what's your name?'
+         name is ask 'what is your name?'
          if name is Hedy print 'nice'
          else print 'boo!'
          ```


coursedata/level-defaults/en.yaml
```diff
@@ -167,7 +167,7 @@
 
          ## Example Hedy Code
          ```
-         name is ask 'what's your name?'
```
Felienne, line 170: Ah interesting, this code works but not in the right way (it creates an assignment) 
This fix is great but we should look at the root cause too!

Felienne (approved):

PR #1829, opened 2022-01-27
Log Search from fixes #1773

### LINKED ISSUE #1773: Add program language to the stats pages log search results

boryanagoncharenko:
**Description**

- Fix styling of the select element in log search form in Safari 
- Fix bug with log searching of empty classes

**Fix for**

#1773 

**How to test**

Locally:
- Log in as admin and go to /admin/stats page. Check that the log search form contains an input(text) for the username and that it is styled correctly in Safari
- Log in as a teacher and go to the /stats/class/$class_id page. Check that the log search form contains a select element that lets you select only the students from the current class. Check that the select is correctly styled in Safari.

On alpha:
- Check that the log search form does not yield errors when the class is empty. Ensure no data is returned.

### DIFF
--- a/app.py
+++ b/app.py
@@ -726,11 +726,12 @@ def query_logs():
 
     class_id = body.get('class_id')
     if not is_admin(user):
-        if not class_id:
+        username_filter = body.get('username')
+        if not class_id or not username_filter:
             return 'unauthorized', 403
 
         class_ = DATABASE.get_class(class_id)
-        if not class_ or class_['teacher'] != user['username'] or body.get('username') not in class_.get('students'):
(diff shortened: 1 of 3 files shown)


TiBiBa (approved):
Looks good to me! I noticed you use `cls` as a variable naming for class-related data. Which I really like! In most places we use `Class` instead of `class` because class is a reserved keyword in Python. But I really don't like the capital variable naming, nice solution! Will open an issue to make our variable naming a bit more consistent.

PR #1838, opened 2022-01-28
[UI IMPROVEMENT] Remove code palette, re-structure adventure view

### LINKED ISSUE #1837: [UI idea] Remove code palette on sidebar
**Idea incl level**
As discussed many times (such as is #1835) the interface is getting crowded, too crowded. With the adventures and all added options we have way to many functionality on a small screen. This can be confusing for children as there are too many options too choice from. One solution might be to hide the code palette on the sidebar. Increasing the size of the adventure tabs and code editor / output.

TiBiBa:
**Description**
In this PR we remove the code palette on the left-side of the screen. Creating more space for the adventures and code editor. The adventures are re-structured to have a structured view: combining an explanation with the corresponding example code on the same row.

**Fix for**
This PR fixes #1837.

**How to test**
Navigate to any Hedy level and notice the missing code palette. Also notice the new structure of the adventures.


Felienne:
Great work @TiBiBa! Let me know if you want me to write a script to update the yamls (we have some code for migration already so it will not be all too hard)

Felienne:
ok @TiBiBa, I updated all adventure yamls according to the new structure!

Felienne:
Ow and @TiBiBa, I guess we want the same layout (and thus change in yaml) for level defaults? Can you make the code for that and I will update the contents

TiBiBa:
> Ow and @TiBiBa, I guess we want the same layout (and thus change in yaml) for level defaults? Can you make the code for that and I will update the contents

Yes, will look into that later today!

Felienne:
Note to self that we just added German adventures that do not yet have the example_code field!

Felienne:
> Note to self that we just added German adventures that do not yet have the example_code field!

Fixed!

Felienne:
@TiBiBa I updated the schema and level defaults too (just for en level 1 for now) so you can use that if you update the level defaults view, to see what it will all look like!

PR #1840, opened 2022-01-28
[FEATURE] Enables filtering on programs page

### LINKED ISSUE #1678: [UI idea] Add filtering to personal programs page
**Idea incl level**
With the implementation of #1656 we have an explore page on which user can filter on level and adventure. It would be nice if we implement this same feature in a user's personal program overview. Especially if a user has a large amount of programs this can be useful to quickly find their desired program. As the code for filtering is already written we will be able to copy large parts, making this implementation relatively easy. 

**Additional context**

TiBiBa:
**Description**
In this PR we implement filtering on the programs page. Similarly to how the `/explore` page works we add a search header for filtering on level and adventure. Using a GET these attributes are sent to the back-end and the page is reloaded with the corresponding programs. We also fix an important bug where a teacher viewing a students page was able to set favorite programs for the public profile. (Not really because we perform an additional check on the back-end if the current user is owner of that program, but still).

**Fix for**
This PR fixes #1678.

**How to test**
Navigate to the `programs` page and notice the new header. Try different filtering options as user as well as teacher of a user (when viewing a student's profile). Make sure that everything works as expected.


Felienne:
Hi @TiBiBa,

This works well! However, it might not be a bad idea to also add some e2e test for all these new routes that we have been adding over the past few weeks, to protect them against future changes (maybe make a separate issue of that?)

TiBiBa:
> Hi @TiBiBa,
> 
> This works well! However, it might not be a bad idea to also add some e2e test for all these new routes that we have been adding over the past few weeks, to protect them against future changes (maybe make a separate issue of that?)

Agree! I will make a separate issue to take a look at the e2e tests!

Felienne (approved):

PR #1764, opened 2022-01-24
[FEATURE] Enables invite to class by username

TiBiBa:
**Description**
In this PR we implement a new feature, enabling teachers to invite students by their username. Currently teachers are only able to add students to their classes by copying and sharing the "join link". As discussed in #1735 some school don't have mail accounts or an easy way of sharing this link.

We add a new button to the class overview for "add student". When pressing this button the prompt modal will pop-up and the teacher is able to enter an username. If this is a valid account (and not already in their class) the username will be stored in a dedicated `invitations` db table. When a user logs in this table is checked if there are any pending records for the user. If so: they are shown in their profile.

**Fix for**
This PR fixes #1735

**How to test**
Make sure you are a teacher and navigate to the overview page of any of your classes. Notice the new "add student" button and verify that you can add another (existing!) account that is not already in your class. When loggin in to this account the profile should contain a new message with the link to join the class.

PR #1841, opened 2022-01-28
[FEATURE] Enables teacher to reset student password

### LINKED ISSUE #1736: [UI idea] Reset password for kid by teacher
**Idea**

Some kids forget their password, but they can't access their email address (f.e. because they can't receive links or can't receive email outside of school) If would be great if teachers can set a password reset for this kid so that they can log in again. Once logged in, they can change their password again.

TiBiBa:
**Description**
In this PR we implement a new feature, enabling teachers too reset a students password. As discussed earlier some children don't have access to a mailadres at the school, making it hard to invite them to classes and/or reset their passwords. With the implementation of #1764 the first one is fixed. In this PR we fix the other issue.

We implement an additional column on the class-overview page called `change_password` pressing it will pop-up the modal for the teacher to enter a new password. Due to the table getting crowded we also implement dynamic column hiding/showing as already implemented within the admin page in #1740.

**Fix for**
This PR fixes #1736.

**How to test**
Navigate to the class-overview and notice the new attribute selection too at the top of the page. Make sure the "change password" column is shown and verify that you can indeed change the password of a student.


MarleenGilsing:
I can't open the class page, @TiBiBa. I can go to the teachers page, but the class page gives me an error.

TiBiBa:
> I can't open the class page, @TiBiBa. I can go to the teachers page, but the class page gives me an error.

Weird! It all works fine for me. @Felienne can you verify if the class page is broken?

MarleenGilsing:
It's probably something with my computer @TiBiBa because I can't open the page either when I'm on Main instead of your PR.

Felienne:
> It's probably something with my computer @TiBiBa because I can't open the page either when I'm on Main instead of your PR.

I know this does not help in any way but for me it works fine locally. What is the error @MarleenGilsing?

MarleenGilsing:
I get this: 

image
image

TiBiBa:
> I get this: 
> 
> 
> 
> image
> 
> image
> 
> 

I think I know the cause! For some reason you don't have the BASE_URL set as an environment variable within your IDE. What IDE are you using? There should be an option to add this value (it should be there, but not sure why it's not found).

Felienne:
Ah ja ik zie het al! Jij hebt de BASE_URL niet ingesteld. Ik denk dat ik je dat wel eens heb laten zien in PyCharm kan je ergens je run configuration aanpassen, daar heb je denk ik al ADMIN_USER staat en daar moet je ook een BASE_URL bij doen, die je dan op "localhost:8080" moet instellen. Als dat niet lukt laat ik het donderdag wel zien.

Felienne:
> > I get this:
> > image
> > image
> 
> I think I know the cause! For some reason you don't have the BASE_URL set as an environment variable within your IDE. What IDE are you using? There should be an option to add this value (it should be there, but not sure why it's not found).

It was not needed before, so I never showed Marleen how to set it. 

Of course if would also be nice if we did not fail this hard for other new contributors, maybe 1) add the needed env vars (also ADMIN_USER) to the docs and 2) throw a nicer error?

MarleenGilsing:
Gelukt :)

TiBiBa:
> > > I get this:
> > > image
> > > image
> > 
> > 
> > I think I know the cause! For some reason you don't have the BASE_URL set as an environment variable within your IDE. What IDE are you using? There should be an option to add this value (it should be there, but not sure why it's not found).
> 
> It was not needed before, so I never showed Marleen how to set it.
> 
> Of course if would also be nice if we did not fail this hard for other new contributors, maybe 1) add the needed env vars (also ADMIN_USER) to the docs and 2) throw a nicer error?

Agree. Propose to change it into the following:

`os.getenv('BASE_URL', 'http://0.0.0.0:8080')`

In case it is not set it falls back to the development url

MarleenGilsing:
Now that I'm on the page, I can easily add the 'change password' column. 
Weirdly in my first few tries the button didn't work, but now it works perfectly! 

Small remark: It asks if I'm REALLY sure of this choice, which sounds a bit dramatic, haha, like I can't change it back. Maybe remove 'really' and 'heel'?

coursedata/texts/en.yaml
```diff
@@ -281,6 +281,11 @@ Auth:
     number_programs: "Number of programs"
     amount_programs: "Saved programs"
     latest_shared_program: "Latest shared program"
+    visible_columns: "Visible columns"
+    enter_password: "Enter a new password for"
+    password_change_prompt: "Are you really sure you want to change this password?"
```
Felienne, line 286:
```suggestion
    password_change_prompt: "Are you sure you want to change this password?"
```

coursedata/texts/nl.yaml
```diff
@@ -273,6 +273,11 @@ Auth:
     highest_level_reached: "Hoogste level"
     number_programs: "Aantal programma's"
     latest_shared_program: "Laatst gedeelde"
+    visible_columns: "Zichtbare kolommen"
+    enter_password: "Vul een nieuw wachtwoord in voor"
+    password_change_prompt: "Weet je heel zeker dat je dit wachtwoord wilt wijzigen?"
```
Felienne, line 278:
```suggestion
    password_change_prompt: "Weet je zeker dat je dit wachtwoord wilt wijzigen?"
```

Felienne (approved):

Felienne:
> Now that I'm on the page, I can easily add the 'change password' column. Weirdly in my first few tries the button didn't work, but now it works perfectly!

This will typically be the case since your browser remembers (in computer terms: caches) the code so it can take a while for it to see the now code.

> Small remark: It asks if I'm REALLY sure of this choice, which sounds a bit dramatic, haha, like I can't change it back. Maybe remove 'really' and 'heel'?

Good point, removed!

PR #1842, opened 2022-01-28
[FIX] Re-implements e-mail changing support from admin page

TiBiBa:
**Description**
In this PR we re-implement an accidentally deleted e-mail changing function on the admin page. We also fix a bug where the programs page re-direct from the admin table was incorrect.

**Fix for**
No issue created

**How to test**
Navigate to admin page, verify that the mail column values are clickable, resulting in a pop-up to change the mail address of a user. Verify that this is indeed working.


Felienne (approved):

Felienne (approved):
Nice, good catch on the redirect too! I had noticed it did not go to the correct page but I was not sure why.

PR #1853, opened 2022-01-31
[BUG FIX] Fixes color off error message on student invite error

### LINKED ISSUE #1851: [BUG] Error message on student invite is green (instead of red)
**Describe the bug**
With the implementation of the new `modal.alert()` we differentiate between success messages and error messages. Showing the message in either a green or red color and template. On the student invite, this setting is missing: making the message green.

Schermafbeelding 2022-01-31 om 12 17 29

**Expected behavior**
Fail message on invite is red.

TiBiBa:
**Description**
In this PR we fix an issue where the color of the error message when inviting a student to a class was incorrect.

**Fix for**
This PR fixes #1851

**How to test**
Make sure you are a teacher, try to invite a non-existing user to a class and notice that the modal alert is red instead of the current green.


static/js/teachers.ts
```diff
@@ -133,7 +133,7 @@ export function invite_student(class_id: string) {
       }).done(function() {
           location.reload();
       }).fail(function(err) {
-          return modal.alert(err.responseText);
+          return modal.alert(err.responseText, 3000, true);
```
Felienne, line 136: Is the 3000 here a color? Maybe we can make it into a constant, or add a comment because like this it is hard to know what it means.

TiBiBa: My bad! The 3000 is the timeout in ms. Will add an inline comment to explain this. However, we use the same structure in a lot of places when calling modals. Will take a look if we can improve the readability.

Felienne: Ah so then I don't know why this change changes color? It is because of the true (Note that I did not read the rest of the code, just this snippet!)

TiBiBa: Yes it's a boolean indicating an error or a success message. When true the message is shown in red, otherwise in green.

Felienne: Ah got it!

Felienne: Does ts allow for named parameters?

 `return modal.alert(err.responseText, timeout=3000, error=true);` would be a lot more understandable?

Felienne: Ah it seems it does not really understand, and I do agree it is overkill to document it at all calls, so let's leave it the way it is.

Felienne (approved):

PR #1855, opened 2022-01-31
[IMPROVEMENT] Implements dedicated timeout for alert modal

### LINKED ISSUE #1852: [UI idea] Give modal alert a non-shared timeout
**Idea incl level**
Currently a timeout is shared by all modals. So when the timeout runs out, all modals are automatically hidden. With the implementation of the alert modal users are able to continue navigating the website without closing the modal. This can result in undesired behavior:

- When trying to add/remove a student from a class -> you receive a alert on success / fail
- When performing this action again (in a input modal) the input modal will disappears when the alert modal timeouts

We should improve this by adding dedicated timeouts for the alert modal and all the other ones.

TiBiBa:
**Description**
Currently the modal timeout is shared between all modals. Due to the re-structuring of the `alert()` modal this is no longer a logical implementation as it is possible for two modals to be shown at the same time. (For example an alert() from the previous interaction and a prompt for the current interaction). In this PR we implement a dedicated timeout for the alert modal to make sure that these don't interfere and close each other at undesired moments.

**Fix for**
This PR fixes #1852.

**How to test**
Find a way to have an alert and any other modal shown at the same time. An easy way to fix this is as follows:

- Invite an invalid user to your class
- Directly after seeing the error message invite another use (stay at the input modal)
- Notice that the input modal does not disappear after the alert does


boryanagoncharenko (approved):
Works as described

PR #1857, opened 2022-01-31
in semantics, remove repeat command from level 11 and further

juliabolt:
**Description/Fix for**

in level 10, the `for` command and the `repeat` command both work, but from level 11 on, the `repeat` command does not work anymore, we have to use `for` there. I missed this before, so fixed it now. So in the semantics, repeat is not in the list of commands anymore in level 11 and further (semantics are up to level 15 now).

**How to test**

See SEMANTICS.md, under commands and types, level 11-15


Felienne:
Ah good catch @juliabolt but I am not sure if we should change the semantics and not the language! Maybe it is ok to keep repeat possible for a while! Or maybe at least give a detailed error message? I have to think about it for a bit.

juliabolt:
Okay, let me know when you've decided what's best :)

juliabolt:
@Felienne I think we decided I don't need to fix anything, only make an issue and this PR can be merged?

We decided the following. (correct me if I'm wrong)
The SEMANTICS.md file will be adjusted to how Hedy works now, so repeat is not possible in level 11 and further, and that should be the same in the language for now, to keep things consistent. For this I made an Issue #1942 .
Later on, repeat can be added in level 11 and further in the language and the semantics, if/when we have that working (but that might take a while).

So now, repeat should be removed from level 11 and further, and this PR can be merged (right?)

Felienne (approved):

Felienne:
> The SEMANTICS.md file will be adjusted to how Hedy works now, so repeat is not possible in level 11 and further, and that should be the same in the language for now, to keep things consistent. 

Yes, as a general rule let's keel the semantics file in line with how Hedy works now, potentially that does mean you will have to make some changes later on.

PR #1859, opened 2022-01-31
fix: JavaScript broken on Safari and old Firefox

rix0rrr:
The lookbehind regex we were using to match a word boundary in non-ASCII
languages is only supported in Chrome, not in other browsers.

Switch to an alternative that isn't as correct (because it must consume
the non-word character instead of just assert it), but should do fine
in most cases.


rix0rrr:
Oops,  looks like Safari doesn't even support `\p{L}` at all.

Felienne:
> Oops, looks like Safari doesn't even support `\p{L}` at all.

Hmmm that is unfortunate. Can we make it not break the whole website on Safari with an if or something, and then give a warning to people that Safari + Arabic do not work together well?

rix0rrr:
I drastically simplified this one. Keywords must now have a space before them (if they have a character at all).

Not 100% ideal, but given the circumstances best we can do I feel.

Felienne (approved):

Felienne:
Works in Chrome and in Safari now!

PR #1863, opened 2022-01-31
[BUG FIX] explore page no longer crashes when filtering

### LINKED ISSUE #1862: [BUG] Explore crashes when filtering without being logged in
**Describe the bug**
Currently the /explore page crashes if you try to filter without being logged in. This is due to the achievements being added without verifying if there is a user. 

**Expected behavior**
Anyone can filter on the /explore page, without being logged in.

TiBiBa:
**Description**
In this PR we fix an issue where the page would crash when filtering without being logged in.

**Fix for**
This PR fixes #1862

**How to test**
Make sure you're logged out and verify you can filter on the /explore page


Felienne:
Didn't we allow exploring only when logged in anyway?

image

TiBiBa:
> Didn't we allow exploring only when logged in anyway?
> 
> image

I just noticed this as well! But when looking at the code I have no indication that we restrict /explore to logged in users. So not sure why this view is shown.

Felienne:
Haha ok, but it is the way I wanted it :D

Felienne:
Ah I see, this is true for hedycode but not for main/alpha. So it figures you did not find it in the code, because it is no longer there. But can we put it back, that seems to be the real issue.

ThomasKluiters (approved):

ThomasKluiters (approved):

Issue #1867, opened 2022-01-31
[BUG] Arabic at random does not select randomly


Felienne:
```
صديقي هو احمد، خالد، حسن
قول صديقي بشكل عشوائي
```

Level 3 example code prints something but does not seem to print a random thing. I assume this is because the list is not separated by the commas that we support `,` but with `،`?

@alaaeddin-swidan I tried to change the ، into , but since I have no clue what the code says it was hard to get right. Can you try this?

And... should we change the grammar to allow ، as list separator? Would that be most natural or could you use , too?

Felienne:
Haha, this code does work!

```
صديقي هو احمد, خالد, حسن
قول صديقي at random
```

So maybe there is an issue with the list separator and also with parsing the words `at random` in Arabic?

alaaeddin-swidan:
Good catch @Felienne! 
I tried the codes and i think you are right, there are two things here: 
- the comma: I think we need to support the RTL comma. Putting myself in the shoes of a child writing using a RTL keyboard it will be difficult to switch keyboards back and forth just for the comma -- and of course we need then extra instructions to let them know that they need to do this. 
- the at random: i translated this into two words (separated by a space) in Arabic, basically with at being a word (بشكل) and random being another word (عشوائي). I am not sure how you work things out in the commands parser, i assume there is something to change there. Otherwise i could come with a one-word translation and it might then be resolved. 
-

Felienne:
Ah just checked and this was fixed by #1868!

PR #1868, opened 2022-01-31
Update at and list separator for Arabic

Felienne:
Fixes (?) #1867

I split at random in Arabic in 2, but maybe this split is not idiomatic? Can you check @alaaeddin-swidan? I also updated the separators int he example which is ok for now but maybe ultimately we need to allow your comma too?


alaaeddin-swidan (approved):
I think this is what we needed -- in line with my comments on #1867. I added missing commas to the demo code level 3.

Issue #1870, opened 2022-01-31
[BUG] Contributing with WSL requires a minor workaround


martenvd:
**Describe the bug**
Making contributions to Hedy while working on Windows Subsystem for Linux (WSL), the wrong local address is exposed. Normally the application would be exposed through 0.0.0.0, but WSL requires that to be either the internal IP address of WSL or localhost (127.0.0.1). 

**Expected behavior**
I would hope to be able to contribute to Hedy without having to alter the `app.py` file.

**What machine and browser you were using (optional)**
- What computer you are using 
**Windows**

PR #1871, opened 2022-01-31
Remove hostname

martenvd:
**Description**
This PR adds a check that will look for whether the host system is a Windows Subsystem for Linux system or not. This is important because 0.0.0.0 is not exposed through WSL. Localhost (127.0.0.1) is though, which means that exposing that makes more sense when trying to contribute to Hedy and also having your workflow on WSL. The thing I strive for is that all developers can contribute to Hedy regardless of the host system they use. 

**Fix for**
#1870 

**How to test**

Follow the `CONTRIBUTING.md` file while working on WSL and having python3 installed. Then run the `app.py` using this branch.

**Checklist**


rix0rrr:
There are two different issues at play here:

* What IP address(es) are we listening on?
* What gets printed in the Hedy server output and what happens when you click on that link?

# What IP addresses are we listening on?

```py
app.run(..., host="0.0.0.0")
```

Means that we will listen on *any* IP address. Doesn't matter how you reach our computer, we will take your traffic. All of these would work:

```
http://localhost:8080/
http://127.0.0.1:8080/
http://192.168.1.128:8080/    <-- from another computer, assuming this was your IP address on your local network
http://my-computer.local:8080/  <-- from another or the same computer
```

# What gets printed and what happens when you click the link

The default behavior of Flask is to print the host we put in back out. So if we listen on `0.0.0.0` (which means: we listen everywhere), it will print out `http://0.0.0.0:8080/`. When you click that link it will open in your browser, and your browser will to connect to that server.

Now, `0.0.0.0` is not actually the address of a computer! IP addresses must have a non-`0` number at the end (and I'm pretty sure the front part also can't be `0`). So `1.0.0.1` would be a valid IP address, but `0.0.0.0` sure isn't. `0.0.0.0` is a symbolic address that means "any IP address" when you use it to specify to a server what IP address to listen on.

Here is where the difference starts to happen! 

* On Linux and Mac, the operating system pretends that `0.0.0.0` just means "local host". This is not correct behavior, but they do it anyway.
* On Windows (or at least on WSL), the operating system goes "nuh-uh, that's not the address of a computer, I can't go there."

Even on WSL, you could still access the Hedy server: you can type `http://localhost:8080` or `http://127.0.0.1/` in your browser's address bar, and it will work. You just can't get there by clicking the link that Hedy prints out during startup.

# Does this all matter?

That's a good question. Maybe it doesn't. I just like typing pedantically correct things 😅.

If we are going to do anything, I would suggest we set the listening to `127.0.0.1` unconditionally. That way:

* The server listens on localhost
* The link that gets printed is clickable on all platforms

Crucially, that would change one thing about the behavior of Hedy: it would no longer be accessible from another computer on your network. If you **listen** on `127.0.0.1` instead of `0.0.0.0`, then it means the server will only accept requests that come from your computer itself (a browser on the same machine), and it will no longer respond to traffic from another computer.

 Is that desirable or not? Hard to say. For development mode, I would say no, but there might be people out there hosting Hedy offline for themselves relying on this behavior?

The best solution is probably to leave the `host` parameter out altogether. From reading the [docs of the Flask.run method](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.run:

image

If we leave the parameter out altogether, it will default to `127.0.0.1`. That would suit most development use cases fine, including those on WSL.

Additionally, the documentation says that users that would like to reinstate the "listen to all traffic, including from other computers" behavior can set the `SERVER_NAME` config variable to listen on `0.0.0.0` instead.

Everybody happy?

Felienne:
> The best solution is probably to leave the `host` parameter out altogether. From reading the [docs of the Flask.run method](https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.run):

Just to clarify @martenvd, if you want to move this PR forward, can you change the code so that it skips the `host` parameter?

Felienne (changes requested):
See conversation for suggestion to leave host out.

martenvd:
Thanks @rix0rrr for the clear explanation! The reason why I opened this PR was that I couldn't connect to Hedy on localhost:8080 without clicking the link in the terminal while on WSL with the host set to 0.0.0.0. Today I tried this again and all of a sudden it did seem to work, so I think I was dealing with a temporary "it doesn't run on my machine" problem 😛. 

@Felienne I altered the PR so that flask now skips the `host` paramter and fills in the default.

Felienne (approved):

das-g:
When following [our docker instructions](https://github.com/Felienne/hedy/blob/87af78c81347c590caa7bfbe6e04f2a86d58ccf0/CONTRIBUTING.md#using-docker) with this change, the Hedy instance running in the docker container is no longer accessible from the host. :slightly_frowning_face:

martenvd:
Ah you're right @das-g, very unfortunate. Seeing as this change didn't matter all that much in the first place, I'll put back in the host parameter in a new PR.

PR #1872, opened 2022-02-01
[FEATURE] Enables teachers to create accounts

### LINKED ISSUE #1386: [UI idea] Allow teachers to load students into class
For example from a csv file with username, email, and password so kids do not have to click. 

After loading, this feature should:

* Generate an account for each email address
* Add the accounts to the class
* Send email to change password to all accounts

TiBiBa:
**Description**
In this PR we implement a new feature, enabling teachers to easily create multiple accounts for their students. Instead of each students having to fill out their own sign up form the teacher can create unlimited account in one go. We implement a new section on the `/for-teachers` page with a "create accounts" button. Directing the teacher to a dedicated "create multiple accounts" page. The teacher only has to fill in the username, mailadres and password for each student. If desired they can also automatically be assigned to a class.

**One important issue to solve**
What do we do with in-active accounts? I can imagine a teacher creating multiple accounts for which some are never used. Do we implemented some timer, that if not used within a week they are automatically removed?

**Fix for**
This PR fixes #1386.

**How to test**
Make sure you are a teacher and navigate to the `/for-teachers` page. Notice the new section with the "create accounts" button. Click it and try to add multiple accounts. Try different combinations and make sure that each field is required. Also try adding an already existing username or mail address. This should result in a warning.

**Important**
This requires some dedicated testing to make sure we don't break anything. Adding multiple accounts add once is a heavy load on the database due to the checking and verifying of each unique username / mail address, the existence of the class and the actual storing of the new user.


TiBiBa:
We're support dynamic field on incorrect fields:
Schermafbeelding 2022-02-01 om 17 47 20 (2)

TiBiBa:
**Food for thought**
I might be better to store all signup code inside `auth.py` as conceptually this is the only place for everything related to account data. This requires some re-writing but might be better on the long run.

das-g:
> **One important issue to solve**
> What do we do with in-active accounts? I can imagine a teacher creating multiple accounts for which some are never used. Do we implemented some timer, that if not used within a week they are automatically removed?

If you implement such a timer for automatic removal of batch-created accounts, make very sure the teachers are aware of it, or they're bound to run into scenarios like the following:

_A teacher plans to begin teaching with Hedy right after the 6-week summer holidays, on the first day after those holidays. The teacher wants to go the last three weeks of those school holidays on vacation themselves. Thus the teacher prepares all the necessary things for teaching with Hedy in advance and also already create the accounts for their pupils. With everything neatly in place, the teacher boards a plane._

_The teacher comes back from their vacation on the last day of the holidays, a Sunday. Their airplane lands at evening, so when back home, they go straight to bed, confident that they're well-prepared for the next day due to their diligent pre-vacation work. Back to school on Monday, some of the kids already in the classroom before the school starts, the teacher checks the preparation one last time and a shocking revelation sets in: All the pupils' accounts are gone!_

_While it's not hard to create them again the same way as before, this makes the teacher very nervous, and they can't do a good job introducing Hedy to their pupils that day._

MarleenGilsing:
Again, another really good feature that is very useful for teachers! Great job!

I was looking for the button for a while in the class overview page. I expected it to be next to the class invite link and the 'Invite student' button. Maybe these buttons should be together, as they have the same function? 

I really like the page lay-out! It's very clear. 

I don't really like the 'amount' function. It works, but not very naturally. To me it would make more sense if I could just type '27' if I wanted to add 27 students. Now I have to add 10, add 10 and add 3 (because there are 4 already), seems inefficient. 
I would prefer if I could just select the amount at once. Or just have a + button under the last line of student accounts. 

The first column says password, which should probably be username. 

E-mail adresses are required and have to be unique. Yet in primary school, students often don't have a school e-mail account. Would it be possible that you register the accounts on the teacher's email adress if the teacher doesn't fill in an email per student? Or maybe have a check-box 'My students don't use e-mail' or something.

The requirements for a password were shown after I filled them in for my whole class. I had to redo them, which in my case was not much trouble because it was 4 fake accounts, but for a real teacher would be a hassle. Maybe show them the requirements upfront?

When I was done creating the accounts, I expected to be redirected to the class overview. Because it now gives you the feeling something went wrong and you have to re-enter all the students' info. 

An extra function that might make it easier for teachers: 
Have a row at the top that you could use for the whole class. For example you would fill in a username in that top row, click a button and that username is filled in for the whole class. Now the teacher only has to add numbers after the username to make them unique (monkey1, monkey2 etc). Same goes for e-mail and for password (giving every student the same password is always handy and letting them change it in the first lesson).

TiBiBa:
> 

Thanks for your really useful feedback! Unfortunately our system doesn't support duplicate e-mail addresses as we allow users to login with either their username or email address. However, it might be worth it to change this to support students without mail accounts. Will make some changes and implement the requested features and let you know when it's ready for review again!

TiBiBa:
Hi @MarleenGilsing ! I made some changes to the UI, would love to hear your feedback. See the screenshot below:
Schermafbeelding 2022-02-03 om 13 51 57

MarleenGilsing:
Yes that looks amazing! I think it's much better than the 'amount' bar and it works naturally too.

website/teacher.py
```diff
@@ -317,6 +319,97 @@ def remove_invite(user):
...
+        usernames = []
+        mails = []
+
+        # Validation for correct types and duplicates
+        for account in body.get('accounts', []):
+            if not isinstance(account.get('username'), str):
+                return g.auth_texts.get('username_invalid'), 400
+            if not isinstance(account.get('mail'), str) or not utils.valid_email(account.get('mail')):
+                return g.auth_texts.get('email_invalid'), 400
+            if not isinstance(account.get('password'), str):
```
ThomasKluiters, line 352: Could we perhaps extract this logic such that we have consistent rules about passwords? It would be easy to "forget" that we duplicated logic here!

website/teacher.py
```diff
@@ -317,6 +319,97 @@ def remove_invite(user):
...
+            usernames.append(account.get('username').strip().lower())
+            if account.get('mail').strip().lower() in mails:
+                return {'error': g.auth_texts.get('unique_emails'), 'value': account.get('mail')}, 200
+            mails.append(account.get('mail').strip().lower())
+
+        # Validation for duplicates in the db
+        classes = DATABASE.get_teacher_classes(user['username'], False)
+        for account in body.get('accounts', []):
+            if account.get('class') and account['class'] not in [i.get('name') for i in classes]:
+                return "not your class", 400
```
ThomasKluiters, line 368: Nitpicky; it would be a bit more consistent to give a 404 here indicating the "class does not exist"

website/teacher.py
```diff
@@ -317,6 +319,97 @@ def remove_invite(user):
...
+            user = DATABASE.user_by_username(account.get('username').strip().lower())
+            if user:
+                return {'error': g.auth_texts.get('usernames_exist'), 'value': account.get('username').strip().lower()}, 200
+            email = DATABASE.user_by_email(account.get('mail').strip().lower())
+            if email:
+                return {'error': g.auth_texts.get('emails_exist'), 'value': account.get('mail').strip().lower()}, 200
+
+        # Now -> actually store the users in the db
+        for account in body.get('accounts', []):
+            hashed = hash(account['password'], make_salt())
```
ThomasKluiters, line 378: I think we should also aim to extract some common logic for account creation here!

ThomasKluiters (commented):
Awesome job, left some minor comments!

website/teacher.py
TiBiBa: Agree! Was already thinking about re-writing all authentication stuff to `auth.py`, will store it in some nice functions which we call on single and multiple account creation.

website/teacher.py
TiBiBa: Agree, will fix

website/teacher.py
TiBiBa: Agree! Was already thinking about re-writing all authentication stuff to `auth.py`, will store it in some nice functions which we call on single and multiple account creation.

TiBiBa:
**Converted to draft until re-structuring is done**

TiBiBa:
Solving some merge issues right now. Will do some extensive testing and than this is ready for review!

TiBiBa:
Ready for review!

Felienne:
> We now support dynamic field styling on invalid input: Schermafbeelding 2022-02-01 om 17 48 13

ow that looks nice, I did not know we supported this!

boryanagoncharenko (commented):
If I log in as a teacher, navigate to the create accounts page and try adding a single student, I am kicked out of my teacher account and I am logged as the user I added. The code explicitly automatically logs the created user and resets the token cookie but this behavoir is nice only when you create your own account. I got many "you are not a teacher" errors before I realized I was logged in as a student.

Another small thing is, if I am working on a small screen, the (red cross) button to remove an account disappears and I cannot scroll horizontally to it. Here is a screenshot:
Screenshot 2022-02-16 at 9 52 41

TiBiBa:
> If I log in as a teacher, navigate to the create accounts page and try adding a single student, I am kicked out of my teacher account and I am logged as the user I added. The code explicitly automatically logs the created user and resets the token cookie but this behavoir is nice only when you create your own account. I got many "you are not a teacher" errors before I realized I was logged in as a student.
> 
> 
> 
> Another small thing is, if I am working on a small screen, the (red cross) button to remove an account disappears and I cannot scroll horizontally to it. Here is a screenshot:
> 
> Screenshot 2022-02-16 at 9 52 41
> 
> 
> 
> 

These are both big mistakes on my side. Thanks for the feedback! Will fix them today.

boryanagoncharenko:
Found one more thing that you should probably have a look into: the student is not invited to a class when I set the class on the create-accounts page.

TiBiBa:
> Found one more thing that you should probably have a look into: the student is not invited to a class when I set the class on the create-accounts page.

Fixed!

boryanagoncharenko (approved):
Everything works as expected now. Thanks for the fixes.

boryanagoncharenko:
I see the PR is still WIP, so I am leaving the merging to you @TiBiBa

PR #1883, opened 2022-02-02
[FEATURE] Shows "last-used" on admin class overview, sort by date

### LINKED ISSUE #1882: [UI idea] Add "last-used" column to class overview
**Idea incl level**
It would be nice to have a "last-used" column on the admin view of the classes. Sorting on this value as well.

TiBiBa:
**Description**
In this PR we improve the admin class overview by adding a new column; "last-used". This value is based on the activity of the teacher and gives insights in the classes activities within Hedy. We also automatically sort the classes by "last-used" date.

**Fix for**
This PR fixes #1882.

**How to test**
Make sure you are an admin and navigate to `/admin`. Notice the new column on the class overview page and the automatic sorting from recent to older.


TiBiBa:
@Felienne I think this one is ready for merging as well! It's a small PR and has no influence on any other work.

templates/admin-classes.html
```diff
@@ -8,23 +8,25 @@ <h1>Classes overview</h1>
...
       </tr>
     </thead>
     <tbody>
       {% for class in classes %}
           <tr>
-            <td class="classname_header">{{class.name}}</td>
-            <td class="teacher_header">{{class.teacher}}</td>
-            <td class="students_header">{{class.students|length}}</td>
-              <td id="statistics_header"><a href="/stats/class/{{class.id}}" target="_blank">Click here</a></td>
+            <td class="classname_cell">{{class.name}}</td>
```
ThomasKluiters, line 25: We're renaming `_header` to `_cell` here, is this intentional? Mostly asking to be sure!

app.py
```diff
@@ -1360,7 +1360,18 @@ def get_admin_classes_page(user):
     if not is_admin(user):
         return utils.error_page(error=403, ui_message='unauthorized')
 
-    classes = DATABASE.all_classes()
+    classes = []
+    for Class in DATABASE.all_classes():
```
ThomasKluiters, line 1364: Nitpicky; it would be a bit more 'pythonic' to do this in a list comprehension:
```python
classes = [{
            "name": Class.get('name'),
            "teacher": Class.get('teacher'),
            "students": len(Class.get('students')) if 'students' in Class else 0,
            "id": Class.get('id'),
            "last_used": utils.datetotimeordate(utils.mstoisostring(DATABASE.user_by_username(Class.get('teacher')).get('last_login')))} for Class in DATABASE.all_classes()]
```

app.py
```diff
@@ -1360,7 +1360,18 @@ def get_admin_classes_page(user):
...
+        current_class = {
+            "name": Class.get('name'),
+            "teacher": Class.get('teacher'),
+            "students": len(Class.get('students')) if 'students' in Class else 0,
+            "id": Class.get('id'),
+            "last_used": utils.datetotimeordate(utils.mstoisostring(DATABASE.user_by_username(Class.get('teacher')).get('last_login')))
```
ThomasKluiters, line 1370: I'm not too familiar with dynamo / performance / load and how many classes there are currently - would it perhaps be a bit better to include `last_teacher_login` as part of the `Class` *class* (😎) and update this on teacher activity?

ThomasKluiters (commented):
Awesome! Some small comments / requests to clarify as I'm a bit new to othe code!

app.py
TiBiBa: Good idea! We currently don't support that many classes; just looked at the database: 45 at the moment. I agree that from a performance perspective it is a good idea to add the `last_log` to the class data. But I want to prevent storing duplicate data in too many places. Especially because of the small scale I think this works fine. The data can only be retrieved by the admin, so we don't have to worry about database load too much. However, if we scale to more classes we maybe don't have a choice to change but for know I suggest to keep it this way. (Will add an inline comment though).

app.py
TiBiBa: Fixed

templates/admin-classes.html
TiBiBa: Yes! We use a similar structure on our other tables where we support dynamic showing / hiding of column. We don't have this feature (yet) on the classes overview, but might be useful in the future. By using the same naming structure we unify the code and, the biggest advantage, can call the same function for the functionality, this one:

```
$(document).ready(function() {
        $('.attribute').change(function() {
            const attribute = $(this).attr('id');
            if(!this.checked) {
                $('#' + attribute + '_header').hide();
                $('.' + attribute + '_cell').hide();
            } else {
                $('#' + attribute + '_header').show();
                $('.' + attribute + '_cell').show();
            }
        });
});
```

app.py
Felienne: Good suggestion @ThomasKluiters and good idea to keep an eye on this! For now I am the only admin so indeed it is probably fine but it might get out of hand one day :)

Felienne (approved):

PR #1885, opened 2022-02-02
[FEATURE] Automatically marks erroneous programs when sharing

### LINKED ISSUE #1807: [UI idea] Don't show erroneous programs in explore
If kids share a program that does not work, it will be shown in explore but it might not be a good example. Can we disallow that? We could either block the public sharing, or (better UX, but more expensive!) filter out erroneous programs when we populate explore.

TiBiBa:
**Description**
In this PR we add a new feature to keep track of erroneous shared programs. From now on programs are parsed before sharing and if an error occurs the user is presented with a confirmation modal. Then we store the parsing result (error : true / false) in the db as addition to the program stored. This way we can differentiate between correct and erroneous programs on the explore page.

**Fix for**
This PR fixes #1807 

**How to test**
This requires some testing as we have re-written some core functionality of our website. Login as any user and save either a correct or incorrect program. This process should work identical as before. When trying to share an incorrect program (either from the code page or the programs page) an additional pop-up is shown to verify that you are sure to share an erroneous program. Navigate to the explore page and notice that your faulty program now has a red header instead of the original blue one.


TiBiBa:
We implemented a visual change to the `/explore` page to indicate erroneous programs:

Schermafbeelding 2022-02-02 om 15 23 30

TiBiBa:
**This needs some extensive testing locally and on alpha to make sure we don't break anything**

As this PR always updates existing programs to add the `error` value we have to make sure that we don't break anything by merging this.

boryanagoncharenko (approved):
Tested all scenarios I could think of and it is working as expected.

PR #1891, opened 2022-02-02
More hindi

### LINKED ISSUE #1887: [BUG] Suspected hash missing for var in range in level 11
**Describe the bug**
Accented vars cannot be used as loop iterators

Level 11:
```
for aaaaá in range 10 to 1
    print aaaaá ' kleine visjes zwommen naar de zee'
```

I suspect this is due to a missing hash of the var given the error message T_OP.

No need for anyone to pick this up, it is just a note to self since I am in the train and can't do it right now :)

Felienne:
This PR does two things (I know, I know, do as I say not as I do children 🤣 ):

- Firstly it restructures the old translated Hindi texts as best as possible
- It also fixes #1887


archive/level21-Additions.lark
```diff
@@ -1,10 +1,10 @@
 // adds length of lists
 
 print: _PRINT (_SPACE* _LEFT_BRACKET _SPACE* (quoted_text | list_access | var_access | sum | length) (_SPACE (quoted_text | list_access | var_access | sum | length))* _RIGHT_BRACKET)?
-input: var _SPACE _IS _SPACE _INPUT (_LEFT_BRACKET  (quoted_text | list_access | var_access | sum | length) (_SPACE (quoted_text | list_access | var_access | sum | length))*  _RIGHT_BRACKET)?
+input: var _SPACE _IS _SPACE _INPUT _SPACE? (_LEFT_BRACKET  (quoted_text | list_access | var_access | sum | length) (_SPACE (quoted_text | list_access | var_access | sum | length))*  _RIGHT_BRACKET)?
```
ThomasKluiters, line 4: Mostly a question for my own understanding; why is adding this space here required for making Hindi work?

Felienne: Ow good catch @ThomasKluiters, this file should not have been part of the PR at all!

Felienne: Note thought that this is in the archive folder that is not actively used but an artefact of changes in the past that still might need to be places somehere

PR #1892, opened 2022-02-03
Repair missing quote with mutations

dolfein:
**Description**
Makes (Hedy) mutants whenever there is a missing quote when trying to print something. Each mutant has a quote inserted at a different position on that line.

Each mutant gets transpiled until a mutant is found without an error, mutants with errors are thrown away.

It uses recursion, that is why the values related to mutations are stored in program_repair.py.

**Fix for**
#1225, point 4

**How to test**
Two tests are added at level 5.
If there is a mutant without an error, it will be saved in `fixed_code` of `UnquotedTextException`. Otherwise `fixed_code` will be None.

**Checklist**


dolfein:
Analysing mutations (in general) are pretty computationally expensive, especially in the worst cases (if the last mutant is a mutant without errors or if there is none). So there is a noticeable time delay in Hedy.

Felienne:
Hi @dolfein!

I will test in more depth but is it correct that this change is not yet visible in the UI?

image

If you would have phrased "how to test" more clearly, I would have known whether this was intentional or not.

dolfein:
> I will test in more depth but is it correct that this change is not yet visible in the UI?

That is correct, it is not visible in the UI yet, the fixed code is saved in the exception (UnquotedTextException).

Felienne:
> That is correct, it is not visible in the UI yet, the fixed code is saved in the exception (UnquotedTextException).

ok! I notice indeed that it makes Hedy slower even while not being visible so I guess it is not wise to merge this now. I guess ultimately we want something interactive where we first (quickly) show the original error, and then ask the user if they want to try to fix it.

Felienne (changes requested):
@dolfein Can you make it such that the code remain in the code base but is not yet executed on transpile to not slow down users while they cannot benefit form the feature?

dolfein:
> @dolfein Can you make it such that the code remain in the code base but is not yet executed on transpile to not slow down users while they cannot benefit form the feature?

You mean like this? We can call the function `mutation_repair(...)` whenever it is fully functional.
I have to comment out the tests though, or should I just delete them for now?

hedy.py
```diff
@@ -2127,18 +2142,6 @@ def is_program_valid(program_root, input_string, level, lang):
             raise exceptions.InvalidSpaceException(level=level, line_number=line, fixed_code=fixed_code, fixed_result=result)
         elif invalid_info.error_type == 'print without quotes':
             # grammar rule is agnostic of line number so we can't easily return that here
-            program_repair.make_mutants(input_string, line)
```
Felienne: so here we would do something like `if repair_enabled: mutation_repair(input_string, line, level, lang)`

Felienne (changes requested):
Hi @dolfein 

Thanks for improving this. What we'd like is a "feature flag" a variable that determines if the feature is on. We can then use the code in the tests (with boolean True) while we leave it at False in the code.

Felienne:
Perfect!!

program_repair.py
```diff
@@ -22,19 +26,42 @@ def replace(input_string, line, column, length, new_string):
     return result
 
 
+def make_mutants(input_string, line):
+    mutants = []
+    length = len(input_string.splitlines()[line - 1]) + 2
+    for i in range(1, length):
+        mutants.append(insert(input_string, line, i, '\''))
```
Felienne, line 33: insert this way is very expensive because split() and join are expensive. Can we rewrite the code so it splits and joins once? So split at the beginning of make_mutants, only pass the one line to insert() and then join all code back in at the end of this?

program_repair.py
```diff
@@ -22,19 +26,42 @@ def replace(input_string, line, column, length, new_string):
     return result
 
 
+def make_mutants(input_string, line):
+    mutants = []
+    length = len(input_string.splitlines()[line - 1]) + 2
```
Felienne, line 31: This feels like a lot of magic, why is the + 2 needed? Can you explain in a comment?

Felienne (changes requested):

Felienne:
Hi @dolfein!

I improved upon your code in a few ways, making it more explicit what happens. I also added one more suggestion for you to work on.

program_repair.py
```diff
@@ -1,15 +1,19 @@
+fixed_code = None
+mutants_made = False
+import hedy
+
 def insert(input_string, line, column, new_string):
```
Felienne, line 5: insert this way is very expensive because split() and join are expensive. Can we rewrite the code so it splits and joins once? So split at the beginning of make_mutants, only pass the one line to insert() and then join all code back in at the end of this?

program_repair.py
```diff
@@ -1,15 +1,19 @@
...
     """"insert new_string at (line, column)"""
     rows = input_string.splitlines()
-    rows[line] = rows[line][:column] + new_string + rows[line][column:]
+    rows[line - 1] = rows[line - 1][:column - 1] + new_string + rows[line - 1][column - 1:]
```
Felienne, line 8: inserting this way is very expensive because split() and join are expensive. Can we rewrite the code so it splits and joins once? So split at the beginning of make_mutants, only pass the one line to insert() and then join all code back in at the end of this? (the is true for your other mutation functions too)

program_repair.py
```diff
@@ -22,19 +26,46 @@ def replace(input_string, line, column, length, new_string):
     return result
 
 
+def make_mutants(input_string, line):
+    mutants = []
+    length = len(input_string.splitlines()[line - 1]) + 2
+    for i in range(1, length):
+        mutants.append(insert(input_string, line, i, '\''))
```
Felienne, line 33: We now insert quotes at all locations from left to right, which is not very effective. For example we generate `pr'int` which we can guess will not work. We can make this more effective by inserting quotes only around spaces present in the erroneous Hedy code?

We can also explicitly disallow inserting quotes inside words.

Felienne (changes requested):

program_repair.py
dolfein: I rewrote it such that split() is only done once, but I do not think we can reduce the amount of join().
The amount of join() calls will always be equal to the amount of mutants.

PR #1902, opened 2022-02-04
[BUG] Support numeric values in list assignment

### LINKED ISSUE #1777: [BUG] drawing turtle using at random gives unexpected response
Given the program should print the char E but in this level we chose to let the program descide the length of the lines.
This program however is drawing something strange, and I do not know why. 
I mention it here just for your information…. Perhaps it is of use for you.

    print We gaan de E tekenen
    a is ask doe je mee?
    lengte is 50, 75, 100
    
    boven is lengte at random
    midden is lengte at random
    onder is lengte at random
    hoogte is onder
    
    turn 180
    forward boven
    turn 270
    forward hoogte
    turn 270
    forward midden
    turn 180
    forward midden
    turn 270
    forward hoogte
    turn 270
    forward onder
    print  🎉🎉🎉🎉🎉🎉🎉

44808548-22FF-48F7-AF96-F72E762B0937

ThomasKluiters:
**Description**

In order to support lists that fully consist of numbers, this commit
will not surround list values with quotes if all the values are in
a numerical format.

This will allow Hedy programs to use values from lists that require
the value to be a number, such as `forward` in turtle programs.

**Fix for**

This commit fixes #1777

**How to test**

Create a list with only numerical values and reference the list value in an expression that accepts a numeric value, such as `forward`.

**Checklist**

### DIFF
--- a/tests/test_level_03.py
+++ b/tests/test_level_03.py
@@ -651,3 +651,15 @@ class TestsLevel3(HedyTester):
       code=code,
       exception=hedy.exceptions.WrongLevelException
     )
+
+  def test_list_assign_numeric(self):
+    code = textwrap.dedent("lengte is 50, 75, 100")
+    expected = textwrap.dedent("lengte = [50, 75, 100]")
+
+    self.single_level_tester(code=code, expected=expected)
+
+  def test_list_assign_alphanumeric(self):
+    code = textwrap.dedent("lengte is a, 75, 100")
+    expected = textwrap.dedent("lengte = ['a', '75', '100']")
+
+    self.single_level_tester(code=code, expected=expected)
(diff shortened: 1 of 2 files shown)


tests/test_level_03.py
```diff
@@ -89,23 +89,23 @@ def test_print_multiple_lines(self):
...
+
     expected = textwrap.dedent("""\
     print(f'*Jouw* favoriet is dus kleur')""")
 
     self.single_level_tester(code=code, expected=expected)
 
   def test_print_quotes(self):
     code = "print 'Welcome to OceanView!'"
-    
+
```
ThomasKluiters, line 108: Whoops - will change this

ThomasKluiters:
Not sure if this is the best solution - we could also force `forward` to work with strings

PR #1905, opened 2022-02-04
Allows assigning a var to a var

### LINKED ISSUE #1777: [BUG] drawing turtle using at random gives unexpected response

Felienne:
Fixes #1777

The issue here was twofold, firstly we did not anticipate assigning a var to a var, so we interpreted the value as a string always (this PR fixes that)

Secondly, `forward` did not complain about getting a string variable, which is maybe something we would want to catch but because this happens at runtime it will be hard to nail down.

Hope this makes sense @ThomasKluiters?

And to your question did it ever work? Maybe before I added the escape code? I don't know! But assigning a var to a var is quite complex and since we don't use it in example code we might just never have observed it before (https://github.com/ericverschoor is just to smart for us haha)


Felienne:
Ah the plot thickens, it seems we explicitly used to allow assigning var names as string values? I don't know what I was thinking but I think something else now haha. The test also does not give me any clue why I thought it would make sense

StephanEggermont:
Do you already want this at a low level? Or make that an explicit step? I tried with my kids at the level where they were introduced, but assumed it was a deliberate limitation

ThomasKluiters:
@Felienne Yeah - I think this makes sense, I'd say it is a bit of a quirk and a gotcha; "If a block of text matches that of a variable, it will be treated as a variable reference". I would think that this behaviour is expected in other places as well as this PR only handles the "var to var" case - maybe a more generic solution is desired for the long term, but this definitely works for the issue we were looking into!

ThomasKluiters (approved):

Felienne:
> @Felienne Yeah - I think this makes sense, I'd say it is a bit of a quirk and a gotcha; "If a block of text matches that of a variable, it will be treated as a variable reference". 

Yeah it is a bit of a gotcha and I think it gets specifically tricky when (as in the case of the issue we address here) the variable is used not in printing but in commanding the turtle because then it remains invisible what the value is.

I suspect because we don't show this explicitly in the example code this is mostly something kids to with programmers parents next to them (supported by the two instances in this thread :) ) so it does not matter that much either way. 

> I would think that this behaviour is expected in other places as well as this PR only handles the "var to var" case - maybe a more generic solution is desired for the long term, but this definitely works for the issue we were looking into!

Agreed there too, if you are interested in looking at it in the more broad sense, let me know!

PR #1864, opened 2022-01-31
[UI IMPROVEMENT] Re-structured class customizations

TiBiBa:
**Description**
In this PR we improve the implementation of the class customizations. Currently we store the customizations as a [class_id, level] pair. This is quite cumbersome because we also keep track of the restriction of previous or next Hedy levels. Resulting in the situation that we have to update three db rows at every update. We also have (a lot!) of duplicate storing of data due to storing the available adventure names in each row. We improve this by implementing the following structure:

```
"class_customizations" {
"class_id": 1234,
"levels": [2,3,4],
"adventures: [
    "Story": [1, 2, 3, 4],
    "Parrot": [1, 2, 3]
]}
```

Using only one db row to store the chosen levels and the chosen levels per adventures. After merging #1812, we will have to change the structure slightly to support personal adventures as well. Overall the code is significantly simplified, on the front-end and well as the back-end. Making it easier to understand and alter the functionality of customizations.

**Fix for**
This PR fixes #1606.

**How to test**
This is mostly a clean-up task so for end-user there should be no visible change. For teachers: log in, navigate to one of your classes and verify that the UI of the customizations is changed. Try different level, adventure combinations and (when being a student of your own class) verify that these are correctly customized on the coding page.

**Important**
Due to the new structure we have to delete all old customizations from the database. This might influence some teachers and classes currently actively working with Hedy. We should notify them to manually copy there customizations to the new format.

PR #1906, opened 2022-02-04
Check if there are customizations for this user

Felienne:
It seems #1864 introduced an issue where levels would be closed for users not in classes since the next level would not be in their levels since that list would be empty. I think this fixes it but I would need confirmation from @TiBiBa (and I need to test a bit more)


Felienne:
ok upon testing a bit more I am pretty sure this fixes the issue! Will merge and test on alpha tomorrow

PR #1909, opened 2022-02-04
reformat license & remove gratuitous quotation marks

### LINKED ISSUE #1908: [CONTENT] gratuitous quotation marks in License.txt
**Describe the issue**
The `License.txt` in the source code repo contains some quotation marks that don't belong there and that aren't in the [EUPL 1.2 versions available from the EU](https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12) (neither [the English PDF](https://joinup.ec.europa.eu/sites/default/files/custom-page/attachment/eupl_v1.2_en.pdf) nor [the English plaintext](https://joinup.ec.europa.eu/sites/default/files/custom-page/attachment/2020-03/EUPL-1.2%20EN.txt) version):
https://github.com/Felienne/hedy/blob/2edeb4b39efaf1fd7853cb032917928b6f48a6e5/License.txt#L20-L28
(at the beginnings of lines 21, 22 and 24–28)

n/a

**Expected text**
We can probably just use the plain text version provided by the EU.

das-g:
**Description**

Re-formats `License.txt` and removes some gratuitous opening quotation marks in it by replacing the file with the plaintext English one available at https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12 as "EN" "TXT-80": [`EUPL-1.2 EN.txt`](https://joinup.ec.europa.eu/sites/default/files/custom-page/attachment/2020-03/EUPL-1.2%20EN.txt). Except for those quotation marks (which didn't belong there) and the title/header section (which is being slightly re-arranged by this change), this doesn't change the textual or semantic content of the license, as can be checked by `git show --word-diff faf87fe91603466e956292af5a896debc53c7421`: Only the line breaks are placed differently within the prose (i.e., the text has been re-flown w.r.t. our previous version) and ASCII's Hyphen-Minus (`-`) is used as bullet list item marker instead of Unicode Em-Dash (`—`).

**Fix for**

fixes #1908

**How to test**

Unless the license is used programmatically in the software somewhere (which I assume isn't the case), no tests are needed.

**Checklist**

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -1,44 +1,53 @@
-European Union Public Licence
-V. 1.2
+                      EUROPEAN UNION PUBLIC LICENCE v. 1.2
+                      EUPL © the European Union 2007, 2016
 
-EUPL © the European Union 2007, 2016
-
-This European Union Public Licence (the ‘EUPL’) applies to the Work (as
-defined below) which is provided under the terms of this Licence. Any use of
-the Work, other than as authorised under this Licence is prohibited (to the
-extent such use is covered by a right of the copyright holder of the Work).
+This European Union Public Licence (the ‘EUPL’) applies to the Work (as defined
+below) which is provided under the terms of this Licence. Any use of the Work,
+other than as authorised under this Licence is prohibited (to the extent such
+use is covered by a right of the copyright holder of the Work).
 
 The Work is provided under the terms of this Licence when the Licensor (as
 defined below) has placed the following notice immediately following the
-copyright notice for the Work: “Licensed under the EUPL”, or has expressed by
-any other means his willingness to license under the EUPL.
+copyright notice for the Work:
+
+        Licensed under the EUPL
+
+or has expressed by any other means his willingness to license under the EUPL.
 
 1. Definitions
 
 In this Licence, the following terms have the following meaning:
-— ‘The Licence’: this Licence.
-— ‘The Original Work’: the work or software distributed or communicated by the
-  ‘Licensor under this Licence, available as Source Code and also as
-  ‘Executable Code as the case may be.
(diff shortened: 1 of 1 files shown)


Felienne:
hmmmm interesting, I am pretty sure I simply took the version that github provided at the time, but I guess they changed things in the mean time.

Felienne (approved):

das-g:
> hmmmm interesting, I am pretty sure I simply took the version that github provided at the time, but I guess they changed things in the mean time.

Interesting indeed, so I now did some digging:

I wanted to check what text GitHub would currently provide (and see wheter that needs fixing, too), but in the current [new-repo form](https://github.com/new), EUPL isn't even offered as an option:
image

Though https://choosealicense.com/ (which that dialog was previously based on, IIRC) had (supposedly at https://choosealicense.com/licenses/eupl-1.2) the exact version we had previously from github/choosealicense.com@9f263936b2fa01a1ed13cb0bab154c0916294c91 till (including) github/choosealicense.com@524ed15f823b4cb0f59c54696f095ce436d264b9. They fixed it with github/choosealicense.com@7a5b314c0472b74f112f91a138482055f564831d but have a slightly differently wrapped version now, because they seem to maintain a shorter max. line length than the EU's plain text version. The excess quotation marks were introduced much earlier when they first line-wrapped their [original version](https://github.com/github/choosealicense.com/blob/bcbf76b178fe9918df52810f5cc44cc10d2425fb/_licenses/eupl-1.2.txt) (that was maybe scraped from a HTML or PDF version or the EU or SPX?) in github/choosealicense.com@0b57486e2cce17ffa8bde636bf591a48fecda83c.

PR #1910, opened 2022-02-04
fix grammar in German Level 4 intro & bring its text closer to the English version

das-g:
**Description**

- Fixes the grammar in the German Level 4 intro text.
- Brings the German Level 4 intro closer in meaning to the current English Level 4 intro.

**Fix for**

(none filed)

**How to test**

1. Set Hedy to German (with the top right language chooser when not logged in or in your profile settings when logged in)
2. Read first tab of [Level 4](https://www.hedycode.com/hedy/4)

**Checklist**

   - well, kinda

### DIFF
--- a/coursedata/level-defaults/de.yaml
+++ b/coursedata/level-defaults/de.yaml
@@ -115,16 +115,16 @@
 4:
     start_code: "print 'Hallo Welt'"
     intro_text: |
-        In Level 4 `print` und `ask` ändern.
+        In Level 4 haben sich `print` und `ask` geändert.
 
         Du musst Text, den du genau so mit `print` ausgeben möchtest, in Anführungszeichen setzen.
 
-        Du kannst du jetzt alle Wörter ausgeben, die du möchtest. Auch die Wörter, mit denen du etwas mit `is` gespeichert hast.
+        Das ist nützlich, denn nun kannst du alle Wörter ausgeben, die du möchtest. Auch die Wörter, mit denen du etwas mit `is` gespeichert hast.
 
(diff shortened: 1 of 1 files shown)


coursedata/level-defaults/de.yaml
```diff
@@ -115,7 +115,7 @@
 4:
     start_code: "print 'Hallo Welt'"
     intro_text: |
-        In Level 4 `print` und `ask` ändern.
+        In Level 4 haben sich `print` und `ask` geändert.
 
         Du musst Text, den du genau so mit `print` ausgeben möchtest, in Anführungszeichen setzen.
```
StephanEggermont, line 121: Want to fix the next line too?

das-g: > Want to fix the next line too?

"Du musst Text, den du genau so mit `print` ausgeben möchtest, in Anführungszeichen setzen." seems already correct to me. What should be fixed there, and why would it belong in this PR rather than a separate one?

StephanEggermont: 122 Du kannst du

das-g: Sorry, I assumed "next" meant the one after the already changed one, rather than the one after the location of your review comment (which the overview UI doesn't even show). I've not also fixed the grammar there and further below in the Level 4 intro, and also added changes to bring the Level 4 intro closer in meaning to its current English version.

das-g:
Sorry for the noise. I had make some mistakes when attempting the additional corrections, which should now be rectified. Please review again @StephanEggermont

StephanEggermont (approved):
Looking good. Thank you!

Felienne (approved):

PR #1911, opened 2022-02-04
VSCode settings added

Mark-Giesen:
Added VSCode .gitignore settings and launch.json.


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Huidige bestand",
```
StephanEggermont, line 8: are you sure you want that to be Dutch?

Felienne:
Hi @Mark-Giesen!

It is great to add the vscode folder to the gitignore, thanks! But we don't want to add the folder itself too, because these two changes will take effect together, so you add the folder before adding the instruction to ignore it. Can you remove your vscode files and the folder from the PR?

.vscode/launch.json
Mark-Giesen: Oops, no! Good catch. The file was generated, didn't check the language.

Mark-Giesen:
The vscode folder is not added to .gitignore, I belive, just the files in it, except the re-included ones. Or am I misreading your comment?

Felienne:
> The vscode folder is not added to .gitignore, I believe, just the files in it, except the re-included ones. Or am I misreading your comment?

Yeah fair enough, what I meant to say is that we shouldn't include the files in the folder because, these are your files.

Felienne:
Removed it for you :)

Felienne (approved):

PR #1921, opened 2022-02-07
Adds semantics level 16

juliabolt:
**Description**

Level 16 has the same commands as level 15, but in level 16 lists are used with [ ]  and it is now also possible to use lists with print and ask.
(I kept repeat in this branch for now, since we haven't decided yet what to do with repeat)

**How to test**

Read SEMANTICS.md level 16


Felienne (changes requested):
Remove the changes for level 11

Felienne (approved):

PR #1928, opened 2022-02-08
Added small new button containing list of commands per level

fRedelaar:
**Description**

In the discussion about removing the palette, I've added some ideas. This PR is the result of the last [idea](https://github.com/Felienne/hedy/discussions/1835#discussioncomment-2132503). 

**How to test**
Should show all the commands available per level, when hovering the "🤔" button.

### DIFF
--- a/templates/level-page.html
+++ b/templates/level-page.html
@@ -14,15 +14,17 @@
       <div class="dropdown inline-block relative green-btn mx-2 cursor-pointer">
             <span class="ltr:mr-1 rtl:ml-1">{{ "🤔" }}</span>
           </button>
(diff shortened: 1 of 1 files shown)


fRedelaar:
image

On hover:
image

Felienne:
Love it and done with such a speed also!!

Could we make it so that if you click an option, you get the code snippet in the editor, like we had in the side bar before? I think that feature is quire powerful.

fRedelaar:
> Love it and done with such a speed also!!
> 
> Could we make it so that if you click an option, you get the code snippet in the editor, like we had in the side bar before? I think that feature is quire powerful.

Sure, I will update it!

Felienne:
Cool, I love this!!

PR #1943, opened 2022-02-11
ADD: re-add host parameter for flask

martenvd:
**Description**

Describe your changes in detail. Please use present tense without a subject to describe your PR, for example: "Fix**es** issue 123" or "Adds translations of levels 1 to 12 for Polish"

Re-adding the host parameter to fix the docker image. Leaving out the host parameter doesn't expose the application from the container, as commented by @das-g

**Fix for**

https://github.com/Felienne/hedy/pull/1871

**How to test**

```
docker build -t hedy .
docker run -p 8080:8080 -d --rm hedy
```

**Checklist**

### DIFF
--- a/app.py
+++ b/app.py
@@ -1848,6 +1848,6 @@ if __name__ == '__main__':
     on_server_start()
 
     # Threaded option enables multiple instances for multiple user access support
-    app.run(threaded=True, debug=not is_in_debugger, port=config['port'])
+    app.run(threaded=True, debug=not is_in_debugger, port=config['port'], host="0.0.0.0")
 
     # See `Procfile` for how the server is started on Heroku.


das-g (approved):
We could add special handling for the case that Hedy runs in a container, like a way to set the host from outside the code (e.g. with an environment variable that the docker image can set, something that could come in handy in other situations, too). But that would make things more complicated and for now this (reverting #1871) is the easiest solution. :+1:

das-g:
Thank you! :heart:

das-g:
> **How to test**
> 
> ```
> docker build -t hedy .
> docker run -p 8080:8080 -d --rm hedy
> ```

Maybe add
> Then access http://localhost:8080/
> The Hedy landing page should show up.

at the end.

(Without that step, one wouldn't notice the problem.)

Felienne (approved):
Thanks for the revert @martenvd and thanks for noticing the issue @das-g!

PR #1944, opened 2022-02-12
Fix english quiz

thjazi39:
empty field that deregulated the display, especially in quiz 6
and in level 6, question 7, there was a typo (this was not part of the question)


Felienne:
Thanks @thjazi39!! Maybe @MarleenGilsing wants to check if this is in order?

Felienne (approved):

Issue #1940, opened 2022-02-11
[BUG] Enable French keywords


Felienne:
Since #1925 we have French keywords so I need to do some manual magic to make them work (or get my shit together and write a proper script haha)

omalaspinas:
Could french keyword contain accents so they are closer to actual french? For example `repeat` would be `répète` (instead of the current repete) and `to` would be `à` (instead of the current `a`).

Felienne:
Yes absolutely @omalaspinas, we support all character sets for keywords! (I didn't realize we did not have them since I don't speak French myself)

If you want to, you can update the keywords file here: https://github.com/Felienne/hedy/blob/main/coursedata/keywords/fr.yaml and I will take that into account when enabling. Otherwise I am sure @thjazi39 can update them.

omalaspinas:
Yes! I'll make a quick PR for the keywords and @thjazi39 as he's the original author to see what he thinks.

Also a second question. Is there any preference for keywords that are in two words? I saw that in english "else if" becomes "elif" but "to list" becomes "to_list" (also the variables seem to be snake_case like). Should the keywords made of two words be simply concatenated ot should the words be separated by a "_"?

Your project is really awesome I can't wait to have the complete translation finished to play with my children :)

omalaspinas:
Should I open an Issue to modify the french keywords and submit a PR? Or should I directly create a PR and connect it to this issue? Also should I make a PR to the add_french branch? or to main?

Felienne:
> Should I open an Issue to modify the french keywords and submit a PR? 

PR is fine without an issue, since we have discussed it here

> Or should I directly create a PR and connect it to this issue? 

Yes please mention this issue with the PR.

> Also should I make a PR to the add_french branch? or to main?

Best to make a fresh PR to main, thanks!!

Felienne:
> Yes! I'll make a quick PR for the keywords and @thjazi39 as he's the original author to see what he thinks.

Superb!

> Also a second question. Is there any preference for keywords that are in two words? I saw that in english "else if" becomes "elif" but "to list" becomes "to_list" (also the variables seem to be snake_case like). Should the keywords made of two words be simply concatenated ot should the words be separated by a "_"? My personal taste would be to separate with the underscore but I'll happily follow the "hedy" way.

I think we always use underscores, but we use elif because that happens to be the Python way. But these are just keys, so you will have to keep them exactly as they are anyway, and only translate (or in your case, change) the right hand side.
 
[...]

PR #1945, opened 2022-02-12
Update for french keywords

omalaspinas:
**Description**

* Updates keywords with accents to be as close as possible to french (see Issue ##1940).
* Updates range keyword to intervalle.
* Adds underscore for two words keywords.

**Fix for**

Some of the french keywords were without accents and no underscores which could make them less understandable for young users.

**How to test**

Not really applicable.

**Checklist**

- [X] Links to an existing issue or discussion (if not, create an issue first)
- [X] Describes changes clear in the format above (present tense, no subject)
- [X] Has a "how to test" section
- [X] Only one thing is done in this pull request (specifically please try to refrain from mixing textual changes to the yamls from code changes)


coursedata/keywords/fr.yaml
```diff
@@ -6,23 +6,23 @@ turn: tourne
 is: est
 sleep: dors
 add: ajoute
-to_list: jusqua
+to_list: jusqu_a
```
Felienne, line 9: Ah now I think I understand your earlier remark about the space versus underscore you meant on this side too.

I reckon this is actually 2 words with a space in between! I think I can make it work so that you can use `jusqu a` cause that will be more authentic I assume. Stay tuned :)

omalaspinas: After the discussion in #1940 with  @thjazi39 I just realized that the also should be an accent on the "a" since it's  "jusqu'à" for consistency. I haven't seen where this to_list is actually used in the language and so it puzzles me on the actual translation this should have.

Felienne: this 'to' is used when adding something to a list, so it is the `to` in:

`add appel to taarten`

It can potentially be different from the `to` for a range : `for i in range 1 to 5`, for Dutch it is for example (toe versus tot)

Felienne:
Yes feel free to update!

thjazi39:
Hi @omalaspinas
Apparently, you also translate hedy from English to French, could we synchronize so that the merges go better?

Currently, I translate the file course_data/quiz/fr.yaml
I also started to modify the coursedata/level-defaults/en.yaml file and the coursedata/adventures/en.yaml file but if you want to work on them, I can restore the files, so you can translate them!

omalaspinas:
Ok. That's great. I finished translating the level-defaults (I wanted to "learn" hedy so I did it from the beginning). As soon as this PR is merged, I will submit a second one for the updated level defaults. I also started the adventure, where I started from scratch since I didn't know what exactly was the state with respect to english. I haven't really touched the quizz though.

thjazi39:
Okay, so I cancelled my modifications that I had made on 
level defaults and on adventure (I hadn't done much)

So, my branch is now only focused on the quiz file (which I'm almost done)

Felienne (approved):
Thanks @omalaspinas!!

I will try to integrate this with and without accents as promised!

PR #1949, opened 2022-02-14
Automatically re-directs teachers to for-teachers page on login

### LINKED ISSUE #1899: [UI idea] Redirect teachers to teacher's page
**Idea incl level**
Teachers on login are not redirected to base_URL/programs but would it not bemore useful to redirect to /for-teachers?

TiBiBa:
**Description**
In this PR we implement a small feature where teachers are re-direct to the for-teachers page instead of the programs page when logging in. For regular users nothing changes. We add an extra boolean called `teacher` to the login response from the back-end. Using this value in the `afterLogin()` function on the front-end to verify where we want to re-direct to.

**Fixes #1899**

**How to test**
Login with both a teacher an a regular account. Notice that in both cases you will be re-directed to a different page. Verify that the re-direct after signup is also still working correctly. A signup re-direct is always set with `teacher = False` as you can't be a teacher account from the start.

**Checklist**
  

static/js/auth.ts
```diff
@@ -362,7 +362,7 @@ $ ('#email, #mail_repeat').on ('cut copy paste', function (e) {
  * - Check if we were supposed to be joining a class. If so, join it.
  * - Otherwise redirect to "my programs".
  */
-async function afterLogin() {
+async function afterLogin(teacher: boolean) {
```
ThomasKluiters, line 365: Nitpicky; I think we should pass an object to `afterLogin` - if we ever want to have different behaviours after logging in it may become a bit confusing in which order we pass the "login state";
 
 I propose:

```typescript
afterLogin({"teacher": response['teacher']})
```

static/js/auth.ts
```diff
@@ -362,7 +362,7 @@ $ ('#email, #mail_repeat').on ('cut copy paste', function (e) {
  * - Check if we were supposed to be joining a class. If so, join it.
  * - Otherwise redirect to "my programs".
  */
-async function afterLogin(teacher: boolean) {
+async function afterLogin(loginData: any) {
```
ThomasKluiters, line 365: I personally really don't care, but, if you want to leverage typescript here, you may want to use type annotations;

```typesccript
loginData: { teacher: boolean }
```

ThomasKluiters (approved):

static/js/auth.ts
Felienne: @ThomasKluiters You made this remark (which has not been changed) but also approved so now I am not sure if I should await this change or merge?

PR #1954, opened 2022-02-14
[BUG FIX] Keeps shared status on program overwrite

### LINKED ISSUE #1893: [BUG] When over-writing an existing shared program it's no longer shared
**Describe the bug**
Currently we enable users to store programs and retrieve them later to continue working on them. However, when storing the same program again we return a "over-write warning". Which in the first place is weird as we loaded an existing program, not started a new one and stored with an already existing name. The second issue is that over-writing a shared program creates a non-shared program, which is weird as well.

**Expected behavior**
When saving an opened shared program the new version is still saved.

TiBiBa:
**Description**
In this PR we implement a fix so we keep the current shared status of a program when overwriting. When a user continues working on a program and tries to save a `duplicate-check` modal is returned. Currently when deciding to overwrite we only keep the `program_id` but nothing else. In this PR we also copy the shared status to the new program.

**Fixes #1893**

**How to test**
Verify that this works correctly in two ways. Make sure you have a saved program, open it and over-write the program with the same name (simply press save). Notice that the shared status remains. The other way around; verify that a non-shared program stays non-shared when overwriting it.

**Checklist**
  

app.py
```diff
@@ -1568,11 +1568,13 @@ def save_program(user):
     # For now, we bring all existing programs for the user and then search within them for repeated names.
     programs = DATABASE.programs_for_user(user['username']).records
     program_id = uuid.uuid4().hex
+    program_public = None
     overwrite = False
     for program in programs:
         if program['name'] == body['name']:
             overwrite = True
             program_id = program['id']
+            program_public = program.get('public', None)
```
ThomasKluiters, line 1577: Nitpicky; do we perhaps want to default to `False`?

TiBiBa: I choose to set it to `None` because this is how we currently handle public programs. The None value is simply not stored in the db (so `public` doesn't exist). I agree that `False` might be better, but this way it's consistent with our current implementation and we make sure that we don't break anything else. (We might simply check if the `public` value exists somewhere instead of a True/False value).

ThomasKluiters: Aha, I see! Makes sense :)

ThomasKluiters (approved):

Issue #1813, opened 2022-01-26
[BUG] types for comparison equal & not equal


juliabolt:
**Describe the bug**
It seems the types allowed for equality (==) are only integer, float and string, while for not-equal (!=), also lists are possible.

**Paste the Hedy code & level**
level 14
In hedy.py from line 107 on, as the comment says, there is a list of commands and their types per level. 
In hedy.py line 126: Command.equality: {1: [HedyType.string, HedyType.integer, HedyType.float]},
In hedy.py line 148: Command.not_equal: {14: [HedyType.integer, HedyType.float, HedyType.string, HedyType.list]}
Maybe this is a mistake?

**Expected behavior**
I expect both comparisons will be possible to use for the same types, so both can compare lists or both cannot.

Felienne:
Thanks @juliabolt!

@boryanagoncharenko this is one for you to dive into I think?

boryanagoncharenko:
Thanks for the report @juliabolt

Let's discuss the what the desired state is. I completely agree that from a programming perspective the 2 operators must be symmetrical. However, with Hedy there are a couple of things to take into account:

- The equals operator is introduced in level 5. Its syntax uses the word `is` and it transpiles to `==` (which is probably confusing for people with Python background, but great for people with no prior programming knowledge). At level 13, the syntax is relaxed to include `==`, which also transpiles to `==`.
- The not equals operator is introduced in level 13. Its syntax is `!=` and it transpiles to `!=`.

@Felienne given the above, how should the operators behave in levels [5..12] and [13..]? I could disallow lists for levels [5..12] and allow lists for both operators in levels [13..]. Alternatively, I could complete allow/disallow lists for all levels.

boryanagoncharenko:
Confirmed with @Felienne that lists will not be supported in comparisons until level 14.

PR #1962, opened 2022-02-14
Allow equality operator to compare lists starting from level 14 #1813

boryanagoncharenko:

< Type _list_ was not allowed to be used with the equality operator. With this change, lists can be compared for equality starting from level 14. >

**Description**

< Add support for lists in equality comparisons starting from level 14 > Use present tense without a subject to describe your PR, for example: "Adds translations of levels 1 to 12 to Polish"

**Fixes < #1813  >**

**How to test**

Run Hedy locally and navigate to level 13. Run the following program and ensure that you get an error message saying what is wrong and giving suggestions how to change the code:
```
a = 1, 2
b = 1, 2
if a is b
    print 'yes'
else
    print 'no'
    
```
 Now navigate to level 14. The following program should not yield errors:
```
a = 1, 2
b = 1, 2
if a == b
    print 'yes'
else
    print 'no'
```

**Checklist**
  

Felienne (approved):

Felienne:
@juliabolt I am not sure whether you already have this in the semantics file?

PR #1973, opened 2022-02-15
[CHORE] Clean up app.py

### LINKED ISSUE #1971: [CHORE] Clean-up app.py
**Idea incl level**
app.py is getting crowded with functions and routes with different functionalities. The file is currently over 1850 lines long! It would be nice to clean this up a bit to make the project structure more readable. A few thinks that came to mind:

- Give all admin routes dedicated python file
- Give all program routes dedicated python file
- Give all quiz routes dedicated python file
- Re-structure authentication routes so all are stored in auth.py

This is a big clean-up task but will make the project more readable and accessible for new contributors. Will pick this up somewhere this week!

TiBiBa:
**Description**
In this PR we clean-up the rapidly growing code in app.py. Routes of the same category are stored in a dedicated python file instead of all together in app.py. We create new files for admin, quiz and programs related routes. We also clean-up the quiz code as for some reason the functionality was divided in app.py, quiz.py and quiz_icons_svg.py. We center all this functionality within `quiz.py`. We also remove some redundant and unused functions.

**Note**
I think this is good to go for now as it does some significant clean-up on the codebase. But, I think on the long-run it might be useful to think about doing a bigger re-structure to make sure the code stays structured as the codebase is getting a big out-of-hand.

**Fixes #1971 **

**How to test**
This is only a clean-up task, if nothing is broken it works.

**Checklist**
  

website/auth.py
```diff
@@ -1,4 +1,5 @@
 import os
+import uuid
```
boryanagoncharenko, line 2: This looks like an unnecessary import

boryanagoncharenko (approved):

website/auth.py
TiBiBa: Good catch! Removed!

Felienne:
Ai this now has conflicts @TiBiBa  😭

TiBiBa:
Will take a look!

TiBiBa:
> Ai this now has conflicts @TiBiBa 😭

This seems to have become a very complex merging issue. Will try to fix it but it might be best to start the clean-up over. A lot of functions that switched files and dependencies are altered on main. I'm afraid I will break something when merging them.

TiBiBa:
@Felienne Merging is fixed! I did some testing and nothing seems broken, but I think we need to do some extra testing to make really really sure that we don't break anything with this re-structuring.

Felienne:
Merged! Maybe you want to check on alpha if it is all in order?

TiBiBa:
Yes, will do some testing on alpha!

PR #1980, opened 2022-02-15
Adds fr keywords

Felienne:
Adds French keywords

**Fixes #1979**

**How to test**

Tests added for weird cases like keywords with spaces.

**Checklist**
  

Felienne:
Maybe you can have a look at whether this is correct @thjazi39?

thjazi39:
ok, I confirm the validations!
I think that if `until` is stuck, while `while` and `if` are not, it's because in one case, we have an apostrophe, while in the 2 other cases, they are 2 distinct words.
If this is done on purpose, I think it's a good idea, and if not, we could standardize it by putting spaces (if supported by hedy)

Felienne:
> ok, I confirm the validations! 

> I think that if `until` is stuck, while `while` and `if` are not, it's because in one case, we have an apostrophe, while in the 2 other cases, they are 2 distinct words. If this is done on purpose, I think it's a good idea, and if not, we could standardize it by putting spaces (if supported by hedy)

If you leave a comment on the code (go to :Files Changed) it will be easier for me to see what you refer to.

I am not sure I follow, we don't have `until`? 

In any case, spaces are supported! Where do you suggest we add them?

thjazi39:
Sorry, I had translated the keywords, but I didn't think it would be a problem for the understanding (I modified the messages to be better!)

And while trying to comment the code, I realized that the `jusqu'à` has been changed to `à` so in the end, there is no problem!

On the other hand, it means that the command `_TO` has the same translation as `_TO_LIST` , and I don't know if this can be a problem

PR #1981, opened 2022-02-15
fix #1978

### LINKED ISSUE #1978: [CONTENT] error quiz 9/10
When translating quiz 9, I noticed several errors:
The first one is a terminal error (10 instead of 3)

and when I wanted to test the code (to know if the upper bound is included) I realized that the for does not exist in level 9!

I think there is a level error for this quiz (the next questions are also about `for` commands) and this command appears only in level 10

thjazi39:
repair of the loop error
and shift the numbers of the last 2 quizzes (9->10 and 10->11)

### DIFF
--- a/coursedata/quizzes/en.yaml
+++ b/coursedata/quizzes/en.yaml
@@ -1170,11 +1170,11 @@ levels:
             hint: "Mind the order of the sentences."
             question_score: 10
 
-    9:
+    10:
         1:
             question_text: "What will be the output from this code?"
             code: |-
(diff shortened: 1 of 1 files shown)


Felienne:
I made the name here a bit more descriptive @thjazi39 so it is easier for others to see what this does (now and in the future) Please also try to use such conventions

Felienne:
Also also if you write "fixes #1978" (instead of fix) the issue will be linked automatically and closed when we merge this PR!

PR #1991, opened 2022-02-15
Make use of keyword is instead of '='

### LINKED ISSUE #1986: [CONTENT] Level 5 example uses = instead of `is`
We is = and not `is` which the tests do not catch because the code also has `_`.

**Hedy code & level**
Level 5: rock,  paper, scissors
```
options = rock, paper, scissors
computer_choice is _
choice is ask What do you choose?
print 'you chose ' _
print 'computer chose ' _
if _ is _ print 'tie!' else print 'no tie'
```

in the first line the "=" should be replaced by "is"

grafik

DerSibi:
**Description**

Fixes issue #1986 by replacing assignment operator by keyword 'is'

**Fix for**

Fixes #1986

**How to test**

Check out branch, open level 5, adventure "rock, papers, scissors". Sample code now show keyword is in line 1

**Checklist**

### DIFF
--- a/content/adventures/de.yaml
+++ b/content/adventures/de.yaml
@@ -1187,7 +1187,7 @@ adventures:
                     ## Example Hedy code
 
                     ```
-                    options = rock, paper, scissors
+                    options is rock, paper, scissors
                     computer_choice is _
                     choice is ask What do you choose?
                     print 'you chose ' _
--- a/content/adventures/en.yaml
+++ b/content/adventures/en.yaml
@@ -1117,7 +1117,7 @@ adventures:
                     ## Example Hedy code
 
                     ```
-                    options = rock, paper, scissors
+                    options is rock, paper, scissors
                     computer_choice is _
                     choice is ask What do you choose?
                     print 'you chose ' _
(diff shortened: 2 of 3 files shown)


DerSibi:
Hi,

I intentionally did not fix it in additional language files. Most of them I cannot read at all, so I wasn't sure whether the issue is present there. How about German? I would upload the fix with the next adventure translation anyway. Or should I add it to this fix?

Felienne:
> I intentionally did not fix it in additional language files. Most of them I cannot read at all, so I wasn't sure whether the issue is present there. 

I will have a look at the other langs (some are a bit behind and might not have all snippets that En/Nl/Deu have)

> How about German? I would upload the fix with the next adventure translation anyway. Or should I add it to this fix?

Ah that is a good question! I think there is something to be said for both, but I would do it in this PR so it links nicely to this issue.

Felienne:
Hi @DerSibi! The process I used to check other languages is to search for rock, since the keys are the same in all languages. Then I scrolled to level 5 and eyeballed for the =. It seems all other langs are fine.

Thanks again for this!

PR #2001, opened 2022-02-16
[UI FIX] Re-directs /explore to /login when not logged in

### LINKED ISSUE #1977: [UI idea] re-direct users to /login when trying to visit /explore
**Idea incl level**
Currently when a non-logged in user is trying to visit the /explore page they get a "non-authorized" error. It might be better to re-direct them to the login page with some message "you need to log in to access this page". Another solution might be too simply throw a modal.alert() "You need to log in to access this page". The second solution might be more user-friendly as they don't have to leave the current page.

TiBiBa:
**Description**
Currently we throw users a "not authorized" error when trying to visit the explore page without being logged in. This works fine, but it suggest that for some reason they are not allowed at all to visit that page. This is incorrect as they can visit the page as soon as they are logged in. In this PR we redirect the user to the /login page instead of throwing an error.

**Fix for**
This PR fixes #1977.

**How to test**
Make sure you are not logged in and attempt to visit the /explore page. Verify that you are correctly re-directed to the login page.

**Checklist**


boryanagoncharenko (approved):
I think it is better to redirect to the /explore page after the user has successfully logged in. In the current case I am redirected to the /programs page.

PR #2017, opened 2022-02-17
[FEATURE] Enables users to set a preferred keyword language on their profile

### LINKED ISSUE #2008: [UI idea] Enable users to set preferred keyword language in profile
**Idea incl level**
With the support of keywords in different languages it would be nice if we enable users to select a preferred language. This way we don't have to show the dropdown overlay menu on each editor and simply always show language. We might have to think about some issues:

- What do we do with all starting code / adventures -> currently all with English keywords
- Do we parse them on each retrieval -> This is really expensive

Do you have a suggestion how we can efficiently implement this? @Felienne

TiBiBa:
**Description**
In this PR we implement the structure to enable users to set and change a preferred keyword language. With the implementation of multi-language keywords users are currently able to switch between their chosen language (if supported) and English through a language switcher on the code editor. With this PR users can set a fixed preferred command language, which is stored on their profile.

**Fix for**
This PR addresses, but does not fully fix #2008.

**How to test**
This is mostly a back-end structural PR and does not influence the user experience much. Navigate to the signup page and notice the new input field on `preferred keyword language`. Verify that this field is working as well as on the profile page where the user settings can be updated. If a language is chosen the language switcher on the front-end is no longer shown. 

**Checklist**


TiBiBa:
@Felienne This PR is ready for review! It creates the structure enabling users to select a preferred keyword language and for us to keep track of this value in the back-end. This should be the starting-point for the multi-language adventures.

Felienne:
Hi @TiBiBa!

I just tested this but it does not seem to work entirely correct yet. I can set a language in my profile and that does remove all switch buttons, however, the code snippets remain in English:

image

I tried setting my language to Spanish and to Dutch bit the snippets remain to use English keywords for both settings.

Interestingly enough, when I set my language to Spanish, and then log out... the options become Es and Nl:

image

Felienne:
Ow maybe the snippets are not yet meant to be shown in the set language? Reading the PR text again I am not sure it was supposed to already work, but at least the weird logging out behaviour is not what we want.

Felienne (changes requested):
See above

TiBiBa:
> Ow maybe the snippets are not yet meant to be shown in the set language? Reading the PR text again I am not sure it was supposed to already work, but at least the weird logging out behaviour is not what we want.

The goal of this PR was to only implement the back-end structure for storing the preferred keyword language as we still have to figure out an efficient way to show all code snippets in the selected language (parsing is expensive, translating complex).

TiBiBa:
The switcher should be fixed! The back-end structure is complete (users are able to set a keyword language), but the snippet are still all loaded in English. I think we need to have some discussion and a separate issue to solve that problem.

Felienne:
> The switcher should be fixed! The back-end structure is complete (users are able to set a keyword language), but the snippet are still all loaded in English. I think we need to have some discussion and a separate issue to solve that problem.

Ok, got it! Then this can be merged (will do so later today)

For the front-end implementation, let's discuss indeed, I have some ideas but they might not scale well :)

Felienne:
Yes the weirs es/nl issue is gone now!

One question, why are Arabic and French not the list, we have keywords for that as well (although not 100% complete for Ar)

image

TiBiBa:
> Yes the weirs es/nl issue is gone now!
> 
> One question, why are Arabic and French not the list, we have keywords for that as well (although not 100% complete for Ar)
> 
> image

Didn't know that! We have a variable `ALL_KEYWORD_LANGUAGES` in `app.py` (line 72). We can simply add them there and it should all work automatically.

Felienne:
> > Yes the weirs es/nl issue is gone now!
> > One question, why are Arabic and French not the list, we have keywords for that as well (although not 100% complete for Ar)
> > image
> 
> Didn't know that! We have a variable `ALL_KEYWORD_LANGUAGES` in `app.py` (line 72). We can simply add them there and it should all work automatically.

Ah good to know, I would have added it there! (adding keywords is still manual work by me and that shows haha)

Felienne (approved):

PR #2020, opened 2022-02-17
[CLEAN-UP] Clean up template globals and g object

TiBiBa:
**Description**

Main goal:

Data used between requests should **not** be stored in the g object!

**Fix for**

**How to test**

**Checklist**


TiBiBa:
We currently store a lot of translations on the `g` object. As this object is thrown away and created again on each request this is expensive. It would be better to store more attributes, such as language as well on a session cookie. As the client-side session cookie can "only" be 4kb, we might want to look into a server-side cookie.

TiBiBa:
Server-side sessions:

https://flask-session.readthedocs.io/en/latest/

PR #2022, opened 2022-02-17
[CLEAN-UP] Clean-up app.py and remove g object dependency

TiBiBa:
**Description**
In this PR we do some small clean-up work to further improve the unnecessary parsing and initializing of values. Currently we use the Flask `g` object to store multiple user-relevant variables such as language, keyword language and translations. However, the `g` object only lives for one request. Afterwards it is deleted and we have to initialize it again for each new request. This is cumbersome and doesn't make any sense as a value such as language (and the translations) is not likely to change between requests.

Here we mark a starting-point of removing some technical dept and shifting to a more efficient way of keeping track of user data. We store the language and keyword language inside the user session and prevent het setting of the value at each request.

**Fix for**
This PR is (another) partly solution for #1179.

**How to test**
This is a clean-up task, if nothing is broken it works.

**Checklist**


TiBiBa:
Close this as the merging errors are getting to complex, will pick this up somewhere later.

PR #1946, opened 2022-02-13
Provide German translation of 'dishes' adventure

DerSibi:
**Please fill out this template by replacing all content between < >**

Provide German translation of 'dishes' adventure

**Description**

Adds a German translation for every level in the 'dishes' adventure.

**Fixes < issue or discussion number >**

< Always link the number of the issue or of the discussion that your pr concerns. >
Notable exception: adding content may be done without an accompanying issue 

**How to test**

< If this is a UI change, describe how to run your code to see it in action. See this https://github.com/Felienne/hedy/pull/880#issue-1016304308 for an example >
< If this is a language change, add a few tests showing the difference >

**Checklist**
Done? Check if you have it all in place using this list*
  
- [ ] Describes changes in the format above (present tense)
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

* If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #2035, opened 2022-02-18
Translations/de/adventures/dishes

Felienne:
Trying to redo #1946


Felienne:
I think this fixes the adventures again @DerSibi, but I had to do it manually :( Let me know if you find other issues!

DerSibi (changes requested):
Suggest that one change to avoid regression. Apart from that, it looks good. Thanks for taking care!

PR #1948, opened 2022-02-13
[CHORE] Rename coursedata to content

ThomasKluiters:
**Please fill out this template by replacing all content between < >**

< Provide a one line summary of your changes in the Title of the PR >

**Description**

< Changes in detail > Use present tense without a subject to describe your PR, for example: "Adds translations of levels 1 to 12 to Polish"

Fixes #1931

< Always link the number of the issue or of the discussion that your pr concerns. >
Notable exception: adding content may be done without an accompanying issue 

**How to test**

< If this is a UI change, describe how to run your code to see it in action. See this https://github.com/Felienne/hedy/pull/880#issue-1016304308 for an example >
< If this is a language change, add a few tests showing the difference >

**Checklist**
Done? Check if you have it all in place using this list*
  
- [ ] Describes changes in the format above (present tense)
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

* If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #2036, opened 2022-02-18
[BUG FIX] Quiz loads correctly again

TiBiBa:
**Description**
With the merging of #1948 the `coursedata` folder was renamed to `content`. But not in all places... In `quiz.py` the function searching for the correct yaml file was still looking for a coursedata folder, making the quiz return an error. In this PR we fix this by changing the path to the correct folder.

**Edit**
Just noticed that the fix was indeed correct in #1948. Will have to take a look which PR reversed this. It might be due to a combination of #1948 and #1973 where in one PR the path was set as correct but in the other PR the quiz was moved to the `/website` folder.

**Fix for**
No corresponding issue

**How to test**
Verify that the quiz is working again correctly.

### DIFF
--- a/website/quiz.py
+++ b/website/quiz.py
@@ -62,7 +62,6 @@ def routes(app, database, achievements):
 
             # Reading the yaml file
         questions = quiz_data_file_for(g.lang, level_source)
-        print(questions)
         if not questions:
             return no_quiz_data_error()
 


Felienne (approved):

TiBiBa:
> Looking at the history I think it was #1973?
> 
> image

Yeah I think it was! I was probably working at the `app.py` clean-up at the same time @ThomasKluiters was working on re-naming the folder. As I moved the file in #1973 no merging error occurred (as it was simply removed from the main folder). Luckily it's fixed!

PR #2040, opened 2022-02-18
leading spaces removed from english texts en quizzes

Mark-Giesen:
Unsure about bidirectional sync between these yaml files and weblate changes.


Felienne:
Let's see what this does on weblate :)

PR #2044, opened 2022-02-19
Adding a preprocessor to remove comments before parsing

### LINKED ISSUE #1705: [BUG] Comments are not allowed in all places
We wrote comments to be mixed with other comments but not to be used *in* commands, like this:

```
choices is ['batu', 'kertas', 'gunting']
print 'Pilihlah'
print 'batu, kertas, gunting'
userc is ask 'Pilihan anda?'
print 'Pilihan anda: ' userc
computerc is choices at random
print 'Pilihan komputer: ' computerc
if userc is computerc:
    print 'SERI'
elif userc is 'batu' and computerc is 'kertas':
    print 'Komputer menang' #prints something <------ does not work now
else:
    print 'Komputer menang'
```

I guess rather than adding comments in the grammar as we do now, it would be easier to write a little preprocessor that removes all code on a line after the #. We already have preprocessing code in place, so that is not all too hard and we don't care about the comments anyway.

Maybe a fun issue for @racertop(when you are done with the input stuff, of course!)?

jpelay:
**Description**

Adds a small function to the preprocessor to be able to parse programs with comments in any place. Up until right now we could only parse comments in the beginning of line. Maybe my solution is a bit slow, since I traverse all the characters in the input string. I tried to use a regex, but it didn't work because # (the character that starts a comment) can be inside of a string. Also, 

Fixes #1705 
**How to test**
I added a few tests in for normal programs and comments after strings.


jpelay:
Why are the tests failing when the run fine in my local? :<

Felienne:
Haha is this why you were talking about regexes @jpelay? 🤣 I will have a look at the failing tests for you. 

I am also not sure this is the best way for efficiency, maybe it is easier to add it to the grammar? In any case this is a good start!!

jpelay:
> 

Yes XD the thing is that I tried with regexes, but since regexes don't manage nesting it was very hard to handle the case when the # is inside a string. I tried with ```(<!')#(<!?.*')``` but this didn't work for those strings that had characters before the #, and look behinds don't accept cuantifiers inside. So basically I wrote a new parser for that rule :p but you're right this is not very efficient!!!! And yep, adding it to the grammar is a good idea I'll try with that.

Felienne:
> And yep, adding it to the grammar is a good idea I'll try with that.

Yeah I have not done any trying or deep thinking haha, but I think making a rule like this:

`comment: "#" text`

should work, since text is more or less anything we support. Maybe we need to explicitly need to also allow , and stuff we disallow in text.

And then we can make `command` in `program` be `command comment?`

Let me know if you need further pointers!

Felienne:
Closing this to go with a grammar-based approach

PR #2049, opened 2022-02-21
[FIX] Enables developers to create css file with all Tailwind classes

### LINKED ISSUE #2039: tailwind/generate-development-css doesn't work
**Describe the bug**

The `build-tools/heroku/tailwind/generate-development-css` script generates an empty css file, instead of a CSS file with all the Tailwind bells and whistles.

It looks like it's stripping all classes, because they're unused, because it's not configured with any files to look at for classes to retain.

This means you have to use `generate-css` to generate the CSS, but that will only retain classes that are actually currently being used.

The upshot of all this is that if you want to experiment with new Tailwind classes, you must:

* First write HTML
* Then regenerate Tailwind CSS
* Then refresh your browser (hard-refresh so it doesn't cache!) to see the results of the new class

It would be better if the checked-in `generated.css` always contains the full Tailwind distribution with all classes. Yes, it'll be large, but at least you can develop to your heart's content.

To make that work, we need to fix `generate-development-css` and `tailwind.full.js` though...

TiBiBa:
**Description**
Before updating to Tailwind v3 the `generate-development-css` file automatically generated all possible Tailwind classes. Making front-end development easier as we don't have to manually generate the .css files after adding classes. However, with the update to Tailwind v3 the `purge` option was deprecated as there was no longer an option to generate the full css file.

As pointed out by @rix0rrr the current script doesn't work anymore. In this PR we implement a work-around to still enable us to generate the complete css file. As this is no longer supported it is a bit hacky-tacky as discussed here:
https://github.com/tailwindlabs/tailwindcss/discussions/6557

**Fixes**
This PR fixes #2039

**How to test**
Run the `generate-development-css` script and verify that a css file is generated containing all Tailwind classes. Note that we copied the `tailwind.config.js` file to `tailwind.full.js` as for some reason the container width is messed up when we use the original full config file.

**Checklist**
  

build-tools/heroku/tailwind/tailwind.full.js
```diff
@@ -1,350 +1,36 @@
...
-  content: [],
-  presets: [],
-  darkMode: 'media', // or 'class'
+  content: [
+      '../../../templates/**/*.html',
+      '../../../main/**/*.md',
+      '../../../coursedata/**/*.md',
+      '../../../static/js/**/*.js',
+  ],
+  safelist: [
```
rix0rrr, line 9: Thanks for picking this up @TiBiBa! 

Question: why do we need to rewrite this whole file? Isn't it enough to add the `safelist` option to the existing config, and that's enough?

Also, wouldn't it make sense to commit the `generated.css` with all the options?

rix0rrr: (To take a step back maybe: do you agree with the premise of issue #2039? 😋)

TiBiBa: From some reason the `tailwind.full.js` file contains some changed to container width. So when generating the complete css file the layout was no longer consistent. I could figure out what caused it so copied the original config and added `safelist` to make sure all classes are generated.

TiBiBa: The changes between Tailwind v1 and v3 where pretty big, so we had to some manual tweaking when switching to v3 to make sure the UI stayed the same. However, this does not seem to work well with `tailwind.full.js`. So this seemed the most suitable solution. Not ideal, but not sure what would be better solution to still generate all classes.

rix0rrr: Right. I thought I regenerated using v3 but maybe I didn’t.

check in the generated.css though?

TiBiBa: Yes only the full generated config file from Tailwind v3 does not align with our UI. The color schemes, fonts and some paddings, margins etc. are different. So in the tweaked config file we reversed these changes by setting these settings to the values of the v1 config file. As these are not set in the full config file the layout is different when using that one to generate the css.

rix0rrr (changes requested):
Thanks for this change.

It needs the full, development version of `generated.css` checked in before it's useful though.

build-tools/heroku/tailwind/tailwind.full.js
rix0rrr: 👍

TiBiBa:
> Thanks for this change.
> 
> It needs the full, development version of `generated.css` checked in before it's useful though.

Just added!

rix0rrr (approved):

PR #2052, opened 2022-02-21
[BUG FIX] Catches error on translating keywords

### LINKED ISSUE #2050: [BUG] _ in code to be translated leads to error
We can't translate code with an _ and that gives an ugly error (and some other things can cause that too like misspelled keywords)

Should we wrap the translate button in a try-catch?

image

TiBiBa:
**Description**
In this PR we fix an issue where the page would return an unclear error when translating a code snippet that for some reason did not translate successfully. We add a try/catch and return an nice error message instead.

**Fixes**
This PR fixes #2050 

**How to test**
Navigate to level 4, select the "rock, paper, scissor" adventure and try to translate. This should no longer result in a page crash and return a nice "something went wrong when translating" error.

**Checklist**
  

Felienne (approved):

Felienne:
Wow that was a speedy fix!

PR #2054, opened 2022-02-21
[FEATURE] Enables dynamic keyword changing from level defaults

TiBiBa:
**Description**
In this PR we start with the implementation of supporting level defaults and adventures with localized keywords. Currently we only support the manual switching of keywords by the user. However, with the implementation of the feature to enable users to set their preferred keyword language it would be nice if all code snippets are automatically shown in this language. This PR is a wip to discuss if we are on the right track.

Currently the level defaults for `nl` and `en` are adjusted to support the dynamic changing of the keywords. We replace each keyword with a placeholder, for example: `{print} hello world!`. On runtime we replace the placeholder with the current keyword language. If there is a keyword language set in the profile we hide the switcher as well.

The `test_z_defaults` have been a bit bucket-fixed to make sure they pass. We now parse the code as well in the test files to make sure the code snippets contain valid code and pass the tests. We might have to re-write the tests somewhere in the future to make sure it all keeps working when we expand on this work.

**How to test**
Make sure you have a preferred keyword language set in your profile. To notice the change set it too `nl` and navigate to level 5. Notice that the level default code is correctly parsed to contain dutch keywords.

**Checklist**
  

TiBiBa:
Things to add / note-to-self:

- Keywords in 'cheatsheet' are not correctly parsed (yet), but they do already contain placeholders as they are extracted from `level-defaults`
- Adventures are still to-be fixed
- We have to run issues on dutch parsed keywords (such as 'op willekeurig')

TiBiBa:
We have to think about een efficient way to implement the keyword language. Maybe remove the editor switcher all together? Currently it is quite complex because we either retrieve a value from the user profile or it is automatically set to English. But in the second case we also have to show the switcher iff the native language is also support with localized keywords. 

To make if even more complex: we can only parse English and the current profile language. So any other language should not be an option to select on profile settings. I think the most straightforward solution would be:

- Automatically set keyword language to English
- Remove the switcher (non-logged users can't switch)
- Let users set commands language to:
   - English
   - Their current profile language (if supported, otherwise: hide option)

What do you think? @Felienne

Felienne:
> make sure the switch changes when copying a language that is not English

I am not sure what this means, can you elaborate?

TiBiBa:
> > make sure the switch changes when copying a language that is not English
> 
> I am not sure what this means, can you elaborate?

For example when the user set keyword language is "es" all level-defaults and adventures will be in "es". But it might be the situation that while programming the users sets the main editor to either "en" or "es". So we have to make sure that the switcher on the main editor always changes (visually) to the keyword language of the snippets from the defaults/adventures.

TiBiBa:
I think this is ready for review! We fixed all level-defaults and adventures for languages that support localized keywords by replacing the keywords with placeholders which are filled on runtime. It all works fine for now, but we have to think about future work on:

- Efficiently store the current keywords list (we keep retrieving the yaml)
- Re-write the level-defaults structure -> separate actual level-defaults and commands (cheatsheet)
   - Together with this: simplify structure as the current dict->list->dict structure is prone to errors
- Re-write adventure retrieval to dedicated Class
   - For some reason this is current done in `app.py` while there is partly(?) a structure in `Hedy_content.py`

These aren't crucial issues for now. But, it would be nice to give the overall structure on retrieving these yamls an update so the code base is better understandable, adjustable and more efficiently accessible when parsing keywords.

Felienne (approved):
thanks for this tour de force!! 🥳

PR #2024, opened 2022-02-18
[FEATURE] Adds 'agree to terms' tickmark to sign-up page

TiBiBa:
**Description**
In this PR we perform an overall UI re-style of the signup page and implement an additional tick mark for agreeing with the privacy agreement. Real-time form validation is implementation using the CSS `input:valid` attribute. A very small re-write is performed in auth.py to better differentiate between functionality within functions and make clear where the mailinglist and teacher account filtering is done.

**Fix for**
This PR fixes #2023 

**How to test**
Navigate to the signup page and notice the new UI and the new tickmark at the bottom of the page. Verify that the tickmark is not required and (when not in testing) an e-mail is sent to the admin when the tickmarked is checked on signup.

**Checklist**

If you're unsure about any of these, don't hesitate to ask. We're here to help!
- [ ] Links to an existing issue or discussion (if not, create an issue first)
- [ ] Describes changes clear in the format above (present tense, no subject)
- [ ] Has a "how to test" section
- [ ] Only one thing is done in this pull request (specifically please try to refrain from mixing textual changes to the yamls from code changes)

PR #2064, opened 2022-02-22
[BUG FIX] Adds accidentally removed keyword language select on signup

TiBiBa:
**Description**
Due to the merging of #2024 we accidentally deleted the earlier implemented option select for a keyword language. In this PR we re-implement and improve this option. The keyword select option is only shown if a non-english keyword supported language is selected. This is done dynamically on language input change. All other keyword languages expect English and the selected language are hidden. When the select language does not support localized keywords the option is hidden and English is automatically selected. We also add a small fix where the keyword language for accounts created through multiple-account creation is automatically set to the teacher language.

**Note**
This dynamic showing/hiding of options does **not** work on Safari. If the user selects a keyword language that is neither their own language or English we return an error from the back-end. I think this is fine for now.

**Fixes**
No related issue.

**How to test**
Navigate to the sign-up page and select a keyword-supported non-english language. Notice the new select option that shows and verify that it works as expected. Also verify that (when on Safari) an error occurs if a non-english non-selected language is chosen. In any other browser only the two valid options (English and the select language) are shown.

**Checklist**
  

Felienne:
Nice that the preferred keywords only appears when the language selected has keywords translated!

Felienne (approved):

PR #2072, opened 2022-02-23
Adds extra placeholder for advance button to support Hindi language

### LINKED ISSUE #1926: [BUG] Extra field needed for Hindi to make advance button look proper
Currently, we have two fields for the advance button:

`{{ ui.advance_button }} {{ next_level }}`

https://github.com/Felienne/hedy/blob/ead6cadb5f68e6b0175318f75dc440024d6fb208/templates/incl-editor-and-output.html#L102

But some languages like Hindi need a different structure, something like: "goto 2nd level" where the number needs to go in between. We need to add an extra field in the UI text, make it empty for all non-Hindi languages and then render it properly. In the new Hindi translation (not yet in a PR) the translators will put a _ where the number should go so I know where the snap the text in half.

TiBiBa:
**Description**
In this PR we add an extra placeholder _after_ the level variable on the advance/regress button to switch between levels. This to support languages that don't automatically show the number and the end of the string. For languages that do we simply level the placeholder empty.

**Fixes**
This PR fixes #1926

**How to test**

**Checklist**
  

Felienne:
Great, this works for this case! 

However, this does not scale very nicely with more languages and forms, in general. I guess once we move to a better localization, we can use placeholders in strings which would be a bit prettier, like we do for error messages. Something like:

`advance_button: "Go to level {level}"` for English  and
`advance_button: स्तर पर {level} जाएँ ` for Hindi

We can either do something like that here now or leave it and take it into account when moving to gettext, whatever you prefer!

TiBiBa:
> Great, this works for this case!
> 
> However, this does not scale very nicely with more languages and forms, in general. I guess once we move to a better localization, we can use placeholders in strings which would be a bit prettier, like we do for error messages. Something like:
> 
> `advance_button: "Go to level {level}"` for English and `advance_button: स्तर पर {level} जाएँ ` for Hindi
> 
> We can either do something like that here now or leave it and take it into account when moving to gettext, whatever you prefer!

Good point, fixed it with a placeholder!

Felienne (approved):

PR #2081, opened 2022-02-23
Add runtime validation and correction to the turtle commands transpiled code #1983

### LINKED ISSUE #1983: [BUG] A valid public Hedy program hangs
**Describe the bug**
I encountered an interesting program in the production Explore page and was curious to see what happens when I run it. The result is that the Hedy page hangs and after a while the browser politely offers to exit the page. Please note that this is a public program which means that it is easily accessible to other users.

Level 1
```
forward 999999999999999999
```

Screenshot 2022-02-15 at 15 35 53

**Expected behavior**
Probably we should give an error when forward is given very large values. However, we should also determine what constitutes as "very large values".

- What computer you are using? Mac
- What browser you were using? Chrome

boryanagoncharenko:
**Please fill out this template by replacing all content between <>**

Add runtime validation and correction to the turtle commands transpiled code

**Description**

- When the turn or forward commands are passed large positive or negative numbers as arguments, their input is capped to a number that does not cause failure. The check is done at runtime which means that the cap should work even when the argument is passed via user input.
- When the turtle commands receive a value that cannot be converted to a number, an error message is displayed. This error message cannot be translated and it is only in English.

**Fixes #1983 **

**How to test**

Run Hedy locally and execute the following programs:
- In level 1 run ```forward 99999999```, ```forward -99999999999```, ```turn 99999999999``` and ```turn -9999999999```. Since the large numbers will be capped to 600 and -600, respectively, the turtle will probably go out of your screen. However, your browser should not hang and the program should only take a couple of secs before it completes successfully.

- In level 4 run:
```
input is 999999999, -999999999
turn input at random
forward input at random
```
Again, the turtle will probably get out of your screen but the program should not hang and it should complete successfully.

- In level 4 execute:
```
input is not, a, number
turn input at random
forward input at random
```
Ensure that you get an error. Is the error clear? Does it give enough directions to fix the program?

**Checklist**
  

hedy.py
```diff
@@ -1235,6 +1230,24 @@ def turn(self, args):
...
+    def make_forward(self, parameter):
+        return self.make_turtle_command(parameter, Command.forward, 'forward', True)
+
+    def make_turtle_command(self, parameter, command, command_text, add_sleep):
+        variable = self.get_fresh_var('trtl')
+        transpiled = textwrap.dedent(f"""\
+            {variable} = {parameter}
+            try:
+              {variable} = int({variable})
+            except ValueError:
```
Felienne, line 1245: Ow smart way to pass runtime errors to the front-end! This can be used on other places too I think (will have to test it more extensively)

Felienne:
Well done, error message is surely clear!

Felienne (approved):

PR #2087, opened 2022-02-24
[FIX] correctly displays copy button on left of editor when RTL language

### LINKED ISSUE #2082: [BUG] Copy button overlays Arabic code because it is on the right hand side
Copy button overlays Arabic code because it is on the right hand side:

image

Same issue that we had with the language switcher (#2063) so maybe it can be solved in the same way?

TiBiBa:
**Description**
In this PR we fix a small bug where the copy button on code snippets was always placed in the top-right corner. However, with RTL languages this overlays the text. We fix this by styling the button using the Tailwind `ltr:` and `rtl:` classes.
 
**Fixes**
This PR fixes #2082

**How to test**
Navigate to the code page and there should be no noticeable difference for LTR languages. Switch to Arabic and verify that the copy button is now correctly displayed in the top-left corner of the code snippets.

**Checklist**
  

Felienne (approved):
Works!

image

PR #2088, opened 2022-02-24
[BUG FIX] Password reset works again

### LINKED ISSUE #2085: [BUG] Reset password doesn't work with reset link
**Describe the bug**
The reset password link works fine, but for some reason we always receive a 'passwords don't match' error. Making it impossible to reset the password. Related to this another weird issue: when clicking the password reset link we open the "reset password" page, but if logged in the current user stays logged in. We should always destroy the current user session if the link is pressed (which makes more sense).

Schermafbeelding 2022-02-24 om 10 46 40

**Expected behavior**
We can reset our password using the password reset link.

TiBiBa:
**Description**
In this PR we fix an issue with password reset where the "passwords do not match" error was always returned. This was due to the repeated password not being sent to the server, causing this error. (This should be added to the e2e tests!). We also change the UI to match the new form UI of the website.

**Fixes**
This PR fixes #2085

**How to test**
We can only test this on Alpha. @Felienne can you validate that the code is correct and merge to Alpha? Than I can test it there.

**Checklist**
  

Felienne:
> (This should be added to the e2e tests!)

Shall we make an issue so we don't forget :)  I know we have an open issue already that it more generic but it is nice to keep it concrete!

TiBiBa:
> > (This should be added to the e2e tests!)
> 
> Shall we make an issue so we don't forget :) I know we have an open issue already that it more generic but it is nice to keep it concrete!

I just took a look, there is already a correct e2e test for this. We simply messed it up on the front-end...

static/js/auth.ts
```diff
@@ -333,6 +340,10 @@ $ ('.auth input').get ().map (function (el) {
 });
 
 // We use GET /profile to see if we're logged in since we use HTTP only cookies and cannot check from javascript.
+// Todo TB -> Why do we do this ?!
```
Felienne, line 343:
```suggestion
// Todo TB Feb 2022 -> Why do we do this ?!
```

website/auth.py
```diff
@@ -734,6 +734,14 @@ def auth_templates(page, page_title):
         public_profile_settings = DATABASE.get_public_profile_settings(current_user()['username'])
         return render_template('profile.html', page_title=page_title, programs=programs,
                                public_settings=public_profile_settings, current_page='my-profile')
+    # Todo TB -> We have to clean this up (a lot!)
```
Felienne, line 737:
```suggestion
    # Todo TB Feb 2022 -> We have to clean this up (a lot!)
```

Felienne (approved):

Felienne:
Looks ok form what I understand :) Let's check on alpha!

PR #2093, opened 2022-02-24
[BUG FIX] Users are now correctly logged in and re-directed to landing page on verification

### LINKED ISSUE #2091: [BUG] Landing page re-direct on account validation doesn't work
**Describe the bug**
This bug is based on code analysis as I'm unable to verify this without deploying the fix to Alpha first. It seems that currently we only re-direct users to the landing page when verifying if the `verification-pending` value does **not** exist. This is exactly the opposite of what we want.

**Expected behavior**
When a user verifies their account they should be re-directed correctly to the landing page.

TiBiBa:
**Description**
Due to a bug our beautiful landing page was never shown to user and the validation process didn't log them in. This wasn't very user-friendly either. Due to this bug **users were not able to verify their account**. In this PR we fix this by successfully re-directing users to the landing page if they verify their profile and even re-direct them to the landing-page if they try to visit it manually. We also fix a missing image.

**Fixes**
This PR fixes #2091

**How to test**
It is hard to verify locally as you will need your verification token. I will give it a go; create a new account and look up the user in the database. It should contain a field called `verification_pending`. Copy the value and navigate to the following url:

`http://0.0.0.0:8080/auth/verify?username=<username>&token=<verification_code>`

You should automatically be logged in and re-directed to the landing page. Verify that the `verification_pending` value is removed from your user in the database. **This should be double-checked when deployed to Alpha**.

**Checklist**
  

Felienne:
wow what a pity this has not worked so far, we made it so lovely!!

It does seem to work now with the steps you describe:

image

PR #2095, opened 2022-02-24
Lock & Unlock

HanshaLeng:

< Add Locks to the adventures and levels, add a Navigation bar of levels>

**Description**

< add locks to the adventures and levels; Associate locks with users; unlock adventure when users run a code without errors; check keywords for commands; add a navigation bar for levels> 

**How to test**

< Register and log in to a user account, start learning from the first adventure in the first level; Run a code without errors, then the next adventure will be unlocked; ask commands for help when stuck in any adventure, if the keyword of command been used, the command will be considered practised; all adventures or commands have been practised, the end quiz will be unlocked; earn at least 60% score and the next level will be unlocked;>

**Checklist**
  

Felienne:
I added the generated css but it does not entirely look right, can you look into that:

image

Otherwise the merge seems ok so you can start testing this version and mix the error up to date with our main.

Felienne:
There is a code error too but that is probably something small:

image

HanshaLeng:
> There is a code error too but that is probably something small:
> 
> image

It seems your merge  added a redundant piece of code from an older version, it cause the fault

TiBiBa (changes requested):

PR #2101, opened 2022-02-25
[CHORE] Cleaned up auth.ts, re-wrote routes to the back-end

### LINKED ISSUE #2089: [CHORE] Clean-up auth.ts, re-write stuff to the back-end
**Description**
Currently we perform a lot of actions on the front-end what should actually be executed on the back-end. For example, when visiting a recover password link we manually retrieve the arguments on the front-end using some complex function. Then we add these arguments to an object and sent them back with a POST when the user actually resets. This doesn't make any sense. 

This is just an example of some coding that should be removed and the functionality should be fixed in the back-end. Resulting in a cleaner (and safer!) codebase and website.

### LINKED ISSUE #2099: [BUG] We can be logged in and reset a password at the same time
**Describe the bug**
Currently we are able (again) to correctly reset a password with the link set by e-mail. However, if a user is logged in they stay logged in while resetting the password of a possible other account. This is very undesirable and unclear behavior.

Schermafbeelding 2022-02-25 om 10 12 47

**Expected behavior**
When visiting the `/reset` we should force log-out the correct user.

TiBiBa:
**Description**
In this PR we try to take a first step into better dividing the functionality of both the back-end and the front-end. Currently `auth.ts` handles a lot of functionality that, in my opinion, should be handled in the back-end. Such as re-directing pages when already logged in, storing GET attributes etc. We re-write the structure of `auth.ts` to only focus on handling the front-end relevant stuff. We also fix some missing error templates, small styling issues and replace the old error modal with our new one.

**Fixes**
This PR fixes #2089 and fixes #2099

**How to test**
This is a clean-up task, if nothing is broken it works. One small think that can be tested is to verify that you are not allowed to visit the `/reset` page when being logged in. You will simply be re-directed to your profile.

**Checklist**
  

Felienne (approved):
As far as I can see, all still works ok (login, logout, create class, join class)

Felienne:
Ah sadly (expected) conflicts! Can you have a look @TiBiBa?

PR #2110, opened 2022-02-25
[CHORE] Switch from YAML to Babel

### LINKED ISSUE #2030: [CHORE] Switch to Babel for translations
Currently we are dependent on YAML files for our translations. This is a great starting point, but it is getting a bit out-of-hand. Whereas yamls are still useful for complete page translations such as the homepage or learn-more page. It would be more efficient to use a dedicated translation library such as Babel for all our other translations.

Babel will automatically extract the correct translation from the corresponding file, based on our set language. Currently we have some homemade dictionary matching to overwrite the default English values with the current language. This works, but on a large scale is slow and in-efficient.

**This is a very very big chore and needs some clear planning before proceeding**

TiBiBa:
**Description**
In this PR we implement a huge chore where we switch from YAML to Babel (gettext) for or text translations. As the current YAML structure is implemented in all parts of our code base this is a very big operation. This is currently a draft as to keep track of the merging conflicts and testing issues.

**Fixes**
This PR fixes #2030.

**How to test**
As this is a (big) clean-up no change should be visible from a user perspective. However, it does need extensive testing to make really really sure we didn't break anything.

**Last things to do before merging**
- Copy `unsaved_changes` string to the `client_messages` YAMLs
- Fix that this is working (as it is currently broken)
- Verify that all pages and error messages are valid

**List of languages processed**
The structure is done, however all languages need to be manually copied  from the YAMLs to the .po files. In the list below we keep track of the languages in process (this is constantly updated):

- ~ar~
- ~bg~
- ~bn~
- ~cs~
- ~de~
- ~el~
- ~en~
- ~es~
- ~fr~
- ~fy~
- ~hi~
- ~hu~
- ~id~
- ~it~
- ~nl~
- ~pt_BR~
- ~pt_PT~
- ~sw~
- ~tr~
- ~zh_Hans~


TiBiBa:
This isn't nearly finished yet, but we should always merge #2101 first!

TiBiBa:
To work with the gettext() library a small tutorial below, this will also be implemented in the documentation later on:

```
  -   We can always get a translation using gettext(u'english string')
      NOTE: We can shorten this notation by simply using _('english string')
  -   We can insert some variable like this: gettext(u'some string %(value)s', value=42)
  -   More interesting for us might be the 'lazy string' the can be defined outside requests, like this:
  -       lazy_gettext(u'Account successfully saved')
  -   This will be really useful when wanting to return translated error messages
  
  - We have to mark ALL translatable string (in english!) with gettext() -> then create a .pot file
  - We create the file as follows: 
      pybabel extract -F babel.cfg -o messages.pot .
  - To add a translation (for dutch): 
      pybabel init -i messages.pot -d translations -l nl
  - To update your files (when adding new strings):
      FIRST create new file:  pybabel extract -F babel.cfg -o messages.pot . 
      THEN:                   pybabel update -i messages.pot -d translations
  LASTLY compile the files:   pybabel compile -d translations
```

TiBiBa:
Not sure why the current state of this PR passes all tests. The TypeScript files give (a lot!) of errors because of the removed `auth.texts` dictionary. We should be extra careful with testing to make sure nothing is broken or we should write additional tests who catch these errors.

TiBiBa:
Current issues:

- Transpire success messages are not shown -> re-write to back-end
- Some promises on the front-end return errors
- `/parse` translations are still based on the YAML files

TiBiBa:
Current big issue:

Re-write Hedy parsing from YAMLS to gettext() values

TiBiBa:
We now allow the following structure:

- Hedy errors are completely handled by `gettext()`
   - We do have to write some work-around (to-do) as dynamic naming of the value works but returns warnings
   - For example, `gettext(u'' + str(code))` is not allowed but does work
   - As we can't find keys that are not used we create a separate `error-messages.txt` with the only reason to find the keys
- I think stored the `ClienErrorMessages` in the YAML for now is the best solution: we can throw them at once to the fe

TiBiBa:
We might be able to improve the client-message structure by using the same trick as with the Hedy errors. Story the keys inside a .txt file and let gettext() handle the translations. We can still use this same textfile to retrieve all translations from gettext(). This reduces the amount of yaml files and directories as this is rapidly growing.

TiBiBa:
@Felienne @Mark-Giesen I think this is ready for a first spin! It's not ready for merging (as we remove a lot of current translations). But it would be nice to discuss the current progress and if we are on the track. `/texts` is completely removed and all keys are now stored with `gettext()`. Currently the translations are complete for English and Dutch, so for these languages no visual change should be found.

Felienne:
ok I tried to merge main in here but because of the weblate updates there are now 1 million conflicts 🤣 , so I guess you need to have a look @TiBiBa

TiBiBa:
> ok I tried to merge main in here but because of the weblate updates there are now 1 million conflicts 🤣 , so I guess you need to have a look @TiBiBa

Woww! I will take a look and try to fix them all, but it seems like a lot a lot of work. Hopefully I will get it fixed before my holiday. One important note on this PR: Afterwards only dutch and English are working as the other languages are not translated yet in Babel. So we don't want to merge this (yet), until we have a suitable solution.

Felienne:
> Woww! I will take a look and try to fix them all, but it seems like a lot a lot of work. Hopefully I will get it fixed before my holiday. 

Yeah it might be easier to just cherry-pick the relevant commits on the current main?

> One important note on this PR: Afterwards only dutch and English are working as the other languages are not translated yet in Babel. So we don't want to merge this (yet), until we have a suitable solution.

Yes I know! Maybe not bad to set it to draft to indicate it should not be merged right away? and maybe part of that is fixed with the weblate merge that translated all missing strings to English?

TiBiBa:
@Felienne merging is fixed, this can be tested!

Felienne:
Guess this might be a day task too hahaha 😭 

image

TiBiBa:
> Guess this might be a day task too hahaha 😭
> 
> image

Wowww... Will take a look to solve this! We might have to decide on what point we want to merge this as keep solving the merging errors is quite some work.

Felienne:
Hi @TiBiBa! I am ok with merging this, @Mark-Giesen tested it and it does all seem to work. Is it ready to be merged again? (not sure since it is a draft)

Felienne:
Ah I see, at least we need to make sure the headers are in En again instead of dev_language :) 

image

TiBiBa:
> Ah I see, at least we need to make sure the headers are in En again instead of dev_language :)
> 
> image

I'm not sure we are ready to merge as currently only the English and Dutch gettext files are translated. As it doesn't support a fallback language, the UI won't be pretty for any non-translated language. We have to think about a smooth transition from the YAMLs to this new structure (and fix this in weblate as well).

TiBiBa:
We have to fix all page titles after the most recent merge from main. As well as check all front-end functionality such as saving, sharing and submitting as a lot has changed on the authentication on that part. Will do that today!

TiBiBa:
@Felienne I think this is ready for review! We still do have to think about only English and dutch being translated at this point, but I'm not sure what would be the best approach. Can we deploy to Alpha and at the same time setup the weblate and make sure a large chunk is fixed before deploying to beta? One issue is that we are unable to merge other fixes in the meantime.

Felienne:
> @Felienne I think this is ready for review! We still do have to think about only English and dutch being translated at this point, but I'm not sure what would be the best approach. Can we deploy to Alpha and at the same time setup the weblate and make sure a large chunk is fixed before deploying to beta? One issue is that we are unable to merge other fixes in the meantime.

We could merge and deploy to alpha but then an issue is that we can't deploy further (f.e. if we need to deploy an important bug fix), but I see we don't want to keep fixing merge conflicts on this one forever. Maybe just prioritize making this in order for a bit?

TiBiBa:
Short overview of to-do's from this point till "ready for merge":

- Fix merging conflicts
- Add compile code to `on-deploy.sh`
- Manually copy all other languages (non English/dutch) from `/texts` to .po files
- Merge!

Felienne (approved):

PR #2112, opened 2022-02-25
Fix-merge-conflicts-weblate

Mark-Giesen:
Poging om mergeconflict binnen Weblate op te lossen.


Mark-Giesen:
the build keeps failing on the missing of start_code in the newly created language files. Those are indead missing, but because they are empty in the english file: start_code: "". What's the purpose of this key being obligatory when it can be empty?

PR #1890, opened 2022-02-02
Syntax highligher improvements

Felienne:
This PR implements syntax highligher support for Arabic names:

Current main:
image

This PR:
image

It also includes a few general improvements towards #1285:
* Allow for `=` in ask
* Moves some comments and commands to the right levels (no visual impact)
* Correctly handles at random #1995 
* Correctly highlights indented lines:

main:
image

This PR:
image

PR #2113, opened 2022-02-25
add hightlighting for forward, turn, sleep in level 4 and higher

### LINKED ISSUE #1793: [BUG] Syntax highlighting in level 4, 6 and 8
Some syntax highlighting issues 
- from level 4 the syntax highlighting for `forward` and `turn` is off
- From level 6 you can use `=` instead of `is`. When `=` is used the syntax highlighting seems to stop after the `=` sign.
- in level 8 and higher the syntax highlighting does not work for `repeat . times`.

See the tabs for each level for the examples

thjazi39:
Partial fix for #1793  :
> from level 4 the syntax highlighting for forward and turn is off

The following problems have been solved in the PR #1890 :
> From level 6 you can use = instead of is. When = is used the syntax highlighting seems to stop after the = sign.
> in level 8 and higher the syntax highlighting does not work for repeat . times

There is also a bug in the sleep syntax highlighting that has been fixed.


Felienne:
> As for the `sleep` one, I am not sure what was changed?

I think, reading the code, it should now work in all levels, and I can confirm it does work!

Felienne (approved):

Issue #2114, opened 2022-02-27
[IMPROVEMENT] Preserve whitespacing during program translation


boryanagoncharenko:
To translate a program from one language to another, we currently parse the user input and then reconstruct the program with the translated keywords. Along the way, we lose information from the original program, e.g. whitespacing.

What we could try instead, is using the keep_all_tokens flag of Lark, so that the parse tree keeps all tokens including whitespacing. From the meta information of the tokens, we could pinpoint their exact location in the user input string and replace them with their translated counterpart.

PR #2115, opened 2022-02-27
Change program translation so that it preserves all tokens lost during parsing #2114

boryanagoncharenko:

< Change program translation mechanism >

**Description**

Instead of parsing and then reconstructing the tree, we now translate programs through extracting tokens' meta information from the parse tree and replacing them in the original user input string with their translated counterparts.

** IMPORTANT NOTE **

Note that this PR contains 3 changes of tests which seem rather weird: 
1) https://github.com/Felienne/hedy/pull/2115/files#diff-f558aa077eea06d131071ab6183c4fccd65dfd56f6974e5e9bce489009a62003R196-R197
2) https://github.com/Felienne/hedy/pull/2115/files#diff-f558aa077eea06d131071ab6183c4fccd65dfd56f6974e5e9bce489009a62003R243-R244
3) https://github.com/Felienne/hedy/pull/2115/files#diff-3f91877484ae8f3c806b4ecad51415810d8cf160f6839037264cffd8e48cabd7R407-R408

After preserving the whitespaces, these tests started failing intermittently. The reason seems to be that when we do not preserve whitespacing, the command following the 'else' clause was put on the same line as the else itself (see the last line):
```
dishwasher is people op willekeurig
als dishwasher is Sophie
print \'too bad I have to do the dishes\'
anders print 'luckily no dishes because' dishwasher 'is already washing up'
```

However, now, the same code is translated to the following. Note that the print command is on the next line of 'anders'
```
dishwasher is people op willekeurig
als dishwasher is Sophie
print \'too bad I have to do the dishes\'
anders
print 'luckily no dishes because' dishwasher 'is already washing up'
```

Apparently there is ambiguity in the grammar which exists regardless of the changes in the current PR. I can replicate the issue by adding the following test to the test_translation_level_05.py of the main branch. Please note that sometimes I have to run it 10 times before it fails. Created issue #2116 to trans this problem.
```
    def test_fail(self):
        code = textwrap.dedent("""\
            dishwasher is people op willekeurig
            als dishwasher is Sophie
            anders
            print 'luckily no dishes because' dishwasher 'is already washing up'""")

        result = hedy_translation.translate_keywords(code, from_lang="nl", to_lang="en", level=self.level)
            if dishwasher is Sophie
            else print 'luckily no dishes because' dishwasher 'is already washing up'""")

```

I created a new issue describing just that problematic test: 

**Fixes < #2114  >**

**How to test**

- All tests should pass
- You could perform manual tests of translating a program back and forth. Note that whitespacing is preserved now. 

**Checklist**
  

Felienne:
Great work @boryanagoncharenko!! I will see if all still works but it looks promising. And I think I know exactly what is up with that one test, so I will pick that up!

Felienne:
Sorry @jpelay this improves but almost destroys your hard work on the translation preservation (#1587) Guess @boryanagoncharenko outsmarts us both with her in-depth Lark understanding 🤣!

But the work was not entirely reversed; the separate rules for input remain, so it was not all for nothing!

Felienne:
@boryanagoncharenko, I removed some duplicate rules that we only had in the grammar so we could preserve some token, so now the grammar is even simpler (that's really nice!)

Felienne (approved):

jpelay:
> Sorry @jpelay this improves but almost destroys your hard work on the translation preservation (#1587) Guess @boryanagoncharenko outsmarts us both with her in-depth Lark understanding 🤣!
> 
> But the work was not entirely reversed; the separate rules for input remain, so it was not all for nothing!

But this is a great change :) Makes things a lot better!!

Felienne:
> > Sorry @jpelay this improves but almost destroys your hard work on the translation preservation (#1587) Guess @boryanagoncharenko outsmarts us both with her in-depth Lark understanding 🤣!
> > But the work was not entirely reversed; the separate rules for input remain, so it was not all for nothing!
> 
> But this is a great change :) Makes things a lot better!!

Thanks for being so understanding! And yeah I do think it is nicer without duplication of the rules

PR #2118, opened 2022-02-28
[CLEAN-UP] Improves error message when a level is not available

### LINKED ISSUE #2108: [BUG] "no such level" is a bit generic
We show "no such level exists" in all cases where the level is not reachable, for various reasons, but this is not always strictly true. 

Two cases come to mind:

**1. Logged in user in a class** 
If I am a logged in user in a class whose teacher hid a level, I get the message that the level does not exist. This is confusing, esp when teachers close levels kids have already seen. Can we change that with a new template saying something like: 

"you are in a class in which this level has been closed"?

**2. Level not translated yet**
If I set the language to Arabic and then navigate to hedy/8/ I also get this message but the real reason is that this level does not exist *in this language*.  Can we change the error with that note? "No such level in your selected language"?

And we want to redirect to En always in that case after a few seconds? Or give the error and a link to the English version with: "Use this to look at the English version instead?" I can make a nice Hedy drawing with different languages :)

TiBiBa:
**Description**
In this PR we fix an issue where a generic error message was returned when a user was not allowed (or unable to) load a specific Hedy level. Instead of simply returning "No such level" we now differentiate between "Not translated" or "Not allowed by class". Making it better understandable for the user why they can't visit this level.

**Fixes**
This PR fixes #2108

**How to test**
Visit either a level that is not yet translated in your current language or restrict a level through class customizations and try to visit it. Notice that in both cases a specific error message (with correct error code either 404 or 403) is returned.

**Checklist**
  

Felienne (approved):

Felienne (approved):

Felienne:
Maybe we want to open an issue to change the text of the error message for untranslated levels with a link to the new weblate interface to people can start to translate straight away?

PR #2121, opened 2022-02-28
[FEATURE] Implements viewing of one specific adventure

### LINKED ISSUE #2096: [UI idea] Link for one adventure
It would be nice if we could make direct link to one adventure, to be used for workshops and for campaigns like de week van het geld.

A link, like hedycode.com/adventure/story would then just link to the Hedy UI, but only with the story tab shown (and level defaults, of course) Somewhere (in the top bar?) 

We might want an extra button to show all adventures (or not, kids could do that by going to Hedy in the top bar?)

@TiBiBa We did have this feature very early when we deployed adventures so maybe there might be some relic code you can use!

TiBiBa:
**Description**
In this PR we implement the functionality to visit one specific adventure. Instead of the normal `/hedy` url users are now able to visit an adventure by using `/adventures/<adventure_name>`. The previous and next level buttons are automatically adjusted to link to the previous next level of the adventure if it exists. This enables teachers, presentations etc with one specific adventure in use.

**Fixes**
This PR fixes #2096.

**How to test**
Navigate to any specific adventure using the URL `/adventure/<adventure_name>`. You will automatically be re-directed to level `1` of this adventure. You can specify a level by adding it to the URL like this `/adventure/<adventure_name>/<level>`. If a adventure/level combination doesn't exist a `no_such_adventure` error is thrown.

**Checklist**
  

Felienne (approved):
Looks good! I was thinking do we need something that tells people they can go to Hedy in the top bar to see all adventures?

PR #2130, opened 2022-03-01
[CHORE] Clean-up mail templates

### LINKED ISSUE #2127: [CHORE] Clean-up process of mail templates
**Describe the issue**
Currently the mail sending template is a bit of a mess. This is relic code that hasn't been touched in a long time. It would be nice to clean this up, give the html template itself a refresh to better align with our UI and improve the multi-language support. Currently we made small improvement in #2126, but that is not really future-proof and more of a "nice to fix" for now. We should do a more extensive clean-up to fix this code.

TiBiBa:
**Description**
In this PR we clean-up the mail templates, give the HTML template a new look and simplify the function handling the e-mail sending. With some dictionary matching we also make sure there is always a correct string available. After merging #2110 we can also re-write this structure to be dependent on `gettext()` instead of YAMLs, but it should be fine for now.

**Fixes**
This PR fixes #2127.

**How to test**
Can't test this locally as we don't support local e-mail sending, we should test this extensively on Alpha.

**Checklist**
  

Felienne (approved):
ready to test on alpha! :)

TiBiBa:
> ready to test on alpha! :)

Will do some testing right away!

PR #2132, opened 2022-03-01
[BUG FIX] Login re-direct on class pre-join works as expected

TiBiBa:
**Description**
In this PR we fix an issue where the login re-direct on the class pre-join was broken. That was due to two separate issues:

- We broken `auth.profile` with #2101
- The saved re-direct item called a `link` value that didn't exist, it should have been `id`

Both are fix in this PR and the temporary storing of the join link is fixed again.

**Fixes**
No related issue

**How to test**
Create a class (or use an already existing one) and make sure you have an account that is not already in this class. Copy the join link, log out and navigate to the join link. Notice the pop-up "you can only join if logged in, do you want to join?". Log in and you should automatically join the class (as the link is temporary stored). You can verify this my navigating to 'My account' -> 'My classes'.

**Checklist**
  

Felienne (approved):
Works! Although I did not get a modal after login that I was enrolled to the class, is that as expected? It did show up in my classes so all is in order.

TiBiBa:
> Works! Although I did not get a modal after login that I was enrolled to the class, is that as expected? It did show up in my classes so all is in order.

This is as expected! It would indeed be nice if a modal was shown but that is quite complex with our current implementation. Another solution is to re-direct the user back to the "join class" page after login in.

PR #2133, opened 2022-03-01
Exclude if/else in all supported languages

### LINKED ISSUE #2116: [BUG] Intermittent program translation failures (probably due to grammar ambiguity)
**Describe the bug**
Add the following test to the test_translation_level_05.py and note that it fails intermittently. Please note that sometimes I have to run it 10 times before it fails.
```
    def test_fail(self):
        code = textwrap.dedent("""\
            dishwasher is people op willekeurig
            als dishwasher is Sophie
            anders
            print 'luckily no dishes because' dishwasher 'is already washing up'""")

        result = hedy_translation.translate_keywords(code, from_lang="nl", to_lang="en", level=self.level)
            if dishwasher is Sophie
            else print 'luckily no dishes because' dishwasher 'is already washing up'""")

```

Level 5

**Expected behavior**
The test should be stable

Felienne:
Fixes #2116 

It is a bit of a hack (it always was!) but this disallows the use of if/else in strings allowing us to fix #2116. We now do so in the grammar of level 5 for all languages, at one point we might want put this in each language because otherwise legal programs become problematic for no good reason because there can be many words for if/else in all languages. But this will do for now. Left in the extra test too.

### DIFF
--- a/grammars/level18-Additions.lark
+++ b/grammars/level18-Additions.lark
@@ -1,11 +1,10 @@
 // adds round brackets in print() and range(), and changes ask to input
 
-command:+= input | input_is_empty_brackets |  input_equals_empty_brackets | print_empty_brackets -= ask > assign
+command:+= input | input_empty_brackets | print_empty_brackets -= ask > assign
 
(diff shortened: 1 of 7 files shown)


Felienne:
(secretly also sneaking in a few more simplifications following #2115)

PR #2134, opened 2022-03-02
Correctly highlights add and remove in all levels

### LINKED ISSUE #2084: [BUG] the add and remove commands don't get syntax highlighting
(Sorry if this issue already exists)

In level 3 the add to and remove from commands are introduced but they don't turn red. 

taarten is aardbei, chocolade
add appel to taarten
remove chocolade from taarten
print taarten at random

thjazi39:
added syntax highlighting for add remove with list
before : 
image

after :
image

fix #2084 

new method added:
```
function rule_listManipulation() {
  return comp(
    recognize('start', {
      regex: "^(" + currentLang._ADD_LIST + ")( )(.*)( )(" + currentLang._TO + ")( )(.*)$",
      token: ['keyword','text','text','text','keyword','text','text'],
      next: 'start',
    }),
    recognize('start', {
      regex: "^(" + currentLang._REMOVE + ")( )(.*)( )(" + currentLang._FROM + ")( )(.*)$",
      token: ['keyword','text','text','text','keyword','text','text'],
      next: 'start',
    }));
}
```


Felienne:
Hi @thjazi39!

I you rephrase the PR text and say "fixes #000" then the issue will close **automatically** when the PR is merged, like magic :)

Felienne:
And very well done on the PR text by the way, the screenshots really help! I just rewrote the PR name a bit to be more precise. That is not only helpful now, but also in the future. Suppose we found a bug that was introduced in this PR, then it helps more if we know it is about add and remove in all levels than if it is just called improvements. Hope that clarifies!

Felienne:
Ah sadly it is not 100% correct, for Dutch, add (voeg ... toe aan) does not highlight:

image

For French I could not find the snippet, it is not translated yet, so that is something to look into also?

thjazi39:
actually, it doesn't work with compound keywords, but it was on the old syntax highlighting system

If the new system is accepted, then this bug will be solved

Felienne (approved):

PR #2136, opened 2022-03-02
[FEATURE] Localized quiz keywords

### LINKED ISSUE #2103: [UI idea] Localize quiz keywords
**Idea incl level**
Currently we already localize the level-defaults and adventures keywords based on the user selected keyword language. It would be nice if we also implement this feature into the quiz.

TiBiBa:
**Description**
In this PR we localize the quiz keywords to align with our implementation of localization for the level-defaults and adventures. Due to the current structure of the quizzes this had to be re-written as well. We now retrieve quiz data similar to level-defaults and adventures by reading them once on setup and then reading from the created dict. This should be more efficient than the current implementation where we read the YAML file again and again (each question!).

**Fixes**
This PR fixes #2103 (but got a bit out-of-hand and fixes the quiz structure as well)

**How to test**
Verify that the quiz is now working for every language. If no quiz template exists the English one is automatically returned. (We might want to do this for level defaults as well?) Keywords are automatically parsed from the selected keyword language, if non is selected these are English. Verify that this is working correct as well.

**Checklist**
  

Felienne:
This works well but.... there are some interesting content issues now, for example:

image

And more subtle (aks is a weird suggestion for vraag!)

image

And this one, ok for Dutch but not ok for languages where `echo` is not `echo`:

image

So I am a bit wary to merge this without having @MarleenGilsing take a peek!

TiBiBa:
I think it would be nice to merge this as the conflicts keep getting more complex. We do have to remove / change some questions to makes sense with dynamic keywords, but maybe fix that in a new issue?

Felienne:
> I think it would be nice to merge this as the conflicts keep getting more complex. We do have to remove / change some questions to makes sense with dynamic keywords, but maybe fix that in a new issue?

Yeah I totally understand but merging this makes the quiz look really bad in localized languages. 

Shall we merge it with a feature flag until we figure out the content issues?

TiBiBa:
> > I think it would be nice to merge this as the conflicts keep getting more complex. We do have to remove / change some questions to makes sense with dynamic keywords, but maybe fix that in a new issue?
> 
> Yeah I totally understand but merging this makes the quiz look really bad in localized languages.
> 
> Shall we merge it with a feature flag until we figure out the content issues?

Sounds good! Will first fix the merge conflicts, than set a flag and make this ready for review again.

TiBiBa:
Merge conflicts seem to be fix, once a flag is always parse English keyword (and don't change anything for the end-user) we are ready for review!

Felienne (approved):

TiBiBa:
We should fix these conflicts and automatically parse all keywords to English. Then we can at least merge this PR so we no longer have to keep fixing these. Once ready with updating the questions we can remove the "always parse to en" flag

Felienne:
Works!

image

Felienne (approved):
Let's hope Weblate does not get all too angry!

Felienne:
That sounds like a plan!

Issue #2111, opened 2022-02-25
[Language idea] quotation marks for string assignments


thjazi39:
From level 2, Hedy gives us the possibility to manipulate strings, and to store them in variables.

However, this can potentially create interpretation problems:

<table>
	<tr>
		<th> Python </th>
		<th> Hedy 2/3 </th>
	</tr>
	<tr>
		<td> var = input('number ?')<br/> print("var equal to :",var) </td>
		<td> var is ask number ?<br/> print var equal to : var </td>
	</tr>
	<tr>
		<td> var = input('number ?')<br/> print(var," equal to :",var) </td>
		<td> var is ask number ?<br/> print var equal to : var </td>
	</tr>
	<tr>
		<td> v = input('x ?') </td>
		<td> v is ask x ? </td>
	</tr>
	<tr>
		<td> v = 'ask x ?' </td>
		<td> v is ask x ? </td>
	</tr>
</table>

In the case of the `ask` command or the `print` command the problem is solved at level 4, by the need to add quotes, but this is not added during assignments, so there are still problems. And later on, in if when comparing strings, this can also be a problem 
(when we want to compare with a random string, with the 2nd element of a list)

I have the impression that some of the bugs are directly related to this impossibility:   #1938  #1800 #1101 #2129

Thus, the introduction of quotation marks for all strings would simplify the interpretation and syntactic coloring:

for interpretation, everything in quotes is considered text, and the rest can be interpreted as commands. this won't solve all the bugs, but for example, in the case of `is` or `is ask` I guess it will make things easier.

for syntax highlighting, it would allow to color all strings in purple, and among what remains, we color all keywords in red.
This will even allow to add an extra color, for example, for variables. 

In both cases, -I think-, that it will simplify the treatments, and that it will avoid many bugs

It should also be easier in the choice of the error message.

Concerning the pedagogical aspect, I don't know too much about it, but I think it could also be interesting. It would allow to explain to the children that there are, on one side the "sentences" / "words" / "text" and on the other side, the "code", the "program", the "commands".
This would also be associated with a color, purple, for everything that hedy will not "look in detail".

In addition, currently, the following code, which could potentially raise questions in children :
<table>
	<tr>
		<td> print 'blabla' </td>
		<td> v is blabla<br/> print v </td>
	</tr>
</table>

So that's the suggestion I wanted to make.
I am not yet very familiar with the way Hedy works, but I imagine that such a change is not to be taken lightly, especially since I would be unable to do it at the moment.

[...]

PR #2137, opened 2022-03-02
Separates keyword rules and adds some other improvements

### LINKED ISSUE #724: syntax highlighting after `echo`
When you use an `echo` command without any text e.g.:
```
ask yes or no?
echo 
print okay
```

The syntax highlighting does not work for the command after `echo`. So in the example above, `print` is not highlighted.

### LINKED ISSUE #1285: Check and improve all syntax highlighting
Since all the changes we made to the levels, not all syntax highlighting is still in order, would be good if we do a thorough check of all levels and update where needed.

### LINKED ISSUE #1686: [BUG] 'at random' works in level 2 where it shouldn't
**Describe the bug**
The `at random` keyword works in level 2 and level 3, however it is only introduced in level 3. Aligned with the syntax highlighting (that also highlights `at random` in level) it should no longer work in level 2.

**Expected behavior**
Receive a `wrong level` warning when using `at random` in level 2

### LINKED ISSUE #1793: [BUG] Syntax highlighting in level 4, 6 and 8

### LINKED ISSUE #1888: [BUG] Range must have 2 numbers in level 18
**Describe the bug**
`range(5)` is not allowed while `range(1,5)` is.

Level 18
```
naam = 'Hedy'
print('ik heet', naam)
for i in range(5):
    print('lala')
```

### LINKED ISSUE #2084: [BUG] the add and remove commands don't get syntax highlighting

### LINKED ISSUE #2131: [BUG] Syntax highlighting for ask without quotes should not highlight
See https://github.com/Felienne/hedy/issues/2129#issue-1155408775

thjazi39:
# Paradigm shift:
This branch was only supposed to experiment with another syntax highlighting paradigm, but since this is conluent, I propose this pull request

The paradigm that this branch will experiment with is the following:
- Levels [1-3]:
In these levels, there are no strings yet, so understanding the lines relies a lot on the context, for example, in `answer is ask what is your name?` you have to color the first `is`, but not the 2nd.
Thus, the syntactic coloring in the first 3 levels works by "valid line" and allows to take into account these specificities.

- Levels [4+]:
In these levels, the following system is applied:
  - Everything between quotes is a string (purple)
  - In what remains, we color the keywords and symbols
  
  We will distinguish the keywords (which need a beginning and an end of word) and the symbols (which are independent), even if both have the same color.

This is a bit simplistic, but it works.

# Example : 
- level 1 :
image
- level 2 : 
image
- level 3 : 
image
- level 5 : 
image

- level 16 :
image

# Questions :

## But in some cases, doesn't this induce over-coloring ?

Indeed, if we do `add times to liste_exemple` we will obtain the following coloring :
image
The word `times` here should not be colored, but I don't think that the fact that the coloring is here is a problem. it allows to tell the child that we use here a "forbidden word" and that using it could mislead

## On the other hand, why in the first levels [2->11] all the strings are not in purple ?
Indeed, in the following examples:

image
image

The strings are not colored, but this comes from the ambiguity due to the absence of quotes (until level 4 for `print` and `ask` until level 12 for `is`)

Indeed, it is impossible to distinguish if the following code 
```
print name is hedy
```
if `hedy` or `name` is the variable. And so we can't distinguish where the string is, it would require remembering the names of the variables that exist before in the code. But this is not possible with an automaton, you need a grammar to do this

this has the consequence that if we do : 
```
add liste1 at random to liste2
```
We will obtain the following coloring :
image

I don't know yet if it will be possible to solve this or not, I would have to look into how strings are interpreted with grammar files.

# Other Changes
- `A-z` -> `A-Za-z` When you do this in 1 time, you also catch brackets (see an [ascii table](https://www.asciitable.com))
- As said in the original file ``b` doesn't seem to work with unicode characters. I found on [this site](https://www.regular-expressions.info/wordboundaries.html) a way to work around the problem. Thus, we get the following code:
image
This allows to simplify the life for the detection of beginning and end of word.

- I unlinked levels 8 and 9 (as well as 11 and 12) to allow more flexibility in syntax processing (obviously, this induces a little redundancy, but not that much)
- Until I have the left and right keyword treatment, in level 1, I color the word after `turn`, no matter the word. So by definition, this includes all languages (unless there are languages that need more than one word to say left and/or right)
- I removed the `is` <=> `=` association in the language files. Indeed, since we are dealing with symbols and keywords, it is no longer necessary.

# Remarks : 
I don't know if this is done expres, but when I was looking at the syntax files, I noticed some stuff, and I don't know if this is normal or not (and if it's not, this should be the subject of other issues) :

- The `repeat` command is replaced by the `for` command in level 9, but still works in level 10 (but no longer works in level 11).
So, the coloring also follows this operation, but I wondered if this was normal. And if it's not, when modifying it, you have to be careful to change the syntax highlighting too.
- The problem is similar with the command `at`, which is replaced by square brackets in level 16, but the command `at` still works in level 16 and in the next level

- At the beginning of the change, I based myself on the following code, found in the `hedy.py` file :
image
I don't know exactly what it is used for, and how it is used, but I noticed that there were some errors. There are some commands that come too early, some that remain but have been removed, and some that don't appear.
If this is important, I could make an update from the lists I redid as part of the syntax highlighting change.

- I didn't manage to test the syntax highlighting with level 18, according to the grammar files, there is the addition of the parentheses, and normally the highlighting works, but it was not tested.

- The following changes have been tested on all the codes available in the statements of the different levels, but only in English.
I will test the other languages but it will normally not be a problem

- In some levels, the statement contains code quotations that don't have the right syntax highlighting, and I don't know if this is normal or if it is due to a bad call of the syntax highlighting engine

image

# bux fix by this update :

fixes #1793 
fixes #2084 
fix #2131 
fix #1888 
fix #1686 
fix #1285
fix #724


Felienne:
Thanks for this extensive work @thjazi39! I really like where you are going with this! 

I feel maybe a discussion would have been a bit of a better place to discuss these but I understand that sometimes you first want to try some code to see if something is possible and will look nice.

Felienne:
Here are some answers to your remarks. 

> * The `repeat` command is replaced by the `for` command in level 9, but still works in level 10 (but no longer works in level 11).
>   So, the coloring also follows this operation, but I wondered if this was normal. And if it's not, when modifying it, you have to be careful to change the syntax highlighting too.
> * The problem is similar with the command `at`, which is replaced by square brackets in level 16, but the command `at` still works in level 16 and in the next level

Yeah I think we did this to ease the step, maybe this should be a warning that it is an old command? I think for now it is ok.

> * At the beginning of the change, I based myself on the following code, found in the `hedy.py` file :
>   image
>   I don't know exactly what it is used for, and how it is used, but I noticed that there were some errors. There are some commands that come too early, some that remain but have been removed, and some that don't appear.
>   If this is important, I could make an update from the lists I redid as part of the syntax highlighting change.

I think this is an artefact of big changes we made in November. If you see issues, do update this list (in a separate PR) The text on the website and the grammars are always leading, if you make changes, maybe check is SEMANTICS.MD also needs an update.

> * I didn't manage to test the syntax highlighting with level 18, according to the grammar files, there is the addition of the parentheses, and normally the highlighting works, but it was not tested.

You can at least open level 18 and see if it looks allright (but maybe you already did that?)

> * The following changes have been tested on all the codes available in the statements of the different levels, but only in English.
>   I will test the other languages but it will normally not be a problem

Great, thanks! Esp. Arabic can be error prone.

> * In some levels, the statement contains code quotations that don't have the right syntax highlighting, and I don't know if this is normal or if it is due to a bad call of the syntax highlighting engine
>   image

Yeah that is as expected. Text between `` is always made pink, where within ``` ``` we use the full highlighter. 
We could sue the full highlighter for `` but because it is incomplete code it will not always highlight well so this is an ok intermediate form I thin.

Felienne:
> * As said in the original file ``b` doesn't seem to work with unicode characters. I found on [this site](https://www.regular-expressions.info/wordboundaries.html) a way to work around the problem. Thus, we get the following code:
>   image
>   This allows to simplify the life for the detection of beginning and end of word.

Great! But... we also want to support Hindi and Bengali, so maybe we need a few more ranges there?

thjazi39:
> I feel maybe a discussion would have been a bit of a better place to discuss these but I understand that sometimes you first want to try some code to see if something is possible and will look nice.

Indeed, at the beginning I was just doing some tests to understand how Ace works, but one thing leading to another, I got to the point!

However, I did start a discussion here #2111 about quotation marks 
This is more or less independent of the work that is done here, although the earlier appearance of quotation marks should solve a lot of problems.
And since here I have adopted the "string first, keyword second" point of view. If ever quotation marks become necessary since level 4, there will be no syntactic modification to do.

> You can at least open level 18 and see if it looks allright (but maybe you already did that?)

I tried, but the level is not accessible (at least, not as easily as with the other 

> Great! But... we also want to support Hindi and Bengali, so maybe we need a few more ranges there?

I took the ranges that were in the original file, so the language porting should not change.
I'll test with the other languages, and if it's a problem, then we can add other ranges.
Another direction we could take would be to define everything that is not a letter ( [^[]/,:; .... ] ) but I don't think that would be very safe, especially with non-displayable characters, etc...

Felienne:
> > I feel maybe a discussion would have been a bit of a better place to discuss these but I understand that sometimes you first want to try some code to see if something is possible and will look nice.
> 
> Indeed, at the beginning I was just doing some tests to understand how Ace works, but one thing leading to another, I got to the point!

Yeah totally understood!

> I tried, but the level is not accessible (at least, not as easily as with the other

https://hedy-alpha.herokuapp.com/hedy/18 works, and for me there is a button in level 17 too?

> > Great! But... we also want to support Hindi and Bengali, so maybe we need a few more ranges there?
> 
> I took the ranges that were in the original file

Yeah I know, this was one of the things I had in mind to improve, so you did not make it worse for sure!

> Another direction we could take would be to define everything that is not a letter ( [^[]/,:; .... ] ) but I don't think that would be very safe, especially with non-displayable characters, etc...

That is something we can try. What I would like is to use proper character classes like we do in the grammar, like \p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo} I don't think these work like that in Typescript/Ace directly (I tried but gave up without depper search) but that seems to be a good task for you to dive into? :) (In a separate PR)

thjazi39:
> > * The `repeat` command is replaced by the `for` command in level 9, but still works in level 10 (but no longer works in level 11).
> >   So, the coloring also follows this operation, but I wondered if this was normal. And if it's not, when modifying it, you have to be careful to change the syntax highlighting too.
> > * The problem is similar with the command `at`, which is replaced by square brackets in level 16, but the command `at` still works in level 16 and in the next level
> 
> Yeah I think we did this to ease the step, maybe this should be a warning that it is an old command? I think for now it is ok.

I guess for the `is` it is the same, it is gradually replaced but is still active until the end?

thjazi39:
I have reviewed the English version, and there doesn't seem to be a problem.
Concerning the other languages
When a keyword is composed of several words. As in the following example, it is well recognized, but there must be only one space, if there is more, it will not be detected.
I see three solutions to solve this problem:
- add a `+` in the language files
- replace the space by an underscore
- add the 2 keywords independently

I think the `+` will be the best solution
image

Concerning the languages using unicode characters, I quickly looked at the Arabic version at level 1
It seems to work well:

image
However, in the statement, we see that the syntactic coloring corresponding to the string appears on both sides of `ask`.
The resulting coloring seems correct, but I don't understand how purple is used on both sides of the `ask` command (this is not supposed to be allowed by the level 1 rules)

thjazi39:
I just realized that I am not taking into account the comments in my syntax coloring.
However, after a few tests, I can't understand if comments are really possible, or I didn't test at the right level.
I saw the PR #2044

I saw that this is only a draft, but I can't figure out if it's an addition of comments or a change in the way the existing comments work

thjazi39:
I have added as agreed a syntax highlighting for the "blanks" in the programs. I chose a color that is maybe a bit too bright, but I can change that.

For the regex, I detect all the underscores alone (see the example) and the string _?_

This allows me to be tolerant to both conventions (one for quizzes, one for adventures)

image

thjazi39:
for comments in the code, I sometimes see `<-` which seem to be comments (because I have the impression that they are interpreted anyway)
image

if needed, I can add them also in the characters that allow to make comments :
image

Felienne:
Ah thanks @thjazi39 for doing the extra work of separating out this improvement from the numbers!!!

Felienne (approved):

PR #2139, opened 2022-03-02
[FIX] User correctly re-directed to landing-page after signup

TiBiBa:
**Description**
Currently the user is only re-directed to the landing page after clicking the verification link sent by mail. However, the user is also automatically logged in after the signup but in this case they are redirected to the programs page. This doesn't make any sense and in this PR we fix this by always redirecting to the landing page on signup. We also fix a crucial bug where the `remember_current_user()` function looked for the `lang` value in the db to get the user language, but this is stored as `language` so nothing is found. It is a miracle that we didn't have issues with this as this function is used by a lot of functionality.

**Fixes**
No related issue.

**How to test**
This can actually be tested locally. Create an account and verify that you are automatically re-directed to the landing page. Also verify that the language of the landing page is according to the user set language.

**Checklist**
  

Felienne (approved):
Works!! 
image

PR #2155, opened 2022-03-03
[BUG FIX] Removes "you must be logged in" pop-up when already logged in

TiBiBa:
**Description**
In this PR we fix an issue with the "you must be logged in to save a program" when the user is already logged in. This is due to being dependent on the front-end `auth.profile` value to determine is the user is already logged in. We rather prefer to do this verification on the back-end so a small re-write is done to check if there is a user on the `/duplicate-check` function. Which automatically fixes the current issue. We now also re-direct the user correctly to the `programs` page if the weren't logged in and save a program.

**Fixes**
No related issue.

**How to test**
Log in and verify that saving of programs is working again. Also verify that when not logged in the pop-up should occur and after logging in the program should still be saved and you should be re-directed to the programs page.

**Checklist**
  

Felienne (approved):
Works!

PR #2159, opened 2022-03-04
update of the list of suggested commands

thjazi39:
As said in PR #2137, I have the impression that there was an error in the `hedy.py` file

Indeed, when I redid the syntax highlighting, I first tried to base it on the list of commands in the `hedy.py` file before I realized that there were lots of errors.

For example :
- the commands `turn`, `forward` and `repeat . times` were present until the end when they are deleted at level 11
- the command `and` appears on level 4 when it only appears on level 13
- the command `ask` no longer exists on level 18
- the `to` and `from` keywords were not present whereas they appear at level 3
- etc..

I don't know if this is important for the functioning of Hedy, according to the comments, it has to do with the suggestion (I guess when the child makes a typo)


Felienne:
Yes indeed this list is used to suggest commands so it does matter it is correct! Thanks for updating.

thjazi39:
However, I didn't pay attention to the order of the orders in the lists, I don't know if this is important or not!

Felienne:
If you want to to make this PR *even* nicer.... you could add a test that fails in main but works after this PR?

For example something like this in test_distance_algo.py?

``
  def test_self_command_ant(self):
    invalid_command = "ant"
    keywords_en_level_4 = hedy.get_suggestions_for_language('en', 4)
    closest = hedy.closest_command(invalid_command, keywords_en_level_4)
    self.assertEqual('ask', closest)
```

In the current main this will suggest and and not ask.

**Review by Felienne — CHANGES_REQUESTED** [review]
Maybe add a few tests?

**Comment by Felienne** [comment]
> However, I didn't pay attention to the order of the orders in the lists, I don't know if this is important or not!

That should not matter, we use string distance to pick the closest command

**Comment by thjazi39** [comment]
I don't understand exactly how the recommendation system works, and what changes it takes into account (deletion and/or addition of a letter, swap of 2 letters, replacement of one letter by another).
And when there is a bigger distance (2 different letters, 3 letters, etc.)
Therefore, I don't know how to manage the distance tests.

However, I made a small python code which, from the list of commands, and a list of modifiers, gives alternate commands:
```
5 : sleep -> slep
13 : from -> frmo
5 : turn -> urn
13 : else -> esle
12 : sleep -> slevp
2 : turn -> tubn
16 : random -> ranom
13 : from -> fromq
6 : print -> prnt
18 : sleep -> sleep
17 : from -> fom
16 : from -> rfom
11 : random -> mandor
10 : from -> form
14 : remove -> removex
11 : for -> fo
3 : add -> adt
10 : from -> rfom
4 : turn -> tunr
14 : add -> addr
```
And from this, I can easily create a large number of tests.

The problem is that since I don't know how the suggestion system is supposed to work, I don't know what to test.

**Comment by Felienne** [comment]
It is supposed to suggest the command with the closest string distance, and I think if you would write some tests, it would very much help you to understand how it works! Tests are a good way to express our expectations about code. You can also read the implementation of course.

Concretely, can you add a few tests of the form above?

```
  def test_self_command_ant(self):
    invalid_command = "ant"
    keywords_en_level_4 = hedy.get_suggestions_for_language('en', 4)
    closest = hedy.closest_command(invalid_command, keywords_en_level_4)
    self.assertEqual('ask', closest)
```

You can generate them with the script you mention above, that would be very efficient. You can also use parametrized tests (we use them in several places in the code base) so you do not have to copy paste all the tests.

**Comment by thjazi39** [comment]
I added some tests for the English version. (I did this in compact mode, but I can also make a function for each, it won't be more difficult for me, it's just that it will take a lot of space in the file)

Concerning the foreign versions, I can also add some tests if needed, I can't understand only from the Dutch version how it works, I'll look at it more in detail.

File: tests/test_distance_algo.py [hunk 1]
```diff
@@ -56,3 +56,51 @@ def test_ask_command_nl(self):
     keywords_nl_level_1 = hedy.get_suggestions_for_language('nl', 1)
     closest = hedy.closest_command(invalid_command, keywords_nl_level_1)
     self.assertEqual('keyword', closest)
+
+  def test_command_en(self):
```
Felienne, line 60: Great!! Buy maybe you can make this into a parameterized test? That is a bit more aligned with the rest of the code base.

Felienne:
I also change the name of the PR, I like it how you added tests there to be complete but tests are not a core ingredient of the PR itself

Felienne (approved):

PR #2171, opened 2022-03-05
Revert "Separates keyword rules and adds some other improvements"

Felienne:
Reverts Felienne/hedy#2137


Felienne:
Hi @thjazi39!

Made a PR to revert the changes, if you want to, you can use this PR and pick only some of the changes to be reverted!

Felienne:
ok since I have to deploy today, I will revert this now @thjazi39 otherwise main has no more highlighting in Safari.

PR #2172, opened 2022-03-05
fix french mistake sur -> au

### LINKED ISSUE #2170: [BUG] at random is not recognized in french
It seems that the keyword `at random` translated as `au hasard` is not recognized in level 3.

As a result, the code provided in the story 'restaurant' does not work.

image

thjazi39:
fix #2170 

replacement of `sur` by `au` in the French grammar file


Felienne:
Would be even nicer if we added a test :) Something like this in level 3?

```
  def test_print_list_random_fr(self):
    code = textwrap.dedent("""\
    animaux est chien, chat, kangourou
    affiche animaux au hasard""")

    animaux = ['chien', 'chat', 'kangourou']
    print(f'{random.choice(animaux)}')""")

    # check if result is in the expected list
    check_in_list = (lambda x: HedyTester.run_code(x) in ['Hond', 'Kat', 'Kangoeroe'])

    self.multi_level_tester(
      max_level=10,
      code=code,
      expected=expected,
      extra_check_function=check_in_list,
      lang='fr'
    )
```

Felienne (approved):

Felienne:
Thanks!!

PR #2122, opened 2022-03-01
Show variables and values live #1329

fRedelaar:
For issue #1329, I've created a box that can display the variables and their values. WIP Draft! **(This work is for a student project at Leiden University)**. Still needs lots of work to make it more fitting in the Hedy theme and overall UX friendly.

**How to test**
Go to any level that uses variables. Level 9, calculator for example. Click on "variables" such as the screenshot below and the list will expand/collapse.
image

PR #2184, opened 2022-03-08
Possible fix for broken login bug

### LINKED ISSUE #2183: [BUG] Login is broken in main
**Describe the bug**
Clicking the login button no longer works.

This seems to be an issue introduced by @fRedelaar in #2122, see:

image

image

fRedelaar:
**Description**
Earlier today a merged PR (#2122) caused a bug (#2183). This bug makes it impossible for users to log into Hedy. In this PR I've added a check whether "variableDiv" is not present. 

("variableDiv" is the yellow box used to present the variables and their corresponding values.)
 
**Fixes #2183 **

**How to test**
Can be tested by logging into Hedy. (I am unsure how to do this locally). 

_Edit 10 march: I've just got taught how to log into Hedy locally. The PR should fix the issue :)_


TiBiBa:
I think we can also solve this with some jquery, making the code easier and the code won't crash if the id doesn't exist. For example:

$('#variableDiv').show() should do the trick. (Or you can use toggle() to switch between hide/show).

Felienne:
> I think we can also solve this with some jquery, making the code easier and the code won't crash if the id doesn't exist. For example:
> 
> $('#variableDiv').show() should do the trick. (Or you can use toggle() to switch between hide/show).

Dude aren't you on vacation? :)

Felienne (approved):
Can confirm this indeed solves the issue, thanks @fRedelaar!

Issue #592, opened 2021-07-21
blank(_) in programs should get its own syntax highlighting


Felienne:

Felienne:
Ow and maybe this one is doable for you too @TiBiBa (would be a bit harder than the update to the existing rules because this needs a new strategy because _ can occur anywhere)

Felienne:
Just though of this @TiBiBa, maybe you can pick this up with #1709

thjazi39:
I'm currently working on syntax coloring, and I think it's quite possible to satisfy this issue.

It does require a little modification though, but it might help to standardize the conventions.
Indeed, in quizzes, since https://github.com/Felienne/hedy/pull/2019/commits/d1a964b0c2655d58e745336ae9ec7cc632213be4

blanks in programs are represented by the string `_?_`.

If it's ok with you, I can modify the `_` in the statements to make them follow the rules, and then adding syntax highlighting will be easier, since this sequence is "more complex" than just a `_`

thjazi39:
I find 140 blank in all files in adventure (with `grep -Re "\b_\b" | wc -l` )
I will replace them with `_?_`

Felienne:
> I find 140 blank in all files in adventure (with `grep -Re "\b_\b" | wc -l` ) I will replace them with `_?_`

Hmm I think I would prefer it the other way around, I find `_?_` a bit heavy in the snippets.

But why not both, they do serve different purposes? I think it is fine for the quizzes to keep `_?_` but snippets to have `_` or is there a reason that thay will make highlighting harder?

thjazi39:
I understand better that the uses of the 2 strings are not exactly the same, so it's not necessarily a problem if there are 2 conventions that coexist (the syntactic coloring only has to recognize the 2 and we will use one or the other according to the context)

However, I still find that the underscore alone is maybe a bit too light or too common.
In the following example, I imagine a quiz where the goal is to find the writing of the `print` command, however, we will not be able to distinguish this case from a compound variable.

![image](https://user-images.githubusercontent.com/48509466/156563888-d723d222-712e-4ab7-b42f-449da0ea675c.png)

(the regex does indeed work with ``b_'', which detects the beginning and end of words, but as long as the underscore is not alone, it will not be detected)

a more 'exotic' symbol could solve this problem because there would be no need to look around

thjazi39:
> > But we can also imagine a preprocessing during the interpretation, and which, if we detect this string, displays a message like "Attention, there are still holes to complete in the program".
[...]

PR #2185, opened 2022-03-08
Uniformisation blank in quizzes and in adventures

thjazi39:
As said in issue #592, a little standardization of syntax files

Uniformity of blanks in the codes:
- In quizzes, there were "\_\_\_" left (whereas we want "\_?\_"
- in adventures, there were "\_\_" or "\_\_\_" (instead of "\_")


coursedata/quizzes/el.yaml
```diff
@@ -62,16 +62,16 @@ levels:
             question_score: '10'
         10:
             mp_choice_options:
```
Felienne, line 64: These texts are generated by Weblate. I guess to minimize conflicts, we want to remove these form the PR and just make the change in En and the actually translated quizzes (such as nl, for example)

thjazi39:
Ok, so I have to report the changes to the site that does the translations?
There are only 115 changes to make, I think it will be quick to do by hand

Felienne:
> Ok, so I have to report the changes to the site that does the translations? There are only 115 changes to make, I think it will be quick to do by hand

No that will go automatically but smaller changes are better. So if you can just remove all quizzes that are 100% copies on En from the PR (for example Greek, but I do not know then all by heart), everything will be fine!

thjazi39:
Ok, I think this is better now

Felienne (approved):
Hope this does not upset Weblate 🤞

PR #2196, opened 2022-03-10
Improved the Dutch levels 1 2 and 3

MarleenGilsing:
I hope I didn't break it by using |- instead of the \n... We'll see...


Felienne (approved):
Let's see if Weblate gets upset :)

PR #2200, opened 2022-03-11
fix typo hounted -> haunted

digitalica:
Just  a typo fix, hounted -> haunted

(hope this wasn't intentional, its just a text change in level 2 of the Haunted house story) )


Felienne (approved):
Thanks for the fix! I am not sure Weblate will be happy about the updated copies but let's see

Felienne:
Thanks for your first contribution @digitalica, welcome to Hedy!!

PR #2201, opened 2022-03-11
Make validate running on Windows dev machine

Mark-Giesen:
**Description**

On a windows machine the validate script produced errors due to \/ and codepage differences.

Instead of filename splitting on /, python now handles this. Atomic writes on windows might go wrong, but error is swallowed. Finally an encoding is passed to the Open() function, this might be smart on all occurences.


Felienne:
@TiBiBa Do you happen to have an old windows machine laying around to test this on?

TiBiBa:
> @TiBiBa Do you happen to have an old windows machine laying around to test this on?

Unfortunately not. I can ask around if I can borrow one somewhere, but might be some work to get the code running. Would be ideal if someone who already has Hedy running locally (and using windows) can try it out.

Felienne (approved):
Seems to not break anything for Mac users, so I assume @Mark-Giesen verified it works on Windows 🤣

Issue #2104, opened 2022-02-25
[BUG] Program count of per user type metrics are incorrect


boryanagoncharenko:
**Describe the bug**
Currently the admin dashboard shows incorrect graphs for the program runs per user type. On the left the graph indicates that there were no students performing program runs and on the right, the number of runs is extremely low.

**Add a screenshot (optional)**
![image](https://user-images.githubusercontent.com/3010723/155708231-53e04ba3-ab73-4d08-98d9-004790bc7b36.png)

**Expected behavior**
The two graphs (per level and per week) should show the same amount of calls per user type.

PR #2202, opened 2022-03-11
Fix incorrect user type metrics #2104

boryanagoncharenko:

< Address incorrect user type metrics >

**Description**

<
Fix bug in user type metrics per week.
Fix incorrect week number format causing wrong comparison in Dynamo Between condition
> 

**Fixes < #2104 >**

**How to test**

Run Hedy locally, login as administrator and open the /admin/stats page. Also, ensure there is sample data that you can see.
- Check that all options for selecting period work: This week, past 2 weeks, past 4 weeks and past year. They should not result in 500 errors and they should return correct values (which highly depends on the sample data you generated).
- Go to the charts with names **User type per level** and **User type per week**. Ensure there are 4 series: anonymous, logged-in, student and unknown user types. The unknown user type refers to the previous data collected before we differentiated user types.
- The graphs **User type per level** and **User type per week** should show equal total number of users.

**Checklist**
  

Felienne:
Seems to be in order using the little local test set I have

Felienne (approved):

PR #2205, opened 2022-03-14
[CLEAN-UP] Removes all existing tokens when creating a new one

### LINKED ISSUE #2165: [CLEAN-UP] Clean-up the current token structure as the functionality is unclear
**Description**
We currently use tokens to automatically log users in on the back-end if we find a still existing cookie on the front-end. Really user-friendly, but in this case the token should be removed from the database. This all doesn't seem to work as expected as we have a lot of token values in the database. We should figure out the current functionality and what can be removed.

TiBiBa:
**Description**
In this PR we clean-up the current tokens structure. As it is (still) unclear to me what the exact functionality purpose of the tokens is I might need some help. If I understand correctly it works as follows: each time a user logs in we create a token. Then if the user opens the browser again and there is still a long-lived cookie (with a token) we verify if this token actually exists in the database. If this is the case we automatically log the user in, this way remembering the logged in user. Lastly, we also use the token as the verification for "password reset" as this token is sent in the mail template and can be used to reset the password.

So far so good. But our database currently exists of `5.096` tokens! We don't have that many users, so there are duplicate entries. But for what purpose? Currently we only delete a token on `logout` or `destroy`. So the following situation occurs: When a users logs in multiple times from multiples devices we have multiples tokens! As I always use incognito mode for developing my account alone creates a lot of tokens. I assume this is undesirable as a lot of them will never be removed. I propose the following change in this PR:

When a users logs in we remove all existing tokens for this user. Keeping track of only one token per user. Which can still be used to serve a the long-lived cookie verification but prevents the overflow of redundant storage in the database. Again, I'm still nog 100% sure on this part of the code base so some pointers of the creator (maybe @fpereiro @rix0rrr ?) would be really helpful!

**Fixes**
This PR fixes #2165

**How to test**
This is clean-up task, on the front-end no changes should be visible. Verify the correct removal of old token by checking the `dev_database.json` file when logging in or out.


rix0rrr:
You got it. I'm not the original author, but I think your analysis of how these tokens are used is spot on.

There are two changes in behavior that this PR would cause, one probably not-so-impactful and one perhaps more impactful:

1. It is no longer possible to have a password reset token as well as a login token at the same time. *Probably* not a big issue, as you probably wouldn't be logged in, or able to log in, if you need your password reset. 
2. It is no longer possible to be logged in on multiple devices. Let's say, a home computer and a work computer. Or a computer and a mobile phone or tablet. The second login would delete the token from the previous login (thereby invalidating the cookie on the other computer, and next time you visit the Hedy website there you would need to login again).

I'm not sure the 2nd behavior is desirable... it runs counter to how persistent logins work on most websites.

What do you think?

rix0rrr:
I see the need to clean the database. 6000 records is not a great deal, but garbage that accumulates in the database needs to be collected eventually.

There is no obviously best solution, but here are two that occur to me:

- Instead of having a token per login (and hence, a token per device), we could instead have a token per user (maybe even store it in the `user` table instead of a separate `tokens` table), and hand out the same token for every machine the user logs in on. That way, users only ever have one token, and there can be no garbage, because there can not be unused tokens. Feels like there is a loss of security here; I'm having a hard time quantifying how big.
- We can garbage collect the tokens. We can add an expiry timestamp to each token (should be the same lifetime as the cookie on the user's machine, which is [1 year](https://github.com/Felienne/hedy/blob/b3d1203ce71eed611dee3e10f0812bb3d35e5d7b/website/auth.py#L329)), and delete them once the timestamp passes. [DynamoDB can do this automatically](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html). The downside of this is that we still accumulate garbage, and can only delete it after 1 year (either that, or we force people to log in more often by decreasing the cookie lifetime).

Neither of these are obvious winners.

TiBiBa:
> You got it. I'm not the original author, but I think your analysis of how these tokens are used is spot on.
> 
> There are two changes in behavior that this PR would cause, one probably not-so-impactful and one perhaps more impactful:
> 
> 1. It is no longer possible to have a password reset token as well as a login token at the same time. _Probably_ not a big issue, as you probably wouldn't be logged in, or able to log in, if you need your password reset.
> 2. It is no longer possible to be logged in on multiple devices. Let's say, a home computer and a work computer. Or a computer and a mobile phone or tablet. The second login would delete the token from the previous login (thereby invalidating the cookie on the other computer, and next time you visit the Hedy website there you would need to login again).
> 
> I'm not sure the 2nd behavior is desirable... it runs counter to how persistent logins work on most websites.
> 
> What do you think?

Thanks for your reply! I think the first change in behavior is actually desirable. I like the idea that the user is logged out everywhere when changing their password. Might be due to someone else knowing the password, an undesirable / forgotten login on a device of another or any similar situation.

I was also thinking about the second situation when making this PR. Not sure what would be the best solution. We might be able to run a script on the server, deleting all tokens older than x days? Or we can delete all tokens except the last x entries (sorted by the timestamp). Both are not ideal, but from a user perspective more friendly I guess. We also have the issue of "never returning" users, keeping their entry in the db (while the cookie in their browser is long-gone). For this situation the first solution might be most feasible.

rix0rrr:
An even better alternative solution would be to not store the token itself. In the ideal situation, we would give a token to the user in a cookie, and not store a copy of the token on the server.

In order to do that, we would probably need to do some cryptographic operation to prove that the token is correct, without storing the actual token or even data per token on the server.

I suppose it could be something like: 

```
SENT TO CLIENT:
token = a pair of (salt, digest)

VALIDATE ON SERVER TO REAUTHENTICATE:
hash == H(salt ++ password)
```

Or something. `H` should be some hashing function, preferably a secure one (like an HMAC).

There are probably already schemes like this we could be re-using; we shouldn't be rolling our own here.

rix0rrr:
>  I like the idea that the user is logged out everywhere when changing their password

Yes, but wait up! In this case, the user would be logged out anywhere *not* "when they change their password", but instead "when they request a password reset".

That is a subtle distinction, and in fact is a possible attack (annoyance) vector, as people can typically request password resets on behalf of other people.

TiBiBa:
> > I like the idea that the user is logged out everywhere when changing their password
> 
> Yes, but wait up! In this case, the user would be logged out anywhere _not_ "when they change their password", but instead "when they request a password reset".
> 
> That is a subtle distinction, and in fact is a possible attack (annoyance) vector, as people can typically request password resets on behalf of other people.

Didn't think about that one! We can fix this by not removing the tokens on the request but on the actual reset itself. The rest of the implementation can stay the same. I agree that the "same token everywhere" solution sounds like a security risk, not sure to what extend but it just doesn't feel safe and I will have to look into the possible risks. Will take a look into your suggested hashing solution! This will work for remembering a logged in user, but we would still need the tokens for password reset. I think the following implementation would be fine (for now):

1. Remove all tokens on login
2. Create a new one and store it

1. Remove all tokens on password reset (not on request)
2. Also store a new one (and automatically login the user)

fpereiro:
Hi @TiBiBa @rix0rrr !

As you both well said, the functionality of tokens serves two purposes:
1) Maintain a list of valid cookies that haven't expired yet.
2) Perform a recover/reset password operation.

If I remember correctly (and I reviewed also the code just now to be sure), we do garbage collect tokens as soon as they expire. No cookie token (the ones created for case 1) will outlive the intended duration of the cookie itself. So, if I understand correctly, there's no risk of the database being never cleaned up. The tokens for password recovery used to have a shorter period of durability (if I remember correctly, 24 hours or so) but it was changed later to last as long as the cookies, which is probably a mistake during a refactor (could probably be mine!).

I would dare suggest that if tokens are currently garbage collected automatically through the ttl setting we use on DynamoDB, no change is better. If we change this, then a user could not be logged in on multiple devices, which seems like an annoyance and it also is not the behavior of most web applications. The cost to us would be relatively minimal, just a few tokens (each of them perhaps 50 bytes?) per user session, which also gets garbage collected after a few weeks. But then again, it might not be necessary to allow for multiple sessions for a given user.

I do suggest not making recover tokens influence cookie tokens, for the reason @rix0rrr outlined above: if anyone (not even a logged in user) requests a password recovery for user's A account, user A will immediately be logged out of all of their current sessions!

Hope that was clear enough, if not let me know!

fpereiro:
I forgot to clarify something important: in the original implementation, multiple login tokens for a given user were allowed; but making a new recovery/reset password token for a user deleted her/his previous recovery token.

TiBiBa:
> I forgot to clarify something important: in the original implementation, multiple login tokens for a given user were allowed; but making a new recovery/reset password token for a user deleted her/his previous recovery token.

Hi @fpereiro ! Thanks for your reply and clarification. If the tokens are indeed automatically cleaned up by Dynamo the best solution might be to simply do nothing. However, as we don't differentiate between normal tokens and recovery tokens I don't like the idea of a user being able to request multiple recovery tokens which will be valid for a whole year. This seems like a security risk. My proposed change is as follows:

- Keep the current structure (if they are indeed automatically removed in Dynamo)
- Remove all user token on password reset (not password request!)

This way the user is logged out everywhere on the moment the password is actually reset. Also solving the issue of possible multiple recovery tokens being around. What do you think? @rix0rrr

Felienne:
> This seems like a security risk. My proposed change is as follows:
> 
> * Keep the current structure (if they are indeed automatically removed in Dynamo)
> * Remove all user token on password reset (not password request!)
> 
> This way the user is logged out everywhere on the moment the password is actually reset. Also solving the issue of possible multiple recovery tokens being around. What do you think? @rix0rrr

Sounds good to me!

rix0rrr:
> However, as we don't differentiate between normal tokens and recovery tokens I don't like the idea of a user being able to request multiple recovery tokens which will be valid for a whole year. 

> Remove all user token on password reset (not password request!)

All sounds good to me as well, but I do want to point out that the remedy you proposed doesn't *necessarily* address your concern about the lifetime of recovery tokens. It only addresses it if the user completes the reset flow. That is *probably* good enough, but just wanting to be thorough here.

As an additional measure, you can also choose to have the lifetime of the recovery tokens be shorter than a year. For example, we could set the TTL back to `now + 1 day` (instead of `now + 1 year`) for recovery tokens, the way it used to be, thereby reducing the exploitation window by a lot. 

As I said, it's a minor thing and I definitely wouldn't spend too much time on it, but it might be a nice cherry on the cake. I agree completely with your other plans.

Thanks for chiming in @fpereiro !

TiBiBa:
> > However, as we don't differentiate between normal tokens and recovery tokens I don't like the idea of a user being able to request multiple recovery tokens which will be valid for a whole year.
> 
> > Remove all user token on password reset (not password request!)
> 
> All sounds good to me as well, but I do want to point out that the remedy you proposed doesn't _necessarily_ address your concern about the lifetime of recovery tokens. It only addresses it if the user completes the reset flow (which they might not). That is _probably_ good enough, but just wanting to be thorough here.
> 
> As an additional measure, you can also choose to have the lifetime of the recovery tokens be shorter than a year. For example, we could set the TTL back to `now + 1 day` (instead of `now + 1 year`) for recovery tokens, the way it used to be, thereby reducing the exploitation window by a lot.
> 
> As I said, it's a minor thing and I definitely wouldn't spend too much time on it, but it might be a nice cherry on the cake. I agree completely with your other plans.
> 
> Thanks for chiming in @fpereiro !

Thanks for your reply! I just took a look at `config.py` to figure out the current `session_length` value. It seems we set this value in the session to `60 * 24 *14 = 20.160 minutes`. But then in `auth.py` we use:

`session_length = config['session']['session_length'] * 60` (I think to get it to seconds?)

The variable `session_length` is then used for all TTL on the tokens by combining it with the current `times()`. So according to this the tokens should only live for 2 weeks, making it a bit unlikely that they are automatically removed regarding the large amount of rows in the db. Making it a bit weirder: the front-end cookie is indeed set to 1 year until expire. I'm getting a bit confused!

TiBiBa:
Ah! Just found the settings in Dynamo:

Schermafbeelding 2022-03-16 om 14 09 17

Enabling this should automatically remove the tokens in 2 weeks, solving our issues. (We might want to align the front-end cookie expire time). The token removal on password reset is nothing more than a "nice to have".

website/database.py
```diff
@@ -149,6 +149,9 @@ def get_token(self, token_id):
 
     def store_token(self, token):
         """Store a token in the database."""
+
+        # But first, remove all already existing tokens for this user
+        TOKENS.del_many({'username': token['username']})
```
Felienne, line 154: I think if we use the expiry date, this is no longer needed?

Felienne (changes requested):

website/database.py
TiBiBa: I will have to take a look at this all together! Will have to make some changes so the reset tokens are handled differently than the login tokens. Will put it into a draft until I figured this out.

TiBiBa: When look at this we can indeed definitely remove this! The removal should happen on the database-side. The only moment this should be called is after the password reset. Please help me remember to index the usernames on the database (and turn on the TTL setting) in Dynamo before we merge this.

TiBiBa:
I think this is ready for merge! When merging we have to take the following two actions into account:

- Add username as index key to tokens table
- Enable TTL on table

Felienne (approved):

PR #2206, opened 2022-03-14
[CLEAN-UP] Simplified saving and sharing

TiBiBa:
**Description**
In this PR we clean-up the structure for saving and sharing programs. Currently we only verify the correctness of programs when sharing one. But, why should we? We simplify our code by always parsing the code on save, independent of sharing or not. This way we reduce the function calls as we can always call the `saveit` function, only with an additional value for sharing or not.

**Fixes**
This fixes nothing, but cleans-up.

**How to test**
Make sure you are logged in and verify that all saving and sharing functionality is still working.


TiBiBa:
@Felienne this one is ready for review!

website/programs.py
```diff
@@ -69,22 +70,32 @@ def save_program(user):
...
         program_id = uuid.uuid4().hex
-        program_public = 0
+        program_public = body.get('shared')
         overwrite = False
         for program in programs:
             if program['name'] == body['name']:
                 overwrite = True
                 program_id = program['id']
-                program_public = program.get('public', 0)
+                program_public = program.get('public', program_public)
```
boryanagoncharenko, line 98: This line means that I cannot share a program that was only saved but not shared. This is a deviation from the current behavior. Here are the steps:
- save a program (regardless if it is with errors or not)
- after the program is saved, click Save&Share and approve the override
- go to the my-programs page and check that the program is updated but not shared

boryanagoncharenko (changes requested):

website/programs.py
TiBiBa: Will take a look!

TiBiBa: Fixed!

TiBiBa:
This should be ready for merging!

PR #2207, opened 2022-03-14
[FEATURE] Implements dynamic (printable) cheatsheets

### LINKED ISSUE #2198: Generate cheat sheets
In #2193 we add really nice cheat sheets but they are made manually, wouldn't it be nice if those can be generated from the examples in the yamls for all languages?

TiBiBa:
**Description**
With #2193 @MarleenGilsing made some great cheatsheet that can be used as reference material for kids. However, as they are .pdf files we are unable to dynamically change them dependent on language and keyword language. In this PR we implement a dynamically created cheatsheet structure, automatically generating them based on the language and keyword language. Currently we use the `level-defaults` to create the files but as they differ from the current cheatsheet and don't give that much extra information it might be useful to create some dedicated cheatsheet YAMLs.

**Fixes**
This PR fixes #2198

**How to test**
Navigate to `/cheatsheet/<level>` and verify that the correct level-defaults are shown. Also verify that the language and keyword language are correct based on the user preferences.

Schermafbeelding 2022-03-14 om 13 09 17


Felienne:
> One small issue: I just noticed when trying to print the page using the built-in print function (in my case Safari), it doesn't look that pretty. If we aim to also make this easily printable we might have to make some changes to this PR.

Safari is becoming our sworn enemy I feel 🤣

Felienne:
I also found no issues in Safari? 

image

(on Mac)

TiBiBa:
> I also found no issues in Safari?
> 
> image
> 
> (on Mac)

Viewing works fine, but printen seems to give an issue:

Schermafbeelding 2022-03-16 om 13 10 35

TiBiBa:
Will solve the merge conflicts and then I think we are ready to merge! Users won't notice anything unless they navigate directly to the url. We do have to think about a nice way to implement and return this to the users. For example with dedicated cheatsheet YAMLs and/or share-able links for teachers.

Felienne (approved):

PR #2210, opened 2022-03-14
Improve UI to let user know they are in single-adventure mode

### LINKED ISSUE #2146: [UI idea] Make it clear to users when they are in one adventure mode
**Idea incl level**
With the merging of #2121 we enable a "single adventure mode". However, we want to make it clear to the user that they can also leave this mode if they want too. It would be nice if we differentiate a bit between the normal and one adventure mode. This is merely a front-end task as the back-end structure is fine as it is.

TiBiBa:
**Description**
In this PR we add a text to the adventure tabs view to inform the user when they are in single adventure mode.

**Fixes**
This PR fixes #2146

**How to test**
Navigate to single adventure mode. For example by navigating to `/adventure/turtle`. Verify that the string given in the adventure tabs is correct and only appears when in this mode. When leaving the mode the string should disappear.

Schermafbeelding 2022-03-14 om 15 30 59


Felienne:
It seems the red text is a link (I get a hand) but it does not seem to link anywhere?

image

And maybe we want to make the text more actionable? Like: "Click Hedy in the top bar to see the full Hedy?" or "Click this text to go to the full Hedy?"

TiBiBa:
> It seems the red text is a link (I get a hand) but it does not seem to link anywhere?
> 
> image
> 
> And maybe we want to make the text more actionable? Like: "Click Hedy in the top bar to see the full Hedy?" or "Click this text to go to the full Hedy?"

Whoops! Will take a look, it shouldn't be clickable. Will add something like click "Hedy" on the menubar to view all adventures.

Felienne (approved):
Better like this!

PR #2211, opened 2022-03-14
Allow text in arithmetic expressions #2067 #2107

### LINKED ISSUE #2067: [BUG] error when using [+-*/] in quoted strings in level 6 and up
This code works fine at level 6, but if you introduce an intermediate variable, then it doesn't work anymore.
```
print '3*10'
```
=> work
```
a is '3*10'
print a
```
=> a is not a Hedy level 6 command. Did you mean ask?

This code works correctly at level 5, but does not work from level 6

Moreover, we get the same error with the symbols + - / *

### LINKED ISSUE #2107: [BUG] Arithmetic operation with a quoted string as an argument cannot be parsed
**Description**
At level 6 the following programs result in an errors because only "var" can be parsed in arithmetic operations.
```
a is 'test'
print a + 'Test'
```
```
a = Hedy + 'Test'
print a
```

Note that this was explicitly removed by this commit: https://github.com/Felienne/hedy/pull/118/commits/4aba6c2f4d566f5e98e72847e616a4a8d5d34752

**Expected behavior**
Instead of a parse exception, the user should receive a type error.

boryanagoncharenko:
Allow text in arithmetic expressions

**Description**

< This PR is am attempt to address 2 issues related to expressions and strings. The reason behind this is that they share a common root cause and resolving them separately will be more challenging them addressing them together. This PR introduces the following set of changes:
- Ensure that arithmetic expressions actually contains an arithmetic operator. A single atom cannot be an expression anymore.
- Relax the grammar of level 6 expressions to contain text-without-spaces, so that programs like `a is '3*10'` can be parsed. The program results in a type error because the program is parsed into a multiplication of the values '3 and 10', which are string values.
- Relax the grammar of level 12 expressions to contain quoted strings.

**Fixes #2067 and fixes #2107**

**How to test**

1. Examine the grammar changes. Do they correctly represent the behavior we want? Can the grammar rules be simplified?
2. Examine the tests. Do they cover all corner-case scenarios?

**Checklist**
  

grammars/level12-Additions.lark
```diff
@@ -1,8 +1,8 @@
 //assigning a string now must be done with quotes
-assign: var (_SPACE _IS _SPACE | _EQUALS) sum | var (_SPACE _IS _SPACE | _EQUALS) text_in_quotes
+assign: var (_SPACE _IS _SPACE | _EQUALS) (atom | expression)
```
Felienne, line 2: nice, this is a lot cleaner and clearer!

grammars/level16-Additions.lark
```diff
@@ -3,7 +3,7 @@
 // TODO for tomorrow: this stills needs tests in level 16 for > and !=
 ?comparison_arg: var | text_in_quotes | NUMBER | list_access
 
-?atom: NUMBER | var | list_access | error_unsupported_number
+?atom: NUMBER | var_access | list_access | text_in_quotes
```
Felienne, line 6: I am not sure I understand why we are dropping `error_unsupported_number`?

Felienne: Ah I see, since we don't have unsupported numbers anymore! Not sure how this ended up in level 16? Bad case of the copy-pasties I guess?

Felienne (approved):
Great work @boryanagoncharenko! Nice cleanup and two bugs fixed with one PR.

PR #2219, opened 2022-03-15
[BUG FIX] syntax highlighting works again on public profile

### LINKED ISSUE #2214: [BUG] Syntax highlighting broken on public profile
**Describe the bug**
The syntax highlighting does not seem to work on the public profiles while the ace editors are correctly rendered. Might have something to-do with the `level` attribute not correctly being set in this case.

TiBiBa:
**Description**
This is a very small fix where we fix a forgotten change on the level attribute for the ace editor. Previously we used the `id` of a `pre` element to indicate the Hedy level for which we needed syntax highlighting. As this is a bit against the HTML best-practices we changed this to added a `level` attribute to the element. However, this change was forgotten on the public profile page.

**Fixes**
This PR fixes #2214

**How to test**
Navigate to a public profile with programs and verify that the syntax highlighting is working.


Felienne (approved):
Fixed indeed!

image

PR #2225, opened 2022-03-15
Reversed default customizations settings

### LINKED ISSUE #2224: Reverse default in teacher customization
I find it a bit confusing that this is the view when I start with my first customization:

image

It is not wrong, I mean the red text says that there are no customizations but it is also weird cause it looks like all levels are closed and all adventures are turned off. Couldn't we start like this?

image

That is more aligned with what the status is now. We can then drop the red text and if the teacher happens to make no changes, just save nothing in the DB.

TiBiBa:
**Description**
In this PR we change the structure of the class customizations. Currently be default alle levels and adventures are not selected, requiring the teacher to manually select these. To provide some additional feedback a red banner was show with a "no customizations yet" warning. We reverse this by setting all levels and adventures selected by default, making the banner redundant.

**Fixes**
This PR fixes #2224

**How to test**
Make sure you are a teacher and have a non-customized class (just create a new one). Navigate to the customizations screen and verify that all levels and adventures are selected by default. The functionality should be identical.


Felienne:
Thanks, this is a lot better. One other small thing, when I make my first class, I get the Achievement (nice) but than it takes quite a while for me to be redirected to the class page, so long that it might make teachers wonder if all is ok. Would it be terribly hard to first redirect and then show the achievement modal (might be in this PR)

TiBiBa:
> Thanks, this is a lot better. One other small thing, when I make my first class, I get the Achievement (nice) but than it takes quite a while for me to be redirected to the class page, so long that it might make teachers wonder if all is ok. Would it be terribly hard to first redirect and then show the achievement modal (might be in this PR)

I think this is pretty hard because the achievement is returned as a response to the creation POST. We have this issue in several places where the achievement is related to an action that requires re-direction or a re-load of the page. I can take a look at solving this, but currently don't have a solution that comes to mind. Might be better to do a big overall achievement showing re-fresh to solve this.

Felienne:
Ah ok, then let's merge this and make an issue for the achievements! it is not a big deal but I can see how it would be confusing for teachers (a very simple way to fix this also is to drop this achievement, it is not all that important to have it?)

Felienne (approved):

PR #2237, opened 2022-03-16
[BUG FIX] Programs age is displayed correctly again

### LINKED ISSUE #2236: [BUG] Program age returns a negative time
**Describe the bug**
For some reason the "last edited" timestamp of a program returns a negative value. Weird!

Schermafbeelding 2022-03-16 om 14 45 49

**Expected behavior**
The "last edited" value of a program should be correctly displayed.

TiBiBa:
**Description**
In this PR we remove a weird YAML key only containing a "-" that made it appear like the shown age of a program was a negative value. Not sure why and how this key only contains the "-" value. For some languages it does contain an actual word, but not for either English or Dutch. We remove it in this PR, but that might not be the best way to go.

@Felienne @Mark-Giesen might did have something to do with the changing in the translations? I can't remember the purpose of this "ago-1" key. This code is from before my time😁. 

**Fixes**
This PR fixes #2236

**How to test**
Navigate to your program page and verify that the unclear "-" symbol is removed from the "last edited" value.


Felienne:
Ah yeah that does make sense, like we had with the "go to level X" text in Hindi!

TiBiBa:
> I vaguely remember we used something like a template there, can we do that here too?

We did use something like a template! If I remember correctly we created the translated string such as:

`go to level {level}`

Whereas we replace `{level}` on runtime with the correct value. Enabling each language to place the `{level}` placeholder wherever they want. We should be able to use the same idea here, such as:

`{hours} ago`

Felienne:
Yes that seems like the way to go here too, thanks for the catch @boryanagoncharenko!

Felienne (changes requested):
Let's go with the template approach

TiBiBa:
> Let's go with the template approach

Fixed!

Felienne:
Can we merge this is light of the Babel stuff?

TiBiBa:
> Can we merge this is light of the Babel stuff?

Unfortunately not... we have to first merge the Babel branch and then fix the merge conflicts on this one. Will take a look after merging!

Felienne:
> > Can we merge this is light of the Babel stuff?
> 
> Unfortunately not... we have to first merge the Babel branch and then fix the merge conflicts on this one. Will take a look after merging!

ok!

Felienne (approved):

Felienne (approved):
Ah interestingly this reverses numbers in weblate, hope it does not get angry!

PR #2238, opened 2022-03-16
[FEATURE] Enables teachers to hide/show the keyword language switcher

### LINKED ISSUE #2009: [UI idea] Enable teachers to enforce keyword language in class
**Idea incl level**

**Additional context**

TiBiBa:
**Description**
In this PR we implement a new feature for class customizations. As requested by @OnnoEbbens this enables teachers to hide/show the keyword language switcher in their class. I like to think of this as a temporary solution as we want to prevent the customizations page from getting to crowded and have to think about a more user-friendly solution which also makes the switcher less intrusive while coding.

**Fixes**
This PR fixes #2009

**How to test**
Make sure you are teacher and student of the same class. Navigate to the customizations page and verify that there is a new option on the "other settings" section. Select the "hide keyword language switcher" and verify that the switcher will always be hidden on the code page. Independent of language, keyword language or any other setting.


boryanagoncharenko:
I stumbled upon a bug that, unfortunately, I cannot reproduce anymore. For some reason, the customisations screen did not pre-create information about any levels. As a result I saved in the DB an empty a structure with levels: [] which later caused an Unauthorized error. Since I could not reproduce it, I assume this is related to the changes to the customisations structure and a customisation state I have saved locally.

Issue #1875, opened 2022-02-01
Translate commands in error messages to the current language


boryanagoncharenko:
Hedy supports the keywords of programs to be in different languages (e.g. `imprimir` instead of `print` when the chosen language is Spanish). However, the error messages always use the English names of the commands. Currently some of the error messages are very specific and they point to parts of the code. When the command name is not translated, it does not match the program code and, therefore, could make the error message hard to understand.

For example the following program uses the command `imprimir` in Spanish but the error says `print`:
![Screenshot 2022-02-01 at 18 54 56](https://user-images.githubusercontent.com/3010723/152015027-1f94b415-26d5-4a4f-ab68-be1f9825089b.png)

Hedy level 3:
```
animales es perro, gato, canguro
imprimir animales
```

boryanagoncharenko:
The type validator visitor contains metadata, so we know where the command appeared in the text. Maybe we could try to correlate the command at hand with the token used in the concrete syntax tree which we can get the same way we do for translating programs? Sounds a bit far fetched and probably too expensive, but I think it is worth trying. I will timebox this to 1 hour and if I don't have anything promising, I will leave this task for later.

Felienne:
Ah thanks for reminding me of this one! It would be great if we can try to get this to work!! Since we have the tokens it might be possible indeed?

Alternatively, profiles can be a source of information maybe; kids can now set the language in their profile so we know what their default is (but they can still always use English so we don't know for sure)

PR #2243, opened 2022-03-16
Use concrete command syntax in invalid type and invalid type combination error messages #1875

boryanagoncharenko:

< Use concrete command syntax in invalid type and invalid type combination error messages #1875 >

**Description**

< 
- Update the text of the invalid type and invalid type combination commands, so that they reflect better their dynamic content.
- Populate the error messages with the concrete keywords used in the program.
- Add type validation for repeat and for-in commands.
- Add tests.

**Fixes < #1875  >**

**How to test**

<
Run Hedy locally, set the language to ES and execute the following code in level 3. Does the error message text give a clear indication of what is wrong and a suggestion of how to address it?
```
animales es perro, gato, canguro
imprimir animales
```
Also run the following in level 6. Is the message text clear when there is no keyword but a sign (+)?
```
name = 'test'
b = 5 + name
```
At level 12, check the error message for invalid type combination. Again, is the message clear?
```
test = 'bla'
print 2.5 + test
```

Now try using commands with multiple keywords, e.g. at random in level 3. Does the error correctly refer to the concrete syntax used? Is the error clear?
```
v is 1
print v at random
print v en aleatorio
print v at aleatorio
print v en random
```
 >

**Checklist**
  

Felienne:
Works, fantastic @boryanagoncharenko!!
image

Felienne (approved):

Issue #1898, opened 2022-02-03
Allowingِ Arabic/RTL (double) quotes 


alaaeddin-swidan:
**Language idea incl level**

The text quotes currently used for Arabic are actually the latin quotes e.g. ' hello '.  The request is to allow Arabic/RTL double quotes (starting from level 3) to overcome issues with quotes' direction and common use in RTL languages. 

 Example:
قول 'مرحبا' -- currently used quotes 
قول “مرحبا” --  latin double quotes
قول "مرحبا"  -- Arabic double quotes

PR #2249, opened 2022-03-17
Allowing double quotes in the grammar and highlighter 

alaaeddin-swidan:

Allows double quotes to be used with text arguments - Draft

**Description**

Adds the double quote to the level1.lark file by editing the _SINGLE_QUOTE value. Additionally, I modify the highlight through the regex value in syntaxModesRules.ts to highlight text surrounded by double quotes. 

**Fixes Issue #1898**

**How to test**

Starting from Level 4, the command 
```
print "hello world" 
```

should run without errors.


thjazi39:
There has been a major change in syntactic coloring.
Thus, the modification to allocate syntax highlighting is no longer valid.
So I took the liberty of adapting the string coloring to the new system:
image
On the other hand, you'll have to follow up on the quizzes too!

In part 4, there are questions about quotation marks, and that will have to be updated, in all languages!

Felienne:
> In part 4, there are questions about quotation marks, and that will have to be updated, in all languages !

You mean the level 4 quiz, right? Can you make an issue so @MarleenGilsing can look into this?

thjazi39:
image
I was thinking about this question in particular, but there are surely others that need to be corrected

Felienne:
> Please keep in mind that support for double quotes requires more changes to the grammars and the transpiler. At a quick glance:
> 
> * At level 4 the following definition `quoted_text: /'((?:[^\\']|\\.)*)'/ ` is used in the print command which means we will not support double quotes there.
> * At level 12 the grammar defines textwithspace by escaping single quotes. Double quotes must be added there too, otherwise the parse tree will not be correct.
> * At level 12 the transpiler strips down single quotes and replaces them with double to ensure the integrity of the transpiled f-string. If we want to support double quotes, we need to add the reciprocal code since now it will produce errors.
> 
> I believe there are more cases to consider and in my opinion the best way ensure the feature is stable is to add a double-quotes counterpart for every test that targets single quotes usage.
> 
> If the issue is more time-consuming than originally anticipated, I'd be happy to assist or even take over the task.

Yes this is a work in progress PR, and if you want to take it over, that would be very welcome @boryanagoncharenko!!

boryanagoncharenko:
I addressed the problematic double-quote cases and extended the tests, so that we have coverage of the main scenarios.
Along the way, however, I noticed something strange which I think we should discuss. @Felienne your input is needed here: https://github.com/Felienne/hedy/issues/2293

@alaaeddin-swidan @thjazi39 Is there another aspect of the PR that is not ready, e.g. highlighting? If not, please let me know and I will remove the Draft status of the PR.

thjazi39:
Concerning the syntax highlighting, I made modifications on this branch (commit 75e89fb and 4a4d62b) in order to allocate double quotes in strings with the new syntax highlighting paradigm.

Felienne (approved):
Seems to work incl highlighting!

image

PR #2251, opened 2022-03-17
Example codes clean up

MarleenGilsing:
All the example codes are on the right part of the screen now. And some general fixes to the adventures and level defaults.


Felienne:
Ah I think this addresses #2257 already!!

Felienne (approved):

PR #2253, opened 2022-03-17
semantics change level 1 turn

juliabolt:

< Provide a one line summary of your changes in the Title of the PR >

**Description**

< In level 1, `turn` could be used with `left` or `right` and then an integer, or with nothing, and if I remember correctly also with just an integer. Now, in level 1, `turn` can only be just with only `left`, `right`, or nothing. So not yet with integers (this changes in level 2 already, when `left` and `right` are not used anymore and `turn` is used with nothing or only an integer). This pull request is to update the semantics. > Use present tense without a subject to describe your PR, for example: "Adds translations of levels 1 to 12 to Polish"

**Fixes < n.a. >**

**How to test**

Documentation change only. See SEMANTICS.md file, under level 1, under commands and types the type for command `turn` changed, and one example of a correct program is deleted, where `turn` was already used with an integer. 

**Checklist**
  
### DIFF
--- a/SEMANTICS.md
+++ b/SEMANTICS.md
@@ -31,7 +31,7 @@ Level 1 supports:
   </tr>
   <tr>
     <td>turn</td>
-    <td>'right' + integer | 'left' + integer | empty</td>
+    <td>'right' | 'left' | empty</td>
   </tr>
   <tr>
     <td>echo</td>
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Good catch, we updated this indeed!

juliabolt:
thanks to tiemen actually ;)

PR #2256, opened 2022-03-18
Differentiaties lifespan of recover token from login token

### LINKED ISSUE #2255: Differentiate ttl between login and reset tokens
Currently we use database stored tokens to support remembered login. These tokens have a ttl (time to live) of 14 days. Dynamo automatically removes these when the time is expired. However, we use the same structure of tokens for the password reset. It would be nicer (and safer!) if we use a different ttl for these tokens such as 4 hours.

TiBiBa:
**Description**
We did some digging and I think we figured out the exact functionality of our login tokens. Currently we set a front-end login cookie to expire in **1 year**. While the database saved cookie only has a lifespan of **14 days**. This is no problem as the front-end cookie is verified by the database if it is still valid. When a user logs out we remove the corresponding token in the database, if not the token is automatically removed by Dynamo in 14 days. (This function was turned off, but fixed again together with #2205.) The same lifespan is used for the reset tokens.

In this PR we change this by given the reset tokens a dedicated lifespan, of only 4 hours. We also update this in the mail template to users are aware that this is happening. One last security issue I'm struggling with: If the user is able to retrieve their current login token from the front-end cookie, this one can also be used to reset a password. Do we think this is a problem?

**Fixes**
This PR fixes #2255.

**How to test**
This is a complete back-end test, nothing in functionality should have changed. You can however test it by emptying your `tokens` table in `dev_database.json` and login. Verify that there is one new token available and copy the `ttl` value to a website such as https://www.unixtimestamp.com and verify that the lifespan is indeed 14 days. Then navigate to the `/recover` page and request a new password. (This should work locally with a token return). Verify that there is a new entry in the db with only a ttl of 4 hours.


website/auth.py
```diff
@@ -21,7 +21,11 @@
 import hashlib
 
 TOKEN_COOKIE_NAME = config['session']['cookie_name']
+# The session_length in the session is set to 60 * 24 * 14 (in minutes) config.py#13
+# We multiply this by 60 to set the "normal" session_length to 14 days
+# We than divide this value again by 84 to set the reset_length to 4 hours
 session_length = config['session']['session_length'] * 60
+reset_length = session_length / 84
```
rix0rrr, line 28: Why not have a separate config value?

TiBiBa: That makes more sense indeed, will fix!

rix0rrr (approved):

PR #2260, opened 2022-03-20
Adds ttl for class invitations to automatically delete after 2 weeks

### LINKED ISSUE #2258: Add ttl to class invitations
Currently class invitations don't have an expiration date. Users can deny these invitations, but otherwise they will live on forever. This is undesirable as one invitation "locks" the account, making it impossible to receive another one. It would be nice to add a ttl (time to live) value to the invitation so they are automatically removed after x days.

TiBiBa:
**Description**
In this PR we add a "time to live" value to the class invitations. Currently class invitations live forever, which can be a problem as a user can only have one pending invitation at the time. We add a dedicated `invite_length` value to the config file, setting it at **7 days**. We also perform some minor clean-up on the code by removing some redundant spacing and indentation.

**Fixes**
This PR fixes #2258

**How to test**
Make sure you are a teacher and navigate to any class overview. Invite a new student (that doesn't have a current invitation and is not in your class). Verify that a new "expiration date" column is shown on the invitations overview. 
**The other part we can only verify on Alpha:** 
We have to make sure that the ttl feature of Dynamo works correctly and automatically removes invitations after these 7 days.


Felienne:
@TiBiBa this has issues I don't dare to merge :)

TiBiBa:
> @TiBiBa this has issues I don't dare to merge :)

Merging conflicts should be solved! We do however have to perform some testing on alpha to make sure the TTL works as expected.

Felienne:
It now seems to have new issues :(

TiBiBa:
> It now seems to have new issues :(

Owh.. this is due to the mega merging of Babel. Will fix!

TiBiBa:
Should be fixed!

Felienne:
@TiBiBa This too can be tested on alpha now!

TiBiBa:
@Felienne I think it would be best if we merge #2276 before this one to make sure all is fixed. Then the merging conflicts can be solved on this branch.

Felienne (approved):

PR #2264, opened 2022-03-21
added syntax highlighting for blanks in programs

### LINKED ISSUE #592: blank(_) in programs should get its own syntax highlighting

thjazi39:
Added blank coloring in blank codes :

image

image

image

fix #592


static/js/syntaxModesRules.ts
```diff
@@ -714,6 +732,18 @@ function rule_string() {
 
 
 
+function rule_blank() {
+  return [{
+    regex: /\_\?\_/,
+    token: 'invalid',
```
Felienne, line 738: I am not really sure I like the word invalid here, can we use blank?

Felienne (changes requested):

static/js/syntaxModesRules.ts
thjazi39: Unfortunately I didn't choose this word at random.
It corresponds to a style defined by monokai, a style that has the property of being flashy.

I can modify the file `./static/vendor/theme-monokai.js` to try to create a style

Issue #1632, opened 2022-01-05
[CHORE] Clean-up TODOs in hedy.py


boryanagoncharenko:

Felienne:
Thanks for opening this @boryanagoncharenko! I see this as an on going issue, something that we continuously need to check upon, because knowing myself I will make more in the future haha!

boryanagoncharenko:
I was planning to resolve it now, but then I see there are more TODOs in hedy.py, so probably we should keep it open?

Felienne:
Yeah and some more in the grammar files that are all mine, so let's keep this open for now! Thanks for your help on your todos!

boryanagoncharenko:
Moved 2 TODOS from hedy.py to #1875 and #1876.
Removed 1 TODO which I don't remember what means :(

@Felienne shall I assign the ticket to you? Or do you want me to clear all of them from hedy.py?

Felienne:
> Moved 2 TODOS from hedy.py to #1875 and #1876. 

Thanks!

> Removed 1 TODO which I don't remember what means :(

Haha I have had a few of those :D 

> @Felienne shall I assign the ticket to you? Or do you want me to clear all of them from hedy.py?

It would be lovely if you can go through what is there and make issues for what you can understand!! I will then try to pick up the remainder. Ow and they are not just in hedy.py but also in the grammars :/

boryanagoncharenko:
Another small batch

Felienne:
What to do about this one @boryanagoncharenko? We can also close it?

PR #2268, opened 2022-03-22
Use var_access where intended in all grammar definitions #1632

boryanagoncharenko:

< Use var_access where intended in all grammar definitions >

**Description**

< 
The grammar definitions contain 2 identical rules var and var_access which are meant to be used for variable definition and variable access, respectively. In many cases, however, the grammar definitions were not using the correct rule, e.g. print is using var instead of var_access.

The changes in this PR ensure that var and var_access are used as intended in all grammar definitions. As a consequence, the check for undefined variables is moved from the Transpiler to the TypeValidator.

**Fixes < TODO in the grammars #1632 >**

**How to test**
< ​
- Validate that the grammar files use var_access and var as intended.
- Because the TypeValidator now has the ability to check for undefined variables, many errors in level defaults and tests were revealed. For example, missing single quotes in if-conditions above level 12 or references to undefined vars which suspiciously look like a typo. It seems that the examples did not comply to the intention of the grammars but worked by accident. Please examine the changed code in levels and check that the changes are valid.
>

**Checklist**
 ​


Felienne:
Ah great catches @boryanagoncharenko!! Does this close the last of #1632 or is there still more to be done?

boryanagoncharenko:
> Ah great catches @boryanagoncharenko!! Does this close the last of #1632 or is there still more to be done?

There are plenty left but I think some of them are "add tests" or "do this like in level X because there it is better". It would be better to address the grammar TODO before using the lark composition rules, so I plan to go through all of them (or at least the most important ones) before I continue with lark.

PR #2281, opened 2022-03-23
Space In String Assignment Fix for levels 6 to 11

### LINKED ISSUE #2073: [BUG] Space in String Affectation
As seen in the issue #1866 
String assignments are problematic when there are spaces:
```
a is ba ae
print a
```
=> We can't run your program. a is not a Hedy level 6 command. Did you mean if?

Adding quotation marks or not does not change the error
Also if you replace the `is` with `=`.

this error appears only at levels [6 ; 11]

aelhamah:
This PR will allow for users to have spaces in their string assignments in levels 6 through 11.

**Description**

Adds `textwithspace` to the assignment in the `grammars/Level6-Additions.lark` file

**Fixes #2073**

**How to test**

Added a unit test to `tests/test_level_06.py` called `test_space_in_string_assignment` that checks for spaces in string assignments with and without quotes both using `is` and `=`. The unit test does not pass without the change in `grammars/Level6-Additions.lark`, and the test uses `multi_level_tester` to assess levels 6 through 11.

### DIFF
--- a/grammars/level6-Additions.lark
+++ b/grammars/level6-Additions.lark
@@ -12,7 +12,7 @@ condition:+= equality_check -= equality_check
 
 list_access_var: var (_SPACE _IS _SPACE | _EQUALS) var _SPACE _AT _SPACE (INT | random)
 assign_list: var (_SPACE _IS _SPACE | _EQUALS) textwithspaces (_COMMA textwithspaces)+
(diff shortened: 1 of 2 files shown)


Felienne:
Thank you so much for your first contribution @aelhamah!

At a first glance, this looks good to me, but I would love for our grammar expert @boryanagoncharenko to also have a look!

boryanagoncharenko (commented):
Awesome! Thanks for fixing the issue.

boryanagoncharenko (approved):

PR #2282, opened 2022-03-23
[BUG FIX] Saving programs works again

TiBiBa:
**Description**
Due to a type mismatch of the `Public` value program saving was no longer possible. This is due to Dynamo expecting a number as index key, but we store the `public` value as a boolean. We fix this in this PR.

**Fixes**
No related issue.

**How to test**
This can only be tested on Alpha as the local db file doesn't care about type mismatching. Deploy to Alpha and verify that storing programs is working once again.


Felienne (approved):
Merging so you can test on alpha!

Issue #2174, opened 2022-03-05
[BUG] Check fix #2173


Felienne:
When you are back, you will want to see if my hotfix is sensible :)

Felienne:
Ah I already found one issue, we can now longer edit shared programs, but I guess that is a lot less bad that not being able to share so I will leave that as a check for later.

TiBiBa:
> Ah I already found one issue, we can now longer edit shared programs, but I guess that is a lot less bad that not being able to share so I will leave that as a check for later.

Weird! Will take a look today to see if we can solve this easily.

TiBiBa:
> Ah I already found one issue, we can now longer edit shared programs, but I guess that is a lot less bad that not being able to share so I will leave that as a check for later.

I'm not able to re-produce this error? I tried to save a program, share it and then open it again and save it again. This all seems to work fine.

Felienne:
Yeah I know how to reproduce it but it is indeed tricky! Will explain after
I am back from lecturing.

TiBiBa:
> Yeah I know how to reproduce it but it is indeed tricky! Will explain after I am back from lecturing.
> […](#)

All right! I just did some thinking on the issue and it might be best to just re-write the current structure of saving. When using the save button we call the `saveit` function whereas when using the save and share button we call the `share_program` function. I think it would make more sense to use the same function but with an additional boolean called `shared`. Will take a look and I think the issue will solve itself when the code is cleaned-up.

Felienne:
> > Yeah I know how to reproduce it but it is indeed tricky! Will explain after I am back from lecturing.
> > […](#)
> 
> All right! I just did some thinking on the issue and it might be best to just re-write the current structure of saving. When using the save button we call the `saveit` function whereas when using the save and share button we call the `share_program` function. I think it would make more sense to use the same function but with an additional boolean called `shared`. Will take a look and I think the issue will solve itself when the code is cleaned-up.

Yeah I agree! The issue that I ran into was that there were two routes that both use this view, one showing just the program without all the rest and one for the edit mode showing the regular view. In the second one, for reasons I forgot, we only allow teachers of kids to view? So I could not edit a program cause the program I was looking at was made by a kid in a class. I think we want those two routes to use the same auth code? If this is not easy to understand, we can look at it together later in the week!

Felienne:
> > > Yeah I know how to reproduce it but it is indeed tricky! Will explain after I am back from lecturing.
> > > […](#)
> > 
> > 
[...]

PR #2284, opened 2022-03-23
[CLEAN-UP] Makes sure errors are always stored as booleans

### LINKED ISSUE #2283: [CLEAN-UP] Unify the use of booleans and numbers on shared / error values
**Description**
Currently we use number values to store programs as shared or containing an error. But, we parse them sometimes as booleans as well. This usage should be unified as it is likely to introduce bugs (and already has!)

TiBiBa:
**Description**
In this PR we unify the type storing of the `error` value of a program. Currently when saving a program we set this value to either `1` or `None`. However, when retrieving public programs for the `/explore` page we look for programs without an error value and parse them, in this situation storing them with either a `True` or `False` value. In this PR we unify this to always use booleans.

Currently we are unable to replace the current `public` structure, as this is relic code and we should very carefully figure out it's dependencies before changing anything.

**Fixes**
This PR fixes #2283

**How to test**
This is only a clean-up, if it works it works.


TiBiBa:
We should merge this one as well before deployed as otherwise the db structure will be messed up! This is most likely introduced by #2206 where we always set the error value on save instead of only on share. This is accidentally performed integer based instead of boolean based and as the value is not an index key Dynamo has no issue with this mix-up.

boryanagoncharenko:
While testing this I discovered an issue: if I am a teacher, I cannot open my own saved programs. This reason seems to be this check: https://github.com/Felienne/hedy/blob/main/app.py#L867 @TiBiBa is this on your radar?

boryanagoncharenko (approved):

TiBiBa:
> While testing this I discovered an issue: if I am a teacher, I cannot open my own saved programs. This reason seems to be this check: https://github.com/Felienne/hedy/blob/main/app.py#L867 @TiBiBa is this on your radar?

Yes, this issue seems to be related to #2174. If correctly, both should be fixed with #2290.

PR #2285, opened 2022-03-23
[CLEAN-UP] Updated class invitations table

### LINKED ISSUE #2280: [UI idea] Clean-up student invitations table
**Idea incl level**
With the addition of the expiration date column the table is getting a bit crowded. The table is currently a `1/x` percentage width, making the UI terrible on smaller screen. The table should be automatically set to content width and the columns should have a `min-width`. Lastly the `expiration_date` header is not correctly translated through Babel.

TiBiBa:
**Description**
In this PR we update the class invitations table. Improve some styling to also work as expected on smaller screens, add a missing Babel key and fix the translation of the "class statistics" button.

**Fixes**
This PR fixes #2280 

**How to test**
Make sure you are a teacher. Navigate to one of your classes and invite a new student, verify that the table is displayed correctly and all values are correctly translated.


boryanagoncharenko:
The invitations table appears correctly. However, I cannot get the expiration_date header to be translated. I followed the steps (extract, update, compile) but it still looks like this:
Screenshot 2022-03-23 at 13 16 01
@TiBiBa am I missing some Babel steps?

Another thing which I discovered while testing is that any currently pending invitations will cause the cIass page to fail because of these changes: https://github.com/Felienne/hedy/pull/2260. I had local invitations pending which were created before the TTL property of the invitation was introduced. Not relevant for this PR but definitely something that needs to be addressed. @TiBiBa is that something on your radar?

TiBiBa:
> The invitations table appears correctly. However, I cannot get the expiration_date header to be translated. I followed the steps (extract, update, compile) but it still looks like this: Screenshot 2022-03-23 at 13 16 01 @TiBiBa am I missing some Babel steps?
> 
> Another thing which I discovered while testing is that any currently pending invitations will cause the cIass page to fail because of these changes: #2260. I had local invitations pending which were created before the TTL property of the invitation was introduced. Not relevant for this PR but definitely something that needs to be addressed. @TiBiBa is that something on your radar?

After compiling you should re-start the server and open a new (incognito) tab before it too take effect. Will take a look if we can have some setting in Flask to always updated when in DEV mode as we have for Tailwind and Typescript. This was on my radar! As there are currently only 2 invitations in the production website database (and the Alpha one is empty) my current idea was to simply truncate the table before merging to production.

boryanagoncharenko:
> > The invitations table appears correctly. However, I cannot get the expiration_date header to be translated. I followed the steps (extract, update, compile) but it still looks like this: Screenshot 2022-03-23 at 13 16 01 @TiBiBa am I missing some Babel steps?
> > Another thing which I discovered while testing is that any currently pending invitations will cause the cIass page to fail because of these changes: #2260. I had local invitations pending which were created before the TTL property of the invitation was introduced. Not relevant for this PR but definitely something that needs to be addressed. @TiBiBa is that something on your radar?
> 
> After compiling you should re-start the server and open a new (incognito) tab before it too take effect. Will take a look if we can have some setting in Flask to always updated when in DEV mode as we have for Tailwind and Typescript.
> 
> This was on my radar! As there are currently only 2 invitations in the production website database (and the Alpha one is empty) my current idea was to simply truncate the table before merging to production. Not sure what would be the best approach for local databases as we are unable to edit these through git.

I did restart the server and opened the class page in a new browser but I keep on getting the expiration_key value. I see in the english .po file contains 2 expiration_date keys: the former is empty while the latter has a value. Probably this is the issue?

TiBiBa:
> > > The invitations table appears correctly. However, I cannot get the expiration_date header to be translated. I followed the steps (extract, update, compile) but it still looks like this: Screenshot 2022-03-23 at 13 16 01 @TiBiBa am I missing some Babel steps?
> > > Another thing which I discovered while testing is that any currently pending invitations will cause the cIass page to fail because of these changes: #2260. I had local invitations pending which were created before the TTL property of the invitation was introduced. Not relevant for this PR but definitely something that needs to be addressed. @TiBiBa is that something on your radar?
> > 
> > 
> > After compiling you should re-start the server and open a new (incognito) tab before it too take effect. Will take a look if we can have some setting in Flask to always updated when in DEV mode as we have for Tailwind and Typescript.
> > This was on my radar! As there are currently only 2 invitations in the production website database (and the Alpha one is empty) my current idea was to simply truncate the table before merging to production. Not sure what would be the best approach for local databases as we are unable to edit these through git.
> 
> I did restart the server and opened the class page in a new browser but I keep on getting the expiration_key value. I see in the english .po file contains 2 expiration_date keys: the former is empty while the latter has a value. Probably this is the issue?

This might be some merging error as I'm only able to find 1 key locally as well as on this branch. I'm also unable to re-produce the bug as it all seems to work fine for me:

Schermafbeelding 2022-03-23 om 12 39 30

boryanagoncharenko:
Yes, I think so too. Something is definitely wrong on my end because what I see (and generate) locally does not match what the PR. The values are correct in the .po files. Thanks for the assistance :)

boryanagoncharenko (approved):

PR #2291, opened 2022-03-23
[FEATURE] Improve comments in grammar

### LINKED ISSUE #1705: [BUG] Comments are not allowed in all places

TiBiBa:
**Description**
In this PR we fix an issue where the comment was not allowed in all places. We add an `%ignore` rule to make sure everything after the `#` is always ignored. We also add the `#` to the `text` rule to make sure something is a text until a newline **or** a #.

**Fixes**
This PR fixes #1705

**How to test**
This is still a WiP, I'm currently figuring out how to write code test cases and verify the rule correctness.


Felienne:
Ah nice, Lark has support for comments, I did not know! Great find. I guess we also want to add a few tests? And maybe @jpelay wants to review? (sorry that this is now being picked up, I know you also wanted to work on it, but there are enough issues open haha)

TiBiBa:
> Ah nice, Lark has support for comments, I did not know! Great find. I guess we also want to add a few tests? And maybe @jpelay wants to review? (sorry that this is now being picked up, I know you also wanted to work on it, but there are enough issues open haha)

Yes apparently there is an `%ignore` option, this way we don't have to parse for the comments as they are automatically ignored. I'm still trying to figure out a lot about Lark and getting familiar with the syntax. For now this seems to simplify the current way we handle comments (but inline comments are still a WiP....). Will keep working on this issue in this branch until I figure it out!

TiBiBa:
> Cool thanks @TiBiBa! If you need help, let me know!

It seems to work, but it also seems too easy a fix to be working. Can you verify?

TiBiBa:
The current idea is expanding on the already existing `text` rule:

- Everything is `text` until a newline
   - Expand with -> everything is a `text` until a newline **or** a #

TiBiBa:
Still have to figure some stuff out as this is working:
Schermafbeelding 2022-03-23 om 14 23 16

And this isn't:
Schermafbeelding 2022-03-23 om 14 23 25

TiBiBa:
The last case is fixed as well by keeping the current comment rule as well:
Schermafbeelding 2022-03-23 om 14 30 12

TiBiBa:
Can verify that the example code in the issue is indeed working as expected:
Schermafbeelding 2022-03-23 om 14 42 03

jpelay:
Hi, Timon! Such a and smart great solution,

jpelay:
Hi,  Timon, this such an elegant and smart solution :) Your test are failing because at level 4 we don't use the grammar merger due to weird interactions in the precedence of the rules, so we have to redefine the command rule again. You just gotta delete the command rule from there, something like this:

`command:print | ask | turtle | assign_list | add | remove | sleep | error_ask_no_quotes| assign | error_print_no_quotes | error_invalid`


Now, the adventures are failing because many programs that are `start_code` consists only of one comment, and programs that consists of only comments raise `EmptyProgramException` because since comments are ignored, there isn't a rule that the `IsValid` filter can grab, and therefore detects it as an empty program.

jpelay:
I took the freedom to make the tests pass, hope it's ok! These were just a few changes:

1. Changed the level 4 and level 5 grammar to remove the comment rule, and in the level 5 I allowed empty programs to be possible (as in the previews levels) this leads to,
2. Since now we allow empty programs it's no longer necessary the emptyProgramException, so I commented,the test on levels 1 that relate to this (if we check empty programs, programs that consitsts of comments alone will not pass, therefore many adventures programs fail) I think these tests are not needed anymore, but Im commenting just in case
3.  I also added the # to some text rules from level 4 on.

jpelay:
Worth checking! comment after repeat fails!
image

jpelay:
Again, sorry I jumped in! Hope it's ok, Timon! I was reviewing the changes and got a bit carried away :sweat_smile: There's a test, however, that stills fails, and it regards the numeric translations, not sure what's happening there yet

TiBiBa:
> Again, sorry I jumped in! Hope it's ok, Timon! I was reviewing the changes and got a bit carried away There's a test, however, that stills fails, and it regards the numeric translations, not sure what's happening there yet

No problem, thanks a lot for helping out! Will take a look at the last failing test later today.

TiBiBa:
Just checking out the last failing test, but not sure why it is failing. Verified the correctness in the code editor itself:

Schermafbeelding 2022-03-24 om 09 57 13

TiBiBa:
Another think to keep in mind:

We should remove the "good job" success message in case of an empty program warning. As we no longer return an actual error we still have the (relic code) "you have no output" warning that catches the empty program.

Felienne:
> Another think to keep in mind:
> 
> We should remove the "good job" success message in case of an empty program warning. As we no longer return an actual error we still have the (relic code) "you have no output" warning that catches the empty program.

Ah good catch, maybe make a separate issue for that? If you want I can also look into why the Chinese test is failing (you can open an issue for me)

Felienne:
And thanks @jpelay for your extensive help on this!!

TiBiBa:
I just fixed the last failing test, but by simply commenting it out in the .yaml file. Not really nice, but other than that it should all work now.

thjazi39:
I have the code snippets ready to add syntax highlighting for comments.
I'll add them on this branch, or I'll do that once the merge #2264 is done (to reduce the risk of conflict?)

Felienne:
I would sa add them here because thematically it goes together

thjazi39:
It's done, I added syntax highlighting for the comments:
image

Felienne:
I think this is good to go! Maybe @jpelay wants to do the final approval?

jpelay:
> I think this is good to go! Maybe @jpelay wants to do the final approval? If you are happy, simply leave an approving review and it will merge automatically

Yes, I think everything is good! Although after the latest merge some tests fail that didn't fail before :< So I don't know if we can still merge. I'm checking them right now, but can't pinpoint the issue just yet.

jpelay:
Not sure why the test fails, but everything works fine on my local in the editor so Im merging.

jpelay (approved):

Felienne:
The issue seems to be with the changes to Missing Command. This test now gives a MissingCommand exception instead of an InvalidArgumentTypeException at level 5. 

```
  def test_add_to_list_with_string_var_gives_error(self):
    code = textwrap.dedent("""\
    color is yellow 
    colors is green, red, blue
    add colors to color""")

    self.multi_level_tester(
      max_level=11,
      code=code,
      exception=hedy.exceptions.InvalidArgumentTypeException
    )
```

Seems to be a task for @boryanagoncharenko?

Felienne:
Converting to draft until @boryanagoncharenko figures out the ambig issues

grammars/level5-Additions.lark
```diff
@@ -1,4 +1,6 @@
-program: _EOL* (command | error_invalid)* (_SPACE)* (_EOL+ (command | error_invalid) (_SPACE)*)* _EOL* //lines may end on spaces and might be separated by many newlines
+program: _empty_program | _non_empty_program
+_empty_program: (_EOL | _SPACE)*
```
Felienne, line 2: Smart fix @boryanagoncharenko!!

Felienne (approved):

boryanagoncharenko:
The level 5 grammar is adapted and the failing zh test is re-added, so there are no intermittent test failures. Changing the PR status from Draft -> Ready For Review.

grammars/level5-Additions.lark
boryanagoncharenko: It is the only fix that actually solved the problem. I thought I understood why there is an ambiguity at level 6, but as I dived deeper I figured the ambiguity should already kick in at level 5. But it did not, which made me wonder whether I understood what's going wrong in the first place... So, I had to take this alternative approach. Grammar ambiguities are hard.

Felienne:
What a team effort @jpelay, @boryanagoncharenko and @TiBiBa <3

PR #2297, opened 2022-03-24
[FIX] Removes en.yaml file that accidentally got added in #2268

TiBiBa:
**Description**
In this PR we remove a file that shouldn't be there. The `/texts` folder with a en.yaml file inside. This seems to have gone wrong in #2268, but the file is empty nevertheless. It is redundant as we switched to Babel and should be removed.

**Fixes**
No related issue.

**How to test**
Just a clean-up, no testing.


Felienne:
Ah yeah of course that branch was branched off before the babel changes.

Felienne (approved):

PR #2300, opened 2022-03-24
[CHORE] Removes mandatory account mail address

### LINKED ISSUE #2075: Make mail address optional on signup
Currently setting up a mail address is mandatory when creating an account. As they have no explicite added value it would be nice if we made this optional.

TiBiBa:
**Description**
In this PR we remove the mail address as mandatory when creating a new account. We received feedback multiple times that students don't have any mail address or don't have access to it at the moment of account creation. As the access is not essential for the creation and using of the account, it is for account validation and password resetting. We start this big chore by first removing all dependencies on a mail address, next we should look at the replacement for accounts without a mail address to still be able to reset their account. It might be best to solve this in a separate issue as the task will grow a bit too big for one PR to be tested otherwise.

**Fixes**
This PR fixes #2075.

**How to test**
Create a new account and verify that you are able to do this without entering a mail account. When entering a mail account it is still validated if the address is valid and doesn't already exist. Also verify that the multiple account creation for teacher still works.


TiBiBa:
Schermafbeelding 2022-03-24 om 11 39 43

TiBiBa:
Back-end structure for signup and profile editing without an mail address is done. We only have to think about a way how to return the key to the user.

templates/recover.html
```diff
@@ -1,14 +1,18 @@
...
     <div class="flex flex-row items-center justify-center mb-2">
         <label class="inline-block w-1/3" for="username">{{_('username')}} / {{_('email')}}</label>
         <input class="inline-block w-1/2 mx-2 px-2 py-1 rounded-lg border-gray-400" id="username" name="username" required>
     </div>
     <button type="submit" class="green-btn my-2">{{_('send_password_recovery')}}</button>
   </form>
+  <div class="mt-4 border-t-2 border-gray-400 pt-4">
+      <h2>Reset wachtwoord met sleutel</h2>
+      <button class="green-btn my-2" onclick="window.location = '/reset_key'">Reset met sleutel</button>
+  </div>
```
boryanagoncharenko, line 16: The content is hardcoded in Dutch. Not sure if this is something we want to address now or in a later PR.

website/auth.py
```diff
@@ -366,6 +362,9 @@ def signup():
             return gettext('keyword_language_invalid'), 400
 
         # Validations, optional fields
+        if 'email' in body:
+            if not isinstance(body.get('email'), str) or not utils.valid_email(body.get('email')):
+                return gettext('email_invalid')
```
boryanagoncharenko, line 367: This line needs to have `, 400` appended to it. Otherwise, the user does not get a proper error when they input an invalid email, but are happily redirected to the /landing-page without being logged it.

boryanagoncharenko (changes requested):
I found a couple of issues:
- When I do not supply an email, I could still request to be a teacher. This causes a failure because we actually attempt to send a welcome-teacher email. Probably this scenario should be disabled altogether as teachers really must have emails. Is that correct?
- If I enter an invalid email, I could still check the subscribe-to-newletter checkbox which leads to an error page. The email for the value I used is: student@student

templates/recover.html
TiBiBa: Forgot about that one! We first have to decide if we want to use the secret key approach, I agree that children will not write it down or simply forget where they wrote it. For the mail validation we use the html `valid` attribute, will take look if we can fix this so it only returns when also a `.nl` for example is added at the end of the mail.

website/auth.py
TiBiBa: Good catch!

templates/recover.html
TiBiBa: Temporary hidden this one

TiBiBa:
> I found a couple of issues:
> 
> * When I do not supply an email, I could still request to be a teacher. This causes a failure because we actually attempt to send a welcome-teacher email. Probably this scenario should be disabled altogether as teachers really must have emails. Is that correct?
> * If I enter an invalid email, I could still check the subscribe-to-newletter checkbox which leads to an error page. The email for the value I used is: student@student

Good catches! Will fix these and also think about the last issues as discussed in #2075. Tying new accounts specifically to a teacher account requires quite some re-writing on the auth back-end and we have to think if this is something we want or keep to the simple view "without an mail address you can't reset a password, if you happen to be in a class your teacher can".

PR #2305, opened 2022-03-24
[UI IMPROVEMENT] No longer shows success message when empty output warning

### LINKED ISSUE #2298: [UI idea] Don't show success message when a warning is returned
**Idea incl level**
Currently when we receive a warning (for example for a program without any output) we also receive the success box with the option to display confetti. This shouldn't be the case. The success message should only be shown iff a program is completely correct.

TiBiBa:
**Description**
In this PR we fix a small QoL feature so we no longer show a success message in the case of an empty output. This can, for example, happen in level 1 when we only use an `ask` but without any `print` or `echo`. There is already a corresponding warning, but as this is front-end calculated this is not caught by the `response` check and the success message is still shown. This is fixed.

This PR fixes #2298

**How to test**
Navigate to level 1 and verify that the following code does give a warning but not a success message:
```
ask How do you do?
```


Felienne:
Maybe a fun PR for you to review @fRedelaar?

fRedelaar (approved):
Tested it and looks really good!

PR #2306, opened 2022-03-24
[FIX] Fuzzy gettext translations are now automatically compiled

TiBiBa:
**Description**
In this PR we fix the on-deploy script to make sure fuzzy translations are also compiled. This is the case for all translations automatically filled with the English version.

**Fixes**
No related issue

**How to test**
Verify that for every language the non-translated keys are shown in English.


Felienne (approved):

Felienne:
Seems to work!

PR #2316, opened 2022-03-25
[UI] Improves language selector

### LINKED ISSUE #2310: [UI idea] Improve language selector (it's getting a bit too big)
**Idea incl level**
New languages are keep getting added to Hedy, which is great! However, the language selector on the menu bar is currently "out-of-view" when selecting another language. This is not very user-friendly and we should think about a new UI choice to let users select another language. This is already partly solved as logged in users select a profile language, but non logged-in user cannot do this.

**Additional context**
Schermafbeelding 2022-03-25 om 11 15 40

TiBiBa:
**Description**
In this PR we clean-up the language selector due to the increasing amount of languages. The current dropdown menu is getting too large and even grows out-of-view. In this PR we fix this issue by splitting the dropdown menu into two columns instead of one. We also re-write the structure so we are able to easily adjust this to 3 or 4 columns in the future.

**Fixes**
This PR fixes #2310

**How to test**
Hover the language selector and verify that it is now two columns instead of one. Also verify that the search is still working and the languages automatically re-position (!) on searching. Making sure that it is always an evenly divided view.

**New UI**
Schermafbeelding 2022-03-25 om 13 21 04


TiBiBa:
Schermafbeelding 2022-03-25 om 13 07 40

Felienne:
Looks good!!

image

But maybe do we want to put the languages in alphabetic order instead of in the order that we added them? Guess it will be hard to find an ordering for Latin words and also Chinese and Hindi but maybe there is a precedent for that?

TiBiBa:
> 

Good idea! As we build the order by simply iterating the dict of supported languages we can either sort the dict or sort on runtime when showing them. Will take a look! Will also take a look what is the best practice for ordering latin and non-latin languages (should be a issue other web developers have as well).

Felienne (approved):

PR #2319, opened 2022-03-25
[FIX] Differentiates between search as placeholder and as button string

### LINKED ISSUE #2318: [BUG] Search button no longer has dedicated translation
**Describe the bug**
Before switching to Babel we differentiated between the "search" as string: used a placeholder like this `search...` and search as a button, used as the button description like this `search`. For some reason this got lost in the transition and currently all "search" occurrences are like the string placeholder, undesirable!

Schermafbeelding 2022-03-25 om 14 03 28

**Expected behavior**
Search buttons should only say "search", differentiating from the placeholders.

TiBiBa:
**Description**
In this PR we add an extra translation to make sure search buttons are no longer labelled as "search..." but simply as "search".

**Fixes**
This PR fixes #2318

**How to test**
Navigate to either the explore or programs page and verify that the search button now indeed only reads "search".


Felienne (approved):

Felienne:
Haha this is a conflict mess too @TiBiBa Guess it is the timestamps but I am not sire

PR #2321, opened 2022-03-25
add color highlight for number

thjazi39:
Added syntax highlighting for integers from level 6. And above level 12, floats are also recognized.

Of course, numbers in strings are not recognized

level < 6  : 
image

6 <= level  < 12 : 
image

12 <= level : 
image


Felienne:
Hi @thjazi39!

I link this idea, esp your implementation of floats to make the difference between 5,2 and 5.2 clear!! But... in a list types can be mixed:

image

I worry that this coloring might give kids the impression that "earthworm" is not allowed here? What do you think?

thjazi39:
indeed, but in the following case, it is necessary to have the recognition of the numbers, so that the children understand that one can make operations with the values :

image

Felienne:
> indeed, but in the following case, it is necessary to have the recognition of the numbers, so that the children understand that one can make operations with the values :
> 
> image

True!

What do you think @OnnoEbbens and @MarleenGilsing? Color numbers in the list or leave it as it is?

thjazi39:
I was going to suggest that the ideal would be to color the numbers if there are only numbers in the list, and if not, not color anything.
image

But even like that, it can be a problem with this code, which only works 5 times out of 6. For the sixth case, no coloring would be needed, but for the other 5, some coloring is needed.

I don't know what will be the best for the comprehension of the children, but from the practical point of view, it is the coloring of the numbers whatever the context that is the most obvious to implement

Felienne:
> I was going to suggest that the ideal would be to color the numbers if there are only numbers in the list, and if not, not color anything. image
> 
> But even like that, it can be a problem with this code, which only works 5 times out of 6. For the sixth case, no coloring would be needed, but for the other 5, some coloring is needed.
> 
> I don't know what will be the best for the comprehension of the children, but from the practical point of view, it is the coloring of the numbers whatever the context that is the most obvious to implement

ok let's leave it like this and see what happens in class!

Felienne (approved):

PR #2323, opened 2022-03-25
fix blank error

### LINKED ISSUE #2320: [BUG] error with blank
not all blanks are highlighted :
image

thjazi39:
fix #2320 
image


Felienne:
Hi @thjazi39,

It would not be terrible if there was a little bit of explanation why the error occurred and what the fix does, cause it it not entirely clear to me

thjazi39:
The problem was that the lookahead was missing, so it captured the next character. And this prevented the new "_" from being recognized.

Let me explain:
We want to recognize the underscores that are alone, but we can't just look for the "_" regex, because in this case, the spaces 1 and 2 will be captured with the underscore A, and so the underscore B will not be recognized, because the space 2 is not available anymore.
```
 _ _ 
1A2B3
```

To solve this problem, I use lookahead and lookbehind which allow to look before and after without selecting what is before or after, and this is very useful to recognize a word in a particular environment, but selecting only the word.
(in the example above, we would have selected only the characters A and B)

The problem is that since Safari doesn't support lookbehind, I tricked myself by using only lookahead. This means that the environment before will be selected with the word, but not what comes after.
( so there will be 1 selected with A, and 2 with B)


And the bug that this PR solves is precisely that, I had forgotten a lookahead in the case of underscores, and that prevented the recognition of an underscore on 2:

image

Felienne:
Works (and thanks for the extra explanation!)

Felienne (approved):

PR #2328, opened 2022-03-27
Adds norwegian to supported languages

### LINKED ISSUE #2327: [CONTENT] Add Norwegian support to code structure
**Describe the issue**
With #2326 there is quickly been made an impressive amount of Norwegian translations. However, as we didn't support this language at all we also have to add this to our code structure. This to make sure it also shows up on the language selector on the menu bar and is a select-able language on profile creation. Not much work, but it should be fixed closely after merging #2326!

TiBiBa:
**Description**
In this PR we add Norwegian as a supported language to our web structure. We also fix a small issue where Chinese wouldn't work with the new Babel structure as we used the `zh` abbreviation whereas Babel expected `zh_hans`. This should all work now.

**Fixes**
This PR fixes #2327 

**How to test**
Navigate to the home page and verify that Norwegian (Norsk) is now an option on the language selector. Also verify that this is working as expected. **Make sure to compile Babel first before checking**.


Felienne:
Making this into a draft since it still needs keyword lang support and syntax highlighting (which will be easier after merging #2329 )

grammars/keywords-nb_NO.lark
```diff
@@ -0,0 +1,30 @@
+_PRINT: "skriv" | "print"
+_ASK: "spør" | "ask"
+_ECHO: "ekko" | "echo"
+_FORWARD: "forover" | "forward"
```
Kegulf, line 4: this was changed from `forover` to `frem`

grammars/keywords-nb_NO.lark
```diff
@@ -0,0 +1,30 @@
...
+_TURN: "snu" | "turn"
+right: "høyre" | "right"
+left: "venstre" | "left"
+_IS: "er" | "is"
+_SLEEP: "sov" | "sleep"
+_ADD_LIST: "legg til" | "add"
```
Kegulf, line 10: This was changed from `legg til` to `legg`

This is to avoid statements like this:
`legg til item til list` which looks silly 😄 
now it will be `legg item til list`

Kegulf: It matches well with the `remove` command as well, which I think is a bonus 😊

hedy.py
```diff
@@ -39,14 +39,15 @@
     'nl': 'Nederlands',
     'es': 'Español',
     'fr': 'Français',
+    'nb_no': 'Norsk',
```
Kegulf, line 42: 'Norsk Bokmål' please 🧡

Kegulf: We have two written languages in Norway (technically three if you count the old native language Samii)

Kegulf (changes requested):
Could you change these small things? 😁 These changes should probably be implemented before the highlight edits is put in place.

Considering changing the `input` keyword as well, but I'll have to confeer with my norwegian co-translators for that one. `innputt` looks a bit silly, not really a word we use in norwegian anyways. Might just stick to `input` here.

grammars/keywords-nb_NO.lark
```diff
@@ -0,0 +1,30 @@
...
+_ELSE: "ellers" | "else"
+_AND: "og" | "and"
+_REPEAT: "gjenta" | "repeat"
+_TIMES: "ganger" | "times"
+_FOR: "for"
+_RANGE: "sekvens" | "range"
+_TO: "til" | "to"
+_STEP: "steg" | "step"
+_ELIF: "elhvis" | "elif"
+_INPUT: "innputt" | "input"
```
Kegulf, line 27: Changed this from `innputt` to `inndata` after confeering with my co-translators 😊

grammars/keywords-nb_NO.lark
```diff
@@ -0,0 +1,30 @@
+_PRINT: "skriv" | "print"
+_ASK: "spør" | "ask"
+_ECHO: "ekko" | "echo"
+_FORWARD: "forover" | "forward"
+_TURN: "snu" | "turn"
+right: "høyre" | "right"
+left: "venstre" | "left"
```
Kegulf, line 7: should these be `_RIGHT` and `_LEFT`?

grammars/keywords-nb_NO.lark
```diff
@@ -0,0 +1,30 @@
...
+_IS: "er" | "is"
+_SLEEP: "sov" | "sleep"
+_ADD_LIST: "legg til" | "add"
+_TO_LIST: "til" | "to"
+_REMOVE: "fjern" | "remove"
+_FROM: "fra" | "from"
+_AT: "på" | "at"
+random: "tilfeldig" | "random"
```
Kegulf, line 15: should this be `_RANDOM`?

Kegulf (changes requested):

grammars/keywords-nb_NO.lark
TiBiBa: Nope! As these are tokens instead of keys they should be lowercase. (correct me if I'm wrong @Felienne)

grammars/keywords-nb_NO.lark
TiBiBa: Same here, also a token!

grammars/keywords-nb_NO.lark
Felienne: You are right! It is linke `random` (also a token)

TiBiBa:
Hi @Kegulf! I just updated the keywords .lark file with your latest suggestions / translations. Let me know if anything is still incorrect!

Felienne:
Ready to deploy Norsk! @Hakon and @Kegulf, I have now used Norsk in the UI instead of Norsk Bokmal because I think it might be easier for kids. 

If we support both of course we can distinguish like we do with Portuguese now. But if you think it is better to add it, we can surely do that!

Felienne (approved):

PR #2329, opened 2022-03-27
Adds Turkish grammars and syntax highlighting

Felienne:
Also adds a script to generate the syntax highlighting automatically, replacing the old (broken) bash script


TiBiBa:
Looks fine to me, will test later today!

PR #2334, opened 2022-03-28
[FIX] Cleans up languages abbreviations to align YAMLs and Babel

### LINKED ISSUE #2332: [BUG] Chinese translations no longer work due to new translation structure
**Describe the bug**
Due to switching to Babel our UI translations for Chines no longer work. This is due to our original implementation expecting `zh` as language keyword, but Babel expects `zh_Hans`. As both are directly requested based on the session language, one in them isn't working. In this case the new one, wit Babel. As we can't change the Babel language keys it might be best to conform our other YAMLs to the Babel key.

**Expected behavior**
All Chinese translations should work again.

TiBiBa:
**Description**
In this PR we clean-up the language abbreviations to align the keys used by the YAMLs and bij Babel. We correctly capitalize the second part of the Portuguese languages and change the Chinese translations from `zh` to `zh_Hans`, also fixing the bug where the Chinese Babel translations where not shown.

**Fixes**
This PR fixes #2332.

**How to test**
Verify that the Chinese UI translations are now working indeed. You can simply verify by noticing the translated menubar options. Otherwise this is a clean-up task, nothing broken: it works!


Felienne:
I will first merge all weblate changes cause I am afraid this might lead to a conflict otherwise!

Felienne:
@TiBiBa This has a complex conflict now, can you take a look?

TiBiBa:
> @TiBiBa This has a complex conflict now, can you take a look?

Will take a look!

Felienne:
Have we missed a file?

image

Felienne (approved):

PR #2336, opened 2022-03-28
[CONTENT FIX] Localizes left/right keyword

### LINKED ISSUE #2335: [CONTENT] Replace 'left' and 'right' by placeholders as they are localized from now on
**Describe the issue**
We localized keyword by surrounding them with curly brackets and parsing the correct language on runtime. However, at the moment of implementation `left` and `right` where not localized (yet). We should still fix this.

**Expected text**
All `left` and `right` occurrences that are actually keywords should be surrounded by curly brackets.

TiBiBa:
**Description**
In this PR we localize the left/right **and** random keywords in both the English and dutch version. As the translations are all using English as a starting point we hope that this fixes ambiguity when translating. The curly brackets should clarify that the keywords should not be translated.

**Fixes**
This PR fixes #2335

**How to test**
Small clean-up, doesn't have to be tested.

### DIFF
--- a/coursedata/adventures/en.yaml
+++ b/coursedata/adventures/en.yaml
@@ -595,7 +595,7 @@ adventures:
                     ## Let's draw
                     You can also use Hedy to draw. By combining turns and lines, you can make a square or stairs!
 
-                    Using `{forward}` you draw a line forwards. The number behind it determines how far the turtle will walk. `{turn} right` turns a quarter turn in clockwise direction, `{turn} left` turns counter clockwise.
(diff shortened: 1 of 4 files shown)


Felienne (approved):

Felienne:
Do wee need to fix this for Norwegian too or is that already in order?

TiBiBa:
> Do wee need to fix this for Norwegian too or is that already in order?

Fixed this already in weblate. We do however have to fix this for other languages as I haven't checked, but as the tests pass we are sure that they are not translated at least (so they will still be left / right). Will take a look!

PR #2341, opened 2022-03-28
[BUG FIX] next level button shows correctly for Norsk

TiBiBa:
**Description**
For some reason the current level-defaults ordering for Norsk is non-chronical. It goes like `[6, 12, 1....]`, this isn't a problem from a code perspective as we retrieve them by dict keys so the ordering is irrelevant. However, it is important for our `max_level` function! This function verifies if the next key in the dict is `current key + 1`. If the ordering is not correct Hedy assumes the current level is the max_level. Due to this issue there was never a next level button when on Norsk while all levels where indeed valid. We solve this issue twice:

- Sort the dict keys before looping to find `max_level`
- Fix ordering of the Norsk level-defaults YAML file

Both fixed seperately should be sufficient however, we want to prevent this in the future and we want the Norsk YAML file to follow the convention of our code structure. So therefore we present: Double fix! 

**Fixes**
No related issue.

**How to test**
Verify that the next level button on the code page now is correctly shown when selected the Norsk language.


Felienne:
> For some reason the current level-defaults ordering for Norsk is non-chronical.

I am pretty sure this is true for other langs too (at least in the adventures) Maybe we should fix that, it is confusing also when searching for something. That is my bad, it is due to reordering. Can you make an issue @TiBiBa?

Felienne (approved):

PR #2347, opened 2022-03-28
[CLEAN-UP] Removes redundant keys from page YAMLs

### LINKED ISSUE #2346: [CLEAN-UP] Move page dependent translation keys from YAMLs to Babel
**Description**
We currently store some page unique UI translations within the page YAMLs. This was useful at the point when we used YAMLs for all our ui translations as this reduced the amount of "dictionary ping-pong" on the server. But now that we switched to Babel this structure is redundant and makes it hard to keep track of our translations. Therefore it would be nice if we move these small translation keys to Babel.

To be clear: This only concerns the short page unique strings, not the actual page content. We still have to discuss that part.

TiBiBa:
**Description**
In this PR we do some clean-up on the resting YAMLs in the pages directory. We use these files for translation as well as structure in some cases. We clean-up all redundant keys that can easily be covered by Babel from now on. We also complete remove the `customize-class` YAML structure as it didn't contain any structure and all translations can be caught by Babel as well.

**Fixes**
This PR fixes #2346

**How to test**
This is a clean-up, not broken? Fixed!


Felienne:
Nothing broken if I run it locally (some conflicts though :( )

Felienne (approved):

TiBiBa:
Will fix these conflicts!

Felienne:
Hope this does not make Weblate angry 🤞

PR #2350, opened 2022-03-29
Adds Norwegian to supported keyword languages

TiBiBa:
**Description**
In this PR we add Norwegian to supported keyword languages. This was already added in the translation file but not yet in the dict on which we base the front-end options. I assume this was an error which make sense as it doesn't make sense to keep track of two identical dictionaries. Therefore in this PR we also remove one of them to include the other one.

**Fixes**
No related issue.

**How to test**
Verify that Norsk is now a keyword language option on profile creation and updating. Verify that Norks only works if your profile language is actually set at Norks. Verify that it is all working correctly and all placeholders are correctly parsed.

### DIFF
--- a/hedy_translation.py
+++ b/hedy_translation.py
@@ -5,8 +5,7 @@ import operator
 import yaml
 from os import path
 
-# TODO, FH mar 2022, why don't we grab ALL_KEYW_LANGS from hedy?
-KEYWORD_LANGUAGES = ['en', 'nl', 'es', 'fr', 'ar', 'tr', 'hi', 'nb_NO']
+KEYWORD_LANGUAGES = hedy.ALL_KEYWORD_LANGUAGES.keys()
 
 # Holds the token that needs to be translated, its line number, start and end indexes and its value (e.g. ", ").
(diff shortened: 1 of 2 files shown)


hedy_translation.py
```diff
@@ -5,8 +5,7 @@
 import yaml
 from os import path
 
-# TODO, FH mar 2022, why don't we grab ALL_KEYW_LANGS from hedy?
-KEYWORD_LANGUAGES = ['en', 'nl', 'es', 'fr', 'ar', 'tr', 'hi', 'nb_NO']
+KEYWORD_LANGUAGES = list(hedy.ALL_KEYWORD_LANGUAGES.keys())
```
Kegulf, line 9: Nice TODO fix 😊 
Compared to the old list, this new list is missing 'tr', is this wanted?

TiBiBa: Indeed! It seems `tr` was also added only in one place, will fix!

TiBiBa: Fixed!

Kegulf (approved):

Felienne (approved):
Thanks for fixing my lazy TODO!

PR #2353, opened 2022-03-29
[CLEAN-UP] Fixes structure of level-defaults and adventures

### LINKED ISSUE #2345: [CLEAN-UP] Make sure level-defaults and adventures are in chronological level ordering
**Description**
Due to some re-structuring, editing and many more operations the order of most YAML files is no longer chronological. It would be nice to fix the order by level once again. This is quite a big clean-up task, but makes working with the file a lot easier later on.

TiBiBa:
**Description**
In this PR we fix the structure of all level-defaults to make sure they are in chronological order. We already fixed the bug where this would result in an invalid `max_level` value, but it is still nice to clean-up the YAMLs a bit. We perform the same clean-up on the English and Dutch adventures as these are still being edited manually, as the other files are languages are all translated through weblate the file structure is less relevant.

We also fix some translation issues, grammar mistakes and other small thingy's in general. Basically this is a starting point of cleaning-up the adventures and bring back some structure.

**Discussion point**
Currently the YAML file decides our adventures order. As weblate does no conform our set chronological order in the English YAML we have to think about this! Do we want to hard-code the order somewhere? If I recall correctly the adventures are supposed to go from easy -> hard as left -> right.

**Fixes**
This PR fixes #2345

**How to test**
Just a clean-up, nothing to test.


TiBiBa:
After merging this we can continue our clean-up with #2371!

Felienne (approved):

Felienne:
Hail Mary for Weblate 🙏

PR #2355, opened 2022-03-29
[BUG FIX] Saved teacher adventures are loaded correctly again

### LINKED ISSUE #2354: [BUG] Saving and loading personally created adventures fails
I made an adventure for one of my classes. When students save the program in this adventure it does appear in the "my programs" list. However when they open the program all the code has disappeared. 

It seems as if the personally created adventures are not properly loaded when you open a program from "my programs". When I load other programs in the same level my personally created tab is not shown at first. Only when I refresh I can see it.

TiBiBa:
**Description**
In this PR we fix an issue where saved teacher adventures where not loaded properly. This was due to a check in the `tabs.ts` code that verified is something was a teacher adventures, in that case: change save name and return. However, the loading of the saved program code in the editor was **below** this return statement, which was never reached when on a teacher adventure tab. This bug is due to the complex structure for which we load and switch tabs, the teacher adventures are handled as a separate adventure. We should create a separate issue to clean the tabs up a bit.

**Fixes**
This PR fixes #2354.

**How to test**
Testing is quite complex as it is best to first encounter the bug to understand the issue:

- Go to either Alpha or production
- Create a class, make sure you have a student account enrolled
- Create a teacher adventure, customize the class and include this adventure
- As the student -> open the adventure and save the program
- When loading verify that the code editor is empty

**Now**
 Repeat the steps above, but verify that the saved program is now indeed loaded correctly in the editor.


Felienne:
Could you check this one @boryanagoncharenko? It seems like a bit of work and I don't have a lot of time this week.

boryanagoncharenko (approved):

PR #2357, opened 2022-03-29
Only highlights keywords from keyword language instead of profile language

### LINKED ISSUE #2356: [UI idea] Only highlight syntax for keyword language instead of profile language
**Idea incl level**
It is requested to only highlight the keyword language opposed to how we currently select the profile language highlighter.

**Additional context**
As discussed in #2279

TiBiBa:
**Description**
In this PR we make a minor change to our syntax highlighting. Our current approach is to always highlight keywords for the profile language as well as the keyword language, enabling users to switch between these. However, as pointed out by @OnnoEbbens this can be confusing when using English keywords and the Dutch page language as some variable names will be highlighted faulty. We fix this by picking the highlighter for the keyword language instead the profile language.

**Fixes**
This PR fixes #2356.

**How to test**
Set your keyword language to English, independent of your profile language and verify that only English keywords are highlighted.


TiBiBa:
Schermafbeelding 2022-03-29 om 14 10 19

Felienne (approved):

PR #2362, opened 2022-03-29
`coursedata/` -> `content/`

### LINKED ISSUE #1931: Rename coursedata folder
The name `coursedata` stems from way back but we don't use the word course anywhere, so maybe `content` is better. Making an issue because it might involve some renaming in the codebase too and some thorough manual testing of all languages.

aelhamah:
Renames the `coursedata/` directory to `content/`

**Description**

Renames the directory and changes any references within the source code to `content` (find and replace)

**Fixes #1931**

**How to test**

No changes to test in particular

```bash
$ python -m pytest
```


content/adventures/zh_Hans.yaml
```diff
@@ -601,7 +601,7 @@ adventures:
...
                 start_code: '{print} What will the die indicate this time?'
-                story_text: "## Dice\nIn this level we can choose from a list. With that we can let the computer choose one side of the die.\nTake a look at the games you have in your closet at home.\nAre there games with a (special) die? You can also copy it with this code.\nFor example, the dice of the game Earthworms with the numbers 1 to 5 and an earthworm on it.\n\n![Die of earthworms with 1 to 5 and an earthworm on it](https://cdn.jsdelivr.net/gh/felienne/hedy@24f19e9ac16c981517e7243120bc714912407eb5/coursedata/img/dobbelsteen.jpeg)\n"
+                story_text: "## Dice\nIn this level we can choose from a list. With that we can let the computer choose one side of the die.\nTake a look at the games you have in your closet at home.\nAre there games with a (special) die? You can also copy it with this code.\nFor example, the dice of the game Earthworms with the numbers 1 to 5 and an earthworm on it.\n\n![Die of earthworms with 1 to 5 and an earthworm on it](https://cdn.jsdelivr.net/gh/felienne/hedy@24f19e9ac16c981517e7243120bc714912407eb5/content/img/dobbelsteen.jpeg)\n"
```
aelhamah, line 604: I wasn't 100% sure about these types of changes, so I figured I would highlight one with this comment :)

TiBiBa:
Thanks for contributing, I will review this right away!

content/adventures/zh_Hans.yaml
TiBiBa: Good catch! I'm not sure where this url refers to, it seems like a copy of an old version of the repo of our project? I can't remember there being an `img` folder in `coursedata`, so this might be a copy of some relic code. We should leave it at `coursedata` for now to make sure the image url is not broken as it doesn't work after the renaming. It would be nice to have a separate issue to figure this out and clean-up the image urls.

@Felienne do you know when this was implemented?

TiBiBa (changes requested):
Reverse url change on images in adventure YAMLs

content/adventures/zh_Hans.yaml
Felienne: That link indeed is old but the file is still there:
image

This link should not be changed!

aelhamah: Thank you @Felienne  and @TiBiBa ! I reverted the changes; However, I made a new PR #2377  because there were too many merge conflicts with the main branch :( Apologies in advance for the confusion!

Felienne: No problem, this is a PR that touches a lot of file, so conflicts are to be expected. I was hoping to merge it when we have not all too many others open to minimize conflicts but I guess that day will never come haha, at least not soon, so I guess we will just have to merge the new one and deal with the fallout.

PR #2363, opened 2022-03-30
[BUGFIX] Allow spaces after square brackets and braces

### LINKED ISSUE #1772: Simplification and relaxing of grammar wrt spaces
In #1771 we surround _COMMA with spaces in the level 1 grammar, simplifying further levels. This change can be made for other characters too, such as `=` and `()` and `[]`. 

There might be cases where we don't want to encourage using spaces (before an opening brace, for example) but I think we should allow it everywhere Python allows it too.

jpelay:
**Description**

As requested in #1772 I added the possibility to write spaces around square brackets and braces, just like you can in Python. So now is possible to write something like: `names [ 1 ]`  instead of throwing a parse exception.

Fixes #1772 

**How to test**

I added a test to show this behavior.
 
- [x ] Describes changes in the format above (present tense)


Felienne:
Great work, thanks @jpelay! Do we also want to do this for `<` and other comparisons?

Felienne:
That will also allow us to simplify, for example in level 14: `smaller: comparison_arg _SPACE? _SMALLER _SPACE? comparison_arg` -> `smaller: comparison_arg _SMALLER comparison_arg`

Felienne (changes requested):
Setting this to request changes to a) keep an overview and b) make sure we con't close the issue while there is more improvement to be made

Felienne:
> > Setting this to request changes to a) keep an overview and b) make sure we con't close the issue while there is more improvement to be made
> 
> Hi @Felienne! I think that's an excellent idea, and it's already done. 

Thanks!!

> A more interesting case would be keywords, do we want to add _SPACE around keywords such as `is` or `in` which are 
> always surrounded by spaces? Maybe also for all of the others?
> 
> This would allow to simplify this `ask: var _SPACE _IS _SPACE _ASK (_SPACE _print_argument)?`
> 
> to this
> 
> `ask: var _IS _ASK (_SPACE _print_argument)?`
> 
> which already looks nice, or to this:
> 
> `ask: var _IS _ASK (_print_argument)?`

Yes, that is a great idea! Do you want to do that in this PR also or in a separate one?

jpelay:
> > > Setting this to request changes to a) keep an overview and b) make sure we con't close the issue while there is more improvement to be made
> > 
> > 
> > Hi @Felienne! I think that's an excellent idea, and it's already done.
> 
> Thanks!!
> 
> > A more interesting case would be keywords, do we want to add _SPACE around keywords such as `is` or `in` which are
> > always surrounded by spaces? Maybe also for all of the others?
> > This would allow to simplify this `ask: var _SPACE _IS _SPACE _ASK (_SPACE _print_argument)?`
> > to this
> > `ask: var _IS _ASK (_SPACE _print_argument)?`
> > which already looks nice, or to this:
> > `ask: var _IS _ASK (_print_argument)?`
> 
> Yes, that is a great idea! Do you want to do that in this PR also or in a separate one?

Thank you! Can I do it in a new one? Because I think that that will involve changing the translation structure a bit!

Felienne:
ok! I made an issue for you: https://github.com/Felienne/hedy/issues/2410

Felienne:
> Thank you! Can I do it in a new one? Because I think that that will involve changing the translation structure a bit!

ok! I made an issue for you: https://github.com/Felienne/hedy/issues/2410

grammars/level14-Additions.lark
```diff
@@ -3,11 +3,11 @@
 condition:+= equality_check_dequals | smaller | bigger | not_equal | smaller_equal | bigger_equal
 
 equality_check: comparison_arg (_SPACE _IS _SPACE | _EQUALS) comparison_arg
-equality_check_dequals: comparison_arg _SPACE? _DOUBLE_EQUALS _SPACE? comparison_arg
-smaller: comparison_arg _SPACE? _SMALLER _SPACE? comparison_arg
-bigger: comparison_arg _SPACE? _LARGER _SPACE? comparison_arg
-smaller_equal: comparison_arg _SPACE? _SMALLER_EQUALS _SPACE? comparison_arg
-bigger_equal: comparison_arg _SPACE? _LARGER_EQUALS _SPACE? comparison_arg
-not_equal: comparison_arg _SPACE? _NOT_EQUALS _SPACE? comparison_arg
+equality_check_dequals: comparison_arg  _DOUBLE_EQUALS  comparison_arg
```
Felienne, line 6: Such a nice cleanup of the grammar!

Felienne (approved):

PR #2365, opened 2022-03-30
Fix typo in Level 11 default message, multiple languages

zonca:
Just fixing a simple typo, removing the extra **the** from:

> `print` works just the same but the we add a new form of the `for`

Noticed it was in multiple language versions, so I fixed them all.


TiBiBa (approved):

Felienne:
@TiBiBa I am a bit weary to merge this cause it might give Webalte conflicts, maybe we should do #2361 first?

PR #2368, opened 2022-03-30
Fix error message in level 2 #2367

### LINKED ISSUE #2367: [BUG] level 2 ask error message says that quotation marks are required
on [level 2](https://hedycode.com/hedy/2), trying to run a program consisting of `ask a` pops up an error message that states that ask requires quotation marks. however, `x is ask a` works just fine without quotation marks, and if i add the quotation marks (`x is ask 'a'`) then they're interpreted literally. i would expect either the words " and quotation marks" to be left out of the error message, or quotation marks to be required

Felienne:
Somehow we had a mistake in the error for level 2 suggesting ask and print should use quotes but they don't! 

**Description**

Fixes #2367


Felienne:
@TiBiBa This fix needs to be copied for all languages, but I am a bit wary to interact with the po files. We can copy this text to all languages that still have En but also have to remove the `'` from the existing translations. Could you fix this further?

TiBiBa:
> @TiBiBa This fix needs to be copied for all languages, but I am a bit wary to interact with the po files. We can copy this text to all languages that still have En but also have to remove the `'` from the existing translations. Could you fix this further?

Will take a look!

TiBiBa (approved):

PR #2372, opened 2022-03-30
Adds support for Persian

TiBiBa:
**Description**
In this PR we add support for Persian as well as simplify the check for right to left languages by storing the list that keeps track of these languages in `hedy_content.py` as well.

**Fixes**
No related issue.

**How to test**
Nothing to test (yet). When we are happy with the amount of Persian translations through Weblate we should merge these first, then merge main into this branch and verify that everything is working as expected. Especially the rtl support.


Felienne:
Is this superseded by #2396?

TiBiBa:
> Is this superseded by #2396?

It is! This PR can be removed, however we might want to have some control as on which point we want to make a language available to the end user. With the implementation of #2396 the language is directly available on the moment some translations from Weblate are merged, also if these are only a few words.

TiBiBa:
Closed due to no longer being relevant.

PR #2388, opened 2022-03-31
[REVERT FIX] Reverts achievements re-structure

TiBiBa:
This reverts commit e676f0d1dd34e7701f9e4ef56e35383f18bc6117.

**Description**
In this PR we reverse the achievements re-structuring as it broke a lot of functionality we didn't take into account (such as the achievements pop-ups). We should first discuss an efficient way to tackle this problem before trying to fix it again.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,7 +121,7 @@ When adding new Babel related translation the implementation is a bit more compl
 2. Next we run the following command to let Babel search for keys:
     * ```pybabel extract -F babel.cfg -o messages.pot .```
 3. We now have to add the found keys to all translation files, with the following command:
-    * ```pybabel update --omit-header -i messages.pot -d translations```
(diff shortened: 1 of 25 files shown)


CONTRIBUTING.md
```diff
@@ -121,7 +121,7 @@ When adding new Babel related translation the implementation is a bit more compl
 2. Next we run the following command to let Babel search for keys:
     * ```pybabel extract -F babel.cfg -o messages.pot .```
 3. We now have to add the found keys to all translation files, with the following command:
-    * ```pybabel update --omit-header -i messages.pot -d translations```
```
Felienne, line 124: Deze change willen we wel houden denk ik? Misschien ff een issue aanmaken?

Felienne (approved):

CONTRIBUTING.md
TiBiBa: Ja zal ik doen!

PR #2390, opened 2022-03-31
Removes .po files headers

### LINKED ISSUE #2389: Remove header from .po files on compile
**Description**
To prevent merge conflict with Weblate we should remove the metadata headers from the .po files. They don't provide any additional value as aren't interested in last updater, timestamp and other info. We should also update this in CONTRIBUTING.md to notify contributors.

TiBiBa:
**Description**
In this PR we remove the redundant headers from the .po files and update CONTRIBUTING.md to notify contributors to do the same when adding translations.

**Fixes**
This PR fixes #2389


Mark-Giesen:
Weblate probably adds these headers again, so the conflicts remain. There are some settings concerning these headers in Weblate, maybe it's better to tweak on that side.

Felienne:
> Note that using this option will lead to an error if the resulting file would not entirely be in ASCII.

This could lead to problems since some of our files are non-ASCII?

TiBiBa:
Pybabel also makes these headers on each update (therefore the `--omit-header` parameter) as currently the main header has the following metadata `"Generated-By: Babel 2.9.1\n"`.  This PR was merely a possible solution to solve the Weblate conflicts, but it is indeed an issue if it introduces new issues with non-ASCII files.

Will take a look if we can stil create the header with relevant info such as content-type and content-transfer-encoding but not all date related metadata as these are introducing the merge conflicts.

PR #2396, opened 2022-03-31
Change creation of ALL_LANGUAGES and rtol/ltor setting to dynamic.

Mark-Giesen:
Change creation of ALL_LANGUAGES and rtol/ltor setting to dynamic.

The available languages were hardcoded in our repo. Since we doe translations outside of Hedy, languages can be added outside our codebase aswell. Therefore the dictionairy use for this (ALL_LANGUAGES) is now filled dynamicly based on the configuration settings of Babel. 
With this the Right-to-Left or Left-to-Right setting of a language is changed dynamicly as well. 
The FALL_BACK_ADVENTURE dictionairy is removed from the codebase, it is no longer used since we have English as default fallback everywhere.


app.py
```diff
@@ -19,24 +19,29 @@
...
+app = Flask(__name__, static_url_path='')
+app.url_map.strict_slashes = False # Ignore trailing slashes in URLs
+babel = Babel(app)
+
 # Hedy-specific modules
 import hedy_content
+hedy_content.fill_all_languages(babel)
 import hedyweb
-from hedy_content import ALL_LANGUAGES, FALL_BACK_ADVENTURE, ALL_KEYWORD_LANGUAGES
+from hedy_content import ALL_LANGUAGES, ALL_KEYWORD_LANGUAGES
```
Felienne, line 40: Is there a reason this part of the header is moved down (I thin kit is good style to have all imports at the top)

app.py
```diff
@@ -166,13 +165,6 @@ def load_adventures_per_level(lang, level):
     level=logging.DEBUG,
     format='[%(asctime)s] %(levelname)-8s: %(message)s')
 
-app = Flask(__name__, static_url_path='')
```
Felienne, line 169: ah nice to move this up!

Felienne:
I tested this in the browser and all seems to work well! But I did not try to add a new (fake) lang to see if that works. And I think @TiBiBa might want to do a final check

app.py
Mark-Giesen: Yes, the line above it: hedy_content.fill_all_languages(babel). Defined and filled those dictionaries. This didn't work for Timon, so I had to declare the dict empty. this code could probably be lowered. The import is an empty dict then, but it's fille later.

app.py
```diff
@@ -303,12 +295,9 @@ def setup_language():
...
 
     # Set the page direction -> automatically set it to "left-to-right"
-    # Switch to "right-to-left" if one of the language in the list is selected
+    # Switch to "right-to-left" if one of the language is rtl according to Locale (from Babel) settings. 
     # This is the only place to expand / shrink the list of RTL languages -> front-end is fixed based on this value
-    g.dir = "ltr"
-    if g.lang in ['ar', 'he', 'ur']:
-        g.dir = "rtl"
-
+    g.dir = Locale(g.lang).text_direction
```
TiBiBa, line 300: What if, for some reason the `Locale` function fails? It is nice to always have a "default" value for this g attribute as it is crucial for our front-end. I propose to change this to:

```
g.dir = "ltr"
g.dir = Locale(g.lang).text_direction
```

hedy_content.py
```diff
@@ -45,6 +18,27 @@
...
+              continue
+
+          if filter(lambda x: x.endswith('.mo'), os.listdir(locale_dir)):
+              locale = Locale.parse(folder)
+              ALL_LANGUAGES[folder] = locale.display_name
+
+  # If not other translations are found, add the default locale.
+  if not ALL_LANGUAGES:
+      locale = Locale.parse(babel._default_locale)
+      ALL_LANGUAGES[babel._default_locale] = locale.display_name
```
TiBiBa, line 40: Not sure if this statement is necessary. We use the value of `ALL_LANGUAGES` for alle our translation retrieval, YAMLs and `gettext` translations. If we don't find the translation file, we also don't want to return this language as an option. In this case we still return the `default_locale` as an option, even if no translations are found at all. But as we can't retrieve any translations for this language the website will crash anyway.

app.py
TiBiBa: I didn't work for our tests either, Python is a bit ambiguous on importing a variabele that isn't there (yet). We might be able to solve this by creating a dedicated Translations class that has an object that keeps track of all our translations, but creating an empty dict first seems to work fine as well.

app.py
TiBiBa: On seconds thoughts, this will still result in the same error if the `Locale` function doesn't return anything. So the best solution might be something like this:

```
g.dir = "ltr"
if Locale(g.lang).text_direction in ["ltr", "rtl"]:
    g.dir = Locale(g.lang).text_direction

TiBiBa (changes requested):

app.py
Mark-Giesen: That's very defensive and calling the Locale getter twice, but yeah, it's save.

hedy_content.py
Mark-Giesen: I think you're right, if the default wasn't found in the for loop, it wil not be found anytime soon.

TiBiBa (approved):

Felienne:
Thanks for the code review @TiBiBa! There is a small import conflict in app.py could you have a look?

TiBiBa:
> Thanks for the code review @TiBiBa! There is a small import conflict in app.py could you have a look?

Whoops, didn't test is after the changing. Only approved based on the last commit, will take a look!

Felienne (approved):
Looks good now!

image

TiBiBa:
Will increase the width a bit of the language selector to improve the UI, then this can be merged.

TiBiBa:
I'm still having the following issue, when performing a server re-start while being logged in the following error occurs:

Schermafbeelding 2022-04-08 om 09 29 42

This is an error thrown when `if g.lang not in ALL_LANGUAGES.keys()`. For some reason we get in a "dead lock" where the current language cannot be found and we return an error. It seems that it's possible that our server handles our requests BEFORE the actual dictionary is filled with the supported languages. And I'm afraid this will happen for our actual users as well when deploying to production. I was able to re-produce this bug multiple times by using the following order:

- Start server
- Visit website
- Re-start server
- Re-load page -> error

TiBiBa (changes requested):
We can get stuck in a dead lock where the languages dict is empty and nothing works (re-loading also doesn't solve the issue as "en" isn't found either).

TiBiBa:
I can't seem to re-product this bug anymore! It will still be nice to make the selector a bit widther to make sure the languages all fit on one line. Will make a suggestion for that!

TiBiBa:
I just made a commit as the `incl-menubar.html` file wasn't edited yet, so I wasn't able to make a suggestion.

TiBiBa (approved):

PR #2405, opened 2022-04-01
Fix weblate conflict

Felienne:
There are now only 4 conflicts related to achievements left, can you pick the right version @TiBiBa?


TiBiBa:
These are the same I just solved in #2402!

Felienne:
> These are the same I just solved in #2402!

Yes sorry! Because Weblate is in conflict they could not be merged in the regular way (and I was not 100% sure how to make them in order)

Felienne:
I promise if you merge them here, Weblate will be happy :)

PR #2413, opened 2022-04-03
Adds spaces to the keywords definitions and removes them from the grammar

### LINKED ISSUE #2410: [GRAMMAR] Allow spaces around keywords
> A more interesting case would be keywords, do we want to add _SPACE around keywords such as `is` or `in` which are always surrounded by spaces? Maybe also for all of the others?

This would allow to simplify this
`ask: var _SPACE _IS _SPACE _ASK (_SPACE _print_argument)?` 

to this

`ask: var  _IS  _ASK (_SPACE _print_argument)?`

 which already looks nice, or to this:

`ask: var  _IS  _ASK (_print_argument)?`

_Originally posted by @jpelay in https://github.com/Felienne/hedy/issues/2363#issuecomment-1083845199_

jpelay:
**Description**

In a previous PR (#2363) we included spaces around the definitions of symbols in the grammar, such as [, (, ==, etc.  This PR extends that idea to the definition of keywords in the grammar, including the surrounding spaces directly to their definitions; with the only exception of else and times.
* Else: The requirement of spaces around else varies from where it is first introduced in level 5, and where blocks are introduced in level 8. So a single rule can't handle these both scenarios.
* Times: Similar case to else, the requirement of spaces after times varies from where it is first introduced to when blocks are introduced. If we make spaces optional since the beginning this would be valid on level 7:

image

Another consequence of this is that print and ask can work even if they don't have surrounding spaces, so a program like `print'Hello World'` would be accepted.

Fixes #2410


grammars/keywords-ar.lark
```diff
@@ -1,46 +1,46 @@
...
-_ECHO: "ردد" | "echo"
-_FORWARD: "تقدم" | "forward"
-left: "left"
-right: "right"
-_IS: "هو" | "is"
-_AT: "بشكل" | "at"
-random: "عشوائي" | "random"
-_IN: "في" | "in"
-_IF: "اذا" | "if"
+_PRINT: ("قول" | "print") _SPACE?
```
Felienne, line 1: I will have to update my generation script to make it so that this gets added automatically (note to self)

tests/test_level/test_level_05.py
```diff
@@ -454,16 +454,6 @@ def test_if_print_has_no_turtle(self):
     if computerkeuze is jouwkeuze print 'gelijkspel!'""")
     result = hedy.transpile_inner(code, self.level)
     self.assertEqual(False, result.has_turtle)
-  def test_no_space_after_keyword_gives_missing(self):
```
Felienne, line 457: I think we have an issue open for this (this too is a note to self)

Felienne (approved):
Wow, this is great @jpelay, cleaning up the grammar A LOT!

Felienne:
I added a few tests @jpelay, not so much to validate your work, more to make sure that future changes don't accidentally undo this work!

tests/test_level/test_level_05.py
Felienne: Ah here is the (closed) issue: https://github.com/Felienne/hedy/issues/857

But I see that for some reason we are not allowing a space after _IS, is that intentional @jpelay (I can imagine it leads to all sorts of issues!)

jpelay: I didn't think about it, I followed closely the grammar that was already in place. I can try changing it and see if it doesn't raise weird issues!

jpelay:
> I added a few tests @jpelay, not so much to validate your work, more to make sure that future changes don't accidentally undo this work!

Thank you!

PR #2416, opened 2022-04-04
[CLEAN-UP] Removes fixed error strings, adds dynamic country selector

### LINKED ISSUE #2287: [CLEAN-UP] Some server errors still depend on fixed strings
**Description**
Currently we still depend on fixed strings in some places, especially the program related server calls. We should replace these all with `gettext()` error messages. While the return of these errors is unlikely as they will only occur if the user deviates from the normal flow it is still a good practice to have all error be returned in the same way.

### LINKED ISSUE #2415: [CLEAN-UP] Remove hard-coded country list
**Description**
It would be nice if we removed the hardcoded country list in `auth.py` and replace it with something a bit more sophisticated such as https://pypi.org/project/pycountry/. We current keep a duplicate list on the back-end and the front-end, this doesn't make any sense and is due for errors. We should dynamically create a list on the back-end and simply parse it to the front-end, for example with a global template function.

TiBiBa:
**Description**
In this PR we fix two small issues. First we clean a few hard-coded strings that where still placed on the `/verify` route, always returning an English string if something went wrong. Secondly we replace the currently hard-coded countries dictionary by a library based one. Cleaning-up the code and making sure we are up-to-date with possible changes in countries, abbreviations and so on.

**Fixes**
This PR fixes #2287 and fixes #2415

**How to test**
This is a clean-up task, if nothing is broken it works. However, it might be nice to verify that the country selector does indeed still work on the signup page as well as the update profile section.


Felienne:
> However, it might be nice to verify that the country selector does indeed still work on the signup page as well as the update profile section.

What does "work" mean here? It does show properly:

image

Should it be localized? Cause if I set the page to Dutch the langs still show in En:

image
 
(but that it as it was as far as I can see)

Felienne:
Ah in that case: it works! 🥳

Felienne (approved):

PR #2419, opened 2022-04-04
Bug fix: variables header would show when running turtle code

### LINKED ISSUE #2370: [BUG] Variables header visible when using Turtle
**Describe the bug**
When using the turtle the variabels header is visible on the output console. This shouldn't be because it is still in production and should be automatically turned off (and visible) by a set boolean in our TypeScript. Slightly related: The dropdown marker is also shortly visible each time we re-load the page. This might have something to do with the hide function being called **after** some `onload()` function. It would be nice to implement this the other way around: set `hidden` in template, only show if set as active. 

Schermafbeelding 2022-03-30 om 11 10 53

**Expected behavior**
All variable view related UI elements should be hidden when the function is turned off.

fRedelaar:
**Description**
Removed the line that caused the bug. When running turtle code, the variable header would pop up. This is **unwanted** since the variable list is still under a feature flag. 

**Screenshot of the bug**
image

**Fixes #2370**

**How to test**
Run any level with turtle code. There should NOT be a blue header saying "variables" in the output window.


TiBiBa (approved):
All works good for me, nice!

PR #2422, opened 2022-04-04
[FIX] Change path of commands to fix includes

### LINKED ISSUE #2418: Add code tests for the new commands YAMLs

TiBiBa:
**Description**
I thought we also had to add some testing for the commands YAML, but apparently that is already done in #2374. We do however fix an issue where the `Commands` class still incorrectly looked for the `/coursedata` folder instead of the `/content` folder. 

**Fixes**
This PR fixes #2418 (but actually a related issue)

**How to test**
Verify that the commands dropdown menu is working again.


Felienne (approved):
Works!

image

PR #2423, opened 2022-04-04
[BUG FIX] Cheatsheets work again

TiBiBa:
**Description**
Due to our separation of commands and level-defaults the cheatsheets where no longer working. This is due to being dependent on the commands being stored within the level-defaults. We fix this as well as add some additional validation to make sure the requested cheatsheet is valid, otherwise we return a 404 error with the "no such level" message. 

**Fixes**
This has no related issue.

**How to test**
Verify that the cheatsheets are working again by navigation to `/cheatsheet/<level>`. Also validate that adding an invalid level to the url results in a 404 error with a "no such level" message.


Felienne (approved):
Works!

PR #2428, opened 2022-04-04
[CHORE] Store level-defaults in adventures YAMLs

### LINKED ISSUE #2417: [CHORE] Store level-defaults within Adventures YAMLs
**Description**
We already did some big clean-up to extract the `commands` from the level-defaults. Making it easier to adjust these as these are conceptually a separate branch of content. However, as the `level-defaults` are no longer used as a separate explanation or attribute (they are simply the firs adventure) it would be nice to implement them as an adventure as well. Cleaning up the current YAML structure.

TiBiBa:
**Description**
In this PR we store the `level-defaults` within the adventures. Reducing the amount of YAML files as well as the overall complexity of our 'tab switching' and back-end structure to merge everything together into tabs. Currently we have the following structure in the back-end which we merge on the front-end:

- Level-defaults
- Adventures
- Teacher adventures

We simplify this by storing the level-defaults as a new adventure named "intro" within the Adventures. Storing them at the top of the YAML file ensures that they are still rendered first. We perform some small front-end tweaking so adventures stored on the "intro" adventure are not actually stored as adventure programs but as "level programs". Simplifying the default name. We also enforce an "adventure structure" making sure that the adventures are always in the same order, independent if they actually exist. This to ensure that the YAML structure does not influence with our front-end experience as Weblate does concern with the file ordering, only the keys themselves.

**Fixes**
This PR fixes #2417

**How to test**
This needs some extensive testing to make sure we didn't break anything. As our current adventure merging magic is quite complex it is likely a bug has been introduced somewhere. Other than this the PR is a clean-up, everything should function exactly as before.


TiBiBa:
We have some failing test on `el.yaml`, but I can't figure out why. It seems some empty code blocks are generated for which the tests try to catch the content with `contents[0]`. For clearly explained, the following element exists:

`<pre><code></code></pre>`

We use a double retrieval to get the actual program with `contents[0].contents[0]`, however the second one doesn't exist! I can't figure out how and why these element blocks are created as the YAML structure seems fine. We now pass them with a try catch if for some reason we are unable to extract the content, but this is of course undesirable.

Mark-Giesen:
After extensive testing we need a very carefull transition to Weblate. In Weblate all components are linked to level-defaults as the "repo-holder". So this needs some extensive thinking as well.

TiBiBa:
> After extensive testing we need a very carefull transition to Weblate. In Weblate all components are linked to level-defaults as the "repo-holder". So this needs some extensive thinking as well.

Agree! Shall we take a look on Thursday how to tackle this problem? This PR doesn't have a very high priority, however the Weblate updates on the current level-defaults folder have to be copied once again right before merging.

Felienne:
Making this a draft so it is clear we are not ready to merge

hedy_content.py
```diff
@@ -49,67 +49,26 @@
...
-
-    return copy.deepcopy(self.levels.get(int(level), {}))
-
-
-class NoSuchDefaults:
-  def get_defaults(self):
-    return {}
-
+ADVENTURE_ORDER = [
+    'default',
```
Felienne, line 53: here it is called default, but in the PR description I also saw "intro" can we keep that to one term? I like defaults because in my mind these things will always be level defaults, but for other people, I think `intro` is clearer.

TiBiBa: This is only the dictionary key in the YAMLs, so end users won't see it. We can change it to 'intro', but do have to change the keys in all YAMLs as well.

static/js/tabs.ts
```diff
@@ -85,23 +85,26 @@ function resetWindow() {
...
+          $('#program_name').val(window.State.level_title + ' ' + window.State.level);
+        } else {
+          $('#program_name').val(adventures [tabName].default_save_name + ' - ' + window.State.level_title + ' ' + window.State.level);
+        }
         theGlobalEditor?.setValue(adventures [tabName].start_code);
       }
     }
 
-    window.State.adventure_name = tabName === 'level' ? undefined : tabName;
+    window.State.adventure_name = tabName === 'intro' ? undefined : tabName;
```
Felienne, line 107: Ah indeed here it is intro and not default

templates/incl-adventure-tabs.html
```diff
@@ -1,81 +1,42 @@
...
-                    </div>
-                {% endif %}
-            {% endfor %}
-            {% for assignment in adventures %}
-              {% if assignment.short_name != 'level' and ('adventures' not in customizations or customizations['adventures'][assignment.short_name] and level|int in customizations['adventures'][assignment.short_name]) %}
-                <div class="tab {% if assignment.short_name == "money" %}money_tab{% endif %} z-10 whitespace-nowrap flex items-center justify-center" tabindex="0" data-tab="{{assignment.short_name}}">
-                    {{ assignment.name }}
+            {% for adventure in adventures %}
+              {% if 'adventures' not in customizations or (customizations['adventures'][adventure.short_name] and level|int in customizations['adventures'][adventure.short_name]) %}
+                <div class="tab {% if loop.index == 1 %}tab-selected{% endif %} z-10 whitespace-nowrap flex items-center justify-center" tabindex="0" data-tab="{{adventure.short_name}}">
```
Felienne, line 6: Is that `loop.index` to do something special for the intro tab? Maybe add a comment?

static/js/tabs.ts
TiBiBa: I think this is due to the dict key being called `default` but the default_save_name being `intro`. Agree that this is confusing, will take a look what would be the best approach to make this clearer!

templates/incl-adventure-tabs.html
TiBiBa: This is independent of the intro tab, we simply want to set the first tab as the selected one by default. When there is a loaded program we overwrite the `tab-selected` with the loaded adventure.

hedy_content.py
Felienne: If possible I would prefer that (can also be done in a separate PR)

Mark-Giesen (approved):

PR #2429, opened 2022-04-04
remove space after commands

### LINKED ISSUE #2424: [SYNTAX HIGHLIGHTER] No more space after keywords needed
Since #2413 a space is no longer mandatory between keywords are arguments:

image

This needs to be updated in the highlighter

thjazi39:
fix #2424 

The space after an order is no longer necessary. For example we have at level 3 :
image

Be careful though, this should also induce quite a bit of over-coloring.
Indeed, to recognize with a regex a keyword, we looked if it was a single word or not. But without spaces, it makes the task even more difficult. And since we don't have the power of a grammar, we're bound to have failures

In particular, short keywords (`is, in, at, to`) will be recognized when they are prefixes of longer words. And preventing this will be difficult


Felienne:
> Be careful though, this should also induce quite a bit of over-coloring. Indeed, to recognize with a regex a keyword, we looked if it was a single word or not. But without spaces, it makes the task even more difficult. And since we don't have the power of a grammar, we're bound to have failures
 
> In particular, short keywords (`is, in, at, to`) will be recognized when they are prefixes of longer words. And preventing this will be difficult

Yeah I understand! Could it be a solution to only color this (without a space) at the beginning of a line? That would reduce overcoloring a lot (although I understand it introduces a bit of a difference between the grammar and the coloring)

thjazi39:
### Changes in keyword management ###
After a reading of the PR, and since it may change, I changed the old system keywords/symbols (a keyword must be alone, while a symbol is independent), By another system :

There are several categories of keywords: 
- SP_K_SP
  These are the keywords that must be "alone" so neither preceded nor followed directly by a word 

- K
  These are the keywords that are independent of the context (formerly the symbols).
  In particular, even if they are between 2 words, the syntax highlighting will select them

- SP_K
  This category of keywords allows you to have keywords that are not preceded
  by another word, but that can be followed immediately by another word. (see the PR #2413)

- K_SP
  This category of keywords allows you to have keywords that can be preceded immediately
  by another word, but that are not followed by another word. (only use with ELIF)

### This leads to a small modification of the file structure : ###

#### Before : ####
image
image

#### And now: ####
image
image

thjazi39:
> Yeah I understand! Could it be a solution to only color these keywords without a space **only** at the beginning of a line? That would reduce overcoloring a lot (although I understand it introduces a bit of a difference between the grammar and the coloring)

Unless I made a mistake reading the grammar, but it seems to me that these keywords still require spaces on both sides 
image
image
image

The system allows you to leave them as before, and this will not include any difference on these keywords

Felienne:
> Unless I made a mistake reading the grammar, but it seems to me that these keywords still require spaces on both sides

Yes that is correct!

Felienne:
@thjazi39 ah of course we now have a conflict since I merged #2434, can you take a look?

thjazi39:
The special cases have been treated, so it should be good !

On the other hand, when I test the PR #2413 example, it doesn't color, and it doesn't work either: So the coloring is consistent with the grammar, but I didn't manage to launch the example, and I don't know if it's normal (yet I'm up to date with the main)

Level 7 : 
image
image

Felienne:
> The special cases have been treated, so it should be good !
> 
> On the other hand, when I test the PR #2413 example, it doesn't color, and it doesn't work either: So the coloring is consistent with the grammar, but I didn't manage to launch the example, and I don't know if it's normal (yet I'm up to date with the main)
> 
> Level 7 : image image

Yeah in the keywords file this is explained: 

```
// not including spaces after times, because the requirement of spaces after times changes after blocks are introduced
```

I am not sure it is needed though but this PR is in line with the grammar as it is so we can merge it!

Felienne (approved):

PR #2434, opened 2022-04-04
Precedence of numbers by a symbol

### LINKED ISSUE #2433: [BUG] Number highlighting doesn't always work correctly
**Describe the bug**
I found that in level 16 for some reason not all numbers are highlighted within the number list:

Schermafbeelding 2022-04-04 om 16 18 33

**Expected behavior**
All numbers within the list should be colored.

thjazi39:
fix #2433 

Following the removal of the lookbehind for Internet Explorer and Safari, the numbers were not displayed correctly.
Indeed, we want to recognize numbers that are alone, that is to say that are not included in a word.
In order to solve this, I had used lookbehind and lookahead to see what is around without selecting the context. (as said here : #2323)
But without lookbehinds, I use a trick that requires the preceding character to be a space or a line break (which works well for keywords)
However, for numbers, since you can have a symbol immediately followed by a number, then it didn't detect the number anymore.
To solve this, I simply modified the recognition of a number by adding the level symbols to make it work.

level 6-11 :
image

level 12-: 
image

the bug (level 16): 
image


Felienne (approved):
Seems to work!

image

PR #2438, opened 2022-04-05
level 1 parsons problem try yaml

xxxpokemon:
first tab for implementing parsons problem


TiBiBa:
Converted to draft to make sure we don't test and merge this before it's ready.

templates/incl-adventure-tabs.html
```diff
@@ -123,19 +123,27 @@ <h2 class="z-10 my-0 ml-auto pt-8 text-red-500 text-base">{{ _('specific_adventu
...
         <div data-tabtarget="{{parsons['tab_name']}}" class="hidden">
-            <div>
-                {{ parsons['text']|commonmark }}
-
-
-                {% for key, line in parsons.code_lines.items() %}
-                    <div id="div1">
+    
+            <div class="flex flex-row">
+                <div class="{% if example_code %}w-1/2{% else %}w-full{% endif %} p-2 ltr:mr-2 rtl:ml-2">
```
TiBiBa, line 128: We only need these two lines if we have some "example_code" and want to split the panel into two views horizontally. An explanation on the left and the code examples on the right (as we do with adventures). Otherwise this doesn't need to be a `flex flex-row` and we don't need the `if` statement as well.

Felienne:
Hi @xxxpokemon!

Just tagging you because quite a lot has changed in the repo, so maybe you want to fix the merge conflict now before it gets worse? :)

Felienne:
Thanks @xxxpokemon! I am making this into a draft for now because we are not ready to merge this at the moment.

Felienne:
Hi @xxxpokemon and @TiBiBa 

I just tested this for  talk I am giving (in which I will be pretending this feature is ready, hahaha) and I notices this weird error:

image

TiBiBa:
> Hi @xxxpokemon and @TiBiBa 
> 
> 
> 
> I just tested this for  talk I am giving (in which I will be pretending this feature is ready, hahaha) and I notices this weird error:
> 
> 
> 
> image
> 
> 
> 
> I guess we are hiding line numbers in the code, accidentally?

Correct! I only fixed the drag/drop syntax highlighting, we still have to alter the way we read the different lines. As previously they where plain text but now they are ace editors so it works a bit different.

Felienne:
Closing this in favor of #2663

PR #2443, opened 2022-04-06
[IMPROVEMENT] Always dynamically parse keywords on /explore and program view

### LINKED ISSUE #2439: [BUG] Dynamic syntax highlighting doesn't work on program view page
**Describe the bug**
Currently the dynamic syntax highlighting seems broken on the program view. This might be something related to the `window.State` value set for the keyword language that is used on the front-end to decide on a highlighter. This is actually part of a larger issue:

- How do we highlight syntax for a program that is not in your language?

We have several options, but non are ideal. As currently the language switcher is also shown, enabling the user to translate the viewed program. We might want to remove this as well? We already have a `readonly` boolean parsed, so we can easily say "only show if ..." The following options are feasible for the program view page:

- Show English syntax highlighting (doesn't highlight a program with non-English keywords)
- Show the user keyword language syntax highlighting (actually doesn't make any sense)
- Show the syntax highlighter for the program language
    - **Note: the program language is not always the same as the keyword language used**
    - We have to remove the language selector in this situation as it doesn't make any sense

Really interested in your views on this issue as it is larger than only the syntax highlighting. We currently show all adventures when searching on `/explore`, do we want to limit this to the user language? @Felienne @boryanagoncharenko 

Schermafbeelding 2022-04-05 om 13 24 13

**Expected behavior**
The syntax should be correctly highlighted, we do however have to discuss in which way.

TiBiBa:
**Description**
In this PR we fix an issue where the syntax highlighting was incorrect when viewing a program using non-user keywords. As on the `/explore` page we show all relevant programs instead of only ones with the same language as the user, this inconsistency can be confusing. We solve this my dynamically parse all keywords to the current user language. We also remove the keyword language switcher from the program view page to remove ambiguity.

**Fixes**
This PR fixes #2439

**How to test**
Create a new (or use a second) account and set this to a non-native language, for example Norsk. Create a program, save and share it to make sure it is visible on the `/explore` page. Make sure language specific keywords are used. Then return to your original account and verify that the keywords of the program are dynamically changed to the current user language when viewing and running the program.


Felienne:
Also, running code does not work now (makes sense I guess, the parser will also be set to Dutch) but I do not get an error message:

image

Felienne:
It seems I always get Spanish, I tried with Dutch and also Norsk (fresh account):

image

TiBiBa:
Weird! Will take a look.

TiBiBa:
@Felienne the bug should be fixed and this is ready for merging!

Felienne:
> @Felienne the bug should be fixed and this is ready for merging!

ok! Testing again now

Felienne:
hmmmmm I still see programs in Spanish independent of my keyword lang:

image

Tried in Chrome and Safari on Mac
sa

TiBiBa:
> hmmmmm I still see programs in Spanish independent of my keyword lang:
> 
> image
> 
> Tried in Chrome and Safari on Mac sa

I issue this is due how we call the local db. For some reason this is done by-reference and the programs are now all actually stored with Spanish keyword. I'm not sure why locally this happens, but we have to do something like:

`code = program['code']`

To prevent his overwriting. This only happens locally and not on the server (will look into this someday). I accidentally wrote some code like `program['code'] = hedy.translate(program['code')` overwriting the original program with the current keyword language translation! This is now fixed and only way to verify this is by creating a new program with non-english keywords and verify that it is now indeed parsed correctly on runtime.

Felienne:
> > hmmmmm I still see programs in Spanish independent of my keyword lang:
> > image
> > Tried in Chrome and Safari on Mac sa
> 
> This issue is due to how we call the local db. For some reason this is done by-reference and the programs are now all actually stored with Spanish keyword. I'm not sure why locally this happens, but we have to do something like:
> 
> `code = program['code']`
> 
> To prevent his overwriting. This only happens locally and not on the server (will look into this someday). I accidentally wrote some code like `program['code'] = hedy.translate(program['code')` overwriting the original program with the current keyword language translation! This is now fixed and only way to verify this is by creating a new program with non-english keywords and verify that it is now indeed parsed correctly on runtime.

Check, it works for a new program!

image

Do we need an issue for the local db issue (don't think it is a very important issue, so we can also leave it)

Felienne (approved):

PR #2446, opened 2022-04-06
fix 2442

### LINKED ISSUE #2442: [SYNTAX HIGHLIGHTING] Overcoloring after turn
Level 1:

image

Would it be nicer if just left and right were highlighted?

thjazi39:
At the time of the change of coloring system Pr # there was not yet the keywords left and right. So the 'keyword' coloring was continued after turn.

When the internationalization of left and right keywords was added, the coloring after turn was not restricted to these keywords.

Here it is ! Note that, from what I understood, Left and Right are only used at level 1 (at level 2, they don't seem to be recognized, so they are not colored. And it is the same for all the following levels)

fix #2442 

- Level 1 :
image

- Level 2 ->11 :
image

- Level 6 -> 10 : 
image

- Level 11 -> : 
image


thjazi39:
I made the merge with the last system change

Felienne (approved):
Works!

image

PR #2456, opened 2022-04-06
[IMPROVEMENT] Reduces db calls by simplifying user data retrieval

### LINKED ISSUE #2445: [CLEAN-UP] Reduce db calls by no longer requesting profile on each request
Currently we request the user profile from the database at each request to verify that there are no new messages. But as we only use messages for "class invitations" this is not so likely. It might be better to only load class invitations when visiting the `profile` page, where they are already loaded in the "messages" section.

TiBiBa:
**Description**
In this PR we make some small fix to improve performance overall. We remove the `@app.context_processor` db calls. These calls where made at each template render to make sure the data was available, however in most cases this data was redundant. We now only make these db calls when rendering the actual relevant pages. We also remove some redundant Babel explanation as that is already mentioned (and improved) in CONTRIBUTING.md. Lastly we remove the "set a preferred language" header, which has become redundant as it's already over 6 months shown to users without a set language. We still catch the possible errors, so this is no longer an issue. And we remove the "Statistics" header on the profile page as it is not really relevant.

**Fixes**
This PR fixes #2445

**How to test**
This is a clean-up task, if nothing is broken it works.

### DIFF
--- a/app.py
+++ b/app.py
@@ -960,11 +960,13 @@ def reset_page():
 @app.route('/my-profile', methods=['GET'])
 @requires_login
 def profile_page(user):
+
+    profile = DATABASE.user_by_username(user['username'])
     programs = DATABASE.public_programs_for_user(user['username'])
     public_profile_settings = DATABASE.get_public_profile_settings(current_user()['username'])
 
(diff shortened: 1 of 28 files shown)


Felienne (approved):
Seems to all still work!

PR #2459, opened 2022-04-06
add quiz level 9

OnnoEbbens:
There is no quiz yet in level 9, therefore some of my students decided to make 2 questions.

I think the students would really like it if they see their collaboration for Hedy. Unfortunately I don't have time to create a whole quiz with 10 questions. So I think this will be a temporary quiz until a complete quiz is created.


Mark-Giesen:
Adding texts to another language than our source language (EN) will be removed by the translation tool.

Mark-Giesen:
there's another PR with quiz questions for level 9 as well. So they are coming soon!

OnnoEbbens:
Ok Thanks! I will cancel this PR now

Issue #2436, opened 2022-04-05
[IMPROVEMENT] Refactor tests


boryanagoncharenko:
The goal of the task is to ensure:
- No dead tests (because they are commented, or they don't have the right indentation and are not picked up)
- No TODOs in tests (it is hard to test something access al levels, so sometimes we leave TODOs for other levels)
- Better visibility of how a single feature is tested across levels (e.g. using inheritance)
- No level gaps and no multi_level_tester with a single level (this is an indication that the test was supposed to be executed for multiple levels but ended up not)

PR #2461, opened 2022-04-06
Refactor tests for level 1 and 2 #2436

boryanagoncharenko:

< Refactor tests for level 1 and 2 >

**Description**

< Refactor tests for levels 1 and 2. Remove Empty Program Exception > Use present tense without a subject to describe your PR, for example: "Adds translations of levels 1 to 12 to Polish"

**Fixes < #2436 >**

**How to test**

< This PR is meant only to refactor tests. Since the changes in the test files are significant, I advise looking at the latest version (as opposed to comparing with the previous versions). Check whether the tests files:
- contain no dead code
- group correctly tests according to the set convention
- share test names to improve traceability

Note that at this point it is not possible to make test classes of different levels inherit one another. Even though using inheritance would immensely improve the navigation of a single test scenario across levels, all public tests of lower levels become visible and, therefore, executed in the current level.>

**Checklist**
  

exceptions.py
```diff
@@ -154,10 +154,6 @@ class UnquotedAssignTextException(HedyException):
     def __init__(self, text):
         super().__init__('Unquoted Assignment', text=text)
 
-class EmptyProgramException(HedyException):
```
Felienne, line 157: I am not sure why we are removing this exception here, how does that relate to the tests?

boryanagoncharenko: This is just cleaning dead code. The exception is not used. @TiBiBa commented out the exception throwing when introducing babel. I also confirmed with him that this exception is not used and not needed. Would you like to reintroduce it?

TiBiBa: Yes this exception has become redundant when introducing the inline comments. If I remember correctly @jpelay removed the throwing of this error when working on fixing the failing tests of the comments.

boryanagoncharenko: sorry, I just realized I did not answered the exact question. I removed this exception together with its tests which were commented out. I only noticed it is dead code because of cleaning the tests. I could also remove the exception in a separate PR.

jpelay: Yes this exception is no longer used. If we include it, then programs that consist exclusevely of comments (of which the adventures have a lot) will fail, because since lark will ignore those tokens, to our transpiler it will be as if they're empty programs.

Felienne (approved):

PR #2474, opened 2022-04-07
Add test for highlighting

thjazi39:
The principle of the syntax colouring test is as follows :
We get the rules of syntax highlighting and we simulate the operation of Ace by a function in python. This simulation returns a "colored" string and we compare this result with what we want.

## Data recovery ##

PR #2527 and PR #2672 now allows to retrieve the syntax coloring rules in the `highlighting/highlighting.json` and  `highlighting/highlighting-trad.json`  files 
This is done through the `getRules` method.

### System 1 ###
You can write tests with this format:

image

This will then be transformed into a character by character coloring. (system 2)
The list of abbreviations used is available in the `Highlighter.py` file

### System 2 ###
But you can prefer, in some cases, to write directly the desired coloring letter by letter, in this case, you can use this format:

image

For the coding of the desired coloring : For each test, we indicate the level concerned, the code we want to test, and the expected result, letter by letter, with the following code:
- T : the text (so in white)
- K : the keywords (in red)
- C : the comments (in grey)
- N : the numbers (in green)
- S : strings (in purple)
- I : blanks (highlighted in pink)
- etc....

Note that if the coloring of a character is not important, as is often the case with spaces, simply put a space in the color code

## Simulation of syntactic coloring: ##

The simulation of Ace's operation takes place in the function `applyRules`, and returns a string, of the same size as the code, with character by character coloring.


static/js/syntaxModesRules.ts
```diff
@@ -8,13 +8,16 @@ import {LANG_tr} from './syntaxLang-tr';
 import {LANG_id} from './syntaxLang-id';
 import {LANG_nb_NO} from './syntaxLang-nb_NO';
 
+// #/#/@/#/# We don't want to keep this block of code for automatic test execution
```
rix0rrr, line 11: How about, instead of post-processing this file at a textual level to remove the parts you don't want to test, you factor out the parts you DO want to test?

So, instead of doing this:

```
┌───────────────────────┐                        ┌───────────────────────┐
│                       │                        │                       │
│  syntaxModesRules.ts  │───────────────────────▶│      testable.ts      │
│                       │      ┌─────────┐       │                       │
└───────────────────────┘      │script.py│       └───────────────────────┘
                               └─────────┘                                
```

Do this instead:

```
            ┌─────────────────────────────────┐            
            │                                 │            
            │        essenceOfRules.ts        │            
            │                                 │            
            └─────────────────────────────────┘            
            ▲                                 ▲            
            │                                 │            
     import │                                 │  import    
            │                                 │            
            │                                 │            
┌───────────────────────┐         ┌───────────────────────┐
│                       │         │                       │
│  syntaxModesRules.ts  │         │      testable.ts      │
│                       │         │                       │
└───────────────────────┘         └───────────────────────┘
```

tests/test_highlighting/highlighter.py
```diff
@@ -0,0 +1,157 @@
...
+# This function allows to check 2 syntactic colorations one desired
+# and the other obtained, and manages some special cases
+
+# Returns a boolean and a value:
+# - If the 2 highlights are consistent, the boolean will be true and the value unused
+# - if there is an error in the highlight, the boolean will be false and
+#   the value will be the number of the 1st character where there is an inconsistency
+def check(result,expected):
+    if len(result) != len(expected):
+        raise ValueError("The desired highlight and the obtained highlight do not have the same length !")
```
rix0rrr, line 30: This stops all tests completely. Shouldn't it just be a type of error that prints examples of the wrong values?

tests/test_highlighting/highlighter.py
```diff
@@ -0,0 +1,157 @@
...
+import re
+
+# Transcription of the used tokens into a symbol (a letter) in order to apply a coloring
+Tokencode = {
+    "text" :               'T',
+    "keyword" :            'K',
+    "comment" :            'C',
+    "variable" :           'N',
+    "constant.character" : 'S',
+    "invalid" :            'I',
```
rix0rrr, line 16: This is clever! I like the idea! I wonder if it's going to be annoying to edit (don't forget that tests also need to evolve over time as the code base changes, they're not set in stone, and readability and maintainability count a lot for tests as well!), but the encoding can always be changed.

A potential alternative encoding could be something like:

```
[print:kw] hello world
```

Which would have the advantage of being more readable and you don't have to fiddle so much with the exact character offsets. Something to think about.

tests/test_highlighting/highlighter.py
```diff
@@ -0,0 +1,157 @@
...
+#######################################################################################
+#######################################################################################
+
+
+
+
+import json
+
+# open data for regex
+file = open('data.json')
```
rix0rrr, line 122: Data driven tests are great, but JSON is an annoying format to edit. Especially once the examples will span multiple lines.

I don't think that there's a lot of benefit to putting tests in JSON vs. putting them in plain Python. Given a couple of well-chosen helpers, tests in Python could look like this:

```py
class HighlighterTests:
  def test_print_is_recognized_as_keyword(self):
    self.assertHighlighted(
      'print hello world',
      'KKKKK TTTTT TTTTT',
    )
```

That's equally as readable and convenient, but allows for more flexibility in the future (and integrates nicely into existing test runner frameworks)

static/js/syntaxModesRules.ts
rix0rrr: Perhaps an even better way, instead of generating `data.json` solely for testing, is to make it the official intermediate format:

- It gets generated at dev time from whatever convenient (Python?) script you all care to write. That way, the syntax highlighting rules don't have to live in TypeScript.
- The tests can load it and simulate the syntax highlighting according to the rules.
- It gets bundled into the `.js` bundle by `esbuild`-- in TypeScript you'll be able to do `import rules_en from './en.json'` and have the JSON object available, ready for passing into Ace.

Like this:

```
┌──────────────────────────┐                                                                                    
│                          │                                                                                    
│ genenerateHighlights.py  │                                                                                    
│                          │                                                                                    
└──────────────────────────┘           ┌─────────────────────────────────┐                                      
              │                        │             en.json             │      import as data                  
              └───────────────────────▶│             nl.json             │◀────────────────────────┐            
                      generate at      │             de.json             │                         │            
                       dev time        └─────────────────────────────────┘             ┌───────────────────────┐
                                                        ▲                              │                       │
                                                        │                              │  syntaxModesRules.ts  │
                                                        │                              │                       │
                                                        │                              └───────────────────────┘
                                          ┌──────────────────────────┐                                          
                                          │                          │                                          
                                          │    testHighlights.py     │                                          
                                          │                          │                                          
                                          └──────────────────────────┘                                          
```

rix0rrr (commented):
Very cool. I like what you're doing. I think we can pull this thread even further and end up putting the syntax highlighting rules into Python, where everyone is much more comfortable (and allows opportunities for further integration and autogeneration etc!)

tests/test_highlighting/highlighter.py
rix0rrr: If we have multiline tests we could make Python helpers that do this:


```py
  def test_multiline_example(self):
    self.assertHighlighted(
      'naam is bert',
      'VVVV KK TTTT',

      'print naam er zit een banaan in je oor',
      'KKKKK VVVV TT TTT TTT TTTTTT TT TT TTT',
    )
```

Which will be impossible to achieve in JSON.

tests/test_highlighting/highlighter.py
thjazi39: Normally not, since expected is generated from the code, which normally has the same size as Result, if we get to this Raise, it's either that the test script doesn't work, or that the tests are inconsistent.

rix0rrr: > Normally not, since expected is generated from the code

I'm not sure I follow. From the code below:

```py
for test in dataTests:
    # get the others data for the tests
    code = test['code']
    expected = test['result']

    # simulation and comparison of the result
    result = simulateRulesWithoutToken(Rules,code)
    valid, indError = check(result,expected)
```

It looks like `expected` comes from the JSON, and so is whatever a human types in there?

thjazi39:
I added in the highlighting tests all the examples that the children will face on the site (that is, all the code present in the adventure/en.yaml file)

I think I'll do the same for the Dutch, (this will multiply the number of tests by 2, but they are very fast tests to run, so it should not be a problem)
The Dutch will allow to test if the compound keywords are not a problem, and it will make sure that the Dutch students who are currently using Hedy don't have any problem

We should also do this with a language that doesn't use the Latin alphabet, and this will make a rather important test base, which will avoid making mistakes

Then we'll have to add, case by case, less realistic examples, in order to test the limits of the syntax highlighting

thjazi39:
We now support 2 systems of writing tests. The first one, with character by character coding. And the second one, with a brace system and a pipe (characters that are never used in Hedy)

You can use several abbreviations, depending on your preferences, and this is summarized in this variable:
image

thjazi39:
Just for information:
I'm currently testing my python function that will simulate Ace (with token). This part is very important, because it is then on this function that we will automatically test the new syntax highlighting rules.

Currently, the function is not yet perfect, including this code: 
```
qmczqdqaqmcqbadq
grbnorbTestananaeqwswsmsazqdqaqmcqbadqcc
grbnorbTestananaeqwswsmsazqdqaqmcqbadqcc

grbnorbTestananaeqwswsmsbzqdqaqmcqbadqcc
grbnorbTestananaeqwswsmsazqdqaqmcqbadqcc

grbnorbTestananaeqwswsmsbzqdqaqmcqbadqcc
```
Which, with this automaton:
image


Where my function does not give the same coloring as Ace:
Ace:
image
My function:
```
CCKNNNNNNN NNNNN
NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN NNNNN  
NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN NNNNN  

NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN NNNNN  
NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN NNNNN  

NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN NNNNN  
```
(There are problems due to the recognition of line breaks)


I found the sources of Ace's Tokenizer: https://github.com/ajaxorg/ace/blob/master/lib/ace/tokenizer.js (long live open-source!) and this should allow me to solve this problem and to have a good basis to have robust and context-based syntax highlighting automata!

tests/Highlighter.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+    # This function allows to check 2 syntactic colorations one desired
+    # and the other obtained, and manages some special cases
+
+    # Returns a boolean and a value:
+    # - If the 2 highlights are consistent, the boolean will be true and the value unused
+    # - if there is an error in the highlight, the boolean will be false and
+    #   the value will be the number of the 1st character where there is an inconsistency
+    def check(self,result,expected):
```
rix0rrr, line 279: Just use `self.assertEquals()` instead of writing your own matcher function.

When you are comparing expected to actual, you want as much information as possible, and the built-in functions do a pretty good job of printing a diff already. No need to rebuild that.

tests/Highlighter.py
```diff
@@ -0,0 +1,303 @@
...
+            cpt += 1
+            if result[i] == "\n" : cpt += 1
+
+        return True,-1
+
+
+
+    #######################################################################################
+    #######################################################################################
+    #######################################################################################
```
rix0rrr, line 303: This looks pretty but doesn't add any value (except set your code apart from other code, which we agreed is a bad thing)

rix0rrr (changes requested):

tests/Highlighter.py
```diff
@@ -0,0 +1,358 @@
...
+                ind_error -= len(code_list[line_cpt]) +1
+                line_cpt += 1
+
+            print("ERROR in this code :")
+            print("In this code :", code_list[line_cpt] )
+            print("We want      :", expected_list[line_cpt] )
+            print("We have      :", result_list[line_cpt] )
+            print("At           :", " " * ind_error + "^")
+
+        self.assertTrue(valid)
```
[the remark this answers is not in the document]
rix0rrr: In general: less code is better, because *code represents a cost*. The cost needs to be justified by the functionality it offers over readily available things.

By using `self.assertEquals` you could get rid of `self.compare` below.

tests/Highlighter.py
```diff
@@ -0,0 +1,358 @@
...
+    # This function allows to compare 2 syntactic colorations one desired
+    # and the other obtained, and manages some special cases
+
+    # Returns a boolean and a value:
+    # - If the 2 highlights are consistent, the boolean will be true and the value unused
+    # - if there is an error in the highlight, the boolean will be false and
+    #   the value will be the number of the 1st character where there is an inconsistency
+    def compare(self, result, expected):
+        if len(result) != len(expected):
+            raise ValueError("The desired highlight and the obtained highlight do not have the same length !")
```
rix0rrr, line 126: Here's another thing to thing about for fitness of purpose:

If we hit this error case, there will be no useful output for a developer to go on. They will only see the error message you wrote above. They will not see the expected code, the actual code, the highlights, not the lengths, nothing.

That's not setting debuggers up to succeed, right?

tests/Highlighter.py
```diff
@@ -0,0 +1,358 @@
...
+
+            cpt += 1
+
+        return True, -1
+
+
+
+    # get rules from json files
+    def get_rules(self, level, lang="en"):
+        os.chdir(os.path.dirname(__file__) +"/..")
```
rix0rrr, line 148: Bad form to `chdir` in general, and definitely in some random utility function here.

Do you think that if some caller code calls `get_rules()`, that they would expect the working directory to be changed out from under them?

The current working directory is like global state: it's scary, anyone can change it and you can't tell what parts of the program changing it might affect. Better not to mess with it at all.

If you can determine the correct directory at all, you can also just open a file directly from that directory without `chdir`ing to it first.

tests/Highlighter.py
```diff
@@ -0,0 +1,358 @@
...
+                end_of_search = True
+
+        # we color the last characters since the last match with the default coloring
+        for c in range(current_position, len(code)):
+            output.append(TOKEN_CODE[default_token])
+
+        return "".join(output), current_state
+
+    # find next match
+    def find_match(self, code, current_state, current_position):
```
rix0rrr, line 336: Please document your functions a little in the style as described [here](https://peps.python.org/pep-0008/#documentation-strings). This will make it a lot easier to review and also reason about it:

For example:

```py
def find_match(self, code, current_state, current_position):
  """Find the next matching rule in the given code.

   If there are multiple rules that match the code in the given state, returns the
   one that matches earliest in the source string.

   Arguments:
     - code (string): the code we're searching
     - current_state (string): name of the highlight state we are in
     - current_position (int): index to start searching

   Returns:
     (did_match, transition): a tuple of a boolean and a transition dictionary. The 
        transition dictionary contains  {rule, match} fields and should only be inspected
        if did_match is True.
  """
```

If you wrote this, we can reflect and say: why does this even need to return a tuple at all?
Why doesn't it return an *optional* Transition? That's even simpler to work with for callers.

```py
  """
   Returns:
     {rule, match} | None: a transition dictionary if a match was found, None otherwise.
  """
```

tests/Highlighter.py
```diff
@@ -0,0 +1,358 @@
...
+
+                if 'next' in current_rule:
+                    current_state = current_rule['next']
+
+
+                if current_position == len(code):
+                    end_of_search = True
+
+            else:
+                end_of_search = True
```
rix0rrr, line 327: Personally I'm not a fan of very long `if` blocks:

```py
if success_condition:
  # a lot of code
  #
  # 
  # no but really, a lot of lines here
  #
  # yep this is happening
  # 
  # scrolling is definitely required
else:
  # short code to deal with failure case
```

Because they hurt readability (it's hard to find the `else` clause from that `if`). I generally write in an early-exit style, and I recommend you get used to doing the same:

```py
if not success_condition: 
  # deal with error case
  return; break; continue; # whatever is necessary to stop execution

# this is where the long happy path starts
#
# can still be long but it's obvious
```

tests/Highlighter.py
rix0rrr: Also I think the argument order makes most sense to put `code` and `current_position` next to each other, and `current_state` later. Conceptually those 2 belong together.

tests/Highlighter.py
rix0rrr: In this case I might even get rid of the boolean flag altogether and do something like this:

```py
next_transition = self.find_match(code, current_state, current_position)
while next_transition:
  start, end = transition_position(next_transition)
  mark_as_text(current_position, start)
  mark_transition(next_transition)
  current_position = end

  next_transition = self.find_match(code, current_state, current_position)

mark_as_text(current_position, len(code))
```

rix0rrr (changes requested):
Thanks for taking my comments to heart on the tester code.

I notice you haven't changed anything about the test cases themselves. What are you planning to do with those?

I would also enjoy it if you could respond to what I'm saying, so I can tell whether you're eye rolling and silently resenting me while being forced to change your code 🙄😡, or whether you agree with me 🤓. Thanks!

tests/Highlighter.py
rix0rrr: Not trying to force you to change anything, just giving you something to think about.

tests/Highlighter.py
rix0rrr: Not trying to force you to change anything, just giving you something to think about.

thjazi39:
> I would also enjoy it if you could respond to what I'm saying, so I can tell whether you're eye rolling and silently resenting me while being forced to change your code , or whether you agree with me . Thanks!

I don't take the remarks personally in any way, I know I know python well, but I know it's also the first time I'm programming in a professional context, so I know I still have a lot to learn!

Concerning the answers to some of your remarks ( the tests, the use of the "assertTrue" function, some loops ) I didn't close the corresponding code comments, and I brought my justification (which is maybe bad ^^).

rix0rrr:
You might need to click "publish review" at the top right in your editor, or something. I can't see your replies for some reason.

thjazi39:
> You might need to click "publish review" at the top right in your editor, or something. I can't see your replies for some reason.

I think this is because the code comments are associated with the codes, so when the code changes, they are marked as "outdated", and are no longer visible.

However, in the Conversation tab, they are still visible (at least to me)

thjazi39:
> Can you send me a link to a comment you responded to?

tests : https://github.com/Felienne/hedy/pull/2474#discussion_r874701980
method private/public : https://github.com/Felienne/hedy/pull/2474#discussion_r874704355
for loop : https://github.com/Felienne/hedy/pull/2474#discussion_r874708122
check : https://github.com/Felienne/hedy/pull/2474#discussion_r874728164
error message in compare : https://github.com/Felienne/hedy/pull/2474#discussion_r881344616
open file : https://github.com/Felienne/hedy/pull/2474#discussion_r881347000
pretty while : https://github.com/Felienne/hedy/pull/2474#discussion_r881361307

Here are the answers (well for some strange reason, I can't get a permalink from my answers, but these are the suggestions you gave me, and the answer is right below)

rix0rrr:
Thank you. This is what I'm seeing:

image

Please confirm you definitely don't need to click another button to publish those comments.

rix0rrr:
I think you might need to click this first:

image

thjazi39:
I think it's good now (actually, I didn't finish my review, I didn't know that was necessary) Sorry !

tests/Highlighter.py
Felienne: But then we could `in`? Something like: `self.assertIn(result, ["KKK TTTTTTTTT", "KKKKTTTTTTTTT", "KKKTTTTTTTTTT"])` so we test if we get one of the 3 valid answers?

thjazi39: Yes, but that means generating a lot of str :

In this example: 
```
if throw is earthworm points = points + 5 else points = points + throw
KK TTTTT KK TTTTTTTTT TTTTTT K TTTTTT K N KKKK TTTTTT K TTTTTT K TTTTT
```
that would be 6^15 possible strings

The solution could be to pass to assert some function of comparison between characters, but the only way I found to do that is to create a new object.

Felienne: ok, I see that might yield too many options, but then we could wrap our own equality function into the assert? Something like:

```
s = "ask something"

result = remove_space_highlighting(highlight(s), 4). (so we'd get "KKKTTTTTTTTT" for "ask something")

self.assert(result,  "KKKTTTTTTTTT")

def remove_space_highlighting(s, space):
   # removes the highlighting symbol at the location of the space
  return s[:space] + s[space+1:]

```

Hope that makes sense!

tests/test_highlighting/test_highlighting_adventure_en_03.py
```diff
@@ -0,0 +1,276 @@
...
+            "KKKKK TTTTTTTTTTTTTTTT",
+            "animals is 🐿, 🦔, 🦇, 🦉",
+            "TTTTTTT KK T  T  T  T",
+            "animal is ask What do you think it is?",
+            "TTTTTT KK KKK TTTTTTTTTTTTTTTTTTTTTTTT",
+            "add animal to animals",
+            "KKK TTTTTT KK TTTTTTT",
+            "print it was a animals at random",
+            "KKKKK TTTTTTTTTTTTTTTT KK KKKKKK",
+            level="level3",lang='en')
```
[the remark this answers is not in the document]
Felienne: I agree with @rix0rrr that there are many many tests. I appreciate that you want to cover a lot, but it should also still be readable. 

Can we make it fewer tests in this way by limiting the test set to one line tests for concepts in each level (print, ask, echo, forward, turn in level 1 etc.)

Add a few multiline ones. We don't need all those multiline test, if we just have a few that is ok. We could even write a little script to generate multiline ones from simpler ones.

tests/Highlighter.py
```diff
@@ -0,0 +1,303 @@
...
+}
+
+
+class HighlightTester(unittest.TestCase):
+
+    def assert_highlighted_chr(self,code,expected,level,lang="en"):
+        
+        rules  = self._get_rules(level,lang)
+
+        result = self.apply_rules(rules,code)
```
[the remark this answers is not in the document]
Felienne: I don't think we use _ for private methods anywhere else in the codebase so I am totally ok with not doing it here too.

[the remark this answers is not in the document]
thjazi39: Ok, so it's done !

tests/Highlighter.py
thjazi39: We can't just delete the undefined characters either, because my function doesn't allow coloring cases for undefined characters.

In particular, an indefinite character can be anything, except a character highlighted in pink as in the following case :

```
if a is 2 else _ "Done!"
KK T KK N KKKK I SSSSSSS
              | |
              1 2
```

Spaces 1 and 2 can be colored with the style "command" or "string" or "text", because it will not be seen, but they should not be colored with the style "error" (which is the one of the pink highlighting)

tests/Highlighter.py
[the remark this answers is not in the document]
thjazi39: I close this channel because it is redundant with https://github.com/Felienne/hedy/pull/2474#discussion_r874728164

tests/Highlighter.py
Felienne: ah ok but then we can make a function that checks whether these spaces are "not error"?

We just really don't want our own assertion function, we want something (does not matter how you divide it precisely) wrapped inside the regular assert.

tests/test_highlighting/test_highlighting_adventure_en_03.py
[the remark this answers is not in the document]
thjazi39: I tested a new solution, it splits the check in 2 different assert (In and Equal) but it can be enough

tests/Highlighter.py
thjazi39: I tested a new solution, it splits the check in 2 different assert (In and Equal) but it can be enough

tests/Highlighter.py
```diff
@@ -134,67 +134,24 @@ def check(self, code, expected, state_machine):
...
-                pass # if it was not fixed at the beginning, everything is tolerated except the pink highlighting
-            else:
-                # in all other cases, an error is returned indicating the number of the problematic character
-                return False, cpt
-
-            cpt += 1
-
-        return True, -1
+        # test between two coloration
+        self.assertEqual(result ,expected)
```
Felienne, line 154: Ah thanks for using the regular assert here!

Felienne:
Hi @thjazi39!

I think we are getting close to being able to merge this now! It would be nice if you could still:

1) Prune the tests a bit as discussed about
2) Add a few (not all too much!) tests for non-latin languages like Arabic and Chinese. It is nice we have Dutch and English now but for non-latin languages things like regexes sometimes do unexpected things.

tests/test_highlighting/test_highlighting_adventure_en_03.py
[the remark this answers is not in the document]
rix0rrr: > These tests allow to cover all the codes displayed on Hedy. This seemed important to me, because it ensures that, no matter what changes are made to the syntax coloring, the codes displayed will always be valid. There is necessarily some redundancy between them, but it allows to have a base of functionality to cover. 

"Some" redundancy is definitely okay. But For example, it would be fine to have 2-3 tests that cover (different aspects!) of the `print` command. But there are degrees here. You easily have 100s of tests here covering the print statement! That seems a bit much.

Can we at least agree that the following principle is a good one to strive for?

**We are looking for the *minimum* amount of tests that covers 100% of the functionality**

If we agree that's the goal, then now of course the question is how do we find that optimum point? 

* What's the minimum amount of tests? Could we remove one? I'm not sure.
* Are we sure we've covered 100% of the functionality? Might there be a bug we're not covering? I'm not sure.

There are no easy mathematical answers here. We cannot *compute* this number. But we can approximate it based on intuition, and approach it over time as we refine our test suite.

Since it's easy to add tests when their need is demonstrated, but hard to know whether a test can be removed because it's not pulling its weight... I would err on the side of conservatism. 2-3 tests for every feature ("behavior"), and let the rest sort itself out over time as we discover bugs.

However you decide to approach it, my intuition says the number you have now is too high.

I appreciate the thoroughness, and you're no doubt approaching this with a mathematical mindset... but you have to consider that software engineering is "people work" as much as it is computer work. Code is never *done*, it will always be updated, by people, and you have to allow for that future change.

tests/Highlighter.py
```diff
@@ -0,0 +1,303 @@
...
+        self.assertTrue(valid)
+
+
+    def assert_highlighted_chr_multi_line(self,*args,level,lang="en"):
+        if len(args)%2 != 0:
+            raise("*args must be even (alternating lines of code and lines of highlighting)")
+
+        code = []
+        expected = []
+        for i,line in enumerate(args):
```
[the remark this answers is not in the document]
rix0rrr: That is fair enough. For what it's worth, I don't agree, I think the following is perfectly readable: 

```py
code = '\n'.join(line for i, line in enumerate(args) if is_even(i))
expected = '\n'.join(line for i, line in enumerate(args) if not is_even(i))
```

But reasonable people may disagree, and it's not a big deal.

tests/Highlighter.py
rix0rrr: > For example, if we have the code :
>
> ```
> "ask something"
> "KKK TTTTTTTTT" (1)
> "KKKKTTTTTTTTT" (2)
> "KKKTTTTTTTTTT" (3)
> ```
>
> We want this coloration (1) But (2) or (3) are valid too

Well that's a bit odd, isn't it? But also, this is nothing a bit of *canonicalization* wouldn't fix.

Using the following rule:

**Whitespace will never be assigned a color**

We could say the one and only colorization of this string is `"KKK TTTTTTTTT"`, and the other 2 variants are not allowed. This seems reasonable, as a human wouldn't be able to perceive the difference anyway.

The canonicalization can be applied in one of 2 ways:

* Either just before testing (requires both the original code string and the colorized string, to find the whitespace locations to strip out)
* Or already during the highlighter simulation. The `AceHighlighter` code could take into account whether or not the input string contains whitespace and remove the highlight tokens there.

The canonicalization function could look something like this:

```py
# Array of booleans
is_whitespace = [c == ' ' for c in code]

# Do a transform on the colorization, retaining the characters only for non-whitespace
# (whitespace characters are replaced with a space)
canonical_colors = ''.join(c if not is_ws else ' ' for c, is_ws in zip(colorization, is_whitespace))
```

tests/Highlighter.py
rix0rrr: > And about the code display, it would be convenient, but I don't have access in the context of this function.

Well. You *could*, since you wrote this function. You can add arguments to also pass the code in, right? 😊

What I'm saying is this:

* First, think about the **user** this code is for. What task are they trying to perform? What would help them with that task? What should they see?
* Then, with that experience in mind, write the code to pass the right information to the right place.

Looks like you did it the other way around: 

* First, write some code that does some computational task
* Then, see what kind of message you can scramble together from the variables you happen to have access to based on the code written in step 1

A lot of people work in the second style. The problem with that style is it works great *if and only if* everything works out. But when things go wrong, it's usually very hard to figure out *where*, and *why*. 

Plan for things to go wrong, always, all the time, and design your software accordingly.

tests/Highlighter.py
rix0rrr: You need to combine the two things you were doing:

```py
# Calculate the directory without actually switching to it
root_dir = os.path.dirname(__file__) +"/.."

# Calculate the filename with respect to the correct directory
filename = f'{root_dir}/highlighting/highlighting-trad.json'

# Finally, open the file
with open(filename) as file_regex_trad:
```

tests/Highlighter.py
thjazi39: This function does not exist anymore. Since the checks are done in the check method, I have access to the code variable (and since I use the assertEqual method, I don't have to write the error message)

tests/Highlighter.py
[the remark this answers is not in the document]
thjazi39: Ok, now it's done !

tests/Highlighter.py
thjazi39: Done !

Felienne:
Thanks for all the changes @thjazi39!! It is a bit hard for me to keep track of all the changes, could you maybe write a little summary of the things you are still working on/planning to work on?

tests/Highlighter.py
Felienne: Haha thanks @rix0rrrm that is a more extensive explanation of what I suggested above :)

thjazi39: I understand better about channeling, but it only works for spaces.

However, trying to adapt the method, there is another problem concerning uncertain characters. 

Indeed, there are the space characters, which can be colored as "keyword", "number" or "string" it will not matter, because it will not be seen on the space, and that's true.
But there are other cases where you want the coloring to be uncertain, and not on spaces!

For example, in this case :
```
print _ the dishes are done by _
KKKKK I                        I
```
because the code is incomplete, the coloring is uncertain.

I don't think that fixing the coloring in this case is a good idea, and that's also why I've taken this view of uncertain coloring.

If you think it's not a good idea to allow tests to be written like this, and the new way of checking tests is not appropriate, then I could delete the spaces afterwards.

thjazi39:
> Thanks for all the changes @thjazi39!! It is a bit hard for me to keep track of all the changes, could you maybe write a little summary of the things you are still working on/planning to work on?

Okay : 
**Task 1**
I'm currently working on rewriting the unit tests, with a lot of parameterized tests.
(and I will add the tests for non-latin characters)


**Task 2**
There is also the problem concerning the validation of uncertain coloring, which is detailed here : https://github.com/Felienne/hedy/pull/2474#discussion_r884159459


There are actually 2 problems, uncertain coloring, and space coloring.

But if we don't want the uncertain coloring, and we only want a certain tolerance of the coloring for the spaces, I could rewrite the test as follows:
1. assert if space are in TOLERATE_COLORATION
2. remove space, and check if coloration are the same with assertEqual

tests/Highlighter.py
Felienne: > because the code is incomplete, the coloring is uncertain.

Just to clarify, we should have a special coloring for incomplete code (very pink):

image

So, I guess you mean here is the part "the dishes are done by" Yes, we do not know the coloring, but then we can just not have these types of tests? Because it makes no sense to test things we don't have the correct answer to.

I would propose not to test these uncertainties at all, also then just proceed with the spaces as we planned earlier. Just test `print _` and verify it is keyword followed by incomplete.

Felienne:
Thanks for the updates!!

> Okay : **Task 1** I'm currently working on rewriting the unit tests, with a lot of parameterized tests. (and I will add the tests for non-latin characters)

Great!

> if we don't want the uncertain coloring, and we only want a certain tolerance of the coloring for the spaces, I could rewrite the test as follows:
> 
> 1. assert if space are in TOLERATE_COLORATION
> 2. remove space, and check if coloration are the same with assertEqual

Yes, this seems like the way to go to me!

Felienne:
Thanks for the updates @thjazi39. There seem to be a few small errors in the Python code now:

image

thjazi39:
I think it's ok, I rewrote the tests and the check function.

There are still the tests for the programs displayed on the website. (I prefer to keep them for now, because when I change the syntax highlighting, it's possible that some command combinations are not covered by the test settings, and this will allow me to detect these errors before they are visible on the website)

Felienne (approved):

Issue #2007, opened 2022-02-16
indentation needs to be checked for all languages


Felienne:
Related to #1213 

Here we only check for English keywords:
https://github.com/Felienne/hedy/blob/ae64eb74face0531bbe89068f1308806af42701d/hedy.py#L1951

We could at use the translation code that knows the equivalent of these for all langs.

Felienne:
Maybe this is a good issue for you @huang-v?

Felienne:
I imagine something like:

for_keywords = <use the hedy_translation to fetch for in all languages>
return first_keyword in for_keywords <and also add that for the other keywords>

Felienne:
Fantastic!

Felienne:
Yes some language are there but a) not all of them and b) if we add new languages we should also update this (which will be forgotten) It would be nicer if we would grab this from the keyword yaml files. Ow and from the cached version, somewhere in the initialization code, because we don't want to open 22 files on each parse :)

PR #2488, opened 2022-04-10
Adds scalable translation when checking for keywords that need indentation

huang-v:
**Description**

< Adds global variable that stores English and language specific commands which is updated when grammar file is created (once per parse). Creates combined list of keywords that need indentation and their translation in the current language selected and checks to see whether the keyword on that line needs is contained in this list >

< Issue #2007  >

**How to test**

<  Tester.py checks to see that translated code also works but I am unsure if this is enough >

**Checklist**
  

hedy.py
```diff
@@ -1966,7 +1971,15 @@ def find_indent_length(line):
...
-    keywords_requiring_indentation = ['if', 'als', 'si', 'for', 'repeat', 'répète', 'repete', 'herhaal']
+    en_keywords_requiring_indentation = ['if', 'for', 'repeat']
+    keywords_requiring_indentation = []
+    global en_commands
+    global lang_commands
+    for keyword in en_keywords_requiring_indentation:
+        if keyword in en_commands:
+            index = en_commands.index(keyword)
+            keywords_requiring_indentation.append(en_commands[index])
+            keywords_requiring_indentation.append(lang_commands[index])
```
TiBiBa, line 1982: This only works if we can guarantee that the English and other language keywords are always in the same order. Is this the case? It might be nice to do some pre-processing somewhere and create a dictionary which contains these translations so that on runtime we can perform this action more efficiently.

For example:

```
# Create some efficient dict like this (English as key -> other language as value):
{'repeat': 'répète', ...}

keyword_requiring_indentation = en_keywords_requiring_indentation
keyword_requiring_indentation.append([k if look_up_dict.get(k, None) for k in en_keywords_requiring_indentation]
```

hedy.py
```diff
@@ -1854,7 +1856,10 @@ def create_grammar(level, lang="en"):
     # ready? Save to file to ease debugging
     # this could also be done on each merge for performance reasons
     save_total_grammar_file(level, result, lang)
-
+    global lang_commands
+    global en_commands
```
TiBiBa, line 1860: global object are fine, but we then we don't want to fill them again for each user (when the function is called). If we create them once and make sure they are cached this will reduce the memory load a lot. You can take a look at some other parts of the code, but we can for example create one global object like this at the start of the file:

```
COMMANDS = []
for language in keyword_languages.keys():
    COMMANDS[language] = .... #get commands for this language
```

Than we can simply call this list of language commands everywhere without the need of creating it again!

TiBiBa (changes requested):
Hi @huang-v ! Thanks for your contribution, I just left some comment on your code. Let me know if you have any questions or need any help!

hedy.py
TiBiBa: Hi! I just did some improvements in #2493. Once this is merged you can easily include `KEYWORDS` from `hedy_content` to make the look-up easier. This is already a dictionary so you can simply iterate over it with the English keywords. Let me know if anything is unclear!

hedy.py
TiBiBa: This has become redundant as I just implemented something very similar in #2493 that you can use for this problem.

huang-v: How do I access what's inside a "website.yaml_file.YamlFile object"?

TiBiBa: Are you referring to a datatype? If it's a reference to a file you can read the content by using something like `.to_dict()`. But I'm not sure if you need this as the `COMMANDS` dictionary already contains the commands for each language as a dictionary.

huang-v:
@TiBiBa Hi! Feel free to take over this issue

TiBiBa:
Closed, issue will be solved in a new PR.

PR #2493, opened 2022-04-11
[CLEAN-UP] Cache keywords yamls

TiBiBa:

< Provide a one line summary of your changes in the Title of the PR >

**Description**

< Changes in detail > Use present tense without a subject to describe your PR, for example: "Adds translations of levels 1 to 12 to Polish"

**How to test**

**Checklist**
  
### DIFF
--- a/app.py
+++ b/app.py
@@ -95,6 +95,7 @@ def load_adventures_per_level(level):
     loaded_programs = {}
     # If user is logged in, we iterate their programs that belong to the current level. Out of these, we keep the latest created program for both the level mode(no adventure) and for each of the adventures.
     if current_user()['username']:
+        # Todo TB -> This is (extremely) inefficient! We should index the level and only retrieve the current level
         user_programs = DATABASE.programs_for_user(current_user()['username'])
         for program in user_programs:
             if program['level'] != level:
@@ -106,9 +107,7 @@ def load_adventures_per_level(level):
                 loaded_programs[program_key] = program
 
     all_adventures = []
-
-    adventure_object = ADVENTURES[g.lang]
-    adventures = adventure_object.get_adventures_per_level(level)
+    adventures = ADVENTURES[g.lang].get_adventures()
 
     # Order the adventures dict by ADVENTURE_ORDER to ensure this is always the same (independent of YAML structure)
     sorted_adventures = {}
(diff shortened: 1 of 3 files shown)


Felienne:
> This might be frustrating when working locally on the code, some we might want to add some `if dev_mode -> don't cache` action here.

First I was like "nah, I will wait" but now that I have waited 3 times already, I think we should open an issue for this haha. Maybe by default when debugging we want to turn the caching off when debugging?

Felienne:
When loaded, it is VERY quick, so that is supernice. Adventures load perfectly, but the level-defaults do not show for me **when logged in**:

image

Very weird! When not logged in, it works fine:

image

Tried for various languages, all have the same behaviour, f.e. French:

image
 
(MacOS, chrome)

Felienne:
> When loaded, it is VERY quick, so that is supernice. Adventures load perfectly, but the level-defaults do not show for me **when logged in**:
> 
> image
> 
> Very weird! When not logged in, it works fine:
> 
> image
> 
> Tried for various languages, all have the same behaviour, f.e. French:
> 
> image
> 
> (MacOS, chrome)

Ow, I just noticed this is an issue in main too, not only in this PR!

Felienne:
> > When loaded, it is VERY quick, so that is supernice. Adventures load perfectly, but the level-defaults do not show for me **when logged in**:
> > image
> > Very weird! When not logged in, it works fine:
> > image
> > Tried for various languages, all have the same behaviour, f.e. French:
> > image
> > (MacOS, chrome)
> 
> Ow, I just noticed this is an issue in main too, not only in this PR, but in a weird way! Locally on main I have the issue, but it does not exist in hedy-alpha. Will try another computer later.

TiBiBa:
Weird! Will take a look to try and fix this bug. Will also add a "don't cache when debug" check because it is indeed pretty annoying to wait every time the server re-starts.

TiBiBa:
Set to draft until the following two issues are fixed:

1. Level-default are shown when logged in
2. Adventures should not be cached when in debug mode

TiBiBa:
Server didn't start on time (once again), so we might have to extend this time even further.

TiBiBa:
> When loaded, it is VERY quick, so that is supernice. Adventures load perfectly, but the level-defaults do not show for me **when logged in**:
> 
> image
> 
> Very weird! When not logged in, it works fine:
> 
> image
> 
> Tried for various languages, all have the same behaviour, f.e. French:
> 
> image
> 
> (MacOS, chrome)

Hi @Felienne ! I'm unable to re-produce this bug, being either logged in or not and trying several languages. Not sure what the cause of this issue could be. On Alpha is all seems to work fine for me as well.

TiBiBa:
Keep this a draft before #2517 is merged as we also want to cache quizzes at this point.

Felienne (approved):

PR #2496, opened 2022-04-11
[SYNTAX HIGHLIGHTER] Small error corrections

thjazi39:
Small error corrections that I noticed while testing Hedy's syntax highlighting (I don't know if the test system will be accepted, nor when, that's why I'm making the change here)

- to escape a usable symbol in a regex, you need 2 backslashes

- For `at random` I was using different groups, but the random was not in a group. It worked with Ace (I don't know how it does it, but normally it shouldn't work) So, I think it's better to fix this error before an eventual update of Ace adds problems

- swap `\` to `'` for regex


thjazi39:
the PR change, for example this :
image
to this
image
(level 3)

PR #2499, opened 2022-04-11
[UI IMPROVEMENT] Implements ace editor on programs page

### LINKED ISSUE #2495: [UI idea] Use ace editor on /programs overview
**Idea incl level**
Currently we use an empty box without styling to show the program of the user. It would be nice if we replace this with the ace editor, as already used in all other spots.

TiBiBa:
**Description**
In this PR we implement the ace editor on the programs and clean-up the scripts there as well. For some reason the client side messages where not included on this page, creating a possible bug where no feedback would be returned to the user on an unexpected error. But most importantly, we now support syntax highlighting on the programs page.

**Fixes**
This PR fixes #2495.

**How to test**
Navigate to the `/programs` page and notice the new ace editors for each stored program. We limit the program shown to 4 lines, the rest is already discarded in the back-end.


app.py
```diff
@@ -637,7 +637,7 @@ def programs_page(user):
     for item in result:
         date = get_user_formatted_age(now, item['date'])
         programs.append(
-            {'id': item['id'], 'code': item['code'], 'date': date, 'level': item['level'], 'name': item['name'],
+            {'id': item['id'], 'code': "\n".join(item['code'].split("\n")[:4]), 'date': date, 'level': item['level'], 'name': item['name'],
```
Felienne, line 640: This is a bit of stringy magic to me, not opposed to it but shall we add a comment to explain?

Felienne (changes requested):
Add comment explaining the [:4]

TiBiBa:
Fixed!

app.py
```diff
@@ -636,10 +636,20 @@ def programs_page(user):
     now = timems()
     for item in result:
         date = get_user_formatted_age(now, item['date'])
+        # We split the code on the newlines, take the last 4 and merge these with newlines again
```
Felienne, line 639:
```suggestion
```

Felienne: Guess it is enough to just say we only want to show the 4st 4 lines

Felienne (approved):
Ah looks nice!

image

PR #2500, opened 2022-04-11
[PERFORMANCE] Implements user-based saved program level filtering

### LINKED ISSUE #2494: [IMPROVEMENT] Index user programs to improve retrieval on level call
**Description**
Currently we retrieve all the user programs when they visit a level. We then filter on the once corresponding with the current level and send these to the front-end. This doesn't make any sense and especially users with a large collection of programs will increase our server and memory load. We should implement an index key on the `level` attribute in Dynamo to improve this implementation.

TiBiBa:
**Description**
In this PR we implement user programs filtering on level by implementation an additional index key in Dynamo. This enables us to improve performance and reduce server load. As we currently retrieve all(!) user programs on each code-page call we have a lot of redundant traffic. We only need the ones for the current level. We also fix a small issue where the programs stored without an adventure wouldn't be loaded correctly when visiting the level, only after switching tabs back and forth.

**Before merging: implement the index key on the actual database!**

**Fixes**
This PR fixes #2494.

**How to test**
This is only a back-end task. Verify that the program retrieval is still working and programs are correctly loaded in their corresponding adventures when visiting the level corresponding to your saved programs.


app.py
```diff
@@ -129,6 +128,7 @@ def load_adventures_per_level(lang, level):
             continue
         # end adventure is the quiz
         # if quizzes are not enabled, do not load it
+        # Todo TB -> Is this still relevant? Teachers can simply "disable" adventures in customizations!
```
Felienne, line 131: I don't think we still need this indeed! And (related) should we just call the tab "quiz" instead of end? No sure how that came to be but it is a bit weird.

Felienne (approved):
Apart from todo issue, this seems to work fine!

TiBiBa:
Alpha is broken because we are missing the index key, will fix this on the Alpha database to verify if it is working.

Felienne:
> Alpha is broken because we are missing the index key, will fix this on the Alpha database to verify if it is working.

Ah when I checked alpha it was behind! Will not deploy until I know we have fixed this!

TiBiBa:
> > Alpha is broken because we are missing the index key, will fix this on the Alpha database to verify if it is working.
> 
> Ah when I checked alpha it was behind! Will not deploy until I know we have fixed this!

I can't seem to get it working. For some reason Dynamo doesn't like two index keys with the date as a sort key. Will try to figure it out tomorrow. We do need the date as we always want to load the users most recent program for a specific adventure/level combination.

Felienne:
> I can't seem to get it working. For some reason Dynamo doesn't like two index keys with the date as a sort key. Will try to figure it out tomorrow. We do need the date as we always want to load the users most recent program for a specific adventure/level combination.

No rush, of course you can look at this tomorrow!

PR #2507, opened 2022-04-12
#2392 fix hard coded text french level 4

### LINKED ISSUE #2392: [CONTENT] Hard-coded text in French level 4
the first words should be `print` and `ask` but `impreso` is not the translation of ask!
Moreover, this is not level 5, but 4!

image

erinrshi:
**Description**

Fixes hard coded "print", "ask" and "4" in yaml file

**Fixes #2392**

**How to test**
< go to: http://0.0.0.0:8080/hedy/4, code should be fixed>


TiBiBa (changes requested):
Thanks for this fix! 

Can you use the corresponding placeholders so the keywords can be parsed dynamically. So like this: `{print}` instead of `print`.

content/adventures/fr.yaml
```diff
@@ -58,7 +58,7 @@ adventures:
                 start_code: "{print} 'Bonjour le monde'"
                 story_text: |4
 
-                    `Print` et `impreso` changent au niveau 5.
+                    `print` et `ask` changent au niveau 4.
```
Felienne, line 61:
```suggestion
                    `{print}` et `{ask}` changent au niveau 4.
```

Felienne (approved):

PR #2479, opened 2022-04-08
[CLEAN-UP] Replaces mail YAMLs with gettext

TiBiBa:
**Description**
In this PR we move the mail translations from YAML files to the gettext structure, simplifying our codebase. This is mostly a clean-up task as we simply move the translations. But as we re-write some back-end structure as well some testing is needed that everything still works.

**Fixes**
This PR fixes #2476.

**How to test**
This cannot be tested locally. When the tests pass we should merge and verify that it's all working with the templates by performing the following actions in multiple languages:

- Create an account (receive a verification mail -> verify that link is working)
- Reset password
- Recover password
- Set account as teacher (only the admin can do this @Felienne) -> also receive mail

PR #2508, opened 2022-04-12
[REVERSE FIX] Removes double Dynamo index key dependency

TiBiBa:
**Description**
In this PR we reverse a fix made in #2479 that depends on two index keys and a sort key to automatically retrieve all user programs for a specific level. This works fine with the local database, but does seem impossible to work around with the actual Dynamo db. Therefore we reverse this and use a list comprehension to filter the correct programs.

**Fixes**
No related issue.

**How to test**
Locally everything should work exactly the same. Verify that on alpha the `/hedy` page is working again after merging this PR.


Felienne (approved):

Felienne:
Works locally, let's check on alpha.

Issue #2457, opened 2022-04-06
fix for level 7 undone by weblate merge


OnnoEbbens:
Recently I submitted a pull request to fix the quiz at level 7 (https://github.com/Felienne/hedy/pull/2379). However some important changes were reversed by a pull request from Weblate (https://github.com/Felienne/hedy/pull/2402). 

I can easily make a new pull request to fix this but maybe it is good to know what is happening first to prevent this from happening again. Is it only a wrongly solved merge conflict or is there some magic in Weblate that causes this?

Also it is good to know that if the quiz answers contain both "option_text" and "code" nothing is shown. Maybe a test can be added to check the quiz yml files for this.

Felienne:
Ah sorry @OnnoEbbens, we are still getting used to Weblate. If it is not to much work for you, could you send the PR again?

Tagging @Mark-Giesen here, he is our Weblate expert and he might have ideas on how to prevent this in the future.

Mark-Giesen:
> Ah sorry @OnnoEbbens, we are still getting used to Weblate. If it is not to much work for you, could you send the PR again?

Unfortunatly that will not work, just like the first time. I've passed this on to @MarleenGilsing on april 2th, she was going to look into this. 
The idea is that we provide English and people translate this into other languages. Weblate keeps the file structures in sync. en.yaml has feedback and code tags in this position , no option_text. Every time you delete the code lines, Weblate will put them back with the orinal English text for translation. 
Just like deleting something that remains in the en.yaml won't work, adding something in a translation that is NOT in the English will likely be deleted (I'm not yet sure why some tags remain)
Usually the proces is we create Englisch and Dutch, so this will not be a problem often, I presume.

Felienne:
> > Ah sorry @OnnoEbbens, we are still getting used to Weblate. If it is not to much work for you, could you send the PR again?
> 
> Unfortunatly that will not work, just like the first time. I've passed this on to @MarleenGilsing on april 2th, she was going to look into this. The idea is that we provide English and people translate this into other languages. Weblate keeps the file structures in sync. en.yaml has feedback and code tags in this position , no option_text. Every time you delete the code lines, Weblate will put them back with the orinal English text for translation. Just like deleting something that remains in the en.yaml won't work, adding something in a translation that is NOT in the English will likely be deleted (I'm not yet sure why some tags remain) Usually the proces is we create Englisch and Dutch, so this will not be a problem often, I presume.

[...]

PR #2510, opened 2022-04-12
fix for #2457

OnnoEbbens:
See #2457


Felienne (approved):
Thanks for the extra work @OnnoEbbens!

PR #2517, opened 2022-04-13
[CLEAN-UP] Removes quiz from adventures

### LINKED ISSUE #2516: [UI idea] Separate quiz from adventures, store default string in gettext
**Idea incl level**
Currently we store the quiz within the adventures. However, this doesn't make any sense as the adventure YAML doesn't have any additional info for the quiz. The string displayed with "you're at the end of the level, take the quiz" is identical for each level and can easily be stored in gettext. Extracting the quiz has the additional advantage that we can easily add other type of adventures to the tabs without having to worry how to fix the quiz tab at the end. By dynamically setting this on the front-end we can als set the quiz at the end.

TiBiBa:
**Description**
In this PR we remove the quiz as a dedicated adventure. Instead we verify if there is quiz data for the specific level on each tab rendering and show the tab (or not) based on this. This has an advantage because the quiz does not belong conceptually in the adventures file. This way we can more easily force it to be the last tab, independent of possible other content. The YAMLs are cleaned up for en, nl and en-devel. @Mark-Giesen Are the rest automatically removed through Weblate because it's removed from the English one?

**Todo (can also be done in a different PR**
After merging this teachers are no longer able to disable the quiz, as it is no longer an adventure. I think it would be best to add this in the section "additional settings" such as "forced dev mode" and "hide cheatsheet".

**Fixes**
This PR fixes #2516.

**How to test**
This is only a clean-up task. Verify that the quiz still works fine on every supported level and language.


Mark-Giesen:
@TiBiBa, Weblate does clean up other languages if it's removed from en.yaml (or en_devel.yaml at this moment). I believe I've seen some instances where it didn't work, but I'm not sure whether this was due to wrong settings at the time, of that it doesn't work with lists for instance. 
In the po files all translations are labeled fuzzy, are there any real text changes, or only context? This fyzzy keyword marks alle these strings in all languages as "need editing". This is wonderfull if the text is realy changed, but maybe not with only the context.

TiBiBa:
> @TiBiBa, Weblate does clean up other languages if it's removed from en.yaml (or en_devel.yaml at this moment). I believe I've seen some instances where it didn't work, but I'm not sure whether this was due to wrong settings at the time, of that it doesn't work with lists for instance.
> In the po files all translations are labeled fuzzy, are there any real text changes, or only context? This fyzzy keyword marks alle these strings in all languages as "need editing". This is wonderfull if the text is realy changed, but maybe not with only the context.

If I understand correctly gettext automatically labels them as `fuzzy` when there is no translation yet. The current translation is a guess based on other keys that are already existing. So all keywords with the fuzzy marker should indeed still be translated.

TiBiBa:
**Important**
In this PR we cache all quizzes on server start. This is fine for deploying but can be annoying when developing as the server startup time takes a bit longer. We can leave it like this in this PR, but should add the "no cache because debug" settings in #2493 similar to adventures and commands.

PR #2527, opened 2022-04-13
Change highlight system

thjazi39:
Currently, syntax highlighting works like this:

image
The core of the syntax highlighting was in typescript (green file) and this made neither modification nor testing obvious.

Moreover, it was necessary to generate the translations by hand, from a script, in order to have `syntaxLang-<LANG>.ts` files from `syntaxLang-template.ts`, and then one had to modify the typescript to include the new language

This PR proposes this:
image

This makes  (propose by @rix0rrr ) coloring more complex, but it has these advantages:
- This allows you to use the `highlighting/highlightingRules/highlighting-<LANG>.json` files to test syntax highlighting, especially with all languages (dark green file).
- The generation of intermediary files containing the regexes allows to debug more easily, because we see directly which regexes will be used by Ace
- Python is more flexible and better known than Typescript, and now the core of the coloring is in the `highlighting/rules_automaton.py` and `highlighting/list_keywords.py` (light green files)
- If we add a new language, by adding the file in content, then the coloring will be automatically added (except in TypeScript)


thjazi39:
Some remarks :
- The blue part on the schema, that is to say what concerns the tests, will come in another PR (I thought it would be better to separate)
- I don't know if the json files containing the regexes should be followed or not
- It might be interesting to add the execution of the `static/js/syntaxModesRules.py` file in the `build-tools/heroku/generate-typescript` file in order not to make the tests longer
- I coded the `syntaxModesRules` file as best as I could, but there may be some conventions or places that need to be changed

Felienne:
Ah thanks for this and for the extensive explanation @thjazi39!!

Felienne:
> * The blue part on the schema, that is to say what concerns the tests, will come in another PR (I thought it would be better to separate)

Yes better to separate!

> * I don't know if the json files containing the regexes should be followed or not

I do not understand this sentence, what does "followed" mean here?

> * It might be interesting to add the execution of the `static/js/syntaxModesRules.py` file in the `build-tools/heroku/generate-typescript` file in order not to make the tests longer

Agreed! 

> * I coded the `syntaxModesRules.py` file as best as I could, but there may be some conventions or places that need to be changed

That's ok

> * Importing the json into typescript was difficult for me, as I don't know typescript. So there are probably some simplifications possible in the `syntaxModesRules.ts` file (a modification that would be nice, but that I don't know how to do, would be to remove the references to the languages, so that the addition of a new language is more transparent)

That's ok too, once it all works we can simplify

> * I placed the `syntaxModesRules.py` file in `static/js` but maybe that's not the best (now that typescript is separated from the syntax highlighting core, maybe we need a special folder for syntax highlighting?)

yes! it is not a good idea to hide python files in a js folder. Can you make that update in this PR?

Felienne:
> > * The blue part on the schema, that is to say what concerns the tests, will come in another PR (I thought it would be better to separate)
> 
> Yes better to separate! Do you plan to finish this first or the tests PR first?
> 
> > * I don't know if the json files containing the regexes should be followed or not
> 
> I do not understand this sentence, what does "followed" mean here?
> 
> > * It might be interesting to add the execution of the `static/js/syntaxModesRules.py` file in the `build-tools/heroku/generate-typescript` file in order not to make the tests longer
> 
> Agreed!
> 
> > * I coded the `syntaxModesRules.py` file as best as I could, but there may be some conventions or places that need to be changed
> 
> That's ok
> 
> > * Importing the json into typescript was difficult for me, as I don't know typescript. So there are probably some simplifications possible in the `syntaxModesRules.ts` file (a modification that would be nice, but that I don't know how to do, would be to remove the references to the languages, so that the addition of a new language is more transparent)
> 
> That's ok too, once it all works we can simplify
> 
> > * I placed the `syntaxModesRules.py` file in `static/js` but maybe that's not the best (now that typescript is separated from the syntax highlighting core, maybe we need a special folder for syntax highlighting?)
> 
> yes! it is not a good idea to hide python files in a js folder. Can you make that update in this PR?

thjazi39:
> > * I don't know if the json files containing the regexes should be followed or not
> 
> I do not understand this sentence, what does "followed" mean here?

I meant followed by git
Indeed, if these files are followed, then at each modification, all the files containing the highlighting regex will be modified, which makes modifications of more than 40 000 lines

and since we don't have to modify the json files directly, but rather modify the `syntaxModesRules.py` file, we could just regenerate the files when we need them

thjazi39:
> > * I placed the `syntaxModesRules.py` file in `static/js` but maybe that's not the best (now that typescript is separated from the syntax highlighting core, maybe we need a special folder for syntax highlighting?)
> 
> yes! it is not a good idea to hide python files in a js folder. Can you make that update in this PR?

Okay, and is there a particular place where I should put the file? in the root? in a Highlighting folder?
And concerning the Json files contained in `keywordTranslation` and in `syntax`, what are the best places for these files

thjazi39:
> > > * I placed the `syntaxModesRules.py` file in `static/js` but maybe that's not the best (now that typescript is separated from the syntax highlighting core, maybe we need a special folder for syntax highlighting?)
> > 
> > 
> > yes! it is not a good idea to hide python files in a js folder. Can you make that update in this PR?
> 
> Okay, and is there a particular place where I should put the file? in the root? in a Highlighting folder? And concerning the Json files contained in `keywordTranslation` and in `syntax`, what are the best places for these files

I left the Json files in their place, and moved the `static/js/syntaxModesRules.py` file to `build-tools/heroku/generate-rules-highlighting.py`

rix0rrr:
image

rix0rrr:
2022-04-21 at 19 19

build-tools/heroku/generate-rules-highlighting.py
```diff
@@ -0,0 +1,452 @@
...
+
+# In the first levels, the strings are not yet well defined,
+# so we have to color them with respect to what is around,
+# so we use particular functions
+
+def rule_level1(keywordLang):
+    return [{
+        'regex': START_LINE + "(" + keywordLang["ASK"] + ")(.*)$",
+        'token': ['text','keyword','text'],
+        'next': 'start',
```
rix0rrr, line 21: What we had in the original TypeScript is a more maintainable abstraction over these raw rules.

What you have now is technically correct, but not as easy to change anymore as what we had. It's not easy to see what the differences between levels are anymore, for example.

Can you introduce a form of abstraction that improves the maintainability here?

(Put another way: yes there is JSON and regexes involved. I would prefer to see and edit as little of either as possible!)

rix0rrr: This file is also pretty large. It would be nice to split out the *mechanism* that allows us to write these rules and renders them to JSON, from the *actual rules* themselves.

The rules themselves probably also should not live in a directory called `build-tools`. They're part of Hedy proper, not just "some script that moves data from A to B".

So I would expect all of this code to live somewhere more prominent, and have the scripts here only invoke files from the other places as needed.

build-tools/heroku/generate-rules-highlighting.py
```diff
@@ -0,0 +1,452 @@
...
+#   In particular, even if they are between 2 words, the syntax highlighting will select them
+# 
+# - SP_K
+#   This category of keywords allows you to have keywords that are not preceded
+#   by another word, but that can be followed immediately by another word. (see the PR #2413)
+# 
+# - K_SP
+#   This category of keywords allows you to have keywords that can be preceded immediately
+#   by another word, but that are not followed by another word.
+KEYWORDS = {
```
rix0rrr, line 292: I see you adding abstraction and convience here. Nice. I don't understand why parts of the rules seem to be autogenerated and parts of it seem to be manually put in?

rix0rrr: Lots of repetition in here as well. That's generally not great for maintainability.

static/js/app.ts
```diff
@@ -1,7 +1,7 @@
 // It's important that this file gets loaded first
-import './syntaxLang-en';
-import './syntaxLang-es';
-import './syntaxLang-nl';
+/*import './syntaxLang-en';*/
```
rix0rrr, line 2: We have git to track history. Just remove these lines, they're easy enough to get back if we change our mind.

static/js/index.ts
```diff
@@ -2,9 +2,9 @@ export * from './modal';
 export * from './app';
 export * from './auth';
 export * from './statistics';
-import './syntaxLang-en';
-import './syntaxLang-nl';
-import './syntaxLang-es';
+/*import './syntaxLang-en';*/
```
rix0rrr, line 5: Same here.

static/js/syntaxModesRules.ts
```diff
@@ -16,845 +17,55 @@ window.State = {};
...
-  _STEP: string;
-  _ELIF: string;
-  _INPUT: string;
-  _OR: string;
-  _WHILE: string;
-  _LENGTH: string;
-};
-
+var CHOICE;
 switch(window.State.keyword_language){
```
rix0rrr, line 21: This switch could turn into a lookup map, couldn't it?

static/js/syntaxModesRules.ts
```diff
@@ -16,845 +17,55 @@ window.State = {};
...
-  return list_rules;
 }
 
 
-
-
-  
-
-
-
```
rix0rrr, line 849: Nice cleanup!

rix0rrr (commented):
Looks good!

I dinged you a little for a lot of repetition in the rule definitions. I noticed at the end that's a little unfair, as you seemed to have basically straight up translated the front-end code into Python.

Still, we *used* to have something more readable (imo), and I'd like to capture that back.

Now, not saying we have to do that in this PR. In fact, it might be beneficial to do the following:

* Commit this PR as is, confirm that the behavior is as intended.
* Then start refactoring the syntax highlighting rules: if all is well, we can verify that the refactoring is behavior-preserving by confirming that the JSON files don't change!

It might also be nice to make the language list a bit more zero-config. Can we detect new languages and translations automatically, and automatically include them? Otherwise, adding new languages will turn into a mess of copy/pasta over time (already heading that way a little).

In general, I'd recommend you try to reduce repetition as much as possible: it will only benefit maintainbility.

Of course, it doesn't all have to happen in the same PR. If you are confident in these changes, then I'm totally fine merging this right now, and doing additional improvements in future PRs. Let me know how you want to proceed.

build-tools/heroku/generate-rules-highlighting.py
Felienne: Yes agreed! We talked about this yesterday and I would prefer all code related to highlighting to live in a folder together.

Felienne (changes requested):
See inline comments above for fixes needed for this to be merged.

static/js/app.ts
thjazi39: Done !

static/js/index.ts
thjazi39: Done !

static/js/syntaxModesRules.ts
thjazi39: Thanks !

build-tools/heroku/generate-rules-highlighting.py
Felienne: ok (maybe make an issues for this assigned to yourself, just so we don't forget?)

build-tools/heroku/generate-rules-highlighting.py
thjazi39: It's done ! I created a folder with all the syntax highlighting stuff and split the python file (and add a readme)

build-tools/heroku/generate-rules-highlighting.py
thjazi39: This exit already exists! (I had the wrong link)

thjazi39:
> 2022-04-21 at 19 19

It's done ! the python file generating the syntax highlighting rules will directly get the translations in the `.yaml` files and this is done in such a way that adding a language doesn't require any modification (except in the Typescript file)

Felienne:
Hi @thjazi39 

I just tried this (Chrome, MacOS, keyword lang Dutch) and I seem to be getting an error:

image

image

Felienne:
> Hi @thjazi39
> 
> I just tried this (Chrome, MacOS, keyword lang Dutch) and I seem to be getting an error:
> 
> image
> 
> image

Ah I see the issue, the argument of JSON.parse should be a JSON string not a filename, so there should be a file open in between.

Felienne:
Hi @thjazi39! This looks pretty good now, can it be merged or are you planning more changes?

thjazi39:
Yes, the PR is ready! It will eventually be necessary to open an issue so that the addition of a new language does not require the modification of typescript, but this is not a priority!

PR #2538, opened 2022-04-14
[PERFORMANCE] Caches quizzes on server start

### LINKED ISSUE #2518: [PERFORMANCE] Cache all quiz YAMLs on server start
**Description**
Currently we manually load the quiz yamls at each tab rendering (as of #2517) and on each quiz loading. We should create global objects such as for the adventures and commands to make sure these are cached on the server. Reducing the memory load and loading times as well.

TiBiBa:
**Description**
In this PR we improve the quiz implementation for dynamic keywords parsing. We follow the new structure implemented for the adventures and commands in #2493. This means that all language/keywords combinations are automatically created and cached on server start. However, when we are developing this caching is prevented to make sure the server starts (and restarts) quickly. This implementation has the advantage that after a longer server start-up the performance of the website will be much quicker and the server and memory load will be lower.

**The keyword language is currently fixed to English until the questions are fixed for dynamic keywords.**

**Fixes**
This PR fixes #2518.

**How to test**
This is just a performance task. Verify that the server is start much faster locally in comparison with the current main branch (as we still always cache on start-up there). Also verify that the quiz is working correctly for all languages.


TiBiBa:
@Felienne this one is ready for review and will hopefully further reduce our server load!

Felienne:
Works, but I did find https://github.com/Felienne/hedy/issues/2583 in this PR (does not seem *caused* by this PR though)

Felienne (approved):

PR #2540, opened 2022-04-14
Tic tac toe adventure for level 13, 14, 15, 17 EN

MarleenGilsing:
I proudly present: THE TIC TAC TOE ADVENTURE
I hope it works since I couldn't test it because I couldn't get my app.py to run the local host. So fingers crossed.


TiBiBa:
This looks awesome, will take a look tomorrow!

Mark-Giesen:
Fixed one type and added this adventure to a Dictionairy for ordering, otherwise it would not show up.
Level 17 is not working due to an error I'm not able to explain, see below.
image
Other levels have no problem with this line, so it's something in the parser at this level?

Some remarks regarding the code: In the levels without while the play might stop prematurely. The loop is from 1 to 9, if you enter a number twice, this results in a "allready taken" message, but this does take one cycle of the game, so a nineth number can never be entered anymore. This is a good reason to move to while, but maybe it should be mentioned somewhere?
If nothing is entered, this results in the same "allready taken" message, this could be smarter maybe?

Mark-Giesen (changes requested):
level 17 doesn't work yet, see comment.

Felienne:
Hi @MarleenGilsing!

I fixed the small issues (a few forgotten colons) but the game ends after one round because... spot_1 is equal to spot_2 and spot_3 because they are all '.'. So we need an extra check for the lines not to be equal to dots (I can show you Thursday if this does not make sense!)

Mark-Giesen:
> Some remarks regarding the code: In the levels without while the play might stop prematurely. The loop is from 1 to 9, if you enter a number twice, this results in a "allready taken" message, but this does take one cycle of the game, so a nineth number can never be entered anymore. This is a good reason to move to while, but maybe it should be mentioned somewhere?
> If nothing is entered, this results in the same "allready taken" message, this could be smarter maybe?

Anything going to be done on these issues, or are these by design?

MarleenGilsing:
@Mark-Giesen, thanks for your remarks. 
I think the code is already quite complex for the students and fixing these problems (which would be caused by playing the game wrongly) would make the code even longer and more complex. So I think it might be better to just leave it.

Mark-Giesen:
> @Mark-Giesen, thanks for your remarks. I think the code is already quite complex for the students and fixing these problems (which would be caused by playing the game wrongly) would make the code even longer and more complex. So I think it might be better to just leave it.

OK, I think I agree. Maybe explain with comments? We'll leave it now and see how it goes. It to nice te keep for ourselves any longer :-)

Mark-Giesen (approved):

PR #2550, opened 2022-04-19
Use inheritance in test_level.py files #2436

### LINKED ISSUE #2436: [IMPROVEMENT] Refactor tests

boryanagoncharenko:
**Use inheritance in test_level.py files**

**Description**

A controversial experiment which aims at improving the traceability of tests among levels through using inheritance of test level classes.
- Make every level test class inherit the previous level test class
- Add decorator which dynamically removes the test methods defined in all base classes

**Fixes #2436**

**How to test**

Often, the transpiler tests in Hedy define full programs but only aim at testing a part of the program. For example, testing the equality condition requires writing an if-statement. When the syntax of the if-statement changes throughout levels, all tests that aim at checking the equality condition have to be redefined accordingly. The issue is that currently there is no mechanism to track how a single equality condition test scenario changes across levels (and even if it is present at all levels). One suggestion to improve traceability is to rely on the built-in functionality of IDEs to show method overloads and this PR explores this idea.

The way to test this PR is to weigh the pros and cons of this approach. Here are a couple of points for reflection:
- Using reflection to dynamically remove base test methods is a dangerous endeavor and it literally goes again the very definition of inheritance. It is useful in this particular case, but it could be the source of confusion in the future.
- Python does not need an indication that the test in the current level is overriding a base method. This means that as long as test cases have the same name across levels, the solution will work out-of-the-box. Additionally, changing a test case in one level will not spill over to other levels because of syntax.
 
**Checklist**
  

Felienne:
> For example, testing the equality condition requires writing an if-statement. When the syntax of the if-statement changes throughout levels, all tests that aim at checking the equality condition have to be redefined accordingly.

I totally understand this problem and agree this could be solved better, but I don't fully understand how this proposal addresses this issue. Maybe you can make that a bit more clear, how this concrete issue is addressed by this approach?

I must admit I do not fully follow all your Python magic, so maybe it is very clear, but not to me :) 

Also, might their be alternative ways to address the same problem? Maybe we can define, at each level, a canonical form of statements, so 

`if {cond} {statement} else {statement_2}` in early levels and 

```
if {cond}:
   {statement}
else:
  {statement_2}
```

in later levels, and then grab the canonical form each time in testing the if, so we can keep the test the same across levels?

PR #2551, opened 2022-04-19
[BUG FIX] Adds try catch to translation to prevent /explore from crashing

TiBiBa:
**Description**
For some reason the translate keywords functionality in `hedy_translation` does seem to crash on some programs, making the explore page unaccessible. In this PR we fix this by surrounding the function with a big ugly try/catch until we figure out the actual issue. This should enable us to update production again as it is already too long behind.

**Fixes**
No related issue.

**How to test**
Deploy to Alpha, verify that the `/explore` page is working again.


Felienne (approved):

Felienne:
Looks like alpha is back in order:

image

Will deploy tonight!

PR #2553, opened 2022-04-19
[CHORE] Clean-up runhedy code

### LINKED ISSUE #2552: [CLEAN-UP] Re-write or remove the run_hedy.py file as it seems redundant
**Description**
We still have a relic file called `runhedy.py` that can be used to run Hedy code locally. However, it is still limited to level 8 and does not seem up to date with our latest codebase. We have to decide to either update or remove the file.

TiBiBa:
**Description**
In this PR we clean-up the runhedy code file that is used when running Hedy locally. We change the fixed level value of `8` to the `hedy.HEDY_MAX_LEVEL` value making sure that this is dynamically based on the actual max level.

**Fixes**
This PR fixes #2552

**How to test**
Verify that Hedy runs locally for all currently existing levels (1 - 18)

### DIFF
--- a/runhedy.py
+++ b/runhedy.py
@@ -25,9 +25,9 @@ def main():
             if args[1] == "--level" or args[1] == "-l":
                 try:
                     level = int(args[2])
-                    if level > 8:
-                        print("Level has been set to 8, because the value specified was to high")
-                        level = 8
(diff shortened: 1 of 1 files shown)


Felienne:
Thanks @TiBiBa, but I think we should remove the whole runhedy, it was a mistake to merge that whole file in retrospect.

PR #2557, opened 2022-04-20
Improves program time-out handling

### LINKED ISSUE #2554: "Program takes too long" error in early levels when using turtle
When working with the turtles in early levels, sometimes users want to make complicated drawings. Because the loops have not been introduced, they are forced to write multiple lines of code containing the same information. This can lead to programs with over 50 lines of code. This is not a problem in and of itself, but when the code gets executed Hedy sometimes thinks the code is taking too long and will exit with an error before the full picture is drawn. This can confuse and upset the user, as they technically did nothing wrong except make a too complicated program for the early levels to handle. 

One way to fix this would be to remove the too long error in early levels, as there is no way to get an infinite loop yet. Another way would be to extend the time before the error occurs. One user commented that the turtle moves at different speeds every run of the program, so sometimes the shape would complete and sometimes it would exit prematurely.

TiBiBa:
**Description**
In this PR we improve the program time-out by making it more dependent on the current level. As users are unable to create very large program from levels 1 till 6 as we don't support loops as the line limit is `100` lines we should take this into account with the time-out. In this PR we remove the time-out for levels 1 till 6 and alter the values for the higher levels.

**Fixes**
This PR fixes #2554.

**How to test**
Running the following program on level 2:

```
print Let's go to sleep!
sleep 30
print Wake up!
```

Then run the following program on level 8:
```
print 'knock knock'
sleep 30
print 'knock knock!'
print 'Who is there?'
```

Verify that the first case won't terminate on a time-out whereas the second one will.


Felienne (approved):
Works!

PR #2559, opened 2022-04-20
[FEATURE] Enables teachers to enable/disable the quiz

### LINKED ISSUE #2558: Enable teacher to enable/disable quiz
**Description**
With the latest clean-up tasks we no longer see the quiz as an adventure but rather as a separate tab integrated into the adventures tabes. This has a lot of advantages but also one minor issue: teachers can no longer enable/disable the quiz for each level as if it is an adventure. We should add this as an "additional setting" to make sure they can still hide/show the quiz.

TiBiBa:
**Description**
In this PR we re-introduce the feature for teachers to enable/disable the quiz in a class. This was already possible when we still view the quiz from an adventure perspective. However, with recent clean-ups we dynamically load the quiz into the tabs; teachers can no longer enable/disable the quiz like an adventure. We fix this by adding the "hide quiz" setting to the other_settings tab within the customizations. We also slightly improve the UI so the buttons are nog longer at the bottom of the page in the case of a lot of open levels.

**Fixes**
This PR fixes #2558.

**How to test**
Make sure you are a teacher and have a class. Navigate to customizations and verify the new UI and the added option for "hiding the quiz". Verify that this is working as expected.


Mark-Giesen (approved):
Works like a charm

PR #2561, opened 2022-04-20
[BUG FIX] Correctly resets customizations on removal

### LINKED ISSUE #2560: [BUG] "Remove customizations" incorrectly removes all checkboxes
**Describe the bug**
When removing customizations we want all checkboxes and levels automatically selected as if nothing has happened. Currently the opposite is happening: all boxes are unchecked. If a teacher accidentally removes the customizations and then saves nothing is available, undesirable! We should fix this.

**Expected behavior**
We removing customizations with "remove customizations" all levels and checkboxes should automatically be selected.

TiBiBa:
**Description**
In this PR we fix a small bug where the customizations page would incorrectly reset on customizations removal. Of course, simply re-loading the page would also have done the trick. But this nicer as it doesn't require a re-load so we are still able to show the feedback message.

**Fixes**
This PR fixes #2560.

**How to test**
Make sure you are a teacher and have some class customizations. Go to the customizations page and remove them, verify that all levels and adventures are correctly checked after the removal. Also verify that all teacher adventures are unchecked as well as all additional settings.


Felienne:
@Mark-Giesen maybe you want to test this one?

Mark-Giesen:
> @Mark-Giesen maybe you want to test this one?

Yes, I would. Didn't work yesterday, like to retry.

Mark-Giesen (commented):
Doesn't seem to work, maybe the testscript is not enough for me. This is what I did. Start hedy local and go to "For teachers", click my class and then the button "customize class". I unchecked some adventures in some levels and checked the "Hide cheatsheet" checkbox and clicked save.
This all went fine, revisiting this page results in the saved settings. Pressing the Remove customization button is what I think should be tested, so I press this, but nothing happens. Not even after reloading the page.

Felienne (dismissed):
Seems not to work yet (I did not check but @Mark-Giesen did)

TiBiBa:
Missed this one, will take a look at it right away!

TiBiBa:
My bad! Missed some edge cases, should all work correctly now.

TiBiBa:
@Mark-Giesen I found and fixed the bug! This was due not correctly escaping the quote in the function call on the template. Should work now!

Mark-Giesen (approved):
Yep, works!

PR #2563, opened 2022-04-20
[BUG FIX] Makes sure submit button is directly aligned correctly on share

### LINKED ISSUE #2562: [BUG] On program share the submit button is aligned incorrectly
**Describe the bug**
When sharing / unsharing a program we dynamically change the program box so we don't require a re-load at each change. So far so good, but for some reason the appearing submit button does not align correctly to the right side of the box.

Schermafbeelding 2022-04-20 om 11 55 01

**Expected behavior**
When sharing the submit button should be aligned to the right side, exactly at the same spot as on a re-load.

TiBiBa:
**Description**
In this PR we fix a very small bug (but it has annoyed me for quite some time). When visiting the programs page and sharing a non-shared program the UI is automatically updated without a re-load, hiding the share button and showing an unshare, copy link and submit button. However, for some reason the alignment of the submit button was incorrect when doing this dynamically. When simply re-loading it worked fine. This resulted in an inconsistent view until a re-load, as can be seen on the screenshot of the related issue. Even weirder: this is due to using Tailwind! When using the `hidden` class this error occurs, however when using `style="display: none;"` it does not. Even when the class code is identical.

**Fixes**
This PR fixed #2562.

**How to test**
Make sure you have a shared and a non-shared program and navigate to the programs page. When sharing the non-shared program verify that the submit button is directly aligned correctly to the right-side of the program box.


Mark-Giesen (approved):
OK

PR #2569, opened 2022-04-20
[BUG FIX] Adds missing 'program is already running' message

### LINKED ISSUE #2568: [BUG] Modal alert is empty when trying to execute a program while there is a waiting input
**Describe the bug**
Currently there is a weird bug where we receive an empty modal alert when we try to run the program again at the moment there is still a program running and we are waiting for the input to be entered by the user.

Schermafbeelding 2022-04-20 om 15 25 35

**Expected behavior**
We should either add a useful error message or remove the run button disabling on input.

TiBiBa:
**Description**
In this PR we add a missing error message, resulting in an empty modal alert. This only occurs in the situation where a user is currently running a program with some user input. While the inline modal is up the run button is disabled, when pressing it an error message should be returned "you already have a running program". This message was missing and is implemented in this PR.

**Fixes**
This PR fixes #2568

**How to test**
Create a program that requires some user input. When the inline modal is up, try running the program again and verify that an error message is correctly display with an explanation why this is not possible.


Mark-Giesen (approved):
Looks good and works.

Issue #1942, opened 2022-02-11
`repeat` level 11 and further


juliabolt:
Right now, in level 10, `repeat` and `for` both work, but from level 11 on, only `for` works.
Maybe, we would want `repeat` to keep working.
After discussing this with @Felienne , we decided the following. (correct me if I'm wrong)
The SEMANTICS.md file will be adjusted to how Hedy works now, so `repeat` is not possible in level 11 and further, and that should be the same in the language for now, to keep things consistent.
Later on, `repeat` can be added in level 11 and further in the language and the semantics, if/when we have that working (but that might take a while).

So now, `repeat` should be removed from level 11 and further.

Felienne:
This is quite doable too I think @huang-v, maybe easier even than the translation of the keywords that need indentation

huang-v:
I am confused on what needs to be changed. Currently "from level 11 on, only for works." which I have tried and observed. If the issue is asking to remove repeat for level 11 and beyond, that is already the current observed behavior. Is the issue asking to reimplement repeat to work for level 11+?

Felienne:
for indeed is fine as it is. The idea was to add repeat to levels over 11.

huang-v:
I am currently working on this issue. 
I was able to successfully enable repeat for levels over 11 but am struggling to get the keyword repeat highlighted. 
![image](https://user-images.githubusercontent.com/102200726/160972202-fcae9c30-9d2c-411e-86b2-87ac15e52c72.png)
Where is highlight information stored?

huang-v:
I noticed that keywords in other languages don't seem to be highlighted either. 
![image](https://user-images.githubusercontent.com/102200726/162607334-fba28b2a-2d28-46cb-811c-e25f085012a1.png)(from level 8)
Do I need to highlight the text? If so, could I get some assistance?

Felienne:
@thjazi39 can probably help.

But note that keywords in other langs will only be highlighted if you enable the corresponding keyword lang in your profile!

Felienne:
Highlighting info is here (f.e. for Spanish): 

https://github.com/Felienne/hedy/blob/main/static/js/syntaxLang-es.ts

thjazi39:
[...]

PR #2572, opened 2022-04-21
Adds repeat keyword to levels 11+

huang-v:
**Description**

Adds ability to use the keyword "repeat" in levels 11+

#1942

**How to test**

Repeat tests were added to level 11 as that is where the keyword is removed from the cumulative command variable and all levels were manually tested 

**Checklist**
  

hedy.py
```diff
@@ -115,14 +115,14 @@ def promote_types(types, rules):
     8 :['ask', 'is', 'print', 'forward', 'turn', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'repeat', 'times'],
     9 :['ask', 'is', 'print', 'forward', 'turn', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'repeat', 'times'],
     10 :['ask', 'is', 'print', 'forward', 'turn', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'repeat', 'times', 'for'],
-    11 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range'],
```
Felienne, line 118: It seems we are adding and not removing repeat here?

Felienne (changes requested):
Hi @huang-v,

This PR seems to be a mix of adding and removing repeat? The title of the PR says adding but the text says removing, and the code seems to mix it too.

I know that the issue itself is also formulated very vaguely, so maybe this is our bad also!

Ultimately we want to have smooth level borders and allow repeat that way, so maybe we should just leave this all as it is now.

huang-v:
Sorry for the confusing wording.
I was told that "The idea was to add repeat to levels over 11."

The list of available keywords is based on the command variable in the lark files. Before, the repeat keyword was removed at level 11. I removed this removal however which essentially adds repeat to levels 11+

Felienne:
ok got it!

PR #2581, opened 2022-04-21
Move remaining yaml of website in one file.

Mark-Giesen:
A lot of text of the website has moved to .po files. The remaining text were left in three folders: start, learn-more and for-teachers. These are now bundled in one file. With this the website can be serviced from one component in Weblate (besides Texts for the po files).
Lock weblate before this merge and recreate the component for this new folder.


hedyweb.py
```diff
@@ -28,7 +28,10 @@ def get_translations(self, language):
 class PageTranslations:
   def __init__(self, page):
     self.data = {}
-    translations = glob.glob('content/pages/' + page + '/*.yaml')
+    if page in 'start, learn-more, for-teachers':
```
TiBiBa, line 31: Shouldn't this be a list like this: `['start', 'learn-more', 'for-teachers']`?

templates/main-page.html
```diff
@@ -1,24 +1,47 @@
...
-{% block body %}
+{% extends "auth.html" %} {% block body %}
 <div class="px-8 py-12 relative text-xl">
-    <a class="github-fork-ribbon left-top z-10" href="https://github.com/felienne/hedy" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a>
+    <a
+        class="github-fork-ribbon left-top z-10"
+        href="https://github.com/felienne/hedy"
+        data-ribbon="Fork me on GitHub"
+        title="Fork me on GitHub"
+        >Fork me on GitHub</a
```
TiBiBa, line 8: This forced line-breaking makes the code harder to read and doesn't change anything. Is it possible to reverse this?

templates/main-page.html
```diff
@@ -1,24 +1,47 @@
...
+                    {{_('main_subtitle')}}
+                </h2>
+            </div>
+            <div class="flex-none">
+                <a href="/hedy"
+                    ><button class="green-btn text-white px-8 py-4">
+                        {{_('try_it')}}
+                    </button></a
+                >
+            </div>
```
TiBiBa, line 37: Same, this HTML code is very hard to read. Especially with the newlines with single `>` symbols.

hedyweb.py
```diff
@@ -28,7 +28,10 @@ def get_translations(self, language):
 class PageTranslations:
   def __init__(self, page):
     self.data = {}
-    translations = glob.glob('content/pages/' + page + '/*.yaml')
```
TiBiBa, line 31: It seems we don't cache this (yet). Might be one reason for the extra memory load we currently have, doesn't have to be fixed in this PR but it would be nice if we cache the page translations as well.

TiBiBa (changes requested):

hedyweb.py
Mark-Giesen: Everythings seems to work in Python :-)

templates/main-page.html
Mark-Giesen: Ah, that's the automatic formatting tool in VSCode. Yes, I can revert that.

hedyweb.py
TiBiBa: That's because it's a string comparison in this case which is dangerous. Because 'a' in ... is also valid as the character can be found in the string. With a in list comparison we only validate to True if it matches one of the items exactly.

templates/main-page.html
Mark-Giesen: better?

templates/main-page.html
```diff
@@ -1,24 +1,23 @@
-{% extends "auth.html" %}
-
-{% block body %}
+{% extends "auth.html" %} {% block body %}
 <div class="px-8 py-12 relative text-xl">
-    <a class="github-fork-ribbon left-top z-10" href="https://github.com/felienne/hedy" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a>
+    <a
+        class="github-fork-ribbon left-top z-10" href="https://github.com/felienne/hedy" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a
+    >
```
TiBiBa, line 5: These lines still have a single-standing `>` symbol, can this be on one line as well?

TiBiBa (dismissed):

translations/en/LC_MESSAGES/messages.po
```diff
@@ -1645,22 +1645,20 @@ msgid "mail_welcome_teacher_body"
 msgstr ""
 "<strong>Welcome!</strong>\n"
 "Congratulations on your brand new Hedy teachers account. Welcome to the "
-"world wide community of Hedy teachers!\n"
+"world wide community of Hedy teachers!\n\n"
```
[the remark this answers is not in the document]
TiBiBa: Not sure how this happens, can you see the author of this commit?

templates/main-page.html
Mark-Giesen: sorry, mist that

templates/main-page.html
```diff
@@ -1,24 +1,23 @@
...
+                <h2 class="font-sans font-light text-white text-shadow-md tracking-wide my-1">{{_('main_subtitle')}}</h2>
+            </div>
+            <div class="flex-none">
+                <a href="/hedy"><button class="green-btn text-white px-8 py-4">{{_('try_it')}}</button></a>
+            </div>
         </div>
-        <div class="flex-none">
-            <a href="/hedy"><button class="green-btn text-white px-8 py-4">{{_('try_it')}}</button></a>
+        <div id="main_page_content">
+            {% for section in content['start-sections'] %}<h2>{{ section['title'] }}</h2><p>{{ section['text'] | commonmark }}</p>{% endfor %}
```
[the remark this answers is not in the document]
Mark-Giesen: Off course, I was just trying to catch a train ;-)

TiBiBa (approved):

PR #2582, opened 2022-04-21
Fix typos in en.yaml

joppiesaus:
Changes the "en"s (Dutch for "and") to "and"s, and renames "mension" to "mansion" in `en.yaml`

**How to test**

This is a language change. Run Hedy and then go to http://localhost:8080/hedy/9#haunted and http://localhost:8080/hedy/16#haunted to see the difference.

lvl 9 haunted house new: (note the and/en at the end of the sentence)
9new

lvl 9 haunted house old:
9old

lvl 16 haunted house new (also note mension->mansion near the top right:
16new

lvl 16 haunted house old:
16old

This PR doesn't reference an issue attached as I forgot to create one and I think this is a really small change. Feel free to reject the PR if there is some good reason to create an issue first.


Mark-Giesen (approved):
Good catch!

PR #2591, opened 2022-04-22
[FEATURE] Implements stop button

### LINKED ISSUE #2570: [BUG] Inline modal promise is not correctly thrown away on modal hiding
**Describe the bug**
This is quite a complex bug but I will try to explain it as shortly as possible. When requiring some user input we show the user the `inline modal` which enables them to enter some text. When the modal is shown we create a TypeScript Promise which waits for the inputs and returns it to the program. However, when we alter the program while waiting for an answer the modal is automatically hidden and we can continue to edit our code, but the Promise isn't forgotten! This creates an unclear and complex situation where the answer is printed multiple times. The bug can be re-produced in the following way:

Execute the following program in level 1:
```
print hello world
ask what is your name?
echo so your name is
```
Now, while the inline modal is still up edit the program resulting in the modal being hidden. Then run the exact program again and actually enter some text to the answer. You will notice that the output will be as follows:
```
hello world
so your name is Timon
so your name is Timon
```
The echo is printed twice! We can repeat this process infinity times and the echo will be repeated infinity times as well. I think this bug can be solved by forcing TypeScript to remove all awaiting Promises at the moment of inline modal hiding. This does however has to be done delicately as we also use some Promises to kill running programs in the background when we run a new one. When we fix this bug the disabling of the run button on the inline modal will also become redundant as we can no longer interfere the program with the modal.

@Felienne Can we remove the run button disabling when the inline modal is up or is this there of another reason? If I remember correctly we implemented this once to prevent weird behavior with the input and background running programs.

**Expected behavior**
When the inline modal is hidden for any reason the related Promises keeping track of the input should be forgotten.

TiBiBa:
**Description**
In this PR we proudly present: a stop button! Currently users are unable to stop programs at any time they want, the only option so to run another program to "overwrite" the current running program. However, this in turn messed with a possible `Promise` created by the inline-modal waiting for input. This prevent this created a bucket-fix to disable to run button as long as there is a modal waiting. Not ideal!

We fix all these issues in this PR. We implement a stop button that shows directly when a program starts running. A user can press this button anything to terminate the program, also when there is a waiting inline-modal. We implemented a work-around by faking some input and directly hiding the program, making it seems as nothing has happened.

**Fixes**
This PR fixes #2570.

**How to test**
Test a variety of programs containing sleep, repeat and possible infinity loops. Verify that the stop button shows as soon as you press the run button and also verify that it terminates the programs successfully in all situations.


TiBiBa:
Schermafbeelding 2022-04-25 om 11 55 16

static/js/app.ts
```diff
@@ -225,6 +223,22 @@ function clearErrors(editor: AceAjax.Editor) {
   }
 }
 
+export function stopit() {
+  // We bucket-fix stop the current program by setting the run limit to 1ms
```
Felienne, line 227: Does this "bucket fix" need an improvement in the future (then maybe add an issue), or are we ultimately happy with it? (in that case, can we remove the comment?)

Felienne (approved):
Are we missing the po text changes(otherwise it works like a charm!)

PR #2604, opened 2022-04-25
[IMPROVE] Always set keyword lang to user lang if not specified in profile

### LINKED ISSUE #2525: [BUG] No default syntax highlighting for Arabic
Ok, this is not really a bug, but a wrong decision previously made (in https://github.com/Felienne/hedy/pull/2357)

image

While it makes a lot of sense for Dutch not to highlight keywords by default, for Arabic (and Hindi, and Greek?) we do want to highlight the language itself because these languages are a lot less likely to use English keywords. The issue is also due to the level defaults in Arabic not using `{print}`, so we also immediately see the English keywords, and see them highlighted.

Maybe we should have categories of languages, "English by default" for Latin languages, and "local by default" for non-latin? Or maybe the highlighter should just be better so there is no confusion (we are working on that)?

But for now, I think we should revert #2357 for Arabic? If those content files also use `{print}`, we have to see what the situation of the highlighter is then and how we want to approach things.

TiBiBa:
**Description**
In this PR we fix some highlighting issues due to incorrect YAML files. Originally we wanted to set the keyword language to English when no keyword language was selected. However, in some languages the keywords themselves are translated in the YAMLs, so when only highlighting English they are no longer highlighted. This is an issue because it makes the UI very unclear. This should be fixed in the YAMLs as the mistake is there, but this is a work-around solution to make sure the language is still highlighted. We now follow this flow:

- If there is keyword language in the session, set that one (this is from a logged in user)
- If not, verify if the `g.lang` is also an existing keyword language
- If so, set this as keyword language
- If not, set `en` as keyword language

We also make a small change on the language switcher. We only show this switcher iff the user has a set keyword language on their profile that is **not** English. Lastly we surround the parsers by some nice try/except statements to make sure all weird error are caught as expected.

**Fixes**
This PR fixes #2525.

**How to test**
Verify that the syntax highlighting for Arabic is working as expected, even though the YAML files don't contain valid placeholders.


Felienne (approved):
Better this way

PR #2616, opened 2022-04-25
[FEATURE] Makes mail optional fur students signup

TiBiBa:

**Description**

_Changes in detail_

**How to test**

**Checklist**
  
### DIFF
--- a/templates/create-accounts.html
+++ b/templates/create-accounts.html
@@ -14,7 +14,6 @@
                 {% for i in range(5) %}
                     <div class="flex flex-row w-full items-center mb-4 account_row {% if i == 0 %}hidden{% endif %}" {% if i == 0 %}id="account_row_unique"{% endif %}>
                         <input id="username" name="username" {% if i != 0 %}required{% endif %} type=text placeholder="{{_('username')}}" class="w-full block appearance-none bg-gray-200 border border-gray-200 text-gray-700 py-3 px-4 ltr:pr-8 rtl:pl-8 rounded ltr:mr-4 rtl:ml-4">
-                        <input id="email" name="email" {% if i != 0 %}required{% endif %} type=email placeholder="{{_('email')}}" class="w-full block appearance-none bg-gray-200 border border-gray-200 text-gray-700 py-3 px-4 ltr:pr-8 rtl:pl-8 rounded ltr:mr-4 rtl:ml-4">
                         <input id="password" name="password" {% if i != 0 %}required{% endif %} minlength="6" type=password placeholder="{{_('password')}}" class="w-full block appearance-none bg-gray-200 border border-gray-200 text-gray-700 py-3 px-4 ltr:pr-8 rtl:pl-8 rounded ltr:mr-4 rtl:ml-4">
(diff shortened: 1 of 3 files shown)


TiBiBa:
Note to self:

We should also write some e2e tests to verify the accounts without mail addresses work as expected. In addition to this it would be nice to also add some tests for the creation of the accounts (we currently only verify single account creation).

TiBiBa:
On second thoughts we have to think about some more flows that change and how we want these to go. I'm still in doubt about the following issues:

- Student users never see the landing page, how do we fix this?
    - Maybe, when on login and you have a related teacher AND a verification pending -> landing page
    - We should also remove the verification pending to prevent this on every login
- What to do with verification pending?
    - We can make the teacher verify this -> send welcome mail to teacher
    - Or use the approach as described above

Love you hear your input on this! @Felienne @Mark-Giesen

TiBiBa:
> On second thoughts we have to think about some more flows that change and how we want these to go. I'm still in doubt about the following issues:
> 
> * Student users never see the landing page, how do we fix this?
>   
>   * Maybe, when on login and you have a related teacher AND a verification pending -> landing page
>   * We should also remove the verification pending to prevent this on every login
> * What to do with verification pending?
>   
>   * We can make the teacher verify this -> send welcome mail to teacher
>   * Or use the approach as described above
> 
> Love to hear your thoughts on this! @Felienne @Mark-Giesen

These thoughts are implemented in this PR together with some additional e2e tests for multiple account creation. It should all work now and is ready for review!

static/js/auth.ts
```diff
@@ -122,15 +122,19 @@ export const auth = {
       }).done (function (response) {
         // We set up a non-falsy profile to let `saveit` know that we're logged in. We put session_expires_at since we need it.
         auth.profile = {session_expires_at: Date.now () + 1000 * 60 * 60 * 24};
-        afterLogin({"admin": response['admin'], "teacher": response['teacher']});
+        // This happens when a student account (without an mail address logs in for the first time
+        if (response['first_time']) {
+          return afterLogin({"first_time": true});
+        }
+        return afterLogin({"admin": response['admin'] || false, "teacher": response['teacher']} || false);
       }).fail (function (response) {
```
Mark-Giesen, line 130: afterLogin expects an "any" type, should be changed into "dict".

TiBiBa: Good catch! Typescript doesn't support the `dict` type as parameter (didn't know that either!). I just added an additional interface creating the `Dict` type and it should all work as expected now.

Felienne:
> These thoughts are implemented in this PR together with some additional e2e tests for multiple account creation. It should all work now and is ready for review!

> These thoughts are implemented in this PR together with some additional e2e tests for multiple account creation. It should all work now and is ready for review!

Thanks, but I am not entirely sure which ones we have chosen (sorry for not answering earlier!) so I need a bit more context.

> * Student users never see the landing page, how do we fix this?
>   
>   * Maybe, when on login and you have a related teacher AND a verification pending -> landing page

You mean the nice "here are things you can do" page? I guess that is a lot less important for student learners cause they have a teacher to help them get started? So I am ok if they end up not seeing it.

PR #2617, opened 2022-04-26
[CHORE] Simplifies quiz yaml

### LINKED ISSUE #2571: [CONTENT] Improve quiz yaml
I know how they came to be but I am not in love with the way we do quizzes right now. Question have as an answer either code or feedback_option (f.e. in Dutch):

This: 

```
                -   code: "{echo}"
                     feedback: "Met {echo} herhaal je een antwoord..."
```

or

```
                -   option_text: "Haydie"
                     feedback: "Deze is het niet !"
```

This is done because we render code differently from text but it is not so pretty (the keys are very differently names but do the same) and leads to weblate problems when quizzes for different langs have different fields (such as in https://github.com/Felienne/hedy/pull/2472) and I suspect that will be inevitably going forward because in some languages you might want to have special questions related to the language (such as: "what is the meaning of this keyword" which you might not need in En)

I would prefer it if we'd make all answers similar, in a field `answer_option` and add a boolean for whether or not an option is code, as such:

```
                -  answer_option: "{echo}"
                    feedback: "Met {echo} herhaal je een antwoord..."
                    code: true*
```

or

```
                -   answer_option: "Haydie"
                     feedback: "Deze is het niet !"
                     code: false
```

* or whatever a boolean is in yaml, maybe it was yes or no? I am too lazy to look that up rn 

Tagging @Mark-Giesen also because it might impact weblate?

TiBiBa:
**Description**
In this PR we simplify the structure of the quiz YAMLs. Currently we differentiate between the `option_text` and `code` as possible answers to a question. We simplify this by making al options `option` and surrounding code blocks by triple backticks so they are automatically parsed as `<pre>` elements in the front-end. And thereby also automatically shown as ace editors.

The translations for all dutch and English YAMLs are fixed and a lot of spelling mistakes and missing placeholders are fixed as well. For French and Norsk the structure is fixed but not (yet) the code backticks. I think it would be best to let Weblate handle this with contributors. For all other languages the English YAML is copied onto the file as not translations where made yet.

**Fixes**
This PR fixes #2571 

**How to test**
This is a clean-up task, if all works it works.


Mark-Giesen:
The quizzes have been translated into chinese as well. @TiBiBa do you want to copy that into this PR of have a go at merge?

TiBiBa:
> The quizzes have been translated into chinese as well. @TiBiBa do you want to copy that into this PR of have a go at merge?

Let me take a look how we can best merge this!

TiBiBa:
> The quizzes have been translated into chinese as well. @TiBiBa do you want to copy that into this PR of have a go at merge?

It seems only a few strings are translated, might not be worth manually copying. I think the best approach would be to merge this as soon as possible too avoid further updating on the old quiz structure.

Mark-Giesen:
> > The quizzes have been translated into chinese as well. @TiBiBa do you want to copy that into this PR of have a go at merge?
> 
> It seems only a few strings are translated, might not be worth manually copying. I think the best approach would be to merge this as soon as possible too avoid further updating on the old quiz structure.

OK, that's doable, but in all language files the leverls 10 and 11 are switched this weekend. Want to fix that after merge, or before?

Mark-Giesen (approved):
Everything in sync

PR #2624, opened 2022-04-28
[FIX] Makes font consistent between global editor and code snippets

### LINKED ISSUE #2622: [UI] Why do we have two different comma views?
I just noticed (while debugging Chinese commas) that our example snippets show a different comma than Ace itself, look:

image

It is 
image
 versus
image

This is purely visual, it seems to be the same ASCII characters (when copy-pasting from above, you get the lower one)

Is that an Ace setting? Can we make it consistent?

TiBiBa:
**Description**
In this PR we fix that both the global editor and the code snippets share the same `fontFamily` to make sure special characters such as comma's are shown identical.

**Fixes**
This PR fixes #2622

**How to test**
Navigate to level 3, notice that the comma's in both the code snippets and the global editor are the same.

**Old view**
Schermafbeelding 2022-04-28 om 14 51 37

**New view**
Schermafbeelding 2022-04-28 om 14 51 22


templates/incl-editor-and-output.html
```diff
@@ -28,7 +28,7 @@
           </div>
       {% endif %}
       <div class="w-full flex-1 flex flex-col">
-        <div id="editor" style="background: #272822; font-family: monospace; color: white" lang="en"
+        <div id="editor" style="background: #272822; font-size:0.95em; color: white" lang="en"
         data-loaded-program="{{not not loaded_program}}"
```
Mark-Giesen, line 32: shouldn't we force a monospaced font-family in code blocks?

Mark-Giesen: Ah, thanks for the clarification.

Mark-Giesen (approved):

PR #2655, opened 2022-05-02
[UI IMPROVEMENT] Differentiates between keyword and non-keyword languages on menubar

### LINKED ISSUE #2618: [UI idea] Show which langs have keywords
We don't really show now in an easy way (for users but also for me as a person that very often gives demos) what languages have keywords, can we make that clear? Maybe in the lang switcher, instead of the whole list, we have two lists, one with and one without keywords? Or something else like a symbol in the list, or maybe an place outside of the list?

TiBiBa:
**Description**
In this PR we more clearly differentiate between supported and non-supported keyword languages of Hedy. Instead of one big language selector we divide them into two rows clearly differentiating the support. This can still be a bit confusing as some keyword supported languages are not translated at all on other aspects, but this should automatically be fixed as more translations are contributed. All is done dynamically and nothing has to be updated when a new (keyword)language is added.

**Fixes**
This PR fixes #2618.

**How to test**
Navigate to the menubar and notice then new structure of the language selector. Notice that this selector is only visible when not being logged in. Verify that alle features such as filtering and selecting are still working as expected.

Schermafbeelding 2022-05-02 om 08 38 27


Felienne (changes requested):
Thanks for this @TiBiBa! Good to show the distinction!

But I fear "other languages" might be not so informative for new users? What if we do "translated content" there? Or "translated UI"? That makes it a bit more clear what people can expect.

TiBiBa:
> Thanks for this @TiBiBa! Good to show the distinction!
> 
> But I fear "other languages" might be not so informative for new users? What if we do "translated content" there? Or "translated UI"? That makes it a bit more clear what people can expect.

Yes good idea, will fix!

TiBiBa:
Fixed!

Felienne (approved):
Nice!

Felienne:
Owno, this has merge conflicts now @TiBiBa. Can you fix them (and then feel free to merge)

PR #2656, opened 2022-05-02
[UI IMPROVEMENT] Improves programs user overview

### LINKED ISSUE #2509: [UI idea] Re-think adventure naming as everything is an adventure
**Idea incl level**
On the programs page we differentiate the naming between adventures and non-adventures. However, with the merging of the level-defaults into the adventures this doesn't make any sense. We should re-think how we want to present users with their programs.

TiBiBa:
**Description**
In this PR we improve the programs overview of a user to clearly differentiate between adventures. Currently the adventure name is printed directly in the program name, which is unclear. We fix this by separately showing the adventure name, if any is there.

**Fixes**
This PR fixes #2509

**How to test**
Navigate to your programs page and verify that the "program box" has a new layout. Also verify that the adventure name is correctly shown (translated) on the corresponding column. When no adventure name is found, "-" is shown.

Schermafbeelding 2022-05-02 om 08 53 19


Felienne (approved):
Nice improvement indeed!

PR #2659, opened 2022-05-02
Better error for repeat statements without a inner commands

### LINKED ISSUE #2590: [BUG] Weird error in level 7 and up when using a block without body
"half" block gives weird error:

image

it is a bit weird, repeat is allowed there of course but it misses a block. Could we make this error message clearer?

boryanagoncharenko:
**Description**

The PR adds a more descriptive error in the cases a repeat statement has no commands to execute

**Fixes #2590 **

**How to test**

Start Hedy locally and navigate to any level between 7 and 18. Run the following snippets `repeat 2 times`,  `repeat 2 times  ` and validate that the error you get back makes sense.
- The error is available only in English atm (please leave a comment with the translation in other languages here and I will update it)
- Note that the text of the error is intentionally made generic so that it can be reused in command-less if-statements and for-statements in the future.

**Checklist**


Felienne (commented):
Nice! 

image

I will add a translation for Dutch and then merge.

PR #2660, opened 2022-05-02
[FEATURE] Auto scrolls output view down with output

### LINKED ISSUE #2464: [UI idea] Automatically scroll output view with content
**Idea incl level**
Current the output view of the code editor doesn't automatically scroll with the content. I think it would be nice to automatically scroll together with the content. However, children would have to scroll up again if they want to read the start of their program. But especially with `sleep` I think automatically scrolling down would benefit most.

TiBiBa:
**Description**
In this PR we make a small improvement to the output view, automatically scrolling down with the loading content. Especially since we introduced the `sleep` command this has become more relevant. Now the view is automatically scrolled to the actual content. However, we might want to make a toggle for this in the future?

**Fixes**
This PR fixes #2464.

**How to test**
Compile the Typescript locally as I seem to have some issues currently with compiling. Therefore for an unknown reason the current app bundle doesn't contain the syntax highlighting. Compile once yourself to fix this. To actual test the PR itself:

Navigate to any level with the `repeat` command (for example level 7). Repeat a random program enough times so the output view is not large enough and verify that the window automatically scrolls down.


Felienne:
It seems this causes an issue in ts: 
image

(maybe related to the merge conflicts?)

Felienne:
> Compile the Typescript locally as I seem to have some issues currently with compiling. Therefore for an unknown reason the current app bundle doesn't contain the syntax highlighting. Compile once yourself to fix this.

Ah  I guess that was the issue I saw earlier, local compile indeed addresses that!

Felienne:
Can't seem to get the scrolling to show, though:

image

TiBiBa:
> Can't seem to get the scrolling to show, though:
> 
> image

Weird, will take a look!

TiBiBa:
Works fine for me on Firefox, Safari and Chrome. But my Typescript is still (partly) broken making it hard to do some debugging as I can't compile..

Felienne:
Ok, this works now and I think this is a great addition!

There is a weird interaction though when you enable read aloud, then:

1. the scrolling goes on the pace of the prints not on the pace of the reading (which is ok, I guess that will be very hard to fix with the async reading?)
2.  but also... the scroll bar remains frozen until the reading is finished, which might be a thing we can fix? I can imagine the kid wants to scroll up while the reading is still happening? Can we release the bar when the printing is done but the reading aloud is still going @TiBiBa

TiBiBa:
> Ok, this works now and I think this is a great addition!
> 
> There is a weird interaction though when you enable read aloud, then:
> 
> 1. the scrolling goes on the pace of the prints not on the pace of the reading (which is ok, I guess that will be very hard to fix with the async reading?)
> 2. but also... the scroll bar remains frozen until the reading is finished, which might be a thing we can fix? I can imagine the kid wants to scroll up while the reading is still happening? Can we release the bar when the printing is done but the reading aloud is still going @TiBiBa

1. Sounds indeed hard to fix as they are async.
2. This should be possible! Will take a look in this branch (as soon as my typescript is fixed...)

TiBiBa:
The output window does not seem scrollable at all at this moment. Will take a look at what might cause this! Until then, will leave it as a draft.

Felienne:
> > The output window does not seem scrollable at all at this moment. Will take a look at what might cause this!
> 
> Weird, cause it did used to scroll! But this is not high prior at all, feel free to leave open for a bit

Although it might be a local issue also for you? For me the branch still works as before: 

image

(but the read aloud still freezes the outputscroller)

TiBiBa:
Closed as the speak aloud issue isn't easily solved. Might pick this up somewhere later.

PR #2663, opened 2022-05-02
Parsons without syntax

xxxpokemon:
parsons without syntax highlighting


TiBiBa:
I leave this here otherwise we might forget. We should remove the language switcher when on the Parsons tab as it might overlay in some cases, see the screenshot below:

Schermafbeelding 2022-05-03 om 16 29 34

Felienne:
Closing in favor of #2761

PR #2672, opened 2022-05-02
Reduction of the size of the syntax highlighting system

rix0rrr:
(edit by @thjazi39)

The PR #2527 introduced a new system for syntax highlighting, in order to have the core of the syntax highlighting in JSON files, which allows to automate the tests.

The problem is that it introduced a lot of redundancy (one rule file per language) and it generated a much too big `static/js/appbundle.js` file.

This PR allows to solve the problem, by separating the translations and the rules.

So now there are only 2 :
- highlighting.json (which contains the core of the coloring rules)
- highlighting-trad.json (which contains the translations)

(more detail here : https://github.com/Felienne/hedy/pull/2672#issuecomment-1117366478)

And it's in the typescript that the 2 are merged according to the language of the user.

How to test :
Check that it doesn't break anything in the front-end, that it still works with the different browsers, that the JS file is of a reasonable size.


thjazi39:
I don't know if this is possible, but instead of preloading all the languages in the appbundle, maybe it would be possible to not preload anything, and only load the language used by the user, when the page is loaded?

rix0rrr:
> maybe it would be possible to not preload anything, and only load the language used by the user, when the page is loaded?

Maybe. Please investigate this. Keeping the following in mind while doing so:

- Loading files from the internet takes time. What will the user observe during that time?
- JavaScript is single-threaded, and the thread must not block, which means file loading will have to be asynchronous. What (if anything) does that change about the code structure?
- What if the user changes the language while the page is already loaded?
- What if the file does not exist?

Felienne:
> I don't know if this is possible, but instead of preloading all the languages in the appbundle, maybe it would be possible to not preload anything, and only load the language used by the user, when the page is loaded?

Sure we might be able to do something with smart loading, but the fact remains that there really is a lot of duplication in all the json files (basically, only the keywords differ and the rest is the same), which we could minimize to save traffic/space. Could you investigate how to do that?

thjazi39:
Concretely, it would be like having a single JSON rule file, and rebuilding the translations in typescript?

This will require some changes, but it is possible.

Regarding the testing of the translations, it will mean that we will have to mimic this process of translations, but since this is not a part that will move a lot, it should not be a problem!

Felienne:
> Concretely, it would be like having a single JSON rule file, and rebuilding the translations in typescript?

Yes that was the idea, that saves a lot of files.

> This will require some changes, but it is possible.

Great! And then indeed we can try to do it as you proposed, only for the language that the user has selected to save processing power.

> Regarding the testing of the translations, it will mean that we will have to mimic this process of translations, but since this is not a part that will move a lot, it should not be a problem!

Great!!

thjazi39:
The translation of regexes is now operational, I hope it does not break anything in the front-end

thjazi39 (approved):

build-tools/heroku/generate-typescript
```diff
@@ -64,6 +64,7 @@ else
         --minify \
         --target=es2017 \
         --global-name=hedyApp \
+        --analyze \
```
TiBiBa, line 67: Would be nice if we remove this parameter when deploying.

TiBiBa (changes requested):
It doesn't seem to work as expected everywhere for me. When on English, level 1 works fine but level 2 and 3 only support partly syntax highlighting. I tried compiling the TypeScript again but this doesn't make any difference. See the screenshot below for clarification:

Schermafbeelding 2022-05-06 om 11 36 12

TiBiBa (approved):
Looks good!

PR #2673, opened 2022-05-03
[CHORE] Improves and simplifies achievements yaml

### LINKED ISSUE #2404: [CONTENT] Add YAML validation to achievements
**Describe the issue**
Our current YAML structure on the achievements is prone to errors due to the files not being validates. The e2e tests only use either the en/nl website so only these files are validated through that process. It should be nice if we add some scheme validation as is already there for the adventures and level-defaults.

TiBiBa:
**Description**
In this PR we simplify the current Achievements structure. Currently we use a dedicated YAML file to keep track of the achievements structure and translatable YAML files to keep track of the achievement translations. This was cumbersome as we where unable to alter stuff without first changing the structural YAML. We fix this by placing the structure as a dict on the top of the achievements page, fixing the structure at that spot. We move some translations to gettext and remove the image URL from the translations (as this is a-non translatable string).

**Fixes**
This PR fixes #2404.

**How to test**
This is a clean-up task, everything should still work as before. Verify that the achievements page as well as the public profile still work in multiple languages. Also verify that the achievement pop-ups still work.


TiBiBa:
**Note**
Before merging we might want to double-check if we don't look any translations this way. (Maybe we can still manually copy some stuff from the YAML to the .po files to prevent this).

PR #2676, opened 2022-05-03
Fix numbers of groups in Highlighting

thjazi39:
After a merge of the main on the test branch, I detected an error in the syntax highlighting which is due to a bad use of groups in level 2

### how to test :
check that it does not break anything in the syntax highlighting, in the front-end or in the tests


TiBiBa:
Can you add a `how to test`. Then I can review this PR and merge.

thjazi39:
There is not much to test, it does not affect the existing coloring (Ace is made so that what is not captured in brackets, when there is only one group, will still be displayed)

This is very strange, because when there are several groups, the text that is not captured in a group will not be displayed.

This is not a problem at the moment. But rather than accommodating this strange behavior, and trying to reproduce it in the syntax highlighting test, I would rather change the rule, so as not to rely on ambiguous behavior

For the test, we can check that it does not change the existing coloring (level 2, with comments)

Felienne:
> For the test, we can check that it does not change the existing coloring (level 2, with comments)

Would be supernice if you can give us a screenshot or description of the expected coloring so it is easier to check.

thjazi39:
We have, before and after this PR :
image

Felienne:
ok, it seems to all still work, incl comments in level 2:

image

Felienne (approved):

PR #2686, opened 2022-05-04
Allow two forms of `is` in Arabic grammar

### LINKED ISSUE #2683: [Language idea] Allowing two forms of "is" in Arabic
**Language idea incl level**
Allowing two forms of "is" in Arabic 

The way "is" is translated can be confusing when used with different gender forms. 
currently "is" is translated into "هو" which is the male form. We need to allow "هي" the female form.

Felienne:
**Description**

Changes the grammar for the male and female form of `is`. 

**Partly fixes #2683**

This is not yet an ideal solution because it is changed in the grammar directly so: it can't be changed in Weblate anymore (ok, I guess) but also will not be taken into account while generating syntax highlighting:

image

This problem is not limited to Arabic, in French we allow (but not highlight) `repete` without accents:

image

Maybe we should have a special command in the Weblate yaml for alternatives, something like a pipe which we then read when making the grammars and the highlighting? (Tagging @thjazi39 so he can form an opinion too)

**How to test**

All tests should pass


thjazi39:
Ok, I see the problem with syntax highlighting:
Before, we used to build "manually" (using a script, but that we only executed once, and then modified by hand).
With the new system, we automatically use the yaml files (content/keywords), and this prevents us from adding executions by hand.

So the question is: Is it better to have a syntax highlighting system that automatically adapts to new translations based directly on the yaml files?
Or a semi-automated system (like before, with frozen translations) but which allows to add manually execptions (accents, multiple forms)?

The ideal would be of course to include multiple forms in the translation files, but I don't know if this is possible

Felienne:
> Ok, I see the problem with syntax highlighting: Before, we used to build "manually" (using a script, but that we only executed once, and then modified by hand). With the new system, we automatically use the yaml files (content/keywords), and this prevents us from adding executions by hand.

> So the question is: Is it better to have a syntax highlighting system that automatically adapts to new translations based directly on the yaml files? Or a semi-automated system (like before, with frozen translations) but which allows to add manually execptions (accents, multiple forms)?

For many reasons, both would be reasonable, but... given that we want to limit file sizes too, we have to make sure we can generate on the fly and thus not save files and override (as discussed here: https://github.com/Felienne/hedy/pull/2672#issuecomment-1115960752)
 
> The ideal would be of course to include multiple forms in the translation files, but I don't know if this is possible

Yes, it is no problem to have people in Weblate type: "repete | répetè" We can document in Weblate that that is allowed. 
I can fix that in the grammars too. So the question for you is: can you also make that work in the generation of the highlighting?

thjazi39:
Ok, I'm redoing the syntax highlighting system to make it lighter.
If the modification is made in the Yaml file, then it should be automatically taken into account in the syntax highlighting

Felienne:
> Ok, I'm redoing the syntax highlighting system to make it lighter. 

Great!!

> If the modification is made in the Yaml file, then it should be automatically taken into account in the syntax highlighting

Ok, I will add it to yaml so you can take that into account too (can be done in a separate pr after which I can merge this one)

Felienne:
> If the modification is made in the Yaml file, then it should be automatically taken into account in the syntax highlighting

I have just pushed the change in the yaml file (also in Fr because that might be easier to test for you) I am not sure it will really be taken into account "automatically", I guess you will have to do a bit of processing to translate the or (written as `|`) into the grammar rules?

Let me know if you need my help.

Felienne:
Ow and this messes up the translation code too @thjazi39 but you do not have to worry about that, just about the highlighting! Not sure if it is best to do that on this PR or on a separate one (both is fine for me)

Felienne:
> On the other hand, it would be necessary to modify the system of replacement of the commands by the 1st occurrence.

Why? What is the issue with the situation we have now? Can you give an example?

thjazi39:
> Why? What is the issue with the situation we have now? Can you give an example?

Here : (sorry I didn't put a screenshot right away, I couldn't change the language for the keywords on the local version)
image


We can see on this screenshot that the repeat keyword is not replaced by the 1st occurrence in several places.

Felienne:
I think I made the right fix, can you check @TiBiBa?

(and could you look at the conflict @thjazi39, I don't dare to touch the highlighting)

Felienne:
And maybe you can check it looks proper @alaaeddin-swidan?

thjazi39 (approved):

thjazi39:
The merge seems to have broken something, but it's at the code translation level, so I don't think I'm the most suitable
(apparently, the tests don't yet tolerate the fact that there are several forms, and the code comparison should take this into account, otherwise the tests might not work every time we add or modify the translations)

Felienne:
Hi @TiBiBa, thanks for pointing out the translation issue, but that already existed before this PR, I made an issue for it here: https://github.com/Felienne/hedy/issues/2711

PR #2691, opened 2022-05-04
[FIX] Only parse keyword language by default on non-latin languages

### LINKED ISSUE #2690: [UI idea] Only show localized keywords by default on non-latin languages
**Idea incl level**
Only shown localized keywords by default on non-latin languages

TiBiBa:
**Description**
In this PR we make a small change so we only parse the keyword when the current language is non-latin. Otherwise we return the English keywords (or the user selected one).

**Fixes**
This PR fixes #2690.

**How to test**
Make sure you are logged out. Verify that for latin languages such as Dutch and French the English keywords are shown. For non-latin languages such as Hindi and Chinese the localized keywords are shown.


Felienne (approved):
Check!

PR #2697, opened 2022-05-05
Make sleep command accept variables, expressions and list access

### LINKED ISSUE #1753: Command sleep cannot be followed by a variable
**Describe the bug**
Given the following program at level 2:

print Ik ben moe jij ook?
moe is ask Ben jij ook moe?
print Wat goed.
sleep
wacht is ask Over hoeveel seconden ga je slapen
print Ok... ik wacht wel even
sleep wacht
print Het is bedtijd  😴🛌

There is an unclear explanation why row 7, sleep wacht, is not not valid.
The  variable is defined in row 5 but is not recognized as a valid input for ‘sleep’

**Expected behavior**
I expect that a variable can be used after the sleep command

Safari on an Ipad Pro

C80B2AFE-A2D8-495C-8589-29C753EEE198

boryanagoncharenko:
**Refactor the sleep command so that it accepts variables, arithmetic expressions and list access**

**Description**

- _Changes sleep command's grammar to support multiple arguments: variables, expressions and list access_
- _Adds type validation, so that the sleep commands accepts only integers_
- _Adds runtime type validation, so that sleep arguments are transformed to a number, e.g. when used with ask command_
- _Automatically updates the sleep argument to fit the range [0, 10], to avoid public programs causing program hanging_

**Fixes #1753 **

**How to test**

1) Consider what would be the acceptable maximum limit for the sleep argument. We do not want users to supply huge values (e.g. 9999999), because this surely will yield an error. The current max value is 10, which means 10 seconds. Is this value too low?

Run Hedy locally and go to level 4.
2) Execute the following snippet and confirm that the error you receive guides you correctly what to correct.
```
n is "test"
sleep n
```
3) Execute the following snippet and supply a non-numerical value. Check that the error makes sense. Note that this error is available only in English atm.
```
n is ask "How long?"
sleep n
```


TiBiBa (approved):
All looks good to me! 

I did notice something strange: When receiving an error from using a string value for `sleep` the stop button is no longer being reversed to the run button. I'm not sure why as the button should automatically reverse on an error and it only seems to occur for this specific situation (have an input modal followed by a runtime error). I don't think this issue is related to this PR but rather the stop button implementation itself. Will take a look!

Also it seems we automatically replace sleep values `> 10`? I tried the following programming in level 3 and it only stopped for 10 seconds and executed fine:

```
print hello!
test is 50
sleep test
print still awake?
```

I do agree that we need some sort of limit to prevent very long programs from running. We have a specific timeout limit for levels 7 and higher (20 seconds with turtle / sleep, 5 seconds without). But we don't have this for lower levels. It might be nice to return a warning in this case? I assume this will be confusing for kids when they don't get feedback why the sleep is shorter then they expect.

boryanagoncharenko:
Thanks for the review.

I noticed that it the program returned errors, the "Stop program" does not transform into "Run program" automatically and perhaps it should. Should I create an issue?

I think that since we have the Run/Stop program feature, we might not need a hard-coded limit on sleep altogether. The idea of this limit is to ensure that if someone shares a program `sleep 99999999`, it will not spread an failing program (which is what happened with other turtle commands: `forward 999999999`). But since the user can now simply stop the program, we don't need it.
Btw I was not able to replicate the warning messages in levels 7 and above. I tried with `sleep 99999999` locally and on production but I got no warnings. Is my test code correct?

TiBiBa:
> Thanks for the review.
> 
> I noticed that it the program returned errors, the "Stop program" does not transform into "Run program" automatically and perhaps it should. Should I create an issue?
> 
> I think that since we have the Run/Stop program feature, we might not need a hard-coded limit on sleep altogether. The idea of this limit is to ensure that if someone shares a program `sleep 99999999`, it will not spread an failing program (which is what happened with other turtle commands: `forward 999999999`). But since the user can now simply stop the program, we don't need it. Btw I was not able to replicate the warning messages in levels 7 and above. I tried with `sleep 99999999` locally and on production but I got no warnings. Is my test code correct?

The following error should occur on level 7 and onwards when the program takes longer than 20 seconds to execute. If I recall correctly the split was made on level 7 as from this level and onwards infinity loops can be created by students. Not an issue from our perspective as the code is executed locally, but the browser will crash at some point making it nog really user friendly.

Schermafbeelding 2022-05-10 om 09 56 07

Issue #2682, opened 2022-05-04
[Language idea] Allowing the extension letter (Tatweel)


alaaeddin-swidan:
**Language idea incl level**
Allowing the extension "letter" (Kashida) in Arabic 
**Additional context**
https://en.wikipedia.org/wiki/Kashida

normal command:  كرر
extended command: كـــــــــــرر

Felienne:
Closed by #2728

PR #2699, opened 2022-05-06
Allow Tatweel for print

Felienne:
First steps in the direction of #2682


Felienne:
Hi @thjazi39!

Another fun challenge for you :D I have changed print in Arabic to allow `ـ` to be placed in between any letter. Can we change the highlighter with this regex? (Ultimately, we would do this in any keyword but I want to see if we can make it happen for print first)

alaaeddin-swidan:
The change works as required. I checked the arabic print in across levels. 
Two comments: 
*The Tatweel letter for the arabic print is only possible between the 1st and the 2nd letter, like this: 
قـول
قــــول
قــــــــــول
and so on. 
 
*It only concerns the print command for now. Do you plan extending the Tatweel to other keywords/commands?

Felienne:
> *The Tatweel letter for the arabic print is only possible between the 1st and the 2nd letter, like this: قـول قــــول قــــــــــول and so on.

You mean that that is the rule in Arabic right (not that that is the rule in Hedy?) Is it the rule that it is only allowed between connected letters? If you have a list of letters I could make it more precise, if you think that is better?

It is now a bit too loose but it will a bit easier to allow it everywhere since I do not know the rule haha.

> *It only concerns the print command for now. Do you plan extending the Tatweel to other keywords/commands?

Yes!

PR #2700, opened 2022-05-06
Refactor tests for levels 7 and 8 #2436

### LINKED ISSUE #2436: [IMPROVEMENT] Refactor tests

boryanagoncharenko:
**Refactor tests for levels 7 and 8**

**Description**

- _Refactors tests for levels 7 and 8_
- _Fixes a bug in repeat statement when there is a variable name collision in levels > 8_
- _Fixes a bug in repeat statement when the passed times argument is a literal non-latin number, e.g. `repeat ५ times` in levels > 8_

**Fixes #2436 **

**How to test**

Examine the changed tests and ensure that:
- they they conform to the rules defined at the beginning of the file
- there are no TODOs or commented code


Felienne:
Ah sorry @boryanagoncharenko, I have worked on the tests a bit too to fix Arabic numberals, can you try to resolve the conflicts?

Felienne:
Haha oops I think we both fixed the same bug :)

Felienne:
I should really keep a better eye on all PR's

Felienne (approved):

PR #2706, opened 2022-05-09
[CLEAN-UP] Creates dedicated route for for-teachers page

### LINKED ISSUE #2705: [CLEAN-UP] Create dedicated route for for-teachers page
**description**
Currently we use a string matching look-up for the rendering of several pages. This doesn't make any sense and makes our functions long and difficult to understand, especially the `for-teachers` page. It would be nice we split this one to a dedicated @app.route.

TiBiBa:
**Description**
In this PR we fix a small clean-up where we shift the for-teachers page route from the global route functions to a dedicated one. This makes our code more readable and is especially useful when implementing the for-teachers tutorial as we do need a dedicated route for that implementation. It would be nice if we split all routes somewhere in the near-future as this would increase readability.

**Fixes**
This PR fixes #2705

**How to test**
This is a clean-up task. Verify that the for-teachers page and all related functionalities are still working. This should also all be caught by the e2e tests, so either they pass or not.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1017,27 +1017,27 @@ def main_page(page):
 @app.route('/for-teachers', methods=['GET'])
 @requires_login
 def for_teachers_page(user):
-  if not is_teacher(user):
-    return utils.error_page(error=403, ui_message=gettext('not_teacher'))
+    if not is_teacher(user):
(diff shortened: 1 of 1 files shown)


TiBiBa:
Closing this one as we made an oopsie with an indentation setting, messing up the whole file

PR #2723, opened 2022-05-12
increasing sorting of translations for syntactic coloring and remove camelCase for snake_case

thjazi39:
In the current syntax highlighting system, there is a file that stores all the keyword translations, the `highlighting/highlighting-trad.json` file.
In this file, the order is not important, but for some unknown reason, the list is not always ordered the same way. This is not a problem for syntax highlighting, but it can make the changes made very unreadable when compared with git.
This Pr therefore orders this file by placing the keys in a fixed order, the alphabetical order.

Look at the file `highlighting/highlighting-trad.json` in the PR#2699 to have an example of the problem.

Another example is this PR, the `highlighting-trad.json` file has not fundamentally changed, it is just sorted in a different way, but since that is not important, we just end up with a lot of changes.
Sorting avoids this problem.

This PR change camelCase to snake_case for variables and functions

**How to test**

Check that nothing is broken in the syntax highlighting or in the front-end


highlighting/generate-rules-highlighting.py
```diff
@@ -80,13 +80,13 @@ def get_Traduction(KEYWORDS_PATH, KEYWORDS_PATTERN):
                 tmp[languageCode][k] = str(yamlFile[k])
 
     Result = {}
```
Felienne, line 82: I know this is present in a lot of you code already, but could you start sticking with the convention that variables start with lower case letters, and use snake_case (so for example `result` and `language_code`)

Felienne (changes requested):

highlighting/generate-rules-highlighting.py
thjazi39: It's done, for all variable (except constant)

PR #2728, opened 2022-05-12
Create lark dynamically

### LINKED ISSUE #2726: [Language idea] Create .lark files for different languages dynamically
**Language idea incl level**
Currently the .lark files containing the grammar for all languages is partly made manually and partly automated. But bottomline is that it's a lot of work to add a new language. It would be nice if we create these files dynamically based on a template file and the YAML file with translations.

TiBiBa:
**Description**
In this PR we enable the dynamic creating of the .lark grammar file for the different languages. Currently we perform this action partly manually and partly automated. We create a template file with placeholders which we overwrite with translation retrieved from the YAML keyword files. We also take into account the edge case where translations might have multiple options.

**Fixes**
This PR fixes #2726 and #2682

**How to test**
Run the `transform_keywords_to_grammar.py` file in the `/content` folder. All .lark files should automatically be created. Notice that they are already created in this PR so running the file is not necessary. Other than that, if the tests pass this is ready to go!


TiBiBa:
@Felienne This (almost) seems to be working. Due to automatically generating the file we also add grammar support for Italian and Pt_BR, but both these languages file on one specific adventure test. Can you maybe take a look? I'm not sure how the back-end structure works when new grammar files are added and if everything should work from now on.

Felienne:
> @Felienne This (almost) seems to be working. Due to automatically generating the file we also add grammar support for Italian and Pt_BR, but both these languages file on one specific adventure test. Can you maybe take a look? I'm not sure how the back-end structure works when new grammar files are added and if everything should work from now on.

I think you have managed to fix this? Or is this still something I should look into specifically?

TiBiBa:
> > @Felienne This (almost) seems to be working. Due to automatically generating the file we also add grammar support for Italian and Pt_BR, but both these languages file on one specific adventure test. Can you maybe take a look? I'm not sure how the back-end structure works when new grammar files are added and if everything should work from now on.
> 
> I think you have managed to fix this? Or is this still something I should look into specifically?

No idea how and why the tests are passing. The did fail earlier but a merge with main seems to have fixed the last two tests. Might be that another bug is already fixed that caused the issue. In that case I think this branch is ready for merge!

TiBiBa:
Nope wait, forgot one important issue! Currently we have to manually run the python file to create the .lark files, do we want to add this to the on-deploy script?

Felienne:
> Nope wait, forgot one important issue! Currently we have to manually run the python file to create the .lark files, do we want to add this to the on-deploy script?

yes, that seems like a good idea!

grammars/keywords-ar.lark
```diff
@@ -1,10 +1,10 @@
-_PRINT:  ( "ـ"* "ق" "ـ"* "و" "ـ"* "ل" "ـ"* | "print") _SPACE?
+_PRINT: ("قول" | "print") _SPACE?
```
Felienne: This overrides the (manual) changes that we made in #2699 (which we don't want)

Since we are now automatically generating anyway... could we add "ـ"* in between all characters automatically in Arabic while generating?

Felienne:
This seems to work, thanks a lot @TiBiBa! The only thing is that we want to allow the tatweel "ـ" in between all letters in the Arabic keywords. We now only do that for print (I added it manually) but now that we generate it, we could "easily" do it for all keywords?

Felienne:
> This seems to work, thanks a lot @TiBiBa! The only thing is that we want to allow the tatweel "ـ" in between all letters in the Arabic keywords. We now only do that for print (I added it manually) but now that we generate it, we could "easily" do it for all keywords?

I just added that, so once we have added this script to the deploy code, it is good to go!

Felienne (changes requested):
Only the call from the deploy script is needed here

TiBiBa:
> Only the call from the deploy script is needed here

This should be fixed!

Felienne (approved):

PR #2733, opened 2022-05-12
Switched order of debugging buttons

### LINKED ISSUE #2732: Reordering debugger buttons
I love these new features! Great job, @fRedelaar! 
I tried them out today and I do have one small remark about the debugger. Wouldn't it make more sense to put the next button on the right and the back button on the left? Like in any type of mediaplayer?

_Originally posted by @MarleenGilsing in https://github.com/Felienne/hedy/discussions/2704#discussioncomment-2736539_

fRedelaar:
**Description**

Switched button order when clicking on the debugger. From  ⏩🔁🛑 to 🔁🛑⏩

**Fixes #2732**

**How to test**
Click on the debugger button, check whether the order of the buttons is: 🔁🛑⏩

### DIFF
--- a/templates/incl-editor-and-output.html
+++ b/templates/incl-editor-and-output.html
@@ -124,9 +124,9 @@
         </div>
         <div id="debug_container">
           <button id="debug_button" class="green-btn" onclick="hedyApp.startDebug()">🐞</button>
-          <button id="debug_continue" class="green-btn hidden" onclick="hedyApp.incrementDebugLine()">⏩</button>
           <button id="debug_restart" class="green-btn hidden" onclick="hedyApp.resetDebug()">🔄</button>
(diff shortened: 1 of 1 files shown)


TiBiBa (commented):
Looks good! One small remark: It might be nice to show the debug buttons at the same spot as the start debug button. This makes the UI feel a bit calmer and users don't have to switch their view. So like this:
Schermafbeelding 2022-05-12 om 13 48 43

Let me know what you think! You can easily implement this by adding the `class="ml-auto"` to the `div id="debug_container">`.

fRedelaar:
Hey @TiBiBa , awesome suggestion! :) Thanks for the tip. I've pushed it right away.

TiBiBa (approved):

PR #2734, opened 2022-05-12
[BUG FIX] Always shows newest users first in admin users overview

### LINKED ISSUE #2641: [UI idea] Sort admin user overview by creation date
**Idea incl level**
Currently we don't sort the admin user overview and simply retrieve "the last 500 users". It would be nice if we indexed the `created` attribute and sort them by the corresponding value. I looked into this but it's a bit more complex than it looks. Currently we have two ways to retrieve multiple entries from our database:

- `get_many()`
- `scan()`

`get_many()` requires a _key_ and also supports a possible related _sort_key_. `scan()` "only" supports `limit` as to limit the amount of entries we want to retrieve. The issue is: We want all entries, with a sort_key but **without** an actual key. This currently does not seem possible so we have to do some re-structuring on either the Dynamo functions or our database to ensure we can retrieve all information and then sort.

TiBiBa:
**Description**
In this PR we make a bucketfix for the issue where the newest users where not automatically shown when loading all users on the admin page. This is due to Dynamo not being sorted automatically on the `created` attribute. The server load increases a bit in this case as previously we only retrieved the first 500 users automatically. Now we always retrieve all users and sort ourselves.

**Fixes**
This PR fixes #2641.

**How to test**
Navigate to the users admin page and verify that the newest users are always shown on top.


Felienne:
This does not seem to work locally, so I guess I will have to merge to alpha?

Felienne (approved):
Works now!

Issue #2735, opened 2022-05-12
Removing variable view for level 1 


fRedelaar:
Also, maybe you can change the variable name in level 1 to echo instead of answer?
Or just diasble the feature in level 1, if changing it is hard.
![image](https://user-images.githubusercontent.com/80678586/168052940-973ae441-afe3-4c59-bef2-e5ccfa5f6e1a.png)

_Originally posted by @MarleenGilsing in https://github.com/Felienne/hedy/discussions/2704#discussioncomment-2736820_

PR #2736, opened 2022-05-12
Removed Variable view in level 1

fRedelaar:
**Description**
Removing the variable view in level by Marleen and Onno's request, since there are no variables yet in level 1.

**Fixes issue #2735 **

**How to test**
Go to level 1, make sure that you are not able to click on the label button. Level 2 and beyond should have the variable view button. 

image


Felienne:
Good work @fRedelaar! 

I was thinking, could we hide the icon/dropdown also in case there are no variables in the program? I think we can do that in the front end entirely (since you need to populate the dropdown?) But if needed, I can easily output an extra boolean from `parse()` to tell you if there are vars in the program.

TiBiBa (changes requested):
I'm still seeing the variable view on level 1, is there something I should update?
Schermafbeelding 2022-05-17 om 16 08 51

fRedelaar:
> I'm still seeing the variable view on level 1, is there something I should update? Schermafbeelding 2022-05-17 om 16 08 51

That is strange! Could you maybe recompile TypeScript?  Or empty cash and do a hard reload? 
`
npx esbuild static/js/index.ts --bundle --sourcemap --minify --target=es2017 --global-name=hedyApp --outfile=static/js/appbundle.js`

fRedelaar:
> 

Hi @Felienne! Nice suggestion, I will look into it :)

TiBiBa (approved):

PR #2744, opened 2022-05-16
[BUG FIX] Removes achievements pop-up from tutorial

### LINKED ISSUE #2721: [BUG] Achievements are handed out in the tutorial
We currently give out achievements in the tutorial might be a bit much in the UI... Can we disable them?

image

TiBiBa:
**Description**
In this PR we fix a small issue where achievements would pop-up for new users when following the tutorial. This can be confusing and we prevent this in this PR by simply removing the `onclick()` function on the run button and replace it by the hard-coded output of the example program. As the user can't alter the tutorial program, this output will always be correct.

**Fixes**
This PR fixes #2721.

**How to test**
Make sure you are logged in with an account that doesn't have the **Getting Started I** achievement and navigate to `/tutorial`. Follow the tutorial and verify that no achievement pops up when running the example program. The run counter of the user will also not be increased.


Felienne:
Works, but we need to look at the conflicts 😅  @TiBiBa

Felienne (approved):

TiBiBa:
> @TiBiBa and some merge conflicts here too :( really annoying these po merge problems, maybe you can show me how to fix them?

I usually solve them by accepting all main changes, re-running the pybabel scripts and manually adding the translations again. Not ideal, but I think the fastest way. Not sure what would be a better approach as all line numbers change as well.

PR #2745, opened 2022-05-16
[CLEAN-UP] Removes quiz iframe structure

TiBiBa:

**Description**

_Changes in detail_

**How to test**

**Checklist**
  
### DIFF
--- a/templates/incl-adventure-tabs.html
+++ b/templates/incl-adventure-tabs.html
@@ -66,9 +66,9 @@
         <!-- Add the tabtarget for the quiz -->
         {% if quiz %}
             <div data-tabtarget="quiz" class="hidden">
-                <h2>{{_('end')}}</h2>
-                <div>{{_('quiz_description')}}</div>
-                <input type="submit" class="green-btn h-3/5 mt-4" value="{{ _('go_to_quiz') }}" onclick="hedyApp.load_quiz ({{level_nr}})">
+                <div id="quiz_container">
+                    {% include 'quiz/startquiz.html' %}
+                </div>
             </div>
         {% endif %}
       </div>
--- a/templates/quiz/startquiz.html
+++ b/templates/quiz/startquiz.html
@@ -1,17 +1,17 @@
-        <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Zilla+Slab:ital,wght@0,300;0,400;0,500;0,600;0,700;1,300;1,400;1,600;1,700&display=swap">
-        <link rel="stylesheet" href="{{static('/css/generated.css')}}">
-        <link rel="stylesheet" href="{{static('/css/additional.css')}}">
-        <link rel="stylesheet" href="{{ static('/css/quiz-css.css') }}">
-        <div class="flex-container">
-            <div class="center-picture p-10"><img class="rounded-full border-green-600 border-8" src="{{ static('/images/quiz_logo.jpg') }}" alt="quiz_logo"/>
-            </div>
-            <div class="title-block p-8 border-t-8 border-blue-600">
-                <h1 class="start-quiz-title font-slab">{{_('start_quiz')}}</h1>
-                <h2 class="start-quiz-subtitle"><i> {{_('level_title')}} {{ level }} </i></h2>
-            </div>
-        </div>
-        <div class="p-10 button-bar border-t-8 border-green-600 ">
-            <button class="green-btn ltr:ml-1 rtl:mr-1" onclick="window.location = '/quiz/quiz_questions/{{level}}/1/1'">
-                {{_('go_to_first_question')}}
-            </button>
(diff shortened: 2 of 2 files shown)


TiBiBa:
Some bugs we are still struggling with:

- Dynamically load answers into either the ace editors or plain text
- Syntax highlighting on the question is broken
- Result page is not working (yet)
- We currently support only one attempt, do we want to change this?
   - If so, we want to track this on the back-end to prevent altering by the user
   - I personally prefer the "one attempt" approach -> its multiple-choice

Felienne:
> We currently support only one attempt, do we want to change this?
> * If so, we want to track this on the back-end to prevent altering by the user
> * I personally prefer the "one attempt" approach -> its multiple-choice

One try for me is fine. What happens when you try the quiz for a 2nd time? Do you see your results page?

Felienne:
> The database structure was so complex I'm still not sure if it is still working as it was. It appears the original structure was to create a `None` placeholder in the database before an answer is given. Then the actual answer is appended to this value inside a list. We preserve this list structure but remove the redundant `None` storing. This does have the disadvantage that I'm not sure if this still works as expected with the stats.
> 
> @boryanagoncharenko maybe you can help me figure this out regarding the quiz stats?

If I recall correctly, @rix0rrr also helped with the quiz/database implementation, can you confirm that it works like above with the None as first value?

boryanagoncharenko:
The quiz stats do not rely on the `None` placeholder. They are quite basic atm - we only track whether the user has started a quiz and their score (if they completed it). The score is taken from the session, not from the DB.

So your change should have no impact on the stats. I also checked your branch and the quiz stats are working as expected.

TiBiBa:
> > We currently support only one attempt, do we want to change this?
> > 
> > * If so, we want to track this on the back-end to prevent altering by the user
> > * I personally prefer the "one attempt" approach -> its multiple-choice
> 
> One try for me is fine. What happens when you try the quiz for a 2nd time? Do you see your results page?

When re-loading the page the quiz is automatically re-started. If I remember correctly we currently keep track of the question the user was at when leaving the quiz, continuing there. We can take a look at implementation this again by checking in the database if there is some incomplete quiz entry for the current level.

TiBiBa:
> The quiz stats do not rely on the `None` placeholder. They are quite basic atm - we only track whether the user has started a quiz and their score (if they completed it). The score is taken from the session, not from the DB.
> 
> So your change should have no impact on the stats. I also checked your branch and the quiz stats are working as expected.

Awesome, thanks for checking!

rix0rrr:
> I guess, given the answer above and the fact that we now only support 1 try, we might also want to rewrite the list to a singular value (can be in a separate PR of course!)

We could, but I'd say why bother.

* If you write scripts to crunch the data you would need to account for both data types (and importantly: **remember** that there are two data types, which will not be obvious if you just look at a recent sampling of the data).
* If you change your mind in the future and want to allow multiple tries again you'd need to change it again

Felienne:
> We could, but I'd say why bother.
> 
> * If you write scripts to crunch the data you would need to account for both data types (and importantly: **remember** that there are two data types, which will not be obvious if you just look at a recent sampling of the data).
> * If you change your mind in the future and want to allow multiple tries again you'd need to change it again

Fair enough, let's leave it as it is then!

Felienne (changes requested):
If possible, I'd like to have 2 tries per question.

TiBiBa:
> If possible, I'd like to have 2 tries per question.

Should be possible! Will require quite some re-writing, but I agree that this is a better implementation from a user perspective.

TiBiBa:
> Nice, this seems to make the quiz a lot smoother!
> 
> But now I have tested, I understand you mean 1 try _of a question_ and not of a quiz! I think there is a lot of value in giving kids another try for a question when they make a mistake, that is a good learning opportunity. But 1 extra try (so 2 in total) is enough I think, 3 is a but much. Would it be complex to keep that part of the functioning the same?
> 
> Also, there is a tiny issue with the visualization of code in quiz test in the answer page:
> 
> image

This should be easy to fix, will take a look!

TiBiBa:
This should work with two attempts!

Felienne (changes requested):
Once we solved the weird text selection issue this should be good to go!

Felienne:
Ehm, the Ace changes seem to have... removed all spaces when code blocks are in answers :) 

image

TiBiBa:
> Ehm, the Ace changes seem to have... removed all spaces when code blocks are in answers :)
> 
> image

Whoops...

TiBiBa:
> Ehm, the Ace changes seem to have... removed all spaces when code blocks are in answers :)
> 
> image

Fixed!

Felienne (approved):
Works now with spaces! 

image

PR #2749, opened 2022-05-16
[FEATURE] Implements achievement rarity feedback on achieving

### LINKED ISSUE #2746: [UI idea] Implement achievement statistics on achieving
**Idea incl level**
It would be nice to return users some data in how rare their achieved achievement is. For example by adding some string "Only x.xx% of the users have this achievement".

TiBiBa:
**Description**
In this PR we implement an exciting addition to the achievement pop-up, rarity! From now on users will see the percentage of users that have already achieved this achievement. Making it even more fun to try to achieve them all. The statistics are calculate once on server start and are updated on each achievements, keeping track of them on the server-side. This has the advantage that it doesn't increase our database load.

**Fixes**
This PR fixes #2746.

**How to test**
Achieve any achievement and verify that there is an extra string on the pop-up containing a percentage of users that already have this achievement. Also verify that it works when you are the first to achieve an achievement.

Schermafbeelding 2022-05-16 om 13 56 08


Felienne:
oooooowww supercool @TiBiBa!!!! This is good for achievement hunters like me!

Felienne:
owno, some conflicts, could you take a look @TiBiBa ?

Felienne (approved):
Works!

TiBiBa:
> owno, some conflicts, could you take a look @TiBiBa ?

Fixed!

PR #2754, opened 2022-05-17
Weblate fix, again

Felienne:
Not sure why #2753 did not do the job, trying again


Felienne:
Ah my bad, I squash merged #2753 and Weblate then (of course) cannot track the commits! All in order now!

PR #2758, opened 2022-05-17
Created an Unparser for my Converter

SlangeF:
**How to test**

To test the code, test files have been added for each level.
 
An example of how to test the code for the first level:
`python runAST.py scriptie_tests/test1.hedy`

After running this command the program will be executed, and the output will be:
- The input code found in the test file
- The resulting AST
- The output code created by the unparser

**Checklist**
  

Felienne:
Thanks @SlangeF!!

Our Python validator does point to a few small issues:

image

Also, can we make your tests more like our tests, in the sense that they can be run with the test runner that we have, in the tests folder, instead of from your code? So a test is structured more or less like this:

```
# setup code (like read a level 1 program, and create the AST)

# call a method (your unparser)

result = ....

# compare the resulting program to an expected program 

```

And maybe the name "scriptie" is a bit too specific, can we generalize that?

Felienne:
Closing this for now (we could later merge smaller versions)

PR #2760, opened 2022-05-18
[IMPROVEMENT] Speed up GitHub code validation

### LINKED ISSUE #2757: [CHORE] Reduce test running time
**Describe the issue**
Tests take a long time, partly because snippets are copied over to non-translated languages and thus tested multiple times. 

**Suggested improvement**
While testing the adventure and level default snippets, we could create a set of snippets per level. Only if the snippet we open to test is *not* in the set, we test it. That could save a lot of time for langs that are only translated partly.

TiBiBa:
**Description**
In this PR we slightly speed up the GitHub code validation by caching the keywords and hashing the code snippets, making sure that we don't run duplicate validations. This doesn't make a very big different (7 minutes in comparison to 8 minutes). But still, every minute is one saved!

**Fixes**
This PR fixes #2757

**How to test**
Nothing to test, just validate that the runtime of the GitHub tests is shorter than usual.


Felienne (approved):
Thanks! Pity it did not speed up the tests but it is better this way in any case!

PR #2762, opened 2022-05-18
Hide variable view when there are no variables

fRedelaar:
**Description**
In PR #2736, Felienne had a suggestion to hide the icon/dropdown also in case there are no variables in the program. This PR will implement that request! 

**Screenshot**

https://user-images.githubusercontent.com/33595265/168983961-e20bb3a5-7722-425e-9daa-b7e2d58030ef.mov

**How to test**
(Recompile the code/Clear cache), and check that the variable view button is hidden until you run a program with variables. Also check whether the variable view dropdown button will be hidden whenever a program is executed without variables. It should work with regular execution and line-by-line execution (🐞 button).


TiBiBa (approved):

TiBiBa (changes requested):
Does not seem to work for me...
Schermafbeelding 2022-05-18 om 13 20 20

TiBiBa (approved):
Never mind, it did after updating the bundles! 

@fRedelaar it would be nice to always have the correct js bundles already in the PR when setting to "Ready for Review". This way the reviewer can directly test the functionality without having to compile themselves.

Felienne:
Thanks for the superquick fix @fRedelaar!!

PR #2766, opened 2022-05-18
[BUG FIX] Gotta Use Them All achievement is achievable again

TiBiBa:
**Description**
The "Gotta Use Them All" achievement was broken due to the commands validation list was not matching the commands list created by the user. This is due to the parser seeing some specific characters such as `*` as commands whereas the command list does not. We fix this by only adding commands to the user commands if there are actually in the list. For users already having non-existing commands saved in their profile we improve the validation to make sure they can still achieve the achievement.

**Fixes**
No related issue.

**How to test**
This takes quite some time but... make sure you use all commands available in Hedy and achieve the "Gotta Use Them All" achievement.


Felienne (approved):
Seem o to me reading the code! But I am a bit too lazy to test this manually haha, so maybe you want to verify this on alpha @TiBiBa?

TiBiBa:
> Seem o to me reading the code! But I am a bit too lazy to test this manually haha, so maybe you want to verify this on alpha @TiBiBa?

Will take a look once merged. (Also pretty excited to see the percentage of users having this achievement!)

PR #2771, opened 2022-05-18
Fix markdown header in ar adventures

Felienne:


Felienne:
This is an interesting PR too for you @Mark-Giesen! Can we somehow force ## to have a space after them?

Mark-Giesen:
I added "## " to the placeholders for the Adventures Component on Weblate. So now these are greyed and more important, you can find the violations. For this language example, open Arabic under Adventures and select "Failing checks: placeholders". There are 25 violations at this moment, the forth is one of these mentioned here. 
If this is what we want, we can extend this to other components.

Felienne:
> I added "## " to the placeholders for the Adventures Component on Weblate. So now these are greyed and more important, you can find the violations. For this language example, open Arabic under Adventures and select "Failing checks: placeholders". There are 25 violations at this moment, the forth is one of these mentioned here. If this is what we want, we can extend this to other components.

Ah great, thanks!! Would be good to extent this indeed to find more issues and warn translators directly

Issue #2783, opened 2022-05-19
[BUG] all_commands() functions doesn't always return all used commands


TiBiBa:
**Describe the bug**
The code parser has a `all_commands()` function that should return all commands used in the code. However, this does not seem to work correct in all cases. Which is problematic for the "Gotta Use Them All" achievement as we use this function to verify if new commands are begin used.

**Paste the Hedy code & level**
To replicate this issue, run the follow code on level 5 and verify that not all used commands are shown:
```
name is ask 'what is your name?'
if name is Hedy print 'nice' else print 'boo!'
```
The `all_commands()` function returns the following list of used commands (missing the `if`!):
```
[Token('RULE', 'ask'), 'else', 'print', 'print']
```

**Expected behavior**
The `all_commands()` function should always return all commands used in the program.

PR #2784, opened 2022-05-19
[BUG FIX] "Gotta Use Them All" achievement is achievable again

TiBiBa:
**Description**
We have to wait for #2783 to be fixed before we can merge this. Otherwise it all seems fixed (and improved in this PR).


TiBiBa:
Will fix this as soon as #2822 is merged!

TiBiBa:
@Felienne it seems I'm still unable to achieve some specific commands. The following are not returned by the `all_commands()` function but are returned when calling the specific commands for each level with `commands_per_level()`:

- at
- from
- in
- range
- times
- to

Felienne:
> @Felienne it seems I'm still unable to achieve some specific commands. The following are not returned by the `all_commands()` function but are returned when calling the specific commands for each level with `commands_per_level()`:
> 
> * at
> * from
> * in
> * range
> * times
> * to
> 
> After a lot of tries I'm stuck with the following commands. The first row is the ones I got and the second row is the ones are all commands needed for the achievements:
> 
> Schermafbeelding 2022-05-30 om 10 06 48

Ah I only looked at the missing `if` I did not verify any others. If you can make some example programs that contain the missing commands, I will have a look later this week!

TiBiBa:
> > @Felienne it seems I'm still unable to achieve some specific commands. The following are not returned by the `all_commands()` function but are returned when calling the specific commands for each level with `commands_per_level()`:
> > 
> > * at
> > * from
> > * in
> > * range
> > * times
> > * to
> > 
> > After a lot of tries I'm stuck with the following commands. The first row is the ones I got and the second row is the ones are all commands needed for the achievements:
> > Schermafbeelding 2022-05-30 om 10 06 48
> 
> Ah I only looked at the missing `if` I did not verify any others. If you can make some example programs that contain the missing commands, I will have a look later this week!

Will do! I also noticed that we need the `color` command but this is not explained in any level introduction or adventure. Not sure how and where we want to introduce this command but might be nice to add this content as well (will make an issue).

TiBiBa:
Hi @Felienne! Below a few short code snippets that should return the missing commands but currently do not:

**Level 3**
```
animals is dog, cat, kangaroo
print animals at random
```
The following code only returns `['is', 'ask', 'random']` but should also return `'at'` as the prefix for random.

```
animals is dog, cat, kangaroo
add penguin to animals
remove cat from animals
print animals at random
```
The following code only returns `['is', 'add', 'remove', 'print', 'random']` but should also return `'from'`.

**Note**
The keyword `in` is correctly found by the function(), I just didn't user the correct code earlier. Note that is the `in` keyword for `if string in list`. The `in` for `range` is not detected as such (but that might be intentional?).

**Level 7**
```
repeat 3 times print 'Hedy is fun!'
```
The following program only returns `['repeat', 'print']` but should also return `'times'`.

**Level 11**
```
for counter in range 1 to 5
    print counter
```
The following program only returns `['for', 'print']` but should also return `range` and `to`.

Felienne:
> Hi @Felienne! Below a few short code snippets that should return the missing commands but currently do not:
> 
> **Level 3**
> 
> ```
> animals is dog, cat, kangaroo
> print animals at random
> ```
> 
> The following code only returns `['is', 'ask', 'random']` but should also return `'at'` as the prefix for random.
> 
> ```
> animals is dog, cat, kangaroo
> add penguin to animals
> remove cat from animals
> print animals at random
> ```
> 
> The following code only returns `['is', 'add', 'remove', 'print', 'random']` but should also return `'from'`.
> 
> **Note** The keyword `in` is correctly found by the function, I just didn't euse the correct code earlier. Note that is the `in` keyword for `if string in list`. The `in` for `range` is not detected as such (but that might be intentional?).
> 
> **Level 7**
> 
> ```
> repeat 3 times print 'Hedy is fun!'
> ```
> 
> The following program only returns `['repeat', 'print']` but should also return `'times'`.
> 
> **Level 11**
> 
> ```
> for counter in range 1 to 5
>     print counter
> ```
> 
> The following program only returns `['for', 'print']` but should also return `range` and `to`.

Thanks for all the clarifications @TiBiBa! I think in fact we should prune the list of all commands, since some of these really go together! `at` is not really a separate thing but always used with `random`, and `from` is always used with `remove`, so it is not a new concept.

Felienne:
I was not really sure where the list of all_commands originates from, maybe from here:

https://github.com/Felienne/hedy/blob/4664dc9cc3321aa1b6b26a1913796fe63c0a82fc/hedy.py#L118

If you can prune the list I think we merge this!

TiBiBa:
> I was not really sure where the list of all_commands originates from, maybe from here:
> 
> https://github.com/Felienne/hedy/blob/4664dc9cc3321aa1b6b26a1913796fe63c0a82fc/hedy.py#L118
> 
> If you can prune the list I think we merge this!

Do we want to prune in `hedy.py`? As suggested by the inline comment the list is used for closest command finding. Otherwise I can manually prune in the Achievements class, but we should remember that is manual and might have to alter it later on.

Felienne:
> Do we want to prune in `hedy.py`? As suggested by the inline comment the list is used for closest command finding. Otherwise I can manually prune in the Achievements class, but we should remember that is manual and might have to alter it later on.

Yeah we don't want to change the original list since indeed that is used for other purposes too. Maybe instead of actual pruning, we just make a manual list of all commands for the achievement? A bit risky I admit when we add commands but we can leave a command in hedy.py to also update the achievements one.

TiBiBa:
Finally got it!
Schermafbeelding 2022-06-14 om 14 53 30

Felienne:
Haha, so much work and yet 50% of users have the achievement :D

website/achievements.py
```diff
@@ -21,7 +21,10 @@ def get_all_commands(self):
         for i in range(1, hedy.HEDY_MAX_LEVEL+1):
             for command in hedy.commands_per_level.get(i):
                 commands.append(command)
-        return set(commands)
+        commands = set(commands)
+        # We manually remove the redundant commands, these are stored in the commands_per_level but not used by the parser
```
Felienne, line 25:
```suggestion
        # We manually remove the redundant commands, these are stored in the commands_per_level but not returned by the parser via AllCommands()
```

Felienne (approved):

Issue #2775, opened 2022-05-19
[BUG] Debug button doesn't align nicely when "edit code" button is shown


TiBiBa:
**Describe the bug**
On the view program page, when the edit code button is shown the debug button has no margin. Small fix, but would be nice to improve!

**Add a screenshot (optional)**
![Schermafbeelding 2022-05-19 om 10 43 47](https://user-images.githubusercontent.com/38155424/169252175-918f8d7f-252f-44fc-a906-4eccf6fa4898.png)

**Expected behavior**
Make sure there is a small margin between all buttons.

fRedelaar:
Wow, good find @TiBiBa  thank you! I will look into it. 

I've also just noticed that the debugger does not work on this page.. So that is a bug on its own. I will create a new issue.

PR #2789, opened 2022-05-19
Adding margin between debugging and edit button

fRedelaar:
**Description**
Added a bit of margin to the debugging and edit button. 

**Fixes issue #2775**

**How to test**
Save & share your code. Go to the link that Hedy provides. Make sure that the is some margin/space between the debug button and the edit button. It should look like:

image


TiBiBa (changes requested):
It would be nice to directly support RTL languages as well (such as Arabic). As the complete site will be turned around `ml-4` should be `mr-4` for RTL languages. You can solve this with Tailwind by adding a prefix to define the page direction. This should work: `ltr:ml-4 rtl:mr-4`.

TiBiBa (approved):

PR #2793, opened 2022-05-20
[IMPROVEMENT] Overall improvements on parsons problem implementation

### LINKED ISSUE #2786: Store parsons problems attempts in database
**Description**
As we now support parson problems on production (really nice @xxxpokemon !) it would be great if we kept track of the attempts of different users. Storing their attempts in the database such as how we store run programs. We do have to think about what is relevant to store as I'm struggling a bit with relevant information. The correct order of the lines is always A, B, C.... Possible information we can store:

- Set user order (A, B, D, C for example)
- Percentage of correct location

Any suggestions on what would be relevant? @MarleenGilsing @Felienne

TiBiBa:
**Description**
In this PR we make several small improvements to the parsons problem implementation to make it more scalable and readable for the future. First of all we split the `get_trimmed_code()` function into one dedicated for "normal" code and one for parsons code. We also verify if there is a mistake in the order, if so: don't parse and only highlight the faulty lines. The prevents weird Hedy correct programs that are not in the correct order of lines as well as difficult to understand error messages.

Lastly we implement another important feature: logging of parsons attempts. Each line order attempt is logged in combination with the level, timestamp and username (or session id). Enabling us to perform some useful data analysis later on.

**Fixes**
This PR fixes #2786.

**How to test**
Verify the following improvements to make sure they are all working:
- When a faulty order is given the program is not parsed
- However, the attempt is logged into the database
- The stop button is correctly reversed into a run button when getting an error


TiBiBa:
We do have to create the table in Dynamo before merging this!

Felienne (changes requested):
Nice work @TiBiBa!!

One small thing, can we disable the debugger in the Parsons mode? Since we can't highlight the lines using it gets a bit confusing.

image

TiBiBa:
Fixed the dynamically hiding/showing of the debug container!

Felienne:
Nice 

> Fixed the dynamically hiding/showing of the debug container!

Great! Have you made the tables in alpha and prod? If so, feel free to merge this!

Felienne (approved):

PR #2798, opened 2022-05-23
Adjust number of groups in regex

thjazi39:
This pr adds an additional constraint on the syntax coloring rules.

Indeed, in order to reduce the risk of bugs related to the number of groups in the regex, I now allow only 2 cases:

- either there is no group in the regex, and the token must be a string
- there are several groups in the regex, and the token must be a list of string (with the same size)

Thus, this case is no longer accepted: 
```
{
     'regex' : '(,)' ,
     'token' : 'keyword'
}
```
Although this is accepted by Ace, it can lead to bugs in some cases

The validate_ruleset function has been modified, and the rules have been adapted accordingly.

**How to test**

Check that the syntax highlighting still works, and that there is nothing broken in the front-end


Felienne:
Thanks @thjazi39! @TiBiBa could you test this? Or is highlighting still broken for you locally?

TiBiBa:
> Thanks @thjazi39! @TiBiBa could you test this? Or is highlighting still broken for you locally?

Highlighting works for me, but only when running the "normale" `/generate-typescript` script, when running it in combination with the `--watch` argument the syntax highlighting breaks. A weird bug and didn't dive into the issue yet, but as I mostly use `--watch` when working on the front-end it's quite annoying.

I assume this PR is unrelated, so will take a look if it still works for me as before!

TiBiBa (approved):
All works for me! I still have to bug as described the comments, but that is unrelated to this PR (would be nice if we have a fix though).

PR #2803, opened 2022-05-24
[BUG FIX] Shows all users on admin users overview

TiBiBa:

**Description**

_Changes in detail_

**How to test**

**Checklist**
  

website/database.py
```diff
@@ -242,14 +242,6 @@ def get_hedy_choices(self):
     def set_program_as_hedy_choice(self, id, favourite):
         PROGRAMS.update({'id': id}, {'hedy_choice': 1 if favourite else None})
 
-    def all_programs_count(self):
```
Felienne, line 245: why are we removing these two functions?

website/admin.py
```diff
@@ -40,11 +40,15 @@ def get_admin_users_page(user):
         keyword_language = None if keyword_language == "null" else keyword_language
 
         filtering = False
-        if substring or start_date or end_date or language or keyword_language or category == "all":
+        if substring or start_date or end_date or language or keyword_language:
             filtering = True
 
         # After hitting 1k users, it'd be wise to add pagination.
```
Felienne, line 46: haha maybe this is the time now that we have more than 4000? Not a bad idea to implement (can be used in other pages also potentially like /explore?

Felienne (changes requested):

website/database.py
TiBiBa: They are no longer used, the user count is already calculated from the length of the users list; reducing the amount of db calls. The programs counts function has never been used in the first place.

Felienne: ok!

website/admin.py
TiBiBa: Good point! Will do this right away.

PR #2809, opened 2022-05-24
[BUG FIX] Fixes quiz lvl 6 question 2

### LINKED ISSUE #2807: [CONTENT] Level 6 question 3 bug
The question shows the wrong characters (dots instead of * + etc) as the answers. Presuming it's a character encoding issue?

TiBiBa:
**Description**
Question 2 of level 6 of the quiz contained some one character answers. Due to using the `|commonmark` parser on the front-end the options are parsed into HTML from the Markdown format. However, for some reason single mathematical symbols are in this case replaced by a black dot marker. To prevent this we create a bit of a bucket-workaround-fix, if and only if the option is only one character: don't use common mark but show as plain text instead. This solves our bug.

**Fixes**
This PR fixes #2807.

**How to test**
Navigate to level 6 -> quiz -> question 2. Verify that the mathematical symbols are shown as expected.

### DIFF
--- a/templates/quiz/feedback.html
+++ b/templates/quiz/feedback.html
@@ -41,7 +41,7 @@
                             <pre class="no-copy-button" level="{{ level_source }}">{{correct_option.option|commonmark}}</pre>
                         </div>
                     {% else %}
-                        <div class="pt-8 px-2 mx-auto mb-2 text-lg">{{ correct_option.option }}</div>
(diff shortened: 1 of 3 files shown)


Felienne:
This works! Unrelated to this PR, but I just noticed not that the phrasing is a bit weird I think:

image

Maybe we should say "what do you use for addition?" Since 'plus' is not really a sign?

Felienne:
> This works! Unrelated to this PR, but I just noticed not that the phrasing is a bit weird I think:
> 
> image
> 
> Maybe we should say "what do you use for addition?" Since 'plus' is not really a sign?

Never mind I made this fix in Weblate do we don't get merge conflicts, so this is now good to go!

Felienne (approved):

PR #2810, opened 2022-05-24
[UI IMPROVEMENT] Starts turtle at bottom of screen going up

### LINKED ISSUE #2730: Change starting position of turtle?
image

This is what happens when the turtle is told to move 200 steps forward. Wouldn't it make more sense if the turtle started at the bottom of the screen (centre) and forward would mean upwards (like a car on a map).

TiBiBa:
**Description**
In this PR we change the starting position and direction of the turtle when making a drawing program. Currently the turtle starts (ommenabij) in the middle of the screen facing to the right. This is changed to started at the middle-bottom of the screen facing up, like car in a game.

**Edit**
To better work with #2826 we decided to start in the middle of the screen facing up instead of at the bottom facing up. So effectively the only change is the arrow pointing another direction at the start of the program.

**Fixes**
This PR fixes #2730.

**How to test**
Navigate to any Hedy level and use the Turtle. Verify that is always start at the middle-bottom of the screen.


Felienne (changes requested):
Works but we need to fix a few more examples

TiBiBa:
> Works but we need to fix a few more examples

Will look into this today! Will take some time as we have to update all adventures for all languages, but should be a lot of copy/paste as most are not translated (yet).

Felienne:
> Will take some time as we have to update all adventures for all languages, 

Haha yeah that is why I did not do it myself 😅

TiBiBa:
We should reset the adventures / commands as we start in the middle again all original snippets work fine.

PR #2812, opened 2022-05-24
[PERFORMANCE IMPROVEMENT] Stores code prefix in front-end instead of parse

### LINKED ISSUE #2811: [PERFORMANCE] Send the code prefix to the front-end only once
**Description**
Currently we send the code prefix to the front-end on each parse request. Either the `TURTLE_PREFIX_CODE` when a turtle is use d in the program or the `NORMAL_PREFIX_CODE` when textual coding is done as well. This isn't very efficient as this code is always identical: creating unnecessary server-load. We can simply stope this in the front-end and only let the user load this once.

TiBiBa:
**Description**
In this PR we make a performance improvement by handling the prefix adding of Python code on the front-end instead of the back-end. This way the same prefix is not sent back from the server each time but instead the prefixes are stored within `app.ts` and appended to the code dependent on if there is a turtle or not. This should reduce traffic a bit.

**Fixes**
This PR fixes #2811.

**How to test**
Everything should still work as expected. Verify that both the turtle and non-turtle programs are still working.


Felienne:
I added the new prefix code and the next feature also seems to work still:

image

Felienne (approved):

PR #2814, opened 2022-05-24
[FEATURE] Adds language filtering on explore page

### LINKED ISSUE #2813: [UI idea] Allow filtering on UI language in /explore
**Idea incl level**
Since https://github.com/Felienne/hedy/pull/2808 all programs are shown in the local keywords, awesome! But I think it would be great if we can also filter on the UI language of the creator, cause a program with a lot of Spanish text is still not so very useful to me. Do we have that information? If so, can we add it as a filtering option?

TiBiBa:
**Description**
We already support level and adventure filtering on the explore page but in this PR add the option to filter on language. Only returning the programs that are stored with a user with a profile language set to the selected language.

**Fixes**
This PR fixes #2813.

**How to test**
Make sure you are logged in, navigate to the `/explore` page and verify that language filtering is working.


Felienne (changes requested):
ok, I know this is minuscule but language has a lowercase letter but Level and Adventure have an uppercase:

image

I wanted to fix it myself but I could not easily find the location, sorry!

Felienne:
Also, if I am being very picky anyway, can we make the order: Adventure, Level and then Language, since the default name is also Adventure and then Level:

image

TiBiBa:
> Also, if I am being very picky anyway, can we make the order: Adventure, Level and then Language, since the default name is also Adventure and then Level:
> 
> image

Fixed!

Felienne:
> This is due to either Babel not being compiled or having your profile set to a non-dutch/English language. The placeholder key is "language" whereas the dutch and English translations are "Taal" and "Language".

Ah this issue again, my bad, sorry! But would it be possible to somehow label those placeholder clearly, so I know I al looking at one (I can imagine other projects would like such a feature too)?

Felienne:
> > This is due to either Babel not being compiled or having your profile set to a non-dutch/English language. The placeholder key is "language" whereas the dutch and English translations are "Taal" and "Language".
> 
> Ah this issue again, my bad, sorry! But would it be possible to somehow label those placeholder clearly, so I know I al looking at one (I can imagine other projects would like such a feature too)?

If not, maybe we can have our own convention, like starting them with a few underscores or something?

Felienne (approved):

PR #2822, opened 2022-05-26
Fix ask and add if to ifelse in AllCommands()

### LINKED ISSUE #2783: [BUG] all_commands() functions doesn't always return all used commands

Felienne:
Fixes #2783

**How to test**

I added a new test that shows if is now properly detected in level 5 and up


Felienne:
Hi @TiBiBa!

With this you should be able to complete #2784!

TiBiBa (approved):
Awesome! All works for me!

PR #2826, opened 2022-05-30
Add option to generate and download dst file for embroidery machines

Felienne:
WIP to output turtle not just to the screen but also to embroidery machines!

480AC386-C755-4016-AE83-9AAF2D2E8BF0
!


Felienne:
Hi @TiBiBa!

I managed to come quite a long way with this in one morning, including actually running generated code on a machine. There are a few tasks left that you could pick up (in this PR should be fine)

1. I made a new /dst route, but we want a button in the UI somewhere that a user can click to download the file (shown only when the turtle canvas is visible I think?)
2. There is an issue with caching. If you generate more dst files in one run of the server, you keep getting the first one. Only a restart of the server fixes this but that seems a bit heavy haha. I tried adding `cache_timeout=0` but that did not seem to help
3. The performance of the current solution is not great, cause when the boolean `dst` is enabled, on every parse, a dst is generated. We might want to only generate the dst file when the download button is clicked? We probably also want to delete/refrain from generating the `time.sleep` output from the turtle?
4. If you have a `ask/input` in the code, the current method does not work cause the exec blocks waiting for input. It would be great if we can save the input so that when generating the input, we know what to embroider. If that is too hard, we should give a warning that programs with an input cannot be embroidered.

Good luck and let me know if you need help/clarification!

Felienne:
Tip: You can upload dst files here to see if they are what you expect: https://stitchpad.io/

TiBiBa:
Will look into this! I do think that making nice drawing will be a lot harder when merging #2810 as the turtle will no longer start in the middel of the screen. We might want to add functionality for pen up/down as well?

Felienne:
> Will look into this! I do think that making nice drawing will be a lot harder when merging #2810 as the turtle will no longer start in the middel of the screen. We might want to add functionality for pen up/down as well?

Ah yeah that is true... Can we maybe change #2810 to start upwards pointing but still in the center? That is a bit of "best possible for both cases"?

Felienne:
And yes, let's at one point add pen up/down but it is a bit complex for lower levels so we want to have something useable without that in any case

TiBiBa:
> > Will look into this! I do think that making nice drawing will be a lot harder when merging #2810 as the turtle will no longer start in the middel of the screen. We might want to add functionality for pen up/down as well?
> 
> Ah yeah that is true... Can we maybe change #2810 to start upwards pointing but still in the center? That is a bit of "best possible for both cases"?

This sounds good to me! Will first adjust this in #2810, then look into this PR.

TiBiBa:
> > > Will look into this! I do think that making nice drawing will be a lot harder when merging #2810 as the turtle will no longer start in the middel of the screen. We might want to add functionality for pen up/down as well?
> > 
> > 
> > Ah yeah that is true... Can we maybe change #2810 to start upwards pointing but still in the center? That is a bit of "best possible for both cases"?
> 
> This sounds good to me! Will first adjust this in #2810, then look into this PR.

#2810 is merged! Will work on this right away.

TiBiBa:
> 4. If you have a `ask/input` in the code, the current method does not work cause the exec blocks waiting for input. It would be great if we can save the input so that when generating the input, we know what to embroider. If that is too hard, we should give a warning that programs with an input cannot be embroidered.

This seems quite complex, but will take a look! Would be nice if Skulpt supported the output of the runned code (it might). But will have to look into this.

TiBiBa:
Hi @Felienne! This all seems to work with one (maybe two) downsides. I will explain the current approach below:

When a turtle program is done running we show the "save" button, giving users the option to save their turtle program as .dst-file. On click a POST is made to the server, generating the .dst file and saving with a random 12-char string. This filename is returned as directly requested with a GET. The file is downloaded and after the GET request automatically deleted. This has two downsides:

- The filenames are not really user-friendly
- It somehow we generate the file but don't download it: it stays on the server

I also don't have a solution (yet) for using user input on the file generating. It might be best to leave this out-of-scope for now as I think this will be quite hard to implement. We do have to think about when and how we want to show the save button as I can imagine a lot of people simply pressing it to figure out what it does. Maybe only show it for logged in users or add some explanatory text?

It would be great if you could verify that the generated (and downloaded) .dst files are indeed valid. Will try to figure this out myself as well.

TiBiBa:
> Hi @Felienne! This all seems to work with one (maybe two) downsides. I will explain the current approach below:
> 
> When a turtle program is done running we show the "save" button, giving users the option to save their turtle program as .dst-file. On click a POST is made to the server, generating the .dst file and saving with a random 12-char string. This filename is returned as directly requested with a GET. The file is downloaded and after the GET request automatically deleted. This has two downsides:
> 
> * The filenames are not really user-friendly
> * It somehow we generate the file but don't download it: it stays on the server
> 
> I also don't have a solution (yet) for using user input on the file generating. It might be best to leave this out-of-scope for now as I think this will be quite hard to implement. We do have to think about when and how we want to show the save button as I can imagine a lot of people simply pressing it to figure out what it does. Maybe only show it for logged in users or add some explanatory text?
> 
> It would be great if you could verify that the generated (and downloaded) .dst files are indeed valid. Will try to figure this out myself as well.

Just verified that the created files are indeed valid!

Felienne:
hmmmm somehow we now get an error? Could you have a look @TiBiBa?

Felienne:
I think I fixed the error, but I now get an error while downloading:

image

TiBiBa:
> I think I fixed the error, but I now get an error while downloading:
> 
> image

Is the file created locally? It might be an issue that the `dst_files` folder is missing not sure if I added that folder to this PR.

TiBiBa:
@Felienne the issue should be fixed with the latest commit!

Felienne:
ok, great!! Indeed it works now, I can download a file but the image seems to be turned?

Output:

image

dst in stichpad:

image

File:
[NgrqBlCbhh4U.dst.zip](https://github.com/Felienne/hedy/files/8890939/NgrqBlCbhh4U.dst.zip)

Felienne:
(this is probably a really tiny issue but still)

TiBiBa:
> ok, great!! Indeed it works now, I can download a file but the image seems to be turned 90 degrees?
> 
> Output:
> 
> image
> 
> dst in stichpad:
> 
> image
> 
> File: [NgrqBlCbhh4U.dst.zip](https://github.com/Felienne/hedy/files/8890939/NgrqBlCbhh4U.dst.zip)

Weird issue, not sure what could cause this. We do turn the turtle once with the code prefix, but that shouldn't be the issue. However, as we now store the prefix on the front-end the back-end version (we use for this code) might no longer be up-to-date, will look into this!

Anotehr issue that I just recalled: we don't support `random` as the code is parsed once again when pressing the download button the drawing will be different from the one the user sees when running the code. Might not be a big issue, but something we should keep in mind.

Felienne:
> Weird issue, not sure what could cause this. We do turn the turtle once with the code prefix, but that shouldn't be the issue. However, as we now store the prefix on the front-end the back-end version (we use for this code) might no longer be up-to-date, will look into this!

Yeah it is weird, I think in my version that did not happen?

> Anotehr issue that I just recalled: we don't support `random` as the code is parsed once again when pressing the download button the drawing will be different from the one the user sees when running the code. Might not be a big issue, but something we should keep in mind.

Yeah that is ok, I guess there is nothing (easy) to be done about that.

TiBiBa:
> > Weird issue, not sure what could cause this. We do turn the turtle once with the code prefix, but that shouldn't be the issue. However, as we now store the prefix on the front-end the back-end version (we use for this code) might no longer be up-to-date, will look into this!
> 
> Yeah it is weird, I think in my version that did not happen?
> 
> > Anotehr issue that I just recalled: we don't support `random` as the code is parsed once again when pressing the download button the drawing will be different from the one the user sees when running the code. Might not be a big issue, but something we should keep in mind.
> 
> Yeah that is ok, I guess there is nothing (easy) to be done about that.

Fixed the 90 degrees issue!
Schermafbeelding 2022-06-13 om 14 35 34

Felienne:
Haha can you approve @TiBiBa? I can't approve my own PR!

TiBiBa:
> Haha can you approve @TiBiBa? I can't approve my own PR!

Will do! I'm still a bit in doubt about the current UI implementation as all users will see the save button (when using the turtle) without any context. I can imagine a lot of users being confused when pressing it and receiving a .dst file. Not sure what would be a good solution for this, but I can imagine users not understanding the functionality.

TiBiBa (approved):

PR #2833, opened 2022-05-31
[BUG FIX] Parsons attempts are correctly stored in database

### LINKED ISSUE #2832: [BUG] Parsons entries in database are overwritten
**Describe the bug**
Due to not having a unique identifier parsons answers in the database are automatically overwritten. It would be nice if we improve this by adding some `attemptID` similar to how we store quiz answers. When we are already working on the database side of this feature it would also be nice to add a boolean `correct` to enable easier analysis later on.

**Expected behavior**
Each Parsons attempt is stored in a dedicated row.

TiBiBa:
**Description**
Currently Parsons attempts in the database are overwritten for each user on each attempt. This is due to the `username` field being the partition_key. We fix this by adding a random 12 char string as id to each entry, setting this as the partition key. We then add `username` as the index key to still enable filtering on username for analysis later on. Lastly we add the `correct` field being either 0 or 1 indicating if the answer was correct.

**Fixes**
This PR fixes #2832.

**How to test**
Deploy to Alpha, verify that new attempts are no longer overwriting the old ones (in the Dynamo database).


Felienne:
Still works in the local version, can you verify on alpha, @TiBiBa?

Felienne (approved):

TiBiBa:
> Still works in the local version, can you verify on alpha, @TiBiBa?

Will take a look!

PR #2838, opened 2022-06-02
[BUG FIX] Parsons attempts no longer overwritten in database

TiBiBa:
**Description**
In this PR we attempt to fix the issue where Parsons attempts are being overwritten in the database when the same user (or session.id) solves a new parsons problem. It is still unclear to me why this happens as locally the entries in the database are appended fine instead of overwriting the old ones. Similarly we use the same structure for storing new programs, which also works fine.

@rix0rrr Do you maybe have a clue what is going wrong? We use the Dynamo `create()` function to create a new entry in the table, accompanied with an `attemptID`. For if the username is the same the previous entry is being overwritten. In this PR we attempt to solve this by removing the username as an indexKey, but I don't know why this could be an issue.

**Fixes**
No related issue.

**How to test**
Deploy to Alpha, verify that new Parsons attempts by the same user are stored correctly.


boryanagoncharenko:
The problem seems to be a mismatch between the definition of the table and the actual table created in AWS. It seems that the intention was to have a table with partition key 'id' and index 'username'. This is what the code defines. But in the AWS console, I see that the partition key is 'username' and there is no index at all.

I am very surprised this works, as the API usually requires you to specify the primary key. But apparently this is not the case for put: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.put_item

TiBiBa:
> The problem seems to be a mismatch between the definition of the table and the actual table created in AWS. It seems that the intention was to have a table with partition key 'id' and index 'username'. This is what the code defines. But in the AWS console, I see that the partition key is 'username' and there is no index at all.
> 
> I am very surprised this works, as the AWS API usually requires you to specify the primary key. But apparently this is not the case for put: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.put_item
> 
> Note that the dynamo.py file contains 2 separate storages (a client and an implementation to be precise): AwsDynamoStorage and MemoryStorage. The former is used on all stages and the latter - only when working locally. It should not surprise you that something would work locally but not when deployed - it is a quite different code path. In your case, everything worked as expected locally because the Parsons table was created on the fly with the right partition key.

Thanks for the clarification! I just updated the table on AWS and it indeed fixed the issue: this PR can be closed.

boryanagoncharenko:
I just want to point out that there is still a mismatch: the definition says there is an index 'username' while the AWS table does not have an index. This is not an issue atm because the index is not used but if you forget the index is not in AWS you end up in a similar situation in the future.

boryanagoncharenko (approved):

boryanagoncharenko:
I approved your PR in case you do not want to add the index. And if you want to add the index, you can close it again.

PR #2841, opened 2022-06-02
[BUG FIX] Tutorial works again and improve e2e tests

### LINKED ISSUE #2840: [BUG] Tutorial is broken due to missing 'quiz_questions' variable
**Describe the bug**
Currently the tutorial is broken due to a missing variable. It is unclear how this went wrong, but we have to look into a fix. Together with this it would be nice if we add the `/tutorial` route to the e2e page tests to verify that nothing breaks the page. This should have been done earlier on as this bug could have been easily prevented.

TiBiBa:
**Description**
In this PR we fix an issue where the tutorial would crash due to a missing variable that was expected by the .html template. However, this mistake should never have been merged in the first place. This is our own mistake by not updating the e2e tests to also verify that the `/tutorial` route is still valid in all languages. To prevent this we also fix the e2e tests for the tutorial.

**Fixes**
This PR fixes #2840.

**How to test**
Login and navigate to `/tutorial`. Verify that it is working again.

### DIFF
--- a/templates/quiz.html
+++ b/templates/quiz.html
@@ -15,13 +15,15 @@
 </div>
 <div id="quiz_container" class="hidden">
     <div class="arrow-steps clearfix">
-        {% for i in range(1, quiz_questions + 1)%}
-           {% if i == 1 %}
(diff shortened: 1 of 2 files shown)


Mark-Giesen (approved):
easy fix!

PR #2842, opened 2022-06-02
[QoL] Adds --watch mode to Tailwind script

### LINKED ISSUE #2839: Enable --watch on generate Tailwind script
**Description**
Currently we support the `--watch` argument when generating Typescript with the corresponding script. When using this argument the script keeps running and when detecting a change in relevant files, compiling the .js files again. Which is really nice! However, Tailwind support the same argument but we don't include this in the generate Tailwind script. Would be nice if we add this as well. Especially with #2776 in mind where we will be doing a lot of generating of new classes.

TiBiBa:
**Description**
In this PR we make a minor QoL fix, adding `--watch` to the Tailwind script. Enabling us to run the Tailwind script in watch-mode and automatically generate a new .css file on each relevant file change. We already supported this for Typescript so it is a nice addition to create support for Tailwind as well.

**Fixes**
This PR fixes #2839.

**How to test**
Verify that the script still workings by simply running `./build-tools/heroku/tailwind/generate-css`. Then run the same script but add the `--watch` argument. Verify that the script stays "stuck" and only updates on every relevant file change (template, ts, styles.css).


build-tools/heroku/tailwind/generate-css
```diff
@@ -5,6 +5,11 @@ cd $scriptdir
 
 staticdir=../../../static
 
+if [[ "${1:-}" == "--watch" ]]; then
+  echo "👀 Running Tailwind compilation in watch mode. 👀"
+  npx tailwindcss build -i styles.css -o $staticdir/css/generated.css --watch
+fi
+
 npx tailwindcss build -i styles.css -o $staticdir/css/generated.css
```
Mark-Giesen, line 13: shouldn't this line then be in an else block?

Mark-Giesen: you'll want the minify to run as well I assume?

TiBiBa: That's not really a problem as the script is run on deploy as well, still minimizing the .css file.

boryanagoncharenko (changes requested):
The script does what is expected. Indeed the --watch argument on line 10 executes until the script is running and when the npx command is terminated, the script is terminated too. 

However, please consider changing the if statement to an if-else statement, not because it will change the output but because it will make it easier for future authors to read and understand this script and, therefore, this will improve the maintainability of the script. Mark's comments are the best argument to my hypothesis. Additionally, please note that the sibling script (generate-typescript) uses an if-else statement too even thought it does need to.

Mark-Giesen (approved):

PR #2849, opened 2022-06-02
[UI Improvements] Simplifies class overview UI

TiBiBa:

**Description**

_Changes in detail_

**How to test**

**Checklist**
  

MarleenGilsing:
It all still works and it looks much better with less words and buttons! Great!

I do think we could still move the 'add students' to the class interface, to keep all of it together and create a nice workflow for the teachers. But I don't know how hard this is to create, so you take that into consideration.

And if I'm gonna be super nitpicky, maybe we can change the dates of last login to 'normal' dates (we don't need to know the seconds haha). But there is obviously not a real need for that change.

TiBiBa:
> I do think we could still move the 'add students' to the class interface, to keep all of it together and create a nice workflow for the teachers. But I don't know how hard this is to create, so you take that into consideration.

I'm exactly sure what you mean. That we remove the "create multiple accounts" option from the teachers overview page and move it into the class overview page? As this is conceptually a different task I would like to keep them separated, but I can imagine that creating the account within the class makes more sense from a user perspective. @Felienne what do you think?

> And if I'm gonna be super nitpicky, maybe we can change the dates of last login to 'normal' dates (we don't need to know the seconds haha). But there is obviously not a real need for that change.

We can for example remove the seconds? Will take a look to make this in a bit more readable format.

MarleenGilsing:
> I'm exactly sure what you mean. That we remove the "create multiple accounts" option from the teachers overview page and move it into the class overview page? As this is conceptually a different task I would like to keep them separated, but I can imagine that creating the account within the class makes more sense from a user perspective. @Felienne what do you think?

Yes that's what I mean. Because a teacher has to create a class first and then add their students. So it makes sense to put it on the class page. And mainly because the other options of adding your students (the invite link and the 'invite students' button) are there as well. Else you can add students on two different pages. 
So maybe we could make a button on the class page to add students and then show the 3 options how to do it.

TiBiBa:
Hi @MarleenGilsing! I just improved the datetime showing making it automatically dependent on the language of choice:

Schermafbeelding 2022-06-09 om 14 51 51
Schermafbeelding 2022-06-09 om 14 52 18

MarleenGilsing (approved):

MarleenGilsing:
> Hi @MarleenGilsing! I just improved the datetime showing making it automatically dependent on the language of choice:
> 
> **Dutch**: Schermafbeelding 2022-06-09 om 14 51 51
> 
> **French**: Schermafbeelding 2022-06-09 om 14 52 18

Neat!

Issue #2855, opened 2022-06-06
[CONTENT] Level 7 has double example code in a few languages


Felienne:
**Describe the issue**
At least in Bulgarian and Spanish:

<img width="1058" alt="image" src="https://user-images.githubusercontent.com/1003685/172177903-0c79d965-9e6a-4bc3-a8c6-26eda244e102.png">

Felienne:
Closed by #2856

Felienne:
Ow no sorry there were a few left!

Felienne:
Ok I think I checked all languages and I did not find any more double ones!

Was very easy, by the way, to check that with Weblate! https://hosted.weblate.org/translate/hedy/adventures/en/?offset=1&q=Level+7%21+Level+7%21&sort_by=-priority%2Cposition&checksum=

PR #2856, opened 2022-06-06
Remove double example codes

Felienne:
Start towards #2855


Felienne:
Ok, this is turning out to be a lot more work than I thought...! @TiBiBa, could you maybe take a look? There are a few languages that still have code in the `story_text` field (translated) and then also in the example_code in En, for example in Italian:

image

I suspect these were not migrated because the story_text was already translated so we did not know what to merge. Could you take a look at the remaining ones? In the final commit, I changed the tests so that it outputs what still is to be done :)

Felienne:
> Hi @Felienne , will take a look! Not sure how to solve this conflict as I'm not familiar with this code. Can you solve the conflict? Than I will fix the resting double examples.

I will have a look at the conflict!

Felienne:
> Hi @Felienne , will take a look! Not sure how to solve this conflict as I'm not familiar with this code. Can you solve the conflict? Than I will fix the resting double examples.

I will have a look at the conflict!

Felienne:
I have time today to make this one in order @TiBiBa!

PR #2859, opened 2022-06-07
[CHORE] Clean up quiz css

### LINKED ISSUE #2797: [CLEAN-UP] Make quiz completely dependent on tailwind classes for styling
**Describe the issue**
We currently have some manual written CSS for the quiz styling. This is not necessary as we can achieve the same styling with Tailwind classes. Might also be a good first issue to get familiar with our code base and Tailwind in general!

TiBiBa:
**Description**
In this PR we remove the dependency on separate .css files for the quiz styling. We re-write most classes, remove some redundant usage and use Tailwind where possible. Dedicated styling is performed in `styles.css` but otherwise within the `quiz.html` template.

**Fixes**
This PR fixes #2797.

**How to test**
This is only a clean-up task. The quiz styling should remain (almost) the same. We did switch some hard-coded hex-colors to Tailwind classes, so minor differences in colors might occur.


Felienne (approved):
Seems to still work!

image

PR #2865, opened 2022-06-09
highlighting of digits translations

### LINKED ISSUE #2835: [BUG] Non-Latin numbers are not colored green
**Describe the bug**
Latin is green:

image

Arabic (for example) is not:

image

Would be nice if we can make that equal!

thjazi39:
Added coloring of digits according to the language translation.

**Description**

A new regex has been added to the file highlighting/definition.py to support the translation of the numbers.

This translation is then used in the definition of the numbers in the file highlighting/rules_list.py.

Concerning the recovery of the translation of the numbers, it could not be done like the classical commands, because we want a translation for all the digits.

Automated tests have been added for this new feature

fix #2835 

**How to test**

check that the front-end is not broken, and that the coloring of the Arabic numbers works

image

image

here are some arabic numbers (this will make the tests easier)
```
    0 : ٠            5 : ٥
    1 : ١            6 : ٦
    2 : ٢            7 : ٧
    3 : ٣            8 : ٨
    4 : ٤            9 : ٩
  283 : ٢٨٣        142 : ١٤٢
  255 : ٢٥٥        269 : ٢٦٩
  217 : ٢١٧        208 : ٢٠٨
  511 : ٥١١        587 : ٥٨٧
  423 : ٤٢٣        368 : ٣٦٨
22.52 : ٢٢.٥٢    28.66 : ٢٨.٦٦
38.72 : ٣٨.٧٢     30.2 : ٣٠.٢
48.66 : ٤٨.٦٦    26.38 : ٢٦.٣٨
30.08 : ٣٠.٠٨    48.54 : ٤٨.٥٤
 40.4 : ٤٠.٤     59.54 : ٥٩.٥٤
```


highlighting/generate-rules-highlighting.py
```diff
@@ -84,18 +86,22 @@ def get_translations(KEYWORDS_PATH, KEYWORDS_PATTERN):
 
         yaml_file = yaml.safe_load(keywords_file)
         tmp[language_code] = {}
+        digits[language_code] = []
         for k in yaml_file:
-            if k in TRANSLATE_WORD:
-                tmp[language_code][k] = str(yaml_file[k])
+            if str(k) in TRANSLATE_WORD:
```
Felienne, line 91: I am not entirely sure what this does, Do we need to exclude numbers form being translated...?

Can we put it in a more meaningful function or add a comment?

thjazi39: The use of digits is different from that of commands.
Digits are used within brackets and that's why I made a special case.

image
I have added some comments and split into functions

PR #2874, opened 2022-06-09
Creates new parsons problems for level 1 - 8

### LINKED ISSUE #2831: [Feature] Multiple parson's problems for a level in UI
We already support Parson's in a level in the yaml, but we will have to make content and also make it clear in the UI that there are more puzzles to do (and maybe also change the DB storing of tries?) 

Maybe something like in the quiz where there is a question 1, 2 etc.

> As well as changing the structure so we can support multiple parsons adventures per level. 

_Originally posted by @TiBiBa in https://github.com/Felienne/hedy/issues/2761#issuecomment-1131606090_

MarleenGilsing:
The inputscreen for these parsons problems still need to be changed. @TiBiBa or @fRedelaar can you do that?

**Fixes**
This PR fixes #2831


Felienne:
> Awesome! Will look into how we can best merge this into the UI as it requires some re-writing on both the back and front-end. We do want to differentiate between naming as they are now both called "Dragging". The ids `dragging` and `dragging2` are fine as these are not visible to the user.
> 
> Maybe we want to step away from the adventure tabs? Introduce some dedicated tab for the parsons? I'm not sure what would be the best approach but as the amount of adventures and the parsons keep growing the current UI is no longer suitable.

Can we do it a bit like quizzes, where you start and then you have a question 1 and a question 2? but then instead it is Parson's 1 and Parson's 2 etc.... I feel that is logical and also compact? (I made an issue along those lines -> #2831)

TiBiBa:
> > Awesome! Will look into how we can best merge this into the UI as it requires some re-writing on both the back and front-end. We do want to differentiate between naming as they are now both called "Dragging". The ids `dragging` and `dragging2` are fine as these are not visible to the user.
> 
> > 
> 
> > Maybe we want to step away from the adventure tabs? Introduce some dedicated tab for the parsons? I'm not sure what would be the best approach but as the amount of adventures and the parsons keep growing the current UI is no longer suitable.
> 
> 
> 
> Can we do it a bit like quizzes, where you start and then you have a question 1 and a question 2? but then instead it is Parson's 1 and Parson's 2 etc.... I feel that is logical and also compact? (I made an issue along those lines -> #2831)

Will look into this! We might want to sort the by difficulty than as well. @MarleenGilsing shall we discuss this in the near future?

TiBiBa:
Will work on this today to also solve #2831 in one go.

TiBiBa:
I'm struggling a bit with how we can nicely display the different parsons problems for each level. Currently I've implemented the following structure but am not really happy with it:

Schermafbeelding 2022-06-14 om 12 11 59

What do you think? @Felienne @MarleenGilsing

Felienne:
> I'm struggling a bit with how we can nicely display the different parsons problems for each level. Currently I've implemented the following structure but am not really happy with it:
> 
> Schermafbeelding 2022-06-14 om 12 11 59
> 
> What do you think? @Felienne @MarleenGilsing

I was thinking something like the quiz for consistency. with numbers on the top?

image

TiBiBa:
> > I'm struggling a bit with how we can nicely display the different parsons problems for each level. Currently I've implemented the following structure but am not really happy with it:
> > Schermafbeelding 2022-06-14 om 12 11 59
> > What do you think? @Felienne @MarleenGilsing
> 
> I was thinking something like the quiz for consistency, with numbers on the top?
> 
> image

So do we want a fixed order like the quiz? The structure is getting a bit complex as we have to find a way to return the lines of a previous exercise to the original state once switching back as well as dynamically changing the amount of empty lines visible. I think we have two options:

- Use a quiz like structure: re-write the whole back-end, retrieve each question separately from the server
- Use the current approach: load all exercises once, do some magic on the front-end to fix the UI dependent on the amount of lines

Felienne:
> So do we want a fixed order like the quiz?

If possible, yes, I would prefer that so we can have them go easier to harder.

Felienne:
> * Use a quiz like structure: re-write the whole back-end, retrieve each question separately from the server
> * Use the current approach: load all exercises once, do some magic on the front-end to fix the UI dependent on the amount of lines

I guess option 2 is preferable for performance reasons? And I guess we need to do a bit of magic anyway since the first puzzle also has a certain amount of lines?

TiBiBa:
> > * Use a quiz like structure: re-write the whole back-end, retrieve each question separately from the server
> > * Use the current approach: load all exercises once, do some magic on the front-end to fix the UI dependent on the amount of lines
> 
> I guess option 2 is preferable for performance reasons? And I guess we need to do a bit of magic anyway since the first puzzle also has a certain amount of lines?

I think as well. The complex issue with option 2 is that we way to find a way to "remember" the lines as currently when dragging/dropping them we simply copy the line content from one to another. However, when going to the next exercise we want the answer lines to be empty again, removing the lines already there. If the users want to go back to the previous exercise we have no clue what lines to re-present to the user. When retrieving the lines on each question we can prevent this.

We do have to re-structure the YAML files, but I think this is the case in both implementations. As they are currently structure by exercise (just like the adventures) where the should be re-written to be structured by level (like the quizzes).

Felienne:
> If the users want to go back to the previous exercise we have no clue what lines to re-present to the user.

We could also not allow going back? Or simply make the lines empty when a user goes back? We don't remember quiz answers either, do we?

Felienne:
> As they are currently structure by exercise (just like the adventures) where the should be re-written to be structured by level (like the quizzes).

yeah that makes sense! let's align with the quizzes as much as possible

TiBiBa:
> > If the users want to go back to the previous exercise we have no clue what lines to re-present to the user.
> 
> We could also not allow going back? Or simply make the lines empty when a user goes back? We don't remember quiz answers either, do we?

The issue is that we also don't remember the exercise lines. As they are removed from the exercise when being dragged/dropped on the answer lines window. We simply switch them with the empty spot. Not allowing to go back might be a solution, in this case I propose to copy the quiz structure and adjust for the parsons.

TiBiBa:
Notes to self:

- Re-write YAML from exercise to level structure
- Re-write database implementation to also keep track of exercise number like this: `{'level': 1, 'exercise': 1}`
- Create dedicated .ts file for the Parsons
- Load exercises dynamically using AJAX

Think to consider:
- Do we want some type of score like with quizzes?
- How many attempts do we want to support?
- Do we want to keep the output window or create one dynamically for each exercise (like the quiz)

TiBiBa:
Current idea:

- Use implementation very similar to quiz
- Show progress bar when `len(exercises) > 1`
- Don't show scores -> always show next / prev level buttons
- Update database storing

TiBiBa:
Hi @Felienne! I think this PR is almost ready for review. We currently support the following structure:

- On retrieving `/hedy` we check if there is a Parsons and how many exercises
- When selecting the "Parsons" tab we dynamically retrieve the exercise
- If there is a next exercise, show the "next exercise" button after finishing the current exercise

The following changes and limitation are made as well:
- We only support a maximum of 10 code lines -> this is hard-coded
    - We might want to improve this by on runtime checking the max amount of line in all exercises of the current level
- We update the database function to also store the exercise number on each attempt

**Screenshots**
Schermafbeelding 2022-06-15 om 13 48 58
Schermafbeelding 2022-06-15 om 13 48 16
Schermafbeelding 2022-06-15 om 13 48 31

TiBiBa:
Just notice a big mistake: we don't update the story text (yet), will look into this!

TiBiBa:
Hi @MarleenGilsing! I can't request a review from you as this is your PR, but can you take a look if everything works as expected?

Felienne:
Not a suer big issue but the output window is not emptied when you go to a new exercise:

image

Felienne:
Also, on some cases, all lines and the explanation are dragging (but that is not caused by this PR) 

Screenshot 2022-06-18 at 15 47 29

Felienne:
I will merge this now to prevent conflicts in the texts later on and to make sure translators are not translating the old version.

Felienne (approved):
Not 100% in order but I want to unlock weblate

PR #2879, opened 2022-06-10
[FEATURE] Enables hiding / showing of parsons through class customizations

### LINKED ISSUE #2878: [UI idea] Enable teachers to show/hide parsons problems in class customizations
**Idea incl level**
It would be nice if we add the option to hide / show parsons with class customizations such as how we handle this with adventures and the quiz.

TiBiBa:
**Description**
In this PR we add the showing and hiding of the parsons problems to class customizations. Currently teachers have no option to hide the parsons problems other than hiding the whole level. We fix this by adding the option to "other settings" in the customizations page.

**Fixes**
This PR fixes #2878.

**How to test**
Make sure you are a teacher and have a class. Navigate to customizations and verify the new "hide parsons" option in the other settings section. Verify that this is working.


Felienne (approved):
Works!

image

PR #2880, opened 2022-06-10
[UI IMPROVEMENT] Centralizes workflow for adding students to classes

### LINKED ISSUE #2869: [UI idea] Create dedicated section for adding students to class
**Idea incl level**
Currently our "adding students to a class" workflow is a bit of a mess. Teachers have three options:

- Share the class join link
- Invite students by username
- Create multiple accounts and add them

The first two options are shown next to each other on the class overview page. However, the multiple accounts creation option is shown as a dedicated section on the for teachers page. As suggested by @MarleenGilsing it might be nice to group these within the class overview page with the following structure:

**Class overview page**
- Class name
- Table with students
- Some buttons
    - Add students
        - Share join link
        - Invite students by username
        - Create accounts and add to class

TiBiBa:
**Description**
In this PR we continue with the clean-up of the class overview page to make it less crowded with a more logical workflow. The multiple accounts creation section is removed from the for-teachers page and instead placed within the class overview page. We create a dedicated "add students" section, which displays multiple options when clicked.

**Fixes**
This PR fixes #2869.

**How to test**
Navigate to the class overview page, verify that all three options are working as expected.

**New view**
Schermafbeelding 2022-06-10 om 12 23 03
Schermafbeelding 2022-06-10 om 12 23 10


Felienne (changes requested):
Works like a charm but... the add page misses a "back to class" button, could we add that?

image

Felienne:
> Works like a charm but... the add page misses a "back to class" button, could we add that?
> 
> image

Like the customize page does have:

image

TiBiBa:
> > Works like a charm but... the add page misses a "back to class" button, could we add that?
> > image
> 
> Like the customize page does have:
> 
> image

Will look into this! As it is a re-direct we might have to add some hidden url to know the page we where coming from (might also be able to get it from the browser history, which will be even nicer).

TiBiBa:
Fixed!
Schermafbeelding 2022-06-21 om 10 57 16

Felienne (approved):
Great, all set now!

PR #2883, opened 2022-06-12
[BUG] Handle non-numerical values for repeat

### LINKED ISSUE #2790: [BUG] Ask gives a runtime error in level 8 with repeat and text
**Describe the bug**
Entering text with ask gives a runtime error in level 8 (and maybe more)

image

```
hoeveel is ask 'hoeveel?'
repeat hoeveel times
    print 'Hello everyone'
    print 'This is all repeated ' hoeveel ' times'
```

Enter f and get the error.

ThomasKluiters:
This commit will introduce a new step to using repeats; we will assign
the amount of times we are repeating the statements to a temporary
variable. First, we try to convert the 'times' value to a number. If
this fails, we set it to 0. Effectively communicating; "we can only
count numbers".

Fixes #2790.

I'm curious if this is how we would like to solve the issue. We could also throw a very specific error instead of repeating zero times; "Hey, I don't understand {times}!".

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -233,7 +233,7 @@ def get_suggestions_for_language(lang, level):
     # if we allow multiple keyword languages:
     en_commands = get_list_keywords(commands_per_level[level], 'en')
     en_lang_commands = list(set(en_commands + lang_commands))
-            
+
     return en_lang_commands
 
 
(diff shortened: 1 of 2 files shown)


ThomasKluiters:
I'll happily fix the failing tests, if you guys think the solution is appropriate :)

ThomasKluiters:
image

@Felienne I have changed it to behave as you suggested!

Felienne:
Closing this for now @ThomasKluiters, I think it is too far behind to still be useful! If you are up dat date with main again, let me know!

PR #2885, opened 2022-06-13
Makes sure to always generate grammar before starting tests

### LINKED ISSUE #2882: [BUG] Run grammar generation before tests
In https://github.com/Felienne/hedy/pull/2877/commits/d9b928df806a11aa1930abd3e0a0b3902cf02c73 I manually generated the grammars because without that, the tests failed. Can we change our tests script to also always generate the grammars, or we will have this issue each time when someone updates the keywords and also adds content (to optimize, we could also store the hashes of the keyword files and only generate the grammar when there is a change?)

TiBiBa:
**Description**
In this PR we make a small improvement to the testing pipeline by first creating the grammar files before running all tests. This to ensure that all tests still pass in the case someone changing something in the grammar as well as the content without manually generating the grammar files.

**Fixes**
This PR fixes #2882 

**How to test**
GitHub should automatically take care of this but to validate you can manually run the `validate-tests` script in `build-tools/github` folder.


Felienne (approved):

Felienne:
Ah that was an easy fix! Thanks.

PR #2887, opened 2022-06-13
[CLEAN-UP] Removes string constant from adventures YAMLs

### LINKED ISSUE #2886: [CONTENT] "Example Hedy code" is duplicated a lot in the yamls
All adventure yamls contain the text "Example Hedy code" above code blocks, which used to make sense before we have separate example_code blocks but now is a lot of duplication that will need to be translated each time adding extra work, and can also lead to differences between translators, f.e. in the Hungarian there are three different versions.

Can we: 
1. Move the text to the po file once
2. Put it into the yamls automatically
3. Remove the text from all yamls

It is a bit of manual work, but ultimately it will look better if it is the same everywhere and less translation work!

TiBiBa:
**Fixes**
This PR fixes #2886.


TiBiBa:
@Felienne this one is ready for review! We do have to wait for Weblate to fix the .po files for all non-dutch/English languages as currently the placeholders don't make any sense. But Weblate should overwrite these with the English one as soon as this PR is merged.

Felienne:
> @Felienne this one is ready for review! We do have to wait for Weblate to fix the .po files for all non-dutch/English languages as currently the placeholders don't make any sense. But Weblate should overwrite these with the English one as soon as this PR is merged.

Ah I see, it is weird now! But... can't we copy in the translations from the files that we already have? In many cases I think we do know the translation? f.e. for Hindi: `## उदाहरण कोड`, for French: `## Exemple de code`. It is a bit of manual work but going to the site, changing the language and then copy pasting 1 header (even though I know, it might not always be the best one) should be doable?

Felienne (changes requested):
Can we add the existing translations to the po files?

Felienne:
Hi @TiBiBa!

I fear this will have a lot of conflicts with #2856, so you can leave this one for now, I will ping when I am ready!

Felienne:
ok.... I think I made it all look nice again!

Felienne (approved):

PR #2889, opened 2022-06-13
add regular prefix also when using turtle

### LINKED ISSUE #2888: [BUG] Arabic variable does not work with turn
**Describe the bug**
This code gives a runtime error:

```
hoek is ٩٠
turn hoek
```

image

It is superweird, because the resulting code is this:

```
hoek = '٩٠'
trtl = hoek
try:
  trtl = int(trtl)
except ValueError:
  raise Exception(f'While running your program the command <span class="command-highlighted">turn</span> received the value <span class="command-highlighted">{trtl}</span> which is not allowed. Try changing the value to a number.')
t.right(min(600, trtl) if trtl > 0 else max(-600, trtl))
```

Which just runs for me, also in Skulpt (tried it on: https://skulpt.org/) Also if we include our own `int` in the prefix. What then goes wrong here? Can you investigate @TiBiBa?

Felienne:
Fixes #2888

**How to test**

I added a test (that passes) but that does not really prove anything since the error was in Skulpt. So to test, verify that the broken code in #2888 now passes and than everything else also still works.


TiBiBa (approved):
Works!

PR #2891, opened 2022-06-13
[BUG FIX] All filtering options on admin users page work again

### LINKED ISSUE #2890: [BUG] Filtering is broken on admin users page
**Describe the bug**
All filtering seems broken on the admin users page.

TiBiBa:
**Description**
In this PR we fix the admin filtering options that where broken due to two new features recently introduced:
- Accounts without mail addresses
- Localized timestamps on the front-end

We didn't take these into account resulting in a crash when trying to filtering on a mail substring as well as a crash when attempting to filter on a date.

**Fixes**
This PR fixes #2890.

**How to test**
Make sure you are an admin (if not, set the `IS_ADMIN` environment variable to your username) and navigate to `/admin/users`. Verify that all filtering options are working again and the UI updates as expected.


Felienne (approved):
Fixed!

image

PR #2892, opened 2022-06-13
[FIX] Verify if keyword needs indentation for all languages

### LINKED ISSUE #2007: indentation needs to be checked for all languages

TiBiBa:
**Description**
Currently we use a fixed list to check if a keywords needs indentation on the next line. This doesn't make sense anymore as keywords are generated dynamically based on the translations provided by contributors. In this PR we fix the static check function by making it dependent on a dynamically created list containing the keywords that need indentation for all supported keyword languages.

**Fixes**
This PR fixes #2007.

**How to test**
Nothing to test, if the tests pass it is ready to merge!


hedy.py
```diff
@@ -33,6 +35,12 @@
 # Python keywords need hashing when used as var names
 reserved_words = ['and', 'except', 'lambda', 'with', 'as', 'finally', 'nonlocal', 'while', 'assert', 'False', 'None', 'yield', 'break', 'for', 'not', 'class', 'from', 'or', 'continue', 'global', 'pass', 'def', 'if', 'raise', 'del', 'import', 'return', 'elif', 'in', 'True', 'else', 'is', 'try']
 
+# Let's retrieve all keywords dynamically from the cached KEYWORDS dictionary
+indent_keywords = []
+for lang, keywords in KEYWORDS.items():
+    for keyword in ['if', 'for', 'repeat']:
```
Felienne, line 41: We might miss `while` here (I know we did not have it in the original but I think that was a mistake!)

TiBiBa: Whoops! Will add.

Felienne: So that was my bad, originally :)

Felienne (approved):
Turns out we missed else too! Added it with tests.

PR #2894, opened 2022-06-14
[BUG FIX] Quiz answer 4 no longer always selected

### LINKED ISSUE #2860: [BUG] Answer 4 is always set as active on new quiz question
**Describe the bug**
A very unique bug as I have no clue how this happens. Currently when switching questions in the quiz answer 4 is automatically marked as selected, which doesn't make any sense. 

Schermafbeelding 2022-06-07 om 13 50 13

**Expected behavior**
When a question is loaded no answers should be selected.

TiBiBa:
**Description**
We fix a weird bug where the answer container of answer 4 would always be selected when switching a quiz question. The cause is still unknown, but by removing the `active` attribute in the answer loading loop instead of globally the issue seems to be fixed.

**Fixes**
This PR fixes #2860.

**How to test**
Navigate to quiz level 1 question 1 and answer, verify that when going to the next question answer 4 is no longer automatically selected.


Felienne:
I have not managed to reproduce the original error but this does not select anything so it seems good!

Felienne (approved):

PR #2899, opened 2022-06-15
add test for the final state during syntactic highlighting test

thjazi39:
added a test for the final state during syntactic highlighting.

While developing the solution to the over-coloring problem, I realized that one case was covered by the old tests that had been generated (`test_WEBSITE.py`) but not by the parametric tests.

The uncovered case concerns the command sequences. We have to check that after highlighting a line, the state that will be given for the beginning of the next line will be fixed.

In the same way, some tests can now be done by starting the line with a particular state (like just after a `if`, where it is impossible to get another `if`)

**How to test**
Verify that the new tests pass and that it does not block other tests


Felienne (approved):
Thanks!!

PR #2906, opened 2022-06-16
[UI IMPROVEMENT] Improve cheatsheet UI

### LINKED ISSUE #2904: [CONTENT] example not used in cheat sheet?
#2903 made me realize that we don't really use `example` in the commands yaml's anymore (not in the dropdown cheat sheets), shall we just remove it, or do we plan to use that for the paper cheat sheets?

TiBiBa:
**Description**
In this PR we improve the UI of the cheatsheets, making a better differentiation between the different commands.

**Fixes**
No related issue (partly discusses and fixes #2904).

**How to test**
Navigate too `/cheatsheet/<level>` and verify that the generated cheatsheet looks as expected and works.

**New view**
Schermafbeelding 2022-06-16 om 10 19 31


Felienne (changes requested):
While checking the Arabic demo I found out that the Ace is not r2l in the cheat sheets:

image

Whoopsie :) 

Not caused by this PR but would be nice to fix.

TiBiBa:
> While checking the Arabic demo I found out that the Ace is not r2l in the cheat sheets:
> 
> image
> 
> Whoopsie :)
> 
> Not caused by this PR but would be nice to fix.

Weird, this should automatically be caught; will look into this!

TiBiBa:
Fixed!
Schermafbeelding 2022-06-21 om 09 25 21

static/js/app.ts
```diff
@@ -78,7 +78,7 @@ def convert_numerals(alphabet, number):
 
 
   // Set const value to determine the current page direction -> useful for ace editor settings
-  const dir = $("#main_container").attr("dir");
+  const dir = $("body").attr("dir");
```
Felienne, line 81: Just because I am curious... why does this fix rtl? Does body have a certain setting so that it works?

Felienne (approved):

PR #2908, opened 2022-06-16
Tests for incomplete code

thjazi39:
Currently, syntactic highlighting and testing is only concerned with highlighting on finished code.

However, during user input, the code will be incomplete most of the time, and highlighting on code during input may mislead the user.

This pr therefore adds a test phase where the codes are tested on incomplete cases. In concrete terms, this means testing the highlighting on the sub-codes, adding the test words one by one.

For example, on the following test :

```python
def test_list_remove(self, level):
        self.assert_highlighted_chr(
            "remove allergies from flavors",
            "KKKKKK TTTTTTTTT KKKK TTTTTTT",
            level=level, lang="en")
```
In addition to the classic test it will also test the following pairs of tests :

```
remove
KKKKKK

remove allergies
KKKKKK TTTTTTTTT

remove allergies from
KKKKKK TTTTTTTTT KKKK
```

This ensures that the highlighting will be consistent for the user on intermediate forms.

**Adaptation**

The addition of these intermediary tests allowed to detect a small bug concerning the highlighting of the list manipulation functions at level 3 :
image

And that is corrected in this PR, and we now get :
image

**How to test**

For the tests : check that the addition of these intermediary tests work and do not block the following tests

For the addition of the syntactic highlighting modification at level 3: Check that it doesn't break the front-end, and that it works on the intermediate forms


Felienne (approved):

Felienne:
Thanks for this addition @thjazi39. We take the stance here that incomplete code like `remove cat from ` still will be highlighted. I don't disagree but I could also see how we would not do that (since it will throw an error) 

I think it would be good to document our design decisions somewhere, including this one. Could you add this to the docs? (potentially part of this is in your thesis too?)

PR #2923, opened 2022-06-19
Fixes #2917 - Improves recognisability of the language switch menus

### LINKED ISSUE #2917: [UI idea] Improve language dropdown
**Language dropdown**
The language dropdown is hard to find if you do not speak the selected language, due to lack of an identifying item.

I also noticed that the dropdown is not keyboard selectable.

arjank:
**Description**
Adds the globe icon to the language selector in the main menu as well as in the profile.
Also changes the deprecated '.keyup()' to ".on('keyup'" and removed the button in link on the main page. There's no reason for it to be both.

**How to test**
Checkout and view in browser

**Checklist**
  

Felienne:
Nice, thanks @arjank!!

This is a tiny issue but the globe seems to be closer to the text that the other icons:

image

image

Could we make the space between the globe and the text a bit larger so it is more consistent (If I'd know how to, I'd do it myself but I don't :) )

templates/incl-menubar.html
```diff
@@ -38,7 +38,7 @@
     {% if not username %}
       <div class="dropdown inline-block relative {% if not username %}ltr:ml-auto rtl:mr-auto{% endif %} py-2 z-20">
           <button class="bg-blue-400 text-white font-semibold py-1 px-2 rounded inline-flex items-center">
-            <span class="ltr:mr-1 rtl:ml-1">{{ current_language().sym }}</span>
+            <i class="fas fa-fw fa-globe"></i> <span class="ltr:mr-1 rtl:ml-1">{{ current_language().sym }}</span>
```
TiBiBa, line 41: The horizontal distance is now always one whitespace, we can automatically align this nicely horizontally by creating a surrounding `div` with the classes `flex flex-row gap-2`.

templates/incl-menubar.html
```diff
@@ -71,7 +71,7 @@
 </nav>
 <script src="{{static('/vendor/jquery.min.js')}}" type="text/javascript" crossorigin="anonymous"></script>
 <script>
-    $("#search_language").keyup(function() {
+    $("#search_language").on('keyup', function() {
```
TiBiBa, line 74: Awesome, good catch!

templates/main-page.html
```diff
@@ -11,7 +11,7 @@ <h1 class="font-bold font-slab text-white text-6xl text-shadow-md tracking-wide"
                 <h2 class="font-sans font-light text-white text-shadow-md tracking-wide my-1">{{_('main_subtitle')}}</h2>
             </div>
             <div class="flex-none">
-                <a href="/hedy"><button class="green-btn text-white px-8 py-4">{{_('try_it')}}</button></a>
+                <a href="/hedy" class="green-btn text-white px-8 py-4 no-underline">{{_('try_it')}}</a>
```
TiBiBa, line 14: Could we re-write this to a button with an `onclick`? I find it confusing to style an `a` element as a button and behave like a button but it isn't an actual button. This is what we use in most places and it would be nice to be a bit consistent (however, it wasn't correct your change either).

TiBiBa (changes requested):

Felienne:
Removing ready label cause I just put it there to test if mergify was still running.

templates/main-page.html
arjank: I think I can do that. It would be better too, because keyboard users might be tempted to use the space bar to active a button, but that doesn't work with links.

templates/incl-menubar.html
Felienne: Ah yeah the globe seems to "float" a bit now vertically? I am not sure why...

TiBiBa: Adding the `items-center` class shoud do the trick!

Felienne:
> This looks great! Schermafbeelding 2022-06-20 om 19 25 02

Yes!

> I'm still a bit in doubt on the added value of the icon on the profile settings page. As the settings will already be in the user selected language they should already be able to understand the options. Also currently the icon is only shown on updating personal settings, but not on signup. Can we either remove the icon or make this consistent? Schermafbeelding 2022-06-20 om 19 24 49

I think I agree that is it not really needed and not really consistent if we use it on the profile page.

arjank:
> This looks great! Schermafbeelding 2022-06-20 om 19 25 02

Yay!

> I'm still a bit in doubt on the added value of the icon on the profile settings page. As the settings will already be in the user selected language they should already be able to understand the options. Also currently the icon is only shown on updating personal settings, but not on signup. Can we either remove the icon or make this consistent? Schermafbeelding 2022-06-20 om 19 24 49

My reasoning was that it would be useful on the profile settings page too, for example in case a student changes the language settings for another student (or the teacher). But it's easy to revert this.

TiBiBa (approved):

PR #2929, opened 2022-06-21
[PERFORMANCE] Cache all public programs once on server start

### LINKED ISSUE #2861: [PERFORMANCE] /Explore filtering can take (very) long
**Description**
It seems that the explore filtering takes very long to load. We have to look into this if we can improve this somehow. One solution might be to cache all programs, also minimizing the database load.

TiBiBa:
**Description**
In this PR we improve the performance of the explore page by caching all explore programs once on server start. This is not ideal, but with the rapidly growing amount of programs we simply can't handle all unique filtering requests on runtime. We can improve this in the future by using some cron job and call this each x seconds, but it should be fine for now.
**Fixes**
This PR fixes #2861.

**How to test**
All should work as expected, however when a program is deleted after server start it will still remain in the explore page. We can fix this with the future work cron job idea.


Felienne:
This appears not to break anything but let's test well on alpha before deploying!!

Felienne (approved):

PR #2931, opened 2022-06-21
[FEATURE] Adds highscores page

### LINKED ISSUE #2317: [UI idea] Create an achievements high-scores page
**Idea incl level**
It might be nice to have some sort of high-scores page to compare your current amount of achievements against other users. It would even be nicer if some achievements would be worth more points than other ones. As this doesn't in any way contribute to Hedy as a programming language or the functionality it is more of a "nice to have" as it might contribute to the community like website we are slowly becoming.

TiBiBa:
**Description**
In this PR we introduce a new and exciting feature: Highscores! Always wondered how much achievements you have in comparison with other users? Now you know! We create a dedicated highscore page, based on data found on the public profiles table. This table is expanded with the `country` and `achievements_count` value to reduce the amount of DB calls needed on request.

**Fixes**
This PR fixes #2317.

**How to test**
Make sure you are logged in and notice the new "Highscores" option on the profile dropdown menu in the menu bar. Verify that it is all working as expected and that the usernames are click-able to redirect to the public profiles. Also verify that the filtering for world, country and class work as expected. Filtering on class should be hidden when not currently in a class.

Schermafbeelding 2022-06-23 om 17 10 13
Schermafbeelding 2022-06-23 om 17 13 30


TiBiBa:
Note to self:

We do have to add some e2e testing to validate all functionality. It would be best to add this into this PR as well, but a separate issue should be a problem either.

TiBiBa:
> Note to self:
> 
> We do have to add some e2e testing to validate all functionality. It would be best to add this into this PR as well, but a separate issue should be a problem either.

All tests are added! Should be ready for merge once they all pass.

TiBiBa:
Forgot: We have to handle users without a country!

TiBiBa:
Now also works with users without a set country (as this is optional):
Schermafbeelding 2022-06-24 om 15 38 32

Felienne:
> With this in mind we also have to keep in mind the following issue:
> 
> * What do we do if a user changes their country?
>   
>   * We have to update the public profile country as well!

Did we figure out a solution to this? (I also think the case is very rare, so if it does not cause crashes, I'd be ok with not updating it)

TiBiBa:
> > With this in mind we also have to keep in mind the following issue:
> > 
> > * What do we do if a user changes their country?
> >   
> >   * We have to update the public profile country as well!
> 
> Did we figure out a solution to this? (I also think the case is very rare, so if it does not cause crashes, I'd be ok with not updating it)

Yes! I've implemented a check for all edge cases. Having no country or not being in a class should both work fine now. (I've also added some e2e tests to validate this).

Felienne (approved):

PR #2933, opened 2022-06-21
[BUG FIX] Drag drop parsons works in Firefox

### LINKED ISSUE #2922: [PARSONS] Drag & drop not working in Firefox
Drag & drop is not working in Firefox (at least on Linux). The drop is registered, but you get redirected immediately. The url you get redirected to contains one letter and depends on the item you have dragged.  

I also tested Chromium on the same computer, that  is working as expected.

TiBiBa:
**Description**
In this PR we fix an issue where the Parsons problems where not solvable on Firefox because on the drag/drop a new url would be opened. We fix this by adding a `e.preventdefault()` to the drag/drop handler. We also fix a typo in an English Parsons exercise of level 2.

**Fixes**
This PR fixes #2922.

**How to test**
Verify that the Parsons drag/drop works on Firefox.


TiBiBa:
@arjank can you verify that this is working for you as well? For some reason I can't select you as a reviewer!

Felienne (approved):

PR #2935, opened 2022-06-21
Adds alt attributes to all images

### LINKED ISSUE #2926: [UI idea] Alt texts for images
**Use alt texts for images**

We should use `alt` texts for images. And while that might be hard in some cases, we should at least provide the Hedy Logo with an `alt` text.

TiBiBa:
**Description**
We add the `alt` attribute to all images to provide an alternative when for some reason the image can't load or to give a user who is unable to see the image still provide with useful information.

**Fixes**
This PR fixes #2926.

**How to test**
Nothing to test, verify that each image has an `alt` attribute

### DIFF
--- a/templates/achievements.html
+++ b/templates/achievements.html
@@ -33,8 +33,8 @@
                         <div class="w-40 rounded-lg ltr:mr-4 rtl:ml-4 mt-4 relative bg-blue-200">
                             <div class="relative h-40">
                                 <div class="flex items-center justify-center h-40">
-                                    <img src="{{static('/images/achievement.png')}}" class="px-2">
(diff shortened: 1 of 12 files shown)


arjank:
Should the alt texts be translatable?

arjank:
> > Should the alt texts be translatable?
> 
> I do like that idea, but let's do it separately, because this PR already adds value and translation might take a bit of time, since I am not sure if we can (easily) do it.

Makes sense :smile:

Mark-Giesen:
> @Mark-Giesen do you know if we can translate alt tags for images in Weblate in an easy way?

I think enclosing the texts in _() will make them available for translation in Weblate.

PR #2944, opened 2022-06-23
[LANGUAGE] Converts Arabic numerals in comparison level 6 to 14

### LINKED ISSUE #2943: [BUG] Calculator adventure in Levels 6 and 9 counts answers entered in Arabic Hindi numbers as wrong
**Describe the bug**
Calculator adventure in Levels 6 and 9 counts answers entered in Arabic Hindi numbers as wrong. 

Level 6 example
The questions asks to enter the result of 7*6. Although 42 is entered (in Arabic Hindi numerals) ...
2022-06-22_15h53_52
We can see that the program incorrectly identifies this as a wrong answer  
2022-06-22_15h54_04

**Expected behavior**
Numbers in Latin or Arabic Hindi digits should be accepted as correct answers. 

- What computer you are using (Windows, Mac, Linux?) Windows
- What browser you were using (Chrome, Edge, Safari) MS Edge

Felienne:
**Description**

`equality_check` used to cast to str which made a comparison like `٢` to `2` inequal. We now use convert_numeral in places where we used to do str. These should work the same for int (2 -> '2') and for string (identity operation: 'x' -> 'x') but casts non-Latin numerals to latin numbers so the comparison works.

**Fixes #2943**

**How to test**

Tests should pass.

**Checklist**
  
### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -8,15 +8,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2022-07-04 11:35+0200\n"
-"POT-Creation-Date: 2022-07-04 13:06+0200\n"
+"POT-Creation-Date: 2022-07-06 10:00+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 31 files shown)


Felienne:
> Could you remove remove the debug print that I slipped at line 2304 of line hedy.py 😓?

NP! I was wondering where the extra output was coming from (but I was too lazy to truly investigate haha)

PR #2946, opened 2022-06-23
[BUG FIX] Lines in goal parsons code boxes no longer appear indented

### LINKED ISSUE #2938: [BUG] Parson's lines always look indented
When dragging a line down, it seems they become "indented" or at least we have a lot of space upfront? I think that can be confusing in higher levels?

image

TiBiBa:
**Description**
In this PR we fix an issue with the lines in the goal parsons code boxes to appear indented. This was due to "spaghetti code". Let me try to explain this weird bug:

For support RTL languages we had to manually adjust the ace editor gutter to be dependent on this page direction change as this is not natively supported. We therefore did some manually styling, changing the gutter dependent on the page direction, also fixing the line number direction and margins and so on. The issue was: We styled this as follows:

```
#editor-area .ace_gutter {
    width: 32px !important;
    position: absolute;
}
```

In natural language: For all elements fix the class `ace_gutter` found within the element with the id `editor-area` force the width to be 32px, make this the most important styling there is and set the position absolute. Skipping to last week: we implemented the parsons within the editor template, but also within this `editor-area` element! Meaning that **all** the ace editors within this area will take this styling (also our parsons boxes!). As we also remove the gutter from parsons goal boxes the following issue occurred:

- We fix a width to 32px
- But we also remove the display of the gutter

This combined resulted in a hidden merging of 32px, which was hard to find as the margin itself was also hidden. The issue is fixed by specifying the hard-coded changes to the gutter to only be for specific the `editor` not the whole area. We also slightly clean-up the `parsons.html` template by replacing repetitive code with a for-loop.

**Fixes**
This PR fixes #2938.

**How to test**
Navigate to any Parsons exercise and verify that the goal parsons box is no longer indented.


Felienne (approved):
Works!

image

Felienne:
And thanks for the extensive explanations @TiBiBa! Sounds like this was a tricky one to figure out!

PR #2951, opened 2022-06-23
All image alt attributes are translatable

### LINKED ISSUE #2945: [UI idea] Localize alt texts for images
**Idea incl level**
Currently the `alt` attributes for images are hard-coded in English. It would be nice to replace these with Babel placeholders so they can be translated by contributors through Weblate.

TiBiBa:
**Description**
In this PR we change the hard-coded `alt` attributes on the images with Babel placeholders. Enabling contributors to translate them in their language of choice. We also add translations for English and Dutch right away.

**Fixes**
This PR fixes #2945.

**How to test**
Nothing to test as it's a back-end task. However, you could verify that when on either English or Dutch; the alt attributes on images are shown correctly in the browser.


TiBiBa:
Hi @arjank! For some reason I'm still unable to select you as a reviewer. @Felienne Do you maybe know what can cause this?

Felienne:
Yeah I think because @arjank is not a collaborator (yet...!) Just sent out an invite.

arjank (approved):
Looks good, the images in templates/translating.html are not translatable, but neither is that page, so that's ok.

PR #2953, opened 2022-06-23
Adds option to quiz question to show output container

### LINKED ISSUE #2941: [BUG] Sometimes quizzes ask kids to match textual output but then the output is also highlighted
Output in quizzes should not be highlighted, f.e. in level 8:

image

Can we make a block that is output but not code? If that is hard, I am also ok with changing the content here although it might be hard to communicate to translators what words not to use.

TiBiBa:
**Description**
In this PR we make a differentiation between `code` and `output` snippets in the quiz question. Currently we only support either a textual question or a textual question in combination with a code snippet. We expand on this by also support a textual question in combination with an output snippet. The YAML is adjusted to support the `output` attribute as wel as the Typescript coding showing this. Currently this is only fixed for the English YAML, Weblate should take care of the rest.

**Fixes**
This PR fixes #2941.

**How to test**
Make sure your set language is English. Navigate to level 8 and start the quiz. Verify that question 6 contains an output example, without any syntax highlighting (as this is simply a plain `span` element). This approach should work identical to the code input/output approach of the main editor.


Felienne (approved):

Felienne:
Works!

image

PR #2960, opened 2022-06-24
Adds parsons exercises to test pipeline for snippet validation

### LINKED ISSUE #2958: [TESTING] Add code validation to Parsons exercises (identical to adventures)
**Description**
Currently we validate all adventures to have valid content, making sure that all snippets returned to the user actually work. It would be nice if we also have this for the parsons exercises. Because when the program is in the correct order it should always execute as expected.

TiBiBa:
**Description**
In this PR we add code snippet testing to the Parsons exercises. We already have this for the adventures and the commands, so it makes sense to also validate the Parsons. The content of the English file is manually copied to all other files as we changed the structure but Weblate does not seem to update the files. This is necessary otherwise all tests will fail. We also add the hash table structure to the level commands YAMLs to speed up the testing pipeline.

**Fixes**
This PR fixes #2958.

**How to test**
Nothing to test, Github should handle this. If the tests pass we are ready to merge.


tests/test_snippets/test_parsons.py
```diff
@@ -0,0 +1,88 @@
...
+                        else:
+                            unique_snippets_table.add(hash(code))
+                        Hedy_snippets.append(Snippet(filename=file, level=level, field_name=f"{exercise_id}", code=code))
+                except:
+                    print(f'Problem reading commands yaml for {lang} level {level}')
+
+
+    return Hedy_snippets
+
+def translate_keywords_in_snippets(snippets):
```
Felienne, line 51: I know I caused the existing duplication cause I was lazy but can we put this in one place (in Tester?)

Felienne:
Merging now since Weblate is calm, I made an issue for the test duplication

Felienne (approved):

Felienne:
Ah oops I think my Farsi merge introduced a problem :( https://github.com/Felienne/hedy/pull/2960/commits/6529bebddf7cfc540a548d677f132560baa200f9

Can you check this tomorrow @TiBiBa? It is getting late :)

PR #2966, opened 2022-06-27
[UI IMPROVEMENT] Displays country name instead of fixed 'your country' header

### LINKED ISSUE #2962: [UI idea - High scores] Show name of country instead of "Your country"
I love https://github.com/Felienne/hedy/pull/2931!

But wouldn't it be nicer if we'd say: "The Netherlands" instead of "Your country"? We have the info for the filter so I guess it is not all too much hassle?

TiBiBa:
**Description**
In this PR we changed the fixed 'your country' header to the actual user country. We also add a template function and call this for each row entry to get the country name as well.

**Fixes**
This PR fixes #2962.

**How to test**
Navigate to `/highscores` and validate that it's working.
Schermafbeelding 2022-06-27 om 09 23 47


Felienne:
This works in English but... for a Dutch profile it also shows "Netherlands" not "Nederland" Is that as expected?

image

TiBiBa:
> This works in English but... for a Dutch profile it also shows "Netherlands" not "Nederland" Is that as expected?
> 
> image

This is due to using the same countries list as the signup and profile page. I think it is best to create a separate issue to look into localizing the countries as this is relevant in multiple spots.

arjank:
> > This works in English but... for a Dutch profile it also shows "Netherlands" not "Nederland" Is that as expected?
> > image
> 
> This is due to using the same countries list as the signup and profile page. I think it is best to create a separate issue to look into localizing the countries as this is relevant in multiple spots.

I agree that it should be a separate issue. And imho we should aim for a solution that does not require additional work for translators. Just translating ~200 country names is not really related to Hedy.

TiBiBa:
I made a bit of a bucket-fix solution, but couldn't find any feasible library or approach online:
```
# Define and load all countries
COUNTRIES = {k: v.name for k, v in iso3166.countries_by_alpha2.items()}
# Iterate through all found country abbreviations
for country in COUNTRIES.keys():
    # Get all spoken languages in this "territory"
    spoken_languages = languages.get_territory_language_info(country).keys()
    found = False
    country_name = None
    # For each language, try to parse the country name -> if correct: adjust in dict and break
    # If we don't find any, keep the English one
    for language in spoken_languages:
        if found:
            break
        try:
            l = Locale(language, country)
            country_name = l.get_territory_name(country)
            found = True
        except:
            pass
    if country_name:
        COUNTRIES[country] = country_name
```

TiBiBa:
On second thoughts it is not that successful as we get the following statistics:
```
We have found a total of 84 languages
We have missed a total of 169 languages
```

TiBiBa:
Made an oopsie, we now find the almost all localized country names:
```
We have found a total of 245 languages
We have missed a total of 5 languages
```

TiBiBa:
All should work now!
Schermafbeelding 2022-06-28 om 13 11 20
Schermafbeelding 2022-06-28 om 13 14 52

Felienne:
> > Works (although the sorting is still, I think, using the English names? Maybe we want to add a search field?)
> > image
> 
> Sort field would be hard as we use the built-in HTML `select` input. Will look into this, but it already supports a bit of a sort field as users can type inside the input to auto-select an option.

Yeah I thought so! Let's leave it as it is!

Felienne (approved):

PR #2967, opened 2022-06-27
[BUG FIX] Highscores are sorted correctly

### LINKED ISSUE #2965: [BUG] Highscore ranking is incorrect
**Describe the bug**
The highscores should be double-sorted. First by amount of achievements, then by the oldest one getting that specific amount. Currently this both doesn't seem to work.
Schermafbeelding 2022-06-26 om 23 34 46

**Edit**
They are only sorted by timestamp. We should find a way to sort by both.

TiBiBa:
**Description**
In this PR we fix the highscore ordering as we want to ordering on two dict values at the same time. The amount of achievements as well as the timestamp of the last achieved one.

**Fixes**
This PR fixes #2965.

**How to test**
Verify that the high scores are always sorted by amount of achievements (high-low) and then timestamp of last update (low-high). This ensures that the user with the most achievements and the lowest timestamp is ranked higher.

Schermafbeelding 2022-06-27 om 09 39 52


Felienne (approved):
Works!

PR #2970, opened 2022-06-27
[FEATURE] Always shows user in highscores, also when not ranked

### LINKED ISSUE #2969: [UI idea] Always show user highscore, even if not in top 50
**Idea incl level**
Currently we only show the top 50 of the highscores. However, it would be nice to always show the user ranking, even if not in this list. For example, we show 1-50 and then 1 row below we display `1234. <current user>`. This is very hard to test as we don't have 50+ profiles with a public profile, either locally or on alpha. So we have to write, test and verify on beta that it works.

TiBiBa:
**Description**
In this PR we make a nice QoL fix where we always show the current user in the highscore, even if not ranked top 50. This is nice when deploying to production as we have 50+ users with a public profile. When not in the top 50 the user is still shown in the list with their own ranking. So for example:
```
1. user1 
2. user2
...
49. user49
50. user50
124. <current_user>
```

**Fixes**
This PR fixes #2969.

**How to test**
Can't be tested. Validate that the code seems logical and deploy to production, there we can validate that it works.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1213,21 +1213,20 @@ def get_highscores_page(user, filter):
     country = user_data.get('country')
 
     if filter == "global":
-        highscores = DATABASE.get_highscores(filter)
+        highscores = DATABASE.get_highscores(user['username'], filter)
     elif filter == "country":
         # Can't get a country highscore if you're not in a country!
         if not country:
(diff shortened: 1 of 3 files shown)


Felienne (approved):
Looks ok, let's check on alpha/beta!

PR #2976, opened 2022-06-27
[BUG FIX] Sharing a program from the programs page now shows correct modal

### LINKED ISSUE #2975: [BUG] Program share from /programs page still uses old modal alert
**Describe the bug**
When sharing a program from the `/programs` page the old alert modal is still used. We should change this to the new one.
Schermafbeelding 2022-06-27 om 13 53 10

TiBiBa:
**Description**
In this PR we change the use of the old deprecated alert modal to the newly used new intrusive one. This was an oversight on the original implementation but should be fix now.

**Fixes**
This PR fixes #2975.

**How to test**
Navigate to the `/programs` page and share any program. Verify that the new alert modal is used instead of the old one.

**Old**
Schermafbeelding 2022-06-27 om 13 53 10
**New**
Schermafbeelding 2022-06-27 om 13 59 59


Felienne (approved):
Check!

image

PR #2979, opened 2022-06-28
[BUG FIX] Programs filtering works again

### LINKED ISSUE #2977: [BUG] Program filtering is broken
**Describe the bug**
For some reason program filtering is broken. After fixing this it would also be nice to add some e2e tests on the filtering to verify the correctness.

Schermafbeelding 2022-06-27 om 17 29 27

TiBiBa:
**Description**
In this PR we fix the programs filtering. This was broken due to a bug where nothing was returned when we filtered without an adventure. This was due to the return statement being nested in the `if adventure:` statement, a big oopsie. This should be fix. We also add an additional e2e test to make sure we catch these mistakes in the future.

**Fixes**
This PR fixes #2977.

**How to test**
Verify that the filtering of `/programs` works again when only selecting a level **without** an adventure.


Felienne (approved):
Fixed!

image

PR #2980, opened 2022-06-28
[BUG FIX] Correctly set keyword language on custom adventure page

### LINKED ISSUE #2504: [UI idea] Set teacher adventure view syntax highlighting to teacher keyword language
**Idea incl level**
Currently we set no keyword language on the teacher view page, meaning that the syntax will always be highlighted in English. This is difficult for teachers who decide to create an assignment in a non-english keyword language. The highlighting on the code page works fine, but we have to find a way to add the `keyword_lang` attribute to the `window.State` on the adventure view page.

TiBiBa:
**Description**
In this PR we fix an issue where the custom adventure page was not set with the correctly keyword language. This was due to the `window.State.keyword_language` not being set on page call. We fix this by overwriting the script block. 

Note that is not a really nice solution with copying and overwriting the whole script block. It works, for now, but we should clean-up the script blocks on the whole website (there is already an issue for that).

**Fixes**
This PR fixes #2504.

**How to test**
Make sure you are a teacher and have your keyword language set to a non-english language. Create a custom adventure and use some localized keywords. Verify that the syntax highlighting works as expected. Both within the preview window as well as on the dedicated `view adventure` page.


Felienne:
@TiBiBa Just pinging you here since I replied yesterday on a really ancient PR so maybe you missed it

TiBiBa:
> It works in the example pupup:
> 
> image
> 
> But viewing the adventure does not seem to give me the highlighting yet:
> 
> image
> 
> Also... Do we expect teachers to hardcode the keywords in their own language? And does that mean kids in the class lose the ability to change them, or do we call `translate` in that case?

We expect the teachers to hardcode them in their own language. It would be difficult to replace these with placeholders to enable dynamic changing of the keywords. Not sure why the viewing page doesn't correctly highlight the keywords, is your keyword language of the logged in user still set to dutch?

TiBiBa:
It works fine for me:
Schermafbeelding 2022-07-08 om 11 00 20

Felienne:
> It works fine for me: Schermafbeelding 2022-07-08 om 11 00 20

ok, will test again, maybe it was caching

Felienne:
Ok I found the difference! If I make the adventure and then change my language, it does not show properly, but if I set it before, it works. Which I guess is ok.

Felienne:
> We expect the teachers to hardcode them in their own language. It would be difficult to replace these with placeholders to enable dynamic changing of the keywords.

Made a fresh issue to discuss this!

Felienne (approved):

PR #2992, opened 2022-06-29
[BUG FIX] Class name displayed correctly on class overview page

### LINKED ISSUE #2990: [BUG] Last class letter is always bold and capitalized
**Describe the bug**
Very weird bug and we have to look into this what might cause this. It is only a visual issue so not really urgent, but for some reason the last letter of the class name on the class overview page is always capitalized and bold. Also see the screenshot below for clarification.

Schermafbeelding 2022-06-29 om 09 50 22

TiBiBa:
**Description**
In this PR we fix an issue where the last character of the class name would be bold and capitalized. While we are already here we also improve the styling a bit by removing all manual margins and replacing them with the `gap` class. We missed these in the PR dedicated for solving that issue.

Proud to say that this was the weirdest bug every solved (till date). The following issue occurred, when injecting the variable into the template using `<h1>{{class_info.name}}</h1>` the name would be displayed as follows "test12 **3**". However, when adding whitespace into the placeholder brackets like this `<h1>{{ class_info.name }}</h1>` it would be displayed correctly as "test123".

**Fixes**
This PR fixes #2990.

**How to test**
Make sure you are a teacher, navigate to the class overview page of any of your classes and verify that the class name on top of the page is displayed correctly.


Felienne:
Ah got it! This problem only occurs:

* On Safari
* For class name that ends in a number

Can confirm this PR fixes it:

image

Felienne (approved):

PR #2993, opened 2022-06-29
[UI IMPROVEMENT] Improve responsiveness of menubar

### LINKED ISSUE #2991: [UI idea] Make menubar responsive and more accessible on mobile
**Idea incl level**
The UI of the menubar is not really responsive (yet) and doesn't look very nice on mobile. Would be nice if we improve this a bit as part of the ongoing project of making the whole project more responsive. This issue can be seen as well as an sub-issue of #2776.

TiBiBa:
**Description**
In this PR we improve the responsiveness of the menubar. We hide some elements when on a smaller screen size and only keep the icons for the programs, language and profile menu option when on table size or smaller. This ensures that our menubar doesn't shift elements on different sizes.

**Fixes**
This PR fixes #2991.

**Old**
Schermafbeelding 2022-06-29 om 11 17 45

**New**
Schermafbeelding 2022-06-29 om 11 17 55

**Note**
This still doesn't look perfect on some languages, for example Arabic. But it is a big step forward into making the website more responsive. When improving small parts in iterations we make sure that we can keep an overview if we break some UI through these changes.


Felienne:
> This still doesn't look perfect on some languages, for example Arabic.

What is the issue with Arabic? And do we need an issue maybe so we don't forget and to detail the current status and needed improvements?

TiBiBa:
> > This still doesn't look perfect on some languages, for example Arabic.
> 
> What is the issue with Arabic? And do we need an issue maybe so we don't forget and to detail the current status and needed improvements?

It's not a big issue, but the headers don't fit on the small screen size as the strings are longer. I'm not sure what would be the best solution to fix this (other than reducing the font size):
Schermafbeelding 2022-06-29 om 11 57 36

TiBiBa:
Fixed!
Schermafbeelding 2022-06-29 om 12 40 54

Felienne:
Looks great!

Felienne (approved):

PR #2995, opened 2022-06-29
[BUG FIX] Highscores country filtering no longer crashes

TiBiBa:
**Description**
In this PR we fix the highscores counter filtering. For unknown reasons there seems to be usernames that return a none value when requesting the user data. This shouldn't be possible, but it makes the page crash. We surround this by a try-except to catch the error. We also improve the code a bit and change a table header to make more sense.

**Possible issue**
The only possible issue I can think of is that a user removed their profile but their public profile remained. This shouldn't be possible, but we have to dive into this later on.

**How to test**
Can't be tested, deploy to beta and verify that it works (Alpha the error didn't happen).

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -8,14 +8,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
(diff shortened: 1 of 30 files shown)


Felienne (approved):

TiBiBa:
This doesn't seem to work yet!

TiBiBa:
Can verify that this is fixed!

PR #2998, opened 2022-06-29
[UI IMPROVEMENT] Adds notifier if user doesn't have a public profile yet

### LINKED ISSUE #2997: [UI idea] Notify users when visiting highscore page without a public profile
**Idea incl level**
It would be nice to notify users if they visit the highscore page without a public profile, showing a link that they can directly create one.

TiBiBa:
**Description**
In this PR we make a small QoL change where we add a notification and explanation of a user doesn't have a public profile yet but visits the highscores page.

**Fixes**
This PR fixes #2997.

**How to test**
Navigate to the highscores page with and without a public profile. Verify that the notification is only shown if no public profile exists. Also verify that the button works and automatically opens the public profile section when clicked.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1207,6 +1207,7 @@ def get_highscores_page(user, filter):
         return utils.error_page(error=404, ui_message=gettext('page_not_found'))
 
     user_data = DATABASE.user_by_username(user['username'])
+    public_profile = True if DATABASE.get_public_profile_settings(user['username']) else False
     classes = list(user_data.get('classes', set()))
     country = user_data.get('country')
     user_country = COUNTRIES.get(country)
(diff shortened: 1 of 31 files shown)


Felienne (changes requested):
Works for users without a profile:

image

But with it looks a bit weird, the text is off and do we also want the button there (I ran Babel locally):

image

TiBiBa:
> Works for users without a profile:
> 
> image
> 
> But with it looks a bit weird, the text is off and do we also want the button there (I ran Babel locally):
> 
> image

I think this is due to how the fuzzy operator of Babel works, as the translations don't many any sense.

TiBiBa:
This should be fixed!

PR #3010, opened 2022-07-01
Fixed Babel update command to stop fuzzy matching and prevent translation confusion

TiBiBa:
**Description**
In this PR we fix a statement in CONTRIBUTING.md to make sure Babel no longer automatically guesses the translations with fuzzy matching. This prevents confusing translations and from now on we should always add the `-N` parameter to the Babel update command.


Felienne:
Thanks! Should we also update this in the on-deploy script?

Mark-Giesen:
> Thanks! Should we also update this in the on-deploy script?

mmm, not sure. It's not in the deploy script at the moment, just the compile step. I think this is right if we ALLWAYS provide EN texts. I guess this is true, but we could decide to add it to this script and provide empty keys in any language and use Weblate for EN aswell. If anyone ever adds a key with only NL or any other language as "default" the proces doesn't work anymore, since the EN text doesn't exist and isn't provided for all other languages. 
So this is a proces thingy maybe?

TiBiBa:
I think for now it is best to add this to on-deploy as well. We still have to think about the best approach on how we align the Weblate process with our manual adding and deleting of keys.

TiBiBa:
No wait, my bad. The fuzzy matching should be prevented on the update command, not on the compile step. So we don't have to change anything to on-deploy.

Felienne (approved):

PR #3011, opened 2022-07-01
Adds new exception to catch repeats with missing parts

### LINKED ISSUE #2458: Unclear error when forgetting `print` in a `repeat` statement in level 7
When you forget to use `print` in the `repeat` statement in level 7 the error is unclear. As this happens quite often I think it is good to have a clear error message.

An example would be:

`repeat 3 times 'Hedy is leuk!'`

jpelay:
**Description**

In issue #2458 it was asked to add a new exception in level 7, regarding the cases where kids miss the `print` keyword after inside a `repeat`. Later it was suggested that we could add a new rule to also take care of the cases where kids forget the `times` keyword.

**Fixes #2458 *

**How to test**
I added a couple of tests for both exceptions. Also this is the message:
image

**Checklist**
  
- [x ] Describes changes in the format above (present tense)


TiBiBa:
> @TiBiBa Could you take a peak at these conflicts (which might get worse again when I merge Weblate, sorry!)

Should be fixed!

Felienne (approved):
Nice work @jpelay!!

image

PR #3014, opened 2022-07-02
[BUG FIX] Signup works again

### LINKED ISSUE #3012: [BUG] Sign up doesn't work
Creating an account doesn't work right. It results in an error, but the account is created.

TiBiBa:
**Description**
Sign-up was broken due to a Babel key within the mail template, crashing on the `format()` function. This is fixed.

**Fixes**
This PR fixes #3012.

**How to test**
Create an account, verify that you are correctly logged in and redirected to the landing-page.


Mark-Giesen (approved):
Weird that adding an alt would fix this.

PR #3018, opened 2022-07-04
Removes email validation field on sign-up

### LINKED ISSUE #3017: [UI idea] Remove double email field
We have two email fields that also disallow copy-pasting since it was really important to have a correct email address to reset passwords, but since teachers can now do that, it is a lot less important. Shall we remove this second field?

TiBiBa:
**Description**
In this PR we remove the need of the mail repeat on sign-up. As suggested in the issue this is no longer as important as before as teachers can also change a student password for example.

**Fixes**
This PR fixes #3017.

**How to test**
Verify that the repeat mail field is removed from the signup page and everything still works as expected.


arjank:
End to end tests pass, so that's good. And the registration page only shows one email address field.

TiBiBa:
@arjank can you also approve this PR? Then we can merge.

Felienne (approved):
Account creation still works, incl. logging in and out!

PR #3029, opened 2022-07-06
[IMPROVEMENT] Only show variables in variable view that have an actual value

### LINKED ISSUE #2948: [BUG] remove 'convert_numerals: undefined' from variables list
image

We should  remove 'convert_numerals: undefined' from the variables list

TiBiBa:
**Description**
In this PR we make a small QoL fix by only showing variables in the variable view when actually having a value. This solves the issue where sometimes variabels with the value `undefined` are returned to the user. Which is redundant and can be confusing for children.

**Fixes**
This PR fixes #2948.

**How to test**
Make sure you are logged in and navigate to Hedy level 2 and run the follow code snippet:
```
name is Hedy
age is 15
print name is age years old
```
Then take a look at the variable view on the output window and verify that the variable `convert_numerals: undefined` is no longer shown. You can verify with multiple programs that only variabels with an actual value are shown in this view.


Felienne (approved):
Great catch!

PR #3031, opened 2022-07-06
[UI IMPROVEMENT] Makes error line highlighting more readable

### LINKED ISSUE #2955: [UI idea] Error highlighting (level 4+)
**Level 4+: error highlighting**
When I try to execute code, it sometimes results in highlighting the entire line in red. But if the keywords on that line are not part of the problem, they will remain in their slightly different shade of red. That means the keywords are very hard to read. Of course this will resolve itself when you start editing the line, but it could be useful to see the entire line before you start editing.

Example when trying to run the code
`hal lo is 'wereld'`
image

TiBiBa:
**Description**
In this PR we make the line error highlighting more readable. Due to the keywords as well as the line being red it was hard to debug when an error was made while there was also keyword highlighting in the same line. We fix this by changing the highlighting from a red background to a red top and bottom border, see the screenshot below for clarification. We also remove the Red Cross on the gutter as it was a bit redundant and overlayed the line counter.

Schermafbeelding 2022-07-06 om 13 37 48

**Fixes**
This PR fixes #2955

**How to test**
Create a program with a mistake and notice the new line error highlighting. Verify that this works in all cases.


arjank (approved):
File changes make sense (but I skipped the generated files).

The red X is gone. And when I tested with a missing quote error (in level 6) I noticed that the marker only marks the first character of the line, instead of the entire line. Verified in production that the same behaviour is present there, so it's not something new. Still might be something to address.
image

TiBiBa:
> File changes make sense (but I skipped the generated files).
> 
> 
> 
> The red X is gone. And when I tested with a missing quote error (in level 6) I noticed that the marker only marks the first character of the line, instead of the entire line. Verified in production that the same behaviour is present there, so it's not something new. Still might be something to address.
> 
> image
> 
> 

I noticed this as well. This is due to the difference between fullLine and text marking. For text we might want to add some regex to highlight until a space or end of line. Will look into this!

TiBiBa:
We can also decide to simply always highlight the whole line containing the error. In combination with the error message this should provide enough information for the user to fix the issue.

TiBiBa:
Hi @arjank! I just changed the code to behave as follows:
- Highlight the whole row of no column is specified
- If a column is specified, highlight from that point in the row until the first whitespace

This should result in the follow highlighting in your example:
Schermafbeelding 2022-07-07 om 11 40 00

If you agree with this approach I think we are ready to merge!

Felienne:
Thanks for this fix @TiBiBa! I do think it is a lot more readable, and removing the cross is good because it also sometimes created weird floating blocks. 

However, I fear this can also be read as two lines (in the example, lines 2 and also 3) being *underlined* rather than one being marked. Could we add a light color (maybe light yellow to contrast with the syntax highlighting) on the line so it is more clear we mean that 1 line and not the 2 around it?

TiBiBa:
> 

Will look into this to make it more readable!

TiBiBa:
Possible solution:
Schermafbeelding 2022-07-08 om 08 46 42

Felienne:
> Possible solution: Schermafbeelding 2022-07-08 om 08 46 42

This is readable but now there is a bit of a disconnect between the message and the line in terms of color (I know I am being really picky here but I think this matters a lot, potentially) Maybe we can make the line the same shade of light pink we have in the error message? This:

image

TiBiBa:
This is no longer readable, will look into a darker color:

Felienne:
> This is no longer readable, will look into a darker color: Schermafbeelding 2022-07-08 om 10 42 47

Ah I see! We can also slightly change the bottom color! As long as they are a bit consistent.

TiBiBa:
I'm not sure I exactly understand the color scheme you want. The following doesn't look really nice either:
Schermafbeelding 2022-07-08 om 11 55 49

Felienne:
> I'm not sure I exactly understand the color scheme you want. The following doesn't look really nice either: Schermafbeelding 2022-07-08 om 11 55 49

I agree this is also not very pretty. I am not sure what I want, I do not have a thing in mind perse. ]

I like the two red lines a lot, the only thing I worry about is it looking like underlines. Can we figure out any way to change?

If not I think 1) the color should be a redish tint and 2) it should be similar to the box below.

TiBiBa:
Follow up proposal, change the background color of the error line to the same color as the top border of the error box. I'm still not completely convinced with this approach, but at least there is a clear connection between both elements:
Schermafbeelding 2022-07-18 om 14 30 23

Felienne:
> Follow up proposal, change the background color of the error line to the same color as the top border of the error box. I'm still not completely convinced with this approach, but at least there is a clear connection between both elements: Schermafbeelding 2022-07-18 om 14 30 23

Ow this looks good!

PR #3032, opened 2022-07-06
[TESTING] Adds e2e test for programs and explore filtering

### LINKED ISSUE #2978: [PERFORMANCE] Add e2e tests for explore and filtering
**Description**
Currently we have e2e tests for all pages, but not for the GET filtering options. We should add these as well as can be verified by #2977 where the filtering is broken but we have no idea why.

TiBiBa:
**Description**
In this PR we add route validation in the e2e tests for the programs and explore page filtering. As these are built using a GET request the filtering options are routes as well. To verify that we don't break anything we should visit these as well through visiting in the e2e tests.

**Fixes**
This PR fixes #2978.

**How to test**
Nothing to test, if the tests pass all should work!


Felienne (approved):
Tests pass!

Issue #2279, opened 2022-03-22
Don't highlight ask/vraag at the beginning of the line in level over 3


Felienne:
In the last days some students noticed something confusing. In case you use Dutch with English commands and you use a Dutch word in your code that is coincidentally the translation of an English command, the word is highlighted in red as if it were a command. This happens more frequently than I initially thought. As an example, see the fortune adventure in level 4 (https://www.hedycode.com/hedy/4/#fortune) with the code `vraag is ask 'Wat wil je weten?'`. Would it be doable to implement an English-command-only option where users cannot switch to Dutch and no highlighting of Dutch translations of English commands?

_Originally posted by @OnnoEbbens in https://github.com/Felienne/hedy/discussions/2080#discussioncomment-2415140_

Felienne:
> The question I have now is, **Is this over-coloring problem serious?**
> Yes, over-coloring means that the syntactic coloring will not quite color to what Hedy is going to interpret, but on the other hand, it will mean to the child that he is using a key word in the language, and that, while this will not necessarily be a problem, it is a dangerous behavior.

I would say, yes, it is very undesired because kids are still learning and the colors help. So if we can get this fixed, please do!

thjazi39:
Concerning problem solving, since Ace (the syntax coloring system) is based on an automaton system, with different states, there is surely a way to do it better, typically, `at random` can only be used in the case where we are looking for a value (for example, following an `is` or a `print`) and it should be possible to use the state system (or token, to use the Ace vocabulary) 

However, this might take a lot of time, because it requires a thorough analysis of the grammar, because the contexts are quite varied, but I'm interested in doing it, and I'll try once I've managed to set up tests for syntax highlighting.

Also, Hedy's grammar will still have potential changes, for example the issue #1938

And if syntax highlighting uses an automaton system, changing the grammar to fix these bugs will require redoing the syntax highlighting automaton system of the syntax highlighting, which is not ideal for maintenance (redoing the automaton, modifying the tests, etc.).
Whereas by freeing ourselves from the context, we are sure that there will be no unrecognized keywords

thjazi39:
Then, apart from these technical considerations, I think it's feasible, and it's done for the first 3 levels.
But if it's feasible for the "simple" levels, for the "complicated" levels, I think there will always be syntax coloring bugs.

[...]

PR #3033, opened 2022-07-07
Highlight firsts levels

thjazi39:
In order to solve the problem of syntactic over-coloring, it is necessary to replace the current system by a new system: a more complex automaton.

This pull request does not solve the problem of over-coloring, but prepares the solution, by starting the replacement of the system of the first levels (the old system was an exhaustive system). This change of system corrects 2 bugs that were present.

For readability reasons, some transitions are not displayed: 
- From each of the states there is a transition for comments and one for the line break, which sets the state to start.
- From each state S there are 2 transitions for invalid characters ( _ and _?_ ) which do not change the state (we return to the start state).

Here is a visual description of the 3 syntax highlighting automata: 

## Level 1 :
image
Description of the automaton:
- The commands ask, echo, print and forward take us from the start of line state to the "Value" state. This state does not recognize anything.
- The turn command takes us to a state where we recognize only the direction commands
- The command color takes us to a state where we recognize color names. 

## Level 2 :
image
Description of the automaton:
for all commands, the state after will not recognize anything (left and right don't exist anymore, and the command color doesn't exist at level 2)

## Level 3 :
image
Description of the automaton:
for all commands, the state after will not recognize anything (left and right don't exist anymore, and the command color doesn't exist at level 2)

### Tests

When modifying the syntactic highlighting of level 3, the highlighting of commas, and the recognition of blanks was added to level 3, and it was therefore necessary to modify some tests:

Before :
image
image

Now :
image
image

# How to test
Check that it doesn't break anything for level 3, both from the front-end and syntactic highlighting point of view.


thjazi39:
This PR is not fundamentally necessary, but allows to understand the way I propose to solve the problem of over highlighting #2279. 

Indeed, in order to reduce the problem of over-coloring in levels 4 and above, it is necessary to replace the previous system (we highlight a set of commands) by a system that introduces several states. There will be for example a state for the beginning of line, a state "value", etc...
This is feasible, and would make it easier to take into account the context, and thus not color the variable with keywords as name.

However, this system has the disadvantage of being less modular and less accessible compared to the previous system.
Where it was enough to change the command sets for each level, which allowed other people to easily add the repeat command after level 10 or the color command. Now we will have to rethink the whole automaton every time we change Hedy, to make sure that it won't add a new bug.

After that, modularity and maintainability is maybe not so much a problem. I have a python script that generates a visualization of the syntactic highlighting automaton, and if Hedy's syntax has reached its final form, it won't take much to redo the syntactic highlighting

thjazi39:
> Thanks for the extensive explanation @thjazi39!!
> 
> I am just not sure what the middle symbol in this is: image
> 
> The half box thingie is not something I am familiar with. Does it mean any whitespace? Making the whole regex mean not any whitespace?

Indeed, I forgot to specify what this symbol meant! 
My script that generates a graphic representation replaces the space characters, invisible by nature, with this half box.
This allows not to confuse a space in the regex with the space character

Felienne (approved):
Can confirm this improves the highlighting of `_` and does not seem to break anything else, thanks @thjazi39

PR #3034, opened 2022-07-07
Adds a preprocessor step in the grammar merger

### LINKED ISSUE #2712: [Language idea] add special rule "needs colon"
In #2701 we remove a comment that had a nice idea which needs to live in an issue and not in a comment:

```
//TODO: this could be an attribute on the rule that we add in the merger
```

The idea was to have some sort of attribute on a rule in a special grammar form (like we have `+=` and `-=` and `|`) that indicates that the rules should now end in `:`. No rush at all on this, just a way to make the grammars nicer that we can think about (maybe fun for @jpelay)

jpelay:
**Description**

Some rules in the grammar change slightly from their previous levels, example: in level 17 we just add the `_COLON` terminal in the same section of several rule. This new preprocessor would clean the grammars a bit by specifying those additions as annotations:
For example:
```
// Definition of the rule
ifs:  _IF _conditions  _COLON _EOL (_SPACE command) (_EOL _SPACE command)* _EOL _END_BLOCK
// How we would define it now in the grammar
ifs<needs_colon>
```
The syntax is: 
`rule_name<processor_rule>`

I also added a dictinoary that maps from the processor rules names to function that handle the part of modifying the grammar rule.

**Fixes #2712 **

**How to test**

You can test it by running all the tests.

**Checklist**
  

Felienne (approved):
Thanks for building this long desired feature @jpelay!!

Now we can use this in other places too, like needs_brakets or needs_indentation!

jpelay:
> Thanks for building this long desired feature @jpelay!!
> 
> Now we can use this in other places too, like needs_brakets or needs_indentation!

Thank you! Oh yes! That's something I need to check out!

PR #3036, opened 2022-07-07
[FEATURE] Adds blur option to code editor

### LINKED ISSUE #3013: [UI idea] Blur code when running
Writing stories and showing them to others, hiding the code helps prevent spoilers. The ACE editor seems to have a blur() function, a test to just add that in runit() did not do the trick. First decide whether or not we want this, then how it's done.

TiBiBa:
**Description**
In this PR we add a nice QoL feature that enables users to blur their code. As suggested this is a nice help when wanting to show your program to someone without already spoiling what is going to happen because the code and output window are next to each other. We use some manual styling to fix this, by adding/removing the `blur` styling attribute to the editor element on click of the corresponding icon. We can simply change the "amount of blur" by editing the value in `app.ts`, which is currently set at `3px`.

**Screenshots**
Schermafbeelding 2022-07-08 om 08 38 33
Schermafbeelding 2022-07-08 om 08 38 39

**Fixes**
This PR fixes #3013.

**How to test**
Notice the new icon in the bottom-right corner of the code editor. When clicked the icon should change and the code editor should become blurred. Pressing again should reverse this step.


Felienne (changes requested):
Lovely done @TiBiBa! I love the subtle but clear icon. 

Two remarks:

1. We also blur the confetti icon, I guess that is not what we want:



2. We are all, for good reasons, worried about a lot of options in the UI. But I love this feature and kids will do too. What about we just show it in the programmers mode so it is easily accessible but not in view for everyone (and it will, I guess, be mostly used when kids create larger programs they want to show to others)

TiBiBa:
> 

Will look into both options today!

Felienne:
Cool!!

PR #3038, opened 2022-07-08
Adds research section to learn-more page, dynamically load papers

### LINKED ISSUE #3037: [UI idea] Create dedicated research section on learn-more page to store related researsch
**Idea incl level**
With all the students working on Hedy related research it would be nice if we could also show this somewhere on our website. In #3002 the Thesis is simply added as a .pdf file to the project, but without Flask routing this file would be inaccessible (except through GitHub). I suggest a dedicated subsection on the learn-more page with dynamic routing to all .pdf files in a specific folder.

TiBiBa:
**Description**
In this PR we add a research section to the learn-more page, containing related research performed on Hedy. The list is dynamically created based on the content in the `/research` folder. When adding a new .pdf file the name is automatically parsed and the corresponding routing is added. This should make it easy to add research for all contributors. We do have a specific naming convention to keep the list structured. One should keep the following approach:

- Name the file the title of your thesis
- Use underscores instead of whitespaces to ensure its a valid filename
- The name is dynamically parsed to look nicely

**Example**
`An_approach_to_describing_the_semantics_of_Hedy.pdf`
Will be shown as:
`An approach to describing the semantics of Hedy`

**Fixes**
This PR fixes #3037.

**How to test**
Navigate to the `/learn-more` page and notice the new research section at the bottom. Verify that the current link (it is only one paper at the moment) works and the .pdf file is shown in a new tab.


Felienne (approved):
Nice improvement!

PR #3039, opened 2022-07-08
[FEATURE] Adds option to send update messages to users

### LINKED ISSUE #3001: [UI idea] Add some notification bar to notify users on new content or features
**Description**
It would be nice if we create some pop-up or notification that notifies users when we have deployed new content or new features. The admin should be able to "push" a message that users will see the first time they start a session (or either login). We do have to find a nice way to keep track of if logged in users have already seen the message.

This way we can notify them on the existence of the public profile, highscores and/or new adventures being added. In combination with this it might (not sure if this is within scope) be nice to enable teacher to sent notifications to their students.

TiBiBa:

**Description**

_Changes in detail_

**Fixes**
This PR fixes #3001.

**How to test**

**Checklist**
  
### DIFF
--- a/app.py
+++ b/app.py
@@ -1027,9 +1027,13 @@ def profile_page(user):
             invite['class_name'] = class_info.get('name')
             invite['join_link'] = class_info.get('link')
 
+    # At this point we want to retrieve all the users non-read updates we created as admin
+    # But, we also want to localize them: how do we fix this?
+    messages = []
+
     return render_template('profile.html', page_title=gettext('title_my-profile'), programs=programs,
-                           user_data=profile, invite_data=invite, public_settings=public_profile_settings,
-                           user_classes=classes, current_page='my-profile')
+                           user_data=profile, invite_data=invite, messages=messages,
+                           public_settings=public_profile_settings, user_classes=classes, current_page='my-profile')
 
 
 @app.route('/<page>')
--- a/app.py
+++ b/app.py
@@ -1029,7 +1029,11 @@ def profile_page(user):
 
     # At this point we want to retrieve all the users non-read updates we created as admin
     # But, we also want to localize them: how do we fix this?
-    messages = []
+    # We also have to think about user interaction:
+    # - Should they be able to delete the message?
+    # - Or should the message disappear when read?
+    # - Can teachers also send messages to their students?
+    messages = ["test123", "nog een bericht..."]
 
     return render_template('profile.html', page_title=gettext('title_my-profile'), programs=programs,
                            user_data=profile, invite_data=invite, messages=messages,
--- a/templates/profile.html
+++ b/templates/profile.html
@@ -3,7 +3,7 @@
 {% block main %}
    <h2 class="mx-0 lg:mx-16 px-2 mb-4">{{_('account_overview')}}</h2>
    <div class="flex flex-col gap-4 mx-0 lg:mx-16">
-       {% if invite_data %}
+       {% if invite_data or messages %}
            <h1 class="section-header" onclick="$ ('#messages').toggle()">{{_('my_messages')}}</h1>
            <div id="messages" class="profile-section-body">
              <h2 class="profile-section-body-header">{{_('my_messages')}}</h2>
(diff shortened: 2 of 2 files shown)


Felienne:
Hi @TiBiBa!

What is the state of this one, it is a lot behind now of course, do we want to pursue this one or make a new one (or give up on this idea for now?)

TiBiBa:
> Hi @TiBiBa!
> 
> What is the state of this one, it is a lot behind now of course, do we want to pursue this one or make a new one (or give up on this idea for now?)

Missed this one! I like the idea of us being able to push some exciting updates to users, but we should have a clear scope before I start working on this. We current support class invitations (I do think they are not used that much though). We can implement something similar, but on a new table `messages`. But how do we want to efficiently handle the user interaction? Two possible solutions:

**Easy options**
- Store all messages in the new `messages` table
- Add a new attribute to the user profile called `messages` as well
   - Let this attribute contain a set of ids of the unread messages
   - When read -> remove from set
   - Big downside is that we have to update all users when we sent a message

**Safer option**
- Create two new tables `messages` and `interactions`
- On `messages` store the actual messages
- On `interactions` store a message id-username combination and a `read` attribute
- When a user calls their profile -> also get their messages
- Much safer as we only handle the `interactions` table when updating / reading

Felienne:
> Missed this one! I like the idea of us being able to push some exciting updates to users, but we should have a clear scope before I start working on this. We current support class invitations (I do think they are not used that much though). 

Probably because they stem from before teachers could create accounts?

> We can implement something similar, but on a new table `messages`. But how do we want to efficiently handle the user 
> interaction? Two possible solutions:
> 
> **Easy options**
> 
> * Store all messages in the new `messages` table
> * Add a new attribute to the user profile called `messages` as well
>   
>   * Let this attribute contain a set of ids of the unread messages
>   * When read -> remove from set
>   * Big downside is that we have to update all users when we sent a message
> 
> **Safer option**
> 
> * Create two new tables `messages` and `interactions`
> * On `messages` store the actual messages
> * On `interactions` store a message id-username combination and a `read` attribute
> * When a user calls their profile -> also get their messages
> * Much safer as we only handle the `interactions` table when updating / reading

Safer option seems better indeed! Will that be a lot of work?

TiBiBa:
> Safer option seems better indeed! Will that be a lot of work?

Should be do-able today I think! We do have to think about the message retrieval as make a db call on each request is a bit server heavy. Maybe only check for new messages on login and when loading the `my-profile` page? We also have to make some changes to the database structure. And another big bottleneck:

- When sending a message we have to make an interaction entry for each user. But as we can't load all users at once (as we are still struggling with #2994) this has to have some work around. Also: do we only want to support global messages send by the admin or also, for example, class messages?

TiBiBa:
This also still doesn't allow us to localize messages. We first have to add some layer to enable contributors to translated them and then sent the message to the end-users. Should all be possible, but might be quite some work.

Felienne:
> This also still doesn't allow us to localize messages. We first have to add some layer to enable contributors to translated them and then sent the message to the end-users. Should all be possible, but might be quite some work.

Ah good point! Not sure that is feasible though, if we want to push info about a new feature we can't wait for all languages to be updated. Can we do it with gettext, so we can add En and Nl?

TiBiBa:
> > * When sending a message we have to make an interaction entry for each user. But as we can't load all users at once (as we are still struggling with [[PERFORMANCE] Change users table structure to enable pagination and filtering #2994](https://github.com/Felienne/hedy/pull/2994)) this has to have some work around. Also: do we only want to support global messages send by the admin or also, for example, class messages?
> 
> Maybe this is naive but can't we check on login and add a record in interactions when someone has seen it (rather than when they _have_ seen it)?

This is also possible! But only when we aim on supporting global messages, otherwise we wouldn't know what message is aimed for what user. We do have some increase in database call in this scenario:
- Retrieve all messages
- Retrieve all interactions from the current user
- If there is a message without an interaction entry -> show

Felienne:
> This is also possible! But only when we aim on supporting global messages, otherwise we wouldn't know what message is aimed for what user. 

I think that is ok, it is meant for us to notify users of new features. Teachers have enough ways of reaching their students. Maybe we can differentiate one thing: between messages for normal users and those for teacher users (or for both)

> We do have some increase in database call in this scenario:
> 
> * Retrieve all messages
> * Retrieve all interactions from the current user
> * If there is a message without an interaction entry -> show

That is ok I think if it just happens as login

TiBiBa:
Getting there...
Schermafbeelding 2022-08-29 om 12 58 39

Issue #2925, opened 2022-06-20
[UI idea] Inaccessible items (for keyboard users)


arjank:
**Language selection and cheatsheet inaccessible for keyboard users**

The language selection menu and the cheatsheet are not accessible for keyboard users. You might be able to get focus to them (using the tab key) but it's not easy to see when an item has focus. And if those items have focus and you want to activate button using the space bar or the enter key, nothing happens. The items only appear when you hover over them using a mouse. (I have not tested tablets. They don't really have a hover state, so that might be an issue too.)

TiBiBa:
Isn't this already fixed? We did a major re-structuring on the menubar to make it better align with the styling rules and also better accessible on touchscreen.

AnneliesVlaar:
When using tab and spacebar the language switcher reacts

PR #3040, opened 2022-07-08
[PERFORMANCE] Makes dropdown menu touch friendly

TiBiBa:
**Description**
In this PR we switch from the `hover` state as feature to show the dropdown to the `click` state. This has several advantages, but the biggest one is that it makes our website more touch friendly. With the current hover state it is hard for touchscreen users to get the language selector or profile options dropdown to shown. With this implementation we remove the hover dropdown and only show the menu on click, keeping it shown until clicked outside of the dropdown menu.

**Fixes**
This PR is a starting point for #2925, with the goal to make the menubar overall beter accessible.
For now this only enhances the touchscreen experience, but in a different PR we can built upon this.

**How to test**
Use a browser that can mimic a touchscreen device, such as Firefox -> developers tools. Verify that the dropdown menu from either the language selector or the profile options now works as expected: dropping down when clicked. Also verify that the menu is only hidden iff clicked outside the dropdown itself (or clicking the button again).


Felienne (changes requested):
Thanks for teaching me about the responsive test mode in Firefox, awesome!

Indeed the menus now work but it seems the cheat sheets remain opened at all times. The 'Probar' buttons work but I can't close the menu itself:

image

TiBiBa:
> Thanks for teaching me about the responsive test mode in Firefox, awesome!
> 
> Indeed the menus now work but it seems the cheat sheets remain opened at all times. The 'Probar' buttons work but I can't close the menu itself:
> 
> image

I forgot about this one! It also uses the "dropdown" class and should be re-written as well. Will look into this.

TiBiBa:
This is a bit harder than expected, will but this into a draft until fixed.

Felienne (approved):
Dropdown issue fixed now!  

image

PR #3046, opened 2022-07-12
Highlight level 4

thjazi39:
Here is the level 4 transition from the keyword list based system (out of context) to a multi-state approach.

Here is a graphical representation of the level 4 highlighting automaton:

image

(with the same legends as in the PR #3033)

In particular, there is a restriction on the use of strings. Apart from the ask and print commands, quoted strings are no longer recognized :
- Before :
Capture d’écran du 2022-07-12 10-51-33
- After :
Capture d’écran du 2022-07-12 10-52-47

The tests had to be adapted:
- Before :
image
- After :
image

# How to test
Check that it doesn't break anything for level 4, both from the front-end and syntactic highlighting point of view.


thjazi39:
As in PR #3033, this solution does take context into account, but at the expense of maintainability. So modifying hedy's syntax by adding or removing commands will now be more difficult (before it was enough to add or remove a line in a command list, now it will be necessary to rethink the automaton and redo it, being careful not to introduce any error)

Felienne (approved):
Nice, thanks!!

PR #3051, opened 2022-07-17
Allow turtle color command above level 1

### LINKED ISSUE #3045: [BUG] Color command only seems to work in level 1
In the higher levels the color command doesn't work

arjank:
**Description**
The color command was only available in level one, but the other turtle commands are available up to level 10. This PR adds support for the color command all the way to level 10, so it matches the other commands.

As of level 2 color works with a variable, and as of level 3 it's also possible to use 'at random'

**Fixes #3045 **

**How to test**
Go to any  level 2-10 and use the color command together with other turtle commands

As of level 7 you could use the code below (and enter a reasonably high number) to get a colorful circleish shape.
```
hoeken = ask 'Hoeveel hoeken krijgt dit figuur?'
hoek = 360 / hoeken
colors = red, green, blue, orange
repeat hoeken times
    color colors at random
    turn hoek
    forward 20
```

**Checklist**
  

Felienne (approved):
Supercool @arjank!!!!

image

PR #3062, opened 2022-07-18
[BUG FIX] All forms work again as expected

### LINKED ISSUE #3061: [BUG] Login is broken
**Describe the bug**
The login is broken. It seems to happen due to an empty body being received by the back-end (so the mistake is in the front-end). This has a high priority as this influences end-users a lot.

TiBiBa:
**Description**
In this PR we fix an app breaking issue where are forms would no longer work as expected. This was due to our front-end template clean-up where we removed the `js-validated-form` class from all forms. This class is essential for some value matching we do before sending the package to the back-end. This is vulnerable and should be removed in the near future, but to fix all functionality we reverse this removal.

This bug was my mistake by not taking the dependency of the `js-validated-form` class in account. But, this should not have been possible in the first place. We already perform all validation on the back-end and should not rely on the front-end to also perform some form validation. All input fields are already validated by the HTML attributes themselves as well.

**Fixes**
This PR fixes #3061

**How to test**
Verify that all forms are working again: login, signup, recover, reset, profile and public_profile.

### DIFF
--- a/templates/login.html
+++ b/templates/login.html
@@ -4,7 +4,7 @@
 {% block message %}{% endblock %}
 <div class="w-full mx-auto flex flex-col items-center gap-4">
   <div class="bg-white shadow-md rounded-lg rounded py-4">
-      <form onsubmit="event.preventDefault (); hedyApp.auth.submit ('login')" class="flex flex-col items-center gap-2">
+      <form onsubmit="event.preventDefault (); hedyApp.auth.submit ('login')" class="js-validated-form flex flex-col items-center gap-2">
         <h2>{{_('login_long')}}</h2>
         <div class="w-full flex px-8 flex-row items-center justify-center">
             <label for="username" class="inline-block w-40">{{_('username')}}</label>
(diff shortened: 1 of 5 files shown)


Felienne:
Oops and I just deployed! But this seems to fix it. Should we do a small post mortem about why the e2e did not catch this?

Felienne (approved):

TiBiBa:
> Oops and I just deployed! But this seems to fix it. Should we do a small post mortem about why the e2e did not catch this?

This is due to our e2e tests already having the valid packages ready to send to the back-end. We are only validating the back-end functionality and response to various packages, but we never validate if the packages sent from the front-end itself are valid. As we broke the front-end there was no way for the e2e test to catch this issue.

I'm not sure if we can adjust our e2e test to actually use the HTML form instead of only making a POST against the server. This would have a lot of advantages as a similar bug might be introduced later on if we make some changes to our front-end. This is very likely to introduce bugs, especially with the value matching we perform before sending the package.

Felienne:
I know there are exist some tool for automated front-ent testing like Selenium but that might be a bit of work to figure out.

PR #3066, opened 2022-07-19
Enable dynamic keywords for teacher adventures

### LINKED ISSUE #3041: Allow teachers to use multilingual keywords in their own adventures
> We expect the teachers to hardcode them in their own language. It would be difficult to replace these with placeholders to enable dynamic changing of the keywords. 

_Originally posted by @TiBiBa in https://github.com/Felienne/hedy/issues/2980#issuecomment-1178722960_

Maybe I am overlooking things here, but why it is hard? Can we ask them to use {ask} if they want to give kids the power to switch? I guess with a little bit of effort we can then run the teacher-authored adventures through the same pipeline as the others to localize?

TiBiBa:
**Description**
In this PR we enable teacher to use dynamic keywords within their custom adventures. By supporting the use of curly brackets we can detect and parse keywords similar to how we approach this for all our other content. We verify that the placeholders are correct when trying to view a preview, making sure the teacher knows if they created a valid adventure.

**Fixes**
This PR fixes #3041

**How to test**
Verify that you can create, preview, view and customize adventures with keyword placeholders. Also verify that these are parsed correctly for the user selected keyword language. A teacher should **not** be able to save an adventure with an invalid placeholder and should receive a modal alert error instead, verify this as well.


TiBiBa:
Some screenshots:
Schermafbeelding 2022-07-19 om 10 51 39
Schermafbeelding 2022-07-19 om 10 51 44
Schermafbeelding 2022-07-19 om 10 51 53
Schermafbeelding 2022-07-19 om 10 52 45

TiBiBa:
This PR seems to fail the tests on a completely unrelated code snippet. I have no clue what could cause this:
```
הדפס 'Drawing Figures'
[288](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:289)
figure הוא שאל 'Do you want a square or a triangle?'
[289](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:290)
אם figure הוא triangle angle הוא 120 אחרת angle הוא 90
[290](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:291)
פנה angle
[291](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:292)
קדימה 25
[292](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:293)
=========================== short test summary info ============================
[293](https://github.com/Felienne/hedy/runs/7406011428?check_suite_focus=true#step:6:294)
FAILED tests/test_snippets/test_adventures.py::TestsAdventurePrograms::test_adventures_0063_he_5_start_code
```

Felienne:
> This PR seems to fail the tests on a completely unrelated code snippet. I have no clue what could cause this:
> 
> ```
> הדפס 'Drawing Figures'
> figure הוא שאל 'Do you want a square or a triangle?'
> אם figure הוא triangle angle הוא 120 אחרת angle הוא 90
> פנה angle
> קדימה 25
> =========================== short test summary info ============================
> FAILED tests/test_snippets/test_adventures.py::TestsAdventurePrograms::test_adventures_0063_he_5_start_code
> ```
> 
> It seems to be caused by the last line of the code snippet, where the command is put **after** the value instead of the other way around. Why is unclear to me as the snippet itself is directly copied from the English one, only the keywords are being parsed. @Felienne do you have a clue how we can solve this?

Yes, I know the cause of this but it is a bit of a story...!

I will 1) fix it here and 2) open an issue to prevent it in the future!

hedy_content.py
```diff
@@ -32,7 +32,7 @@
 ALL_KEYWORD_LANGUAGES = {}
 
 # Todo TB -> We create this list manually, but it would be nice if we find a way to automate this as well
```
Felienne, line 34: @TiBiBa might not be a bad idea to create an issue for this?

Felienne:
Hi @TiBiBa!

I am not sure if I can verify that this works as I intended it. It seems as a student, I always get the keyword lang of the teacher, even if that is not my language:

image

However, my intention (which maybe I did not express clearly enough!) was to give students the power to choose. I guess that is not what you added in this PR?

TiBiBa:
> Hi @TiBiBa!
> 
> 
> 
> I am not sure if I can verify that this works as I intended it. It seems as a student, I always get the keyword lang of the teacher, even if that is not my language:
> 
> 
> 
> image
> 
> 
> 
> However, my intention (which maybe I did not express clearly enough!) was to give students the power to choose. I guess that is not what you added in this PR?

Hi! This works as intended, however because the local database calls the values by reference they are overwritten once the keywords are parsed. You can verify this by loading the adventure from the teachers page again and notice that the placeholders are replaced by the keywords. This shouldn't be an issue once deployed. Another solution is to make a additional variable when parsing to prevent this local overwrite.

app.py
```diff
@@ -820,9 +820,15 @@ def index(level, program_id):
     commands = COMMANDS[g.lang].get_commands_for_level(level, g.keyword_lang)
 
     teacher_adventures = []
+    # Todo: TB It would be nice to improve this by using level as a sort key
```
Felienne, line 823: @TiBiBa Do we need to turn this in an issue?

Felienne (approved):

PR #3071, opened 2022-07-19
[TESTS] Adds random tests generator

### LINKED ISSUE #988: [TESTING] Generate test cases from separate lines
Once we have reorganized the test cases #841, we could consider extracting the list of single line programs that we already test and generate (random?) combinations of lines to test.

This has as benefits: 

- We do not have to manually combine lines (saving lines of test code)
- We cover more cases

We would have to disallow some cases (like no echo before print in level 1). Maybe that ends up being very complex but we could see where it takes us?

jpelay:
**Description**

A while ago @Felienne had the idea of combining the different commands in a level, rather randomly, in order to be able to test a wider set of possible programs. In this PR, I add a new random case generator that works like this:

In order to be able to generate programs, we can supply individual blocks of self contained programs in a dictionary that has this structure:

```
programs = {'test_name': ['hedy program', 'python_program']}
```

Then among all of the tests in the program dictionary we pick one randomly. If it needs indentation we handle it using a stack of indented blocks, and the number of lines that are still available to use.

**Fixes #988**

**How to test**
I'm still working on adding the tests for the all of the other levels, that's why this PR is a draft, but I'd like if you like can have a look :) 

**Checklist**
  

Felienne:
Hi @jpelay, do we want to keep this open or try the same approach in a new PR (since this is a bit behind?)

PR #3074, opened 2022-07-20
[FEATURE] Show amount of pending messages on menu dropdown menu

### LINKED ISSUE #3068: [UI idea] Show messages in profile icon
When I have a message, it is not clear unless I navigate to messages in my profile. Would it be useful to add a little 1 to my profile icon so it is more clear:

image

(I vaguely remember we had something like that once?)

TiBiBa:
**Description**
In this PR we re-introduce a feature showing the amount of unread messages for a user on the menubar. This is a first step towards the integration of a message system which would allows us (and teachers) to send updates to users. We introduce an additional session value, `messages`, that tracks the amount of unread messages. In the case of invites this is either one or zero. We update this value on profile retrieval, class join and if the session value is not initialized yet.

Funny note: we also found a bug that would remove a class invite from a user if they decide to join another class. This should not happen. The invite should only be removed on joining the same class of the invite, when a teacher removes in the invite or when the invitation expires (7 days). This bug is fixed as well.

**Fixes**
This PR fixes #3068.

**How to test**
Make sure you are a teacher and invite yourself to a class, verify that the menubar is updated in one of the following situations:

- You don't have a session value yet (on login)
- You navigate to the profile page
- You accept or decline an invitation


TiBiBa:
Schermafbeelding 2022-07-20 om 12 22 34

Felienne (approved):
Thanks, hope this does not degrade performance too much but it is worth it I think

PR #3077, opened 2022-07-20
[CLEAN-UP] Removes front-end dependency on javascript input validation 

### LINKED ISSUE #3063: [CLEAN-UP] Remove front-end dependency on javascript validation
**Description**
Currently we use some front-end validation of input fields before making the corresponding POST to the server. This is redundant as we already perform all validation on the back-end. This extra validation is likely to introduce bugs (and already has!). We should look into a nicer way to make sure we get the correct input from the HTML form and POST them with AJAX to the server.

TiBiBa:
**Description**
In this PR we perform a major clean-up on how we approach and handle form data. In the current situation the form data is extracted with TypeScript by a function call `onsubmit`. This data should always conform the fixed `User` interface defined by TypeScript. This data is than sent to the back-end where the actual handling and verification is performed.

However, this can be simplified. The interface layer on TypeScript can be removed as the handling and checking of the data is already performed on the back-end. Therefore we replace this structure by an `onsubmit` trigger on the forms like this:

```
$('form#login').submit(function(e) {
 // Perform all stuff
});
```

Within this function we convert all form inputs found within the triggered submit form into a JSON object and sent this to the back-end. This process change has the advantage that we don't have to adjust an additional layer when changing the input fields. Currently we have to change the front-end input, the TypeScript interface and the back-end. In this PR we reduce this to changing the template and back-end.

Lastly we also clean-up the `auth.ts` overall. By removing redundant front-end session values that we no longer use and simplifying the redirecting process on several actions.

**Fixes**
This PR fixes #3063.

**How to test**
As this is a clean-up task everything should still work. Do **not** rely on the tests as they don't test the actual forms and templates. Therefore verify that all forms still work as expected with corresponding expected error messages. These are login, signup, profile update, public profile, logout, destroy, change password.

**Note**
We should test the reset and recover of the password both on Alpha. Due to the mail dependency we can't test these locally.


app.py
```diff
@@ -1493,17 +1493,19 @@ def modify_query(**new_values):
     return '{}?{}'.format(request.path, url_encode(args))
 
 
+# Todo TB: Re-write this somewhere sometimes following the line below
 # We only store this @app.route here to enable the use of achievements -> might want to re-write this in the future
 @app.route('/auth/public_profile', methods=['POST'])
 @requires_login
 def update_public_profile(user):
     body = request.json
+    print(body)
```
boryanagoncharenko, line 1502: is the print intentional?

website/auth.py
```diff
@@ -493,6 +493,7 @@ def change_student_password(user):
     @requires_login
     def change_password(user):
         body = request.json
+        print(body)
```
boryanagoncharenko, line 496: is this one intentional?

website/auth.py
```diff
@@ -543,22 +544,18 @@ def update_profile(user):
 
         # Validations, optional fields
         if 'birth_year' in body:
-            if not isinstance(body.get('birth_year'), int) or body['birth_year'] <= 1900 or body['birth_year'] > datetime.datetime.now().year:
+            try:
+                body['birth_year'] = int(body.get('birth_year'))
```
boryanagoncharenko, line 548: The built-in method int() will try to transform a given value to an int, e.g. int(2.5) becomes 2 and int("1") becomes 1. This means that here we are relaxing the constraint that 'birth_year' is of type int (which was previously there). Is this intentional?

boryanagoncharenko (changes requested):
I encountered the following issues while testing:
- I cannot sign up. I am getting an error that the repeated password does not match.
- When I try to update my public profile, my picture is not saved. No matter what I change it to, when I reload it will be back to the default one.
- When I try to update my profile I am getting an error that the birth_date is not correct, even though I do not supply a birth date at all. It seems that the field is passed to the backend as an empty string. 
Screenshot 2022-07-21 at 16 22 48

app.py
TiBiBa: This should be removed, thanking for catching this issue!

website/auth.py
TiBiBa: We already have a number constraint on the HTML input field. With the new way of handling the data on the front-end all data is send as strings to our back-end, which handles are conversions and checks. So a user shouldn't be able to select a value other than a whole number, if for some reason they do; the back-end catches this.

TiBiBa:
> I encountered the following issues while testing:
> 
> * I cannot sign up. I am getting an error that the repeated password does not match.
> * When I try to update my public profile, my picture is not saved. No matter what I change it to, when I reload it will be back to the default one.
> * When I try to update my profile I am getting an error that the birth_date is not correct, even though I do not supply a birth date at all. It seems that the field is passed to the backend as an empty string.
> 
> Screenshot 2022-07-21 at 16 22 48

- signup is fixed by re-introducing the password repeat field (not sure why this was removed, but we did this once)
- birth year and gender are fixed when updating with an empty value

website/auth.py
boryanagoncharenko: Making the backend validation complementary to the frontend validation is a dangerous endeavor. But the most important thing here is to ensure that you are aware of the details around change you did.

Speaking of this validation. The screenshot indicates that there is a problem with the error message. It is appended instead of injected. I know this is not the main focus of your PR but perhaps you could address that?

TiBiBa: It's not complementary. The back-end still completely ensures that the data is valid (an integer between 1900 and 2022), however if the user follows the "normal" flow they are unable to enter any value other than an integer. The frontend validation is merely to help the user and reduce server load as users are initially stopped by the browser when entered an invalid value.

I noticed this incorrect error message as well and just committed a fix.

TiBiBa:
All stated issues should be fixed!

Felienne (approved):

Felienne:
Thanks for all the comments @boryanagoncharenko!!

PR #3080, opened 2022-07-20
[LANGUAGE] Adds support for empty lists starting from level 15

### LINKED ISSUE #3050: [CONTENT] No ability to initialise empty lists
Some people were trying to make a shopping list, but there is no way to initialise an empty list to add elements to as the program continues. Even in level 18, which is supposed to be normal Python, it's not possible.

A solution could be to introduce them in level 16 when brackets are introduced, as that is an easy way to recognise when a variable is a list and `emptyList = []` could be possible.

jpelay:

**Description**

_Changes in detail_
Changes the grammar for lists in level 15 to allow both empty lists and lists containing a single element in it. Since starting from this level we ask users to use brackets around lists, we now can allow them to create this types of lists.

**Fixes #3050*

**How to test**

I added a couple of tests to showcase this behavior.

### DIFF
--- a/grammars/level16-Additions.lark
+++ b/grammars/level16-Additions.lark
@@ -6,4 +6,4 @@
 ?atom: NUMBER | var_access | list_access | text_in_quotes
 list_access: var_access _LEFT_SQUARE_BRACKET (INT | random | var_access) _RIGHT_SQUARE_BRACKET
 change_list_item: var_access _LEFT_SQUARE_BRACKET (INT | var_access) _RIGHT_SQUARE_BRACKET _EQUALS (var_access | textwithoutspaces)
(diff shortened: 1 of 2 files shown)


Felienne:
Ah wow, just a little addition to the grammar and nothing even in the transpiler! DId I accidentally write code that generalizes so easily?? 🤣

boryanagoncharenko (approved):
Your change made me realize that up until now Hedy did not allow empty lists, but it also disallowed lists with a single item in them (e.g. friends = ['Ashli']). Perhaps you want to add a test that covers this scenario too.

jpelay:
> Ah wow, just a little addition to the grammar and nothing even in the transpiler! DId I accidentally write code that generalizes so easily?? 🤣

Yes! It was very easy hahaha I was pleasantly surprised!!

> Your change made me realize that up until now Hedy did not allow empty lists, but it also disallowed lists with a single item in them (e.g. friends = ['Ashli']). Perhaps you want to add a test that covers this scenario too.

Yes, already done :)

PR #3085, opened 2022-07-21
[BUG FIX] Keyword placeholders are no longer overwritten at the moment of adventure saving

TiBiBa:
**Description**
In this PR we fix an issue where we would overwrite the keyword placeholders in the teacher adventures at the moment of saving due to accidentally storing the adventure **after** parsing the keywords voor validating if it worked as expected. However, we should still perform this check but no longer save the outcome. Only catch the exception if thrown.

**Fixes**
No related issue.

**How to test**
Verify that the dynamic placeholders no work as expected on teacher adventures.


website/teacher.py
```diff
@@ -473,7 +473,7 @@ def update_adventure(user):
                 return gettext('adventure_duplicate'), 400
 
         try:
-            content = body['content'].format(**hedy_content.KEYWORDS.get(g.keyword_lang))
+            body['content'].format(**hedy_content.KEYWORDS.get(g.keyword_lang))
```
Felienne, line 476: maybe add a comment here why this check is needed? This looks a bit magical :)

TiBiBa: Added!

Felienne (approved):
Works!!

image

PR #3087, opened 2022-07-21
[FEATURE] Enables teachers to directly add adventures to classes

### LINKED ISSUE #3083: [UI idea] Add classes to adventure UI
**Idea incl level**

While testing #3066 I realized it is a bit of a hassle to add adventures to classes because you will have to go back to the generic teachers UI.

What if we add a checklist of all classes below a new adventure and says "Add this adventure to this class right now"?

image

TiBiBa:
**Description**
In this PR we add a nice feature to how class customizations work. From now on teachers are able to directly store their custom adventure in their classes. This seems like a simply task, but due to our data structuring was quite complex to code. We have to retrieve all classes as well as all customizations on each adventure call; which is quite an expensive operation. However, we expect that this has not much influence on performance as this page will not be loaded often.

Adventures can be directly checked or unchecked on classes to automatically add them to the customizations. If customizations do not exist (yet) they are created with only the teacher adventures with a value. This is automatically handled correctly by the back-end when the end user retrieves the coding page. Extra tests are written to verify that everything works as expected.

**Fixes**
This PR fixes #3083.

**How to test**
Make sure you are a teacher and create (or edit an already existing) adventure. Notify the new buttons on the bottom of the page, representing all your current classes. Verify that you can select and de-select these classes which will automatically customize the classes with your current adventure.


TiBiBa:
Converted to draft until I'm happy with the UI, functionality is already done and can be tested.

Felienne:
Ow no.... conflicts 😭 (cc @TiBiBa)

TiBiBa:
All conflicts should be fixed!

Felienne (approved):
Great, this will make it a lot easier for teachers!!

Felienne:
@TiBiBa, this can be merged if the conflicts are fixed

PR #3090, opened 2022-07-21
[FEATURE] Enables users to toggle password visibility on signup

### LINKED ISSUE #3082: [UI idea] Enable users to toggle visibility of password
**Idea incl level**
It would be nice if we enable user to toggle the visibility of the password input on signup and password change.

TiBiBa:
**Description**
In this PR we make a nice QoL change where we add a feature to enable users to toggle the visibility of the password when signing up. As we don't have a password repeat anymore this is a nice addition to enable users to verify that their entered password is what they want it to be.
Schermafbeelding 2022-07-21 om 14 15 51
Schermafbeelding 2022-07-21 om 14 16 01

**Fixes**
This PR fixes #3082.

**How to test**
Navigate to the signup page. Notice the new icon on the password input field and verify that pressing it toggles the symbol itself as well as the type of the field, either showing or hiding the password.


boryanagoncharenko (changes requested):
When I try to sign up, I get the error "The repeated password does not match". My guess is that the /auth/signup endpoint would require changes too.
Screenshot 2022-07-21 at 15 43 05

I noticed one more thing, super tiny and probably not worth the effort, but just to let you know. The lastpass generate password icon overlaps with the show-hide-password icon in the password field. The show-hide-password icon is the one that works, so I can show and hide the password. I am mentioning it because usually, the lastpass generate icon appears before the show-password icon. This is what I see:
Screenshot 2022-07-21 at 15 48 09

Felienne (approved):
Nice improvement to the signup process!

PR #3091, opened 2022-07-21
[BUG FIX] Still reload or redirect when closing the achievement pop-up manually

### LINKED ISSUE #3069: [UI idea] Class page does not refresh after "strafwerk" achievement
Ok this is tiny but if you remove a student for the first time you get an achievement but then the page does not refresh and the student remains in view (but is deleted correctly)

TiBiBa:
**Description**
In this PR we fix an issue where the reloading or redirecting of the window would not happen after an achievement pop-up if the user decided to close the pop-up before the timeout. We fix this by setting an attribute on the pop-up itself and when closing verifying if there is an attribute. If so we perform the action stored there, making sure that we still perform the requested action.

**Fixes**
This PR fixes #3069.

**How to test**
Make sure you get an achievement that contains a re-direct or re-load. Most teacher related achievements have this, but the first one that comes to mind is "detention!" (remove a student from your class). Verify that the reload still happens when closing the achievement before the timeout runs out.


TiBiBa:
Hi @Felienne! This should nicely fix the issue, without a force-reload but only reloading/redirecting if that was already the expected flow of the user.

Felienne (approved):

Felienne (commented):
Perfect, thanks!

PR #3092, opened 2022-07-21
rename due to the addition of the color command

thjazi39:
Rename the "color" variable to "favorite_color" in order not to interfere with the color command that has been introduced.
The change was made only in the English version (weblate will take care of the other languages). As well as in the test file.

- Before : image

- After : image

**How to test**
check that everything is working normally


Felienne:
Great catch @thjazi39!! I will merge this when weblate is merged

Felienne (approved):

PR #3093, opened 2022-07-21
Upgrade font awesome from version 4 to version 6

### LINKED ISSUE #3075: [CHORE] Update the font-awesome dependency
**Description**
Currently we are still on version 4 where there is already a version 6. I'm not sure how we can update this library, but it would be nice to stay up-to-date as it has happened multiple times that symbols are not available for use.

TiBiBa:
**Description**
In this PR we upgrade Font Awesome from version 4 to version 6. As we already had a manually created directory for this we decided to simply overwrite these files. We remove some redundant files that we don't need to minimize the file sizes.

**Fixes**
This PR fixes #3075.

**How to test**
There should be no difference from a user perspective. Verify that all icons on the menu bar are still as expected.


Felienne (approved):
Does not break the icons!
image

PR #3102, opened 2022-07-23
Fix quiz choice using backtick which is not rendered correctly.

Rezwin:
**Description**

This PR fixes single backtick on quiz options that are not rendered correctly.

https://github.com/Felienne/hedy/issues/2269

**How to test**

Go to `http://0.0.0.0:8080/hedy/4#quiz` and check the rendered options.

**Checklist**
  

content/quizzes/en.yaml
```diff
@@ -663,7 +663,7 @@ levels:
             mp_choice_options:
             -   option: |
                     ```
-                    {print} `hello`
+                    {print} (hello)
                     ```
                 feedback: "This quotation mark is skewed, you need a straight one."
```
arjank, line 668: This feedback doesn't match the updated code.

arjank (changes requested):
I'm not sure whether changing from backticks to parentheses is the best option here. True, backticks are not shown, but there could be solutions where the backticks are still used and also shown. It might be confusing to explicitly show them as incorrect here, even though they are required later. But I'll leave that call to @Felienne .

arjank:
The backticks in the answers are removed in line 94 of `quiz.ts`. If that's changed to only remove sequences of 3 backticks it solves the issue.

Felienne:
> I'm not sure whether changing from backticks to parentheses is the best option here. True, backticks are not shown, but there could be solutions where the backticks are still used and also shown. It might be confusing to explicitly show them as incorrect here, even though they are required later. But I'll leave that call to @Felienne .

Yeah I agree that the backticks are a better wrong answers since they are closer to `'`, but I am not an expert in the rendering and can't really estimate if it will be hard to fix that (and @TiBiBa is on vacation)

arjank:
> > I'm not sure whether changing from backticks to parentheses is the best option here. True, backticks are not shown, but there could be solutions where the backticks are still used and also shown. It might be confusing to explicitly show them as incorrect here, even though they are required later. But I'll leave that call to @Felienne .
> 
> Yeah I agree that the backticks are a better wrong answers since they are closer to `'`, but I am not an expert in the rendering and can't really estimate if it will be hard to fix that (and @TiBiBa is on vacation)

I can tell you that this is easy to fix. And I know, because I already did the research.
> The backticks in the answers are removed in line 94 of `quiz.ts`. If that's changed to only remove sequences of 3 backticks it solves the issue.

Felienne:
Cool then let's close this PR in favor of a new one that correctly renders the ticks. Do you want to take a stab at that @arjank?

Rezwin:
I've been trying to fix it with the solution @arjank proposed before.

Following line in `quiz.ts` seems to be the culprit

`editor.setValue($.trim(options[i-1].option.replace(new RegExp('`', 'g'),"")));`

I tried replacing the single backtick with three backtick as @arjank proposed but for some reason I still see the same issue on my local build.

It is my first time with Hedy so I might be doing something wrong. I've made the changes and ran following commands but that doesn't seems to be enough

```
build-tools/heroku/generate-typescript
python app.py
```

Is there anything else I need to try local changes ?

Felienne:
> ```
> build-tools/heroku/generate-typescript
> python app.py
> ```
> 
> Is there anything else I need to try local changes ?

That should work but sometimes the browser caches things. Have you set your browser to no caching (you can do this in Chrome in the developer tab)

arjank:
> ```
> build-tools/heroku/generate-typescript
> python app.py
> ```
> 
> Is there anything else I need to try local changes ?

That should be enough, but browser caching can be annoying. You also might want to use `build-tools/heroku/generate-typescript --watch` in development, so it regenerates the js files immediately when you save changes. (And you could deliberately save a broken file to see that it fails building that.)

You should use `build-tools/heroku/generate-typescript` (without the `--watch`) before you commit app.js.

Rezwin:
Thanks. It seems to work well! 

Do I need to include the generated files - `appbundle.js` and `appbundle.js.map` in the commit ? Or just changes to `quiz.ts`

arjank:
> Thanks. It seems to work well!
> 
> Do I need to include the generated files - `appbundle.js` and `appbundle.js.map` in the commit ? Or just changes to `quiz.ts`

You need to include all three files in the commit. And yes, appbundle.js and appbundle.map.js could give conflicts. In that case, merge main into your branch and regenerate those files.

Felienne:
Cool! The generated files will be added automatically.

Felienne:
> > Thanks. It seems to work well!
> > Do I need to include the generated files - `appbundle.js` and `appbundle.js.map` in the commit ? Or just changes to `quiz.ts`
> 
> You need to include all three files in the commit. And yes, appbundle.js and appbundle.map.js could give conflicts. In that case, merge main into your branch and regenerate those files.

So, yes and no. It is not really needed since they will be generated automatically, but we typically do include them as a service to reviewers so they don't have the regenerate the js, some contributors might be Python only/mainly and might not have the tools installed to generate themselves.

arjank:
> So, yes and no. It is not really needed since they will be generated automatically, but we typically do include them as a service to reviewers so they don't have the regenerate the js, some contributors might be Python only/mainly and might not have the tools installed to generate themselves.

Thanks for the clarification :smiley: We could add a git hook to generate those files on checkout, but that still requires people to have all the tools installed (and probably keep them up-to-date), so I understand that it's easier to keep them included.

Rezwin:
Thanks @Felienne and @arjank. I've updated the PR with the required changes. Please review when possible.

arjank (approved):

Issue #3103, opened 2022-07-24
[HIGHLIGHTING] Arabic comma is not highlighted


Felienne:
Arabic comma is not highlighted:

<img width="1281" alt="image" src="https://user-images.githubusercontent.com/1003685/180663480-b6a2c1fc-b286-4daf-9d7c-2c384768dd3a.png">

PR #3104, opened 2022-07-24
highlights Arabic comma?

Felienne:
Hi @thjazi39!

I think this is a right fix for #3103, but: 

1) can you check it 
2) can we make it more robust so that we have a list of comma's somewhere that can easily be changed in one place (more languages will have more comma's) 
3) can we make this file be uft8, it is really hard to read Arabic characters like this. I understand this is easier for you, but it is not easier for everyone else that reads a bit of Arabic.


highlighting/highlighting.json
```diff
@@ -429,7 +429,7 @@
                     ]
                 },
                 {
-                    "regex": ",",
+                    "regex": ",|\u060c",
```
Felienne, line 432: this is really tricky to read!

Felienne:
ok re point 3, I changed the encoding but I guess we will manually have to make the \u060c back into the arabic comma and while we are added also change the Tatweel. Can you do that?

thjazi39:
> 2. can we make it more robust so that we have a list of comma's somewhere that can easily be changed in one place (more languages will have more comma's)

I think the best solution will be to add a key in the yaml files.

> 3. can we make this file be uft8, it is really hard to read Arabic characters like this. I understand this is easier for you, but it is not easier for everyone else that reads a bit of Arabic.

The file is already in UTF8, so it seems to me that if we put directly the symbol instead of the utf8 code it will still work (python works with it after)

thjazi39:
> ok re point 3, I changed the encoding but I guess we will manually have to make the \u060c back into the arabic comma and while we are added also change the Tatweel. Can you do that?

ok

Felienne:
> > 2. can we make it more robust so that we have a list of comma's somewhere that can easily be changed in one place (more languages will have more comma's)
> 
> I think the best solution will be to add a key in the yaml files.
> 
> > 3. can we make this file be uft8, it is really hard to read Arabic characters like this. I understand this is easier for you, but it is not easier for everyone else that reads a bit of Arabic.
> 
> The file is already in UTF8, so it seems to me that if we put directly the symbol instead of the utf8 code it will still work (python works with it after)

Yes I added an extra commit to add the encoding, because earlier file converted the symbol.

Felienne:
> I think the best solution will be to add a key in the yaml files.

There already is a comma symbol in the yaml files of the keywords! 

https://github.com/Felienne/hedy/blob/6c42e7e46476a338f1d488e9e41282ec411b0088/content/keywords/ar.yaml#L21

You could gather those, and be sure to always allow for the En comma in all languages as we do for keywords.

Felienne:
> > I think the best solution will be to add a key in the yaml files.
> 
> There already is a comma symbol in the yaml files of the keywords!
> 
> https://github.com/Felienne/hedy/blob/6c42e7e46476a338f1d488e9e41282ec411b0088/content/keywords/ar.yaml#L21
> 
> You could gather those, and be sure to always allow for the En comma in all languages as we do for keywords.

And clearly it would be ideal to do this for other symbols too but we can do that in a separate PR.

Felienne:
I don't really understand this change, if it needs to go in, can you make a
PR?

Felienne:
Closing this in favor of #3107

PR #3106, opened 2022-07-25
Solving the problem of over-coloring

### LINKED ISSUE #2279: Don't highlight ask/vraag at the beginning of the line in level over 3

thjazi39:
To solve the problem of over highlighting I added a system that allows to add some rules, which allows to avoid highlighting a word before `ìs`
This is to be used with caution, because it can interfere with other rules that are generated, and create highlighting problems.

fix #2279 

**Result**
- before : image
- after : image

**How to test**
Check that the `ask` keyword is not recognized before `is`

**Addition in the test system**
Because you have to wait until you have an `is` keyword for the `ask` keyword not to be colored, I added an option in the test functions to disable the intermediary tests (consistency during word input)

image


Felienne (changes requested):
The coloring of `ask` at the beginning of the line seems to indeed work but it is now also highlighted as a variable:

image

Also, probably a really small thing but we get an error in the js:

image

thjazi39:
Now it should work!

It even works during line input!
image

Felienne:
> Now it should work!
> 
> It even works during line input! image

Cool, will test this tomorrow!!

Felienne (approved):
Cool, this seems to work well now, tested for Dutch and English:

image

image

PR #3107, opened 2022-07-25
add translation of comma

### LINKED ISSUE #3103: [HIGHLIGHTING] Arabic comma is not highlighted

thjazi39:
This PR adds the translation of commas in the commands
Fixes #3103, improvement of #3104


thjazi39:
The K() function allows to use the translation of a keyword if it exists.

So I added in the list of commands to translate (TRANSLATE_WORD) the comma (changing `'` to `comma` because that's how the key is named in yaml files).

Felienne:
> The K() function allows to use the translation of a keyword if it exists.
> 
> So I added in the list of commands to translate (TRANSLATE_WORD) the comma (changing `'` to `comma` because that's how the key is named in yaml files).

Ah I guess we merge this the js error will be solved since K will be defined :)

highlighting/highlighting-trad.json
```diff
@@ -8,6 +8,7 @@
         "blue": "\u0640*\u0627\u0640*\u0632\u0640*\u0631\u0640*\u0642\u0640*|blue",
         "brown": "\u0640*\u0628\u0640*\u0646\u0640*\u064a\u0640*|brown",
```
Felienne, line 9: Can we please make this uft8 and regular Arabic letters also this is so very hard to read!

highlighting/rules_automaton.py
```diff
@@ -150,8 +150,8 @@ def rule_level3():
         'regex': START_WORD + K("at") ,
```
Felienne, line 150: Can we call this K() function something more descriptive like `translate` or `get_translated_keyword`?

Felienne (changes requested):
Cool, this indeed works and is a lot more flexible than my attempt!!

image

I would just like the code to be a bit more readable in terms of the function name and the Arabic unicode characters

highlighting/definition.py
```diff
@@ -60,7 +60,8 @@
 	"input",
 	"or",
 	"while",
-	"length"
+	"length",
+	"comma"
```
Felienne, line 64: Can't comment on line 71 and 77 but it is unclear to me shy we have a difference with and without groups, can you add that in a comment?

highlighting/highlighting-trad.json
thjazi39: yes, but since these 2 files are generated by python, I will try to see if there is an option for that

highlighting/rules_automaton.py
thjazi39: I can do it, but it should be a short name so as not to make the lines too long (which will reduce readability)

I'll try with `translate`

highlighting/definition.py
thjazi39: I will add comments on this function to improve the readability.

To make a long story short, there is an option `withoutGroup` which allows you to specify if you want the result to be a group (with parenthesis) or not.

Most of the time, since we handle keywords separately, we want a keyword (and its translations) to form a group (hence the default value)

But in very rare cases, we want to make the groups manually (like for colors), and that's what the option is for.

highlighting/highlighting-trad.json
Felienne: Yes there is, something like this probably: https://www.tutorialspoint.com/How-to-read-and-write-unicode-UTF-8-files-in-Python

highlighting/definition.py
```diff
@@ -66,15 +66,21 @@
 
 
 TOKEN_CONSTANT = "text"
+# 
+# 
+def translate(word, withoutGroup = False):
+	""" Function that allows to add tags around the keywords to be translated.
```
Felienne, line 72: tags is `__`? (called dunderscore in Python)?

highlighting/highlighting-trad.json
thjazi39: It's done !

highlighting/definition.py
thjazi39: yes !

highlighting/rules_automaton.py
thjazi39: it's done !

Felienne:
Sadly now this has conflicts @thjazi39, could you take a peek?

thjazi39:
image

The merge is done, and I added tests for the Arabic comma

highlighting/definition.py
```diff
@@ -60,20 +60,27 @@
...
+	"comma"
 ]
 
 
 TOKEN_CONSTANT = "text"
 
-def K(word, withoutGroup = False):
+
+def get_translated_keyword(word, withoutGroup = False):
+	""" Function that allows to add tags around the keywords to be translated.
```
Felienne, line 72:
```suggestion
	""" Function that allows to add double underscores around the keywords to be translated.
```

highlighting/definition.py
```diff
@@ -60,20 +60,27 @@
...
 ]
 
 
 TOKEN_CONSTANT = "text"
 
-def K(word, withoutGroup = False):
+
+def get_translated_keyword(word, withoutGroup = False):
+	""" Function that allows to add double underscores around the keywords to be translated.
+		The "__" tags are added before and after only if the keyword belongs to the list.
```
Felienne, line 73:
```suggestion
		The "__" are added before and after only if the keyword belongs to the list.
```

Felienne (approved):
image

highlighting/highlighting-trad.json
Felienne: I still seem to see the unicode numbers but we can do that in a separate PR. Might be a tool issue.

PR #3108, opened 2022-07-26
Refactor tests for level 12 #2436

### LINKED ISSUE #2436: [IMPROVEMENT] Refactor tests

boryanagoncharenko:
**Refactors tests for level 12**
_Adds missing tests from previous levels and removes duplicate tests_
_Adapts the test file to conform to the repo standards_

**Fixes #2436**

**How to test**

* All tests should pass 
* There should be no duplication. There are 2 tests which are commented out because they will be addressed in a separate PR, but otherwise there should be no dead code. The naming should be in sync with the rest of the tests.


jpelay (approved):

boryanagoncharenko:
> @boryanagoncharenko I loved the changes! There are a few tests where you didn't use the new `input_transpiled` abstraction, specifically in the first few ones of ask, maybe this was intentional! Either way, I think this is ready to go!

Thanks for the review! Not using `input_transpiled` was not intentional and I will address together with the commented out tests. Thanks for pointing it out.

PR #3112, opened 2022-07-26
[BUGFIX] Fixes list access using indices on level 5+

### LINKED ISSUE #3110: [BUG] Can't access lists using indices on level 5+
**Describe the bug**
Starting from level 5, it is allowed to access a list using an int, but the code rule is not working due to a problem in the function that process the production rule.

For example, the following code
```
friend is Hedy, Lola, Frida
x is friend at 2
print x
```
 Gives the following error:
 
image

**Expected behavior**
The code should compile down to a list access

jpelay:
**Description**

Changes de functions that process the `list_access_var` rule to be able to parse correctly lists access using indices.

**Fixes #3110 **

**How to test**

I added a test on level 5 to test this behavior.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -1469,10 +1469,10 @@ class ConvertToPython_4(ConvertToPython_3):
 class ConvertToPython_5(ConvertToPython_4):
     def list_access_var(self, args):
         var = hash_var(args[0])
-        if args[2].data == 'random':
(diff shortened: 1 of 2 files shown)


Felienne (approved):

Felienne:
Thanks for yet another fix to the grammar @jpelay!!!

PR #3114, opened 2022-07-27
Fix failing translations for hedy exceptions #1804

boryanagoncharenko:
**Fix failing translations for all HedyExceptions**

* _Changes the translations of exceptions which did not have correct keywords and failed with KeyError when translated_
* _Adds automated tests that ensures no subclasses of HedyException fails because of wrong arguments. The tests are parametrized to automatically include newly added languages and exceptions._

**Fixes #1804 **

**How to test**

* To ensure the tests are passing, some of the already translated error messages had to be changed back to English. Is this ok? Maybe there is a nicer way to adapt the translation so that we do not completely lose it?
* Check how the tests are created. The idea of the tests is that they will be automatically created when a new language or exception is created but is the code easy to understand? Is the name of each individual test clear?

PR #3116, opened 2022-07-28
[LANGUAGE] Changes Github Workflow to execute deployment script

### LINKED ISSUE #3070: [Language] Prevent tests breaking on main with new else keyword
Ok here is an interesting bit of process knowledge that we need to document somewhere. Recently, @Mark-Giesen merged a PR (#3035) for which all tests passed, so he did not do anything wrong! However... it broke main. How?

Because... this commit in #3035 translated `else` to Hebrew: https://github.com/Felienne/hedy/pull/3035/commits/1560b109a8d47e99df9c74509c50e23754dd9b06

At that point in time, all is well. But, once we run the deploy script, we regenerate the grammars and suddenly, we start to parse Hebrew examples and that causes an issue (more on the concrete issue later-on, in point 2)

To fix this:

1) We should change https://github.com/Felienne/hedy/blob/main/.github/workflows/unittests.yml so that the `on-deploy.sh` script is ran before the tests run (or at least that `python ../../content/yaml_to_lark_utils.py` runs) so the PR fails, instead of main.

2) So... why does this even break? Because the grammars need magic for else to work!

We need to escape all else keywords here:
https://github.com/Felienne/hedy/blob/12031e693fc0cfcaa7bcba12404dc5330a35d0c3/grammars/level5-Additions.lark#L34

And in a slightly different from here:
https://github.com/Felienne/hedy/blob/12031e693fc0cfcaa7bcba12404dc5330a35d0c3/grammars/level5-Additions.lark#L37

Ideally, we add these additions to the yaml_to_lark_utils script so it goes automatically! But maybe that is not worth in, since we do not add languages all too often and if it breaks on the PR, I can easily fix by hand (as I have been doing for now)

To give more context, a concrete example change of the fix is here: https://github.com/Felienne/hedy/pull/3054/commits/911c02476db30eff76ae66f7da3c7e13cc169efa

jpelay:
**Description**

Adds, as requested by @Felienne on issue #3070, a extra step on the Github Workflow to execute the creation of the grammars before the tests run. That way, main won't fail if a new language is added, since we need to modify several parts of the grammar manually.

**Fixes #3070**

**How to test**

**Checklist**
  

jpelay:
There's an e2e test failing, but I don't understand why, doesn't seem to have anything to do with the deployment script?

Felienne:
Weird! I don't know the e2e tests so well, @TiBiBa mostly does but he is on vacation. He can check this when he is back.

.github/workflows/unittests.yml
```diff
@@ -28,5 +28,8 @@ jobs:
     - name: Install Nodejs dependencies
       run: |
         npm ci
+    - name: Run Deployment Script
+      run: |
+        ./build-tools/heroku/on-deploy.sh
```
Felienne, line 33: Do we need the whole script? For the bug we are addressing, maybe `python ../../content/yaml_to_lark_utils.py` is enough? (We can try to see if that fixes the e2e test, even though I am also curious why that happens)

jpelay: That solved the problem indeed!

Felienne:
I would still be curious about the e2e tests failing but this is nice!

Felienne:
> I would still be curious about the e2e tests failing but this is nice!

@TiBiBa I think we have the same e2e issue here as in #3114?

Felienne (approved):

PR #3118, opened 2022-07-31
Improve program reset on stop

### LINKED ISSUE #820: Double print when re-running `ask` command
There is a specific pattern that causes my code to run twice. 
1. I run code with an `ask` command in level 1-4
2. I do not answer the question, instead I modify the code a bit so I can re-run it
3. I run the code again, this time I do answer the question
4. Now every print statement in my code is executed twice.

This is not a huge issue but it is confusing for students.

arjank:
**Description**

Improves reset when programs are stopped. When the stop button is clicked, all running timeout timers are cancelled. And when the user starts editing while a program is running, it is stopped and we return to the normal state.

**Fixes #820**

**How to test**

Samples taken from the issue ticket:
```
name is ask What is your name?
print hi, name
```
Execute the code, but when the question input field pops up, don't enter anything. Instead update the code, (stop and) restart the program, now do answer the question, and verify that the output is produced exactly once

Second example code:
```
print line 1
sleep 4
print line 2
sleep 4
print line 3
```
Run the program, stop and restart it (no changes needed) and verify that the output contains exactly 3 lines, in the correct order

**Checklist**
  

boryanagoncharenko (approved):
I tested all scenarios I could think of and they are all working! Thanks for this fix!

PR #3123, opened 2022-08-08
[FEATURE] Allows teachers to duplicate classes

### LINKED ISSUE #3115: [UI idea] Allow teachers to duplicate classes
**Idea**

Teachers will probably want to create a class with certain assignments that are enabled, but then teach that in multiple groups (like 3 times a grade 7). Could we make a feature that allows teachers to duplicate a class?

TiBiBa:
**Description**
In this PR we introduce a new feature, enabling teachers to duplicate classes. This can be useful when having an already customized classes and wanting to copy this for another class. When duplicating all customizations are automatically copied to the new class.

**Fixes**
This PR fixes #3115.

**How to test**
Make sure you are a teacher and navigate to the `/for-teachers` page. Notice the new duplicate option and verify that it works and copies all customizations.


Felienne:
Nice work @TiBiBa! I do find the icons a bit confusing, I seem to see two:

image

Also, the text is a bit confusing, maybe we can say: "Please enter a name of the new class" instead of "Please enter a name of the duplicated class"? 

And maybe "duplicate" could be "copy customizations to new class" because a teacher might think they can copy customizations to an existing class, but when I enter a duplicated name, I get a second class:

image

I guess we don't want that and we want to check for a duplicate name in any case and give the regular error message.

Felienne:
Also @TiBiBa conflicts 😭

Felienne (changes requested):
See comments above

TiBiBa:
> Nice work @TiBiBa! I do find the icons a bit confusing, I seem to see two:
> 
> image
> 
> Also, the text is a bit confusing, maybe we can say: "Please enter a name of the new class" instead of "Please enter a name of the duplicated class"?
> 
> And maybe "duplicate" could be "copy customizations to new class" because a teacher might think they can copy customizations to an existing class, but when I enter a duplicated name, I get a second class:
> 
> image
> 
> I guess we don't want that and we want to check for a duplicate name in any case and give the regular error message.

I couldn't find a "duplicate" emoji so used the "copy" and "paste" ones. Will look into this, maybe Fontawesome has a nice alternative. will also change the strings.

Felienne:
> I couldn't find a "duplicate" emoji so used the "copy" and "paste" ones. Will look into this, maybe Fontawesome has a nice alternative. will also change the strings.

Maybe something like this:

image

Ow and also make sure we check for duplicate class names (might have been lost in all my remarks above, sorry was a bit chaotic)

TiBiBa:
Should all be fixed!
Schermafbeelding 2022-08-09 om 10 20 24

TiBiBa:
Didn't change the header name (yet). I agree that "duplicate" can be a bit confusing, but making it an explanatory string will mess up the table layout.

Felienne:
> Didn't change the header name (yet). I agree that "duplicate" can be a bit confusing, but making it an explanatory string will mess up the table layout.

Ok I see! Maybe we can put a tiny (i) icon there that pops up an explanatory text a la achievement modals (in a separate PR)

Felienne:
Better!! But now maybe we also want an icon for "view" so the whole line is icons?

TiBiBa:
> Better!! But now maybe we also want an icon for "view" so the whole line is icons?

Fixed!

translations/en/LC_MESSAGES/messages.po
```diff
@@ -671,7 +671,8 @@ msgstr "Hedy logo"
 msgid "back_to_class"
 msgstr "Go back to class"
 
-#: templates/class-overview.html:14 templates/for-teachers.html:39
+#: templates/class-overview.html:14 templates/for-teachers.html:34
+#: templates/for-teachers.html:41
 msgid "class_name_prompt"
 msgstr "Please enter the name of the class"
```
Felienne, line 677: Might be a tad more clear?

```suggestion
msgstr "Please enter the name of the new class"
```

Felienne (approved):
Great!

PR #3125, opened 2022-08-08
[UI FIX] Improves readability of achievements overlay

### LINKED ISSUE #3122: [UI idea] Change achievements overlay class to only make background transparant
**Idea incl level**
While working on a personal project I found that Tailwind also supports the `bg-opacity-<value>` class. This enables us to only make the background of an element transparant instead of the whole element. We currently use the "normal" opacity class for our achievements overlay, making the text a bit transparent as well. This should be changed to using the `bg-opacity-<value>` class

TiBiBa:
**Description**
In this PR we improve the readability of the achievements overlay by only making the background transparant instead of the whole element. Currently the overlay text is also partly transparant, making it harder to read. After this fix we only set this value to the background making the text still good readable.

**Fixes**
This PR fixes #3122.

**How to test**
Make sure you are logged in and navigate to the achievements page. It might be a bit hard to see, but the text should be better readable on the overlay.


Felienne (approved):
Clear difference:

Old:

image

New:

image

PR #3126, opened 2022-08-08
[BUG FIX] Language switcher dropdown works again

### LINKED ISSUE #3109: [UI] Language switcher does not fold out anymore
Not sure why this happened (I think maybe with the rewriting of the front-end accessibility stuff?) but the language switcher has stopped working, it does not fold out anymore:

image

There are no errors shown in the developer output in Chrome so I have no hints to why this happens.

TiBiBa:
**Description**
In this PR we fix the language switcher dropdown menu. Which was broken due to the re-structuring of the `dropdown` class we performed for improving the responsiveness of the main menubar. We forgot the dependency on this class by the language switcher and so on broke it.

**Fixes**
This PR fixes #3109

**How to test**


Felienne (approved):
Works!

PR #3132, opened 2022-08-08
[BUG FIX] Auto selects first tab when intro tab doesn't exist

### LINKED ISSUE #3088: [BUG] First available adventure not auto-selected after class customizations
**Describe the bug**
When several adventures are disables through class customizations the first available adventure is not automatically selected once a student visits the `/hedy` page. See the screenshot below for clarification.
Schermafbeelding 2022-07-21 om 12 45 27

**Expected behavior**
The first available adventures should always be auto-selected.

### LINKED ISSUE #3131: [BUG] Parsons tab does not have a direct link
All tabs have one such as: https://hedycode.com/hedy#next but Parson's done not get one:

image

(if I click it is still stuck on #next):
 
image

TiBiBa:
**Description**
In this PR we improve the tab handling on load. We currently have the following flow:
- Auto select the `intro` tab
- If there is a `window.State.adventure_name` -> select that one
- If there is direct URL -> select that one

In this PR we add an additional option if all the above are not accounted for. Because in the situation of class customizations it can occur that a level does not have the intro tab, making it that there is no tab selected. In this case we add the following check:
- If all of the above are not accounted for: switch to the first tab that is there

We also fix a minor issue where the parsons tab was not added as the hash value correctly in the URL.

**Fixes**
This PR fixes #3088 and fixed #3131.

**How to test**
Customize a class and make sure the intro tab is not there. Then navigate to the class directly without selecting an adventure or a direct hash. In this case the first visible tab should always be selected.


Felienne (approved):
I can confirm the selection is working:

image

Works for Parsons tab too!

image

Issue #3128, opened 2022-08-08
[UI idea] Better display and explain localized keywords within adventures


TiBiBa:
**Idea incl level**
A description of your idea, for example: "I would like to be able to save and freeze programs so they cannot be edited"

**Additional context**
Add any other context about the idea here.

TiBiBa:
Current idea: Add a toggle to the new dropdown enabling the user to toggle between both languages.

PR #3135, opened 2022-08-09
Adds keywords translations dropdown to code page UI

TiBiBa:
**Description**
In this PR we add a new dropdown element to the code page UI to make the translatable keywords more visible. The dropdown menu contains the English keywords as well as the translated keywords (if they exist). In the case the user selects English or a non-keyword language the dropdown is hidden.

**Fixes**
This PR addressed (but does not fully fix) #3128

**How to test**
Navigate to the code page and notice the new dropdown menu. Select both a supported and non-supported keyword language. Verify that the dropdown is hidden when no keywords are translated. Verify that this works for multiple languages and levels. Also verify RTL support by selecting Arabic (or any there RTL language).


TiBiBa:
Don't understand the failing e2e tests (yet), but otherwise this PR is good to go.

TiBiBa:
Just found why the e2e tests failed, should be fixed now!

TiBiBa:
@Felienne we do have to discuss if we want to further improve the visibility of the translated keywords. If not we can close #3128 with this PR.

Felienne:
> @Felienne we do have to discuss if we want to further improve the visibility of the translated keywords. If not we can close #3128 with this PR.

Yeah I think I would like to have a "switch" like the Developer's mode switch to switch between Dutch or English keywords. What do you think about putting that switch in the fold out menu that this PR adds?

TiBiBa:
> > @Felienne we do have to discuss if we want to further improve the visibility of the translated keywords. If not we can close #3128 with this PR.
> 
> Yeah I think I would like to have a "switch" like the Developer's mode switch to switch between Dutch or English keywords. What do you think about putting that switch in the fold out menu that this PR adds?

Sounds good! Will look into this

Felienne (approved):
Looks good, will help kids find this!!

(apart from the conflicts of course hahaaaaa, not sure weblate is making our life better)

TiBiBa:
Will fix the conflicts and merge this!

PR #3137, opened 2022-08-09
[LANGUAGE] Remove punctuation in grammar and class

### LINKED ISSUE #279: Small issue with symbol + quote
A small quirk that happens after merging #275:
`print hello world!`' yields `print hello world! '` (note the space between the exclamation mark and the quote). This happens on levels 1 & 2 and only with the combination symbol + quote. Quote + symbol works fine, as well as normal character + quote. It happens both with single and double quotes.

### LINKED ISSUE #2549: [BUG] Level 2 variables don't parse correctly when using punctuation in print statements
When trying to use punctuation in print statements in level 2, the variables will not parce properly. The variable name will print instead of its contents when a comma or other punctuation is directly attached to the variable.

Example code:
`name is test
print name, heya!`

Expected output:
`test, heya!`

Actual output:
`name, heya!`

Could maybe be fixed by ignoring punctuation when parcing variables

Felienne:
Simplify the grammars, which I have wanted to do since I introduced the silly separate rule in early 2020 haha!! 

This makes internationalization a lot easier since we don't rely anymore on specific symbols and fixes #279 and fixes #2549.


Felienne:
In addition to verifying https://github.com/Felienne/hedy/issues/2549 I should also look at the importing of the new regex lib (this is a note to self!)

Felienne:
Want to document these interestings bit of non-accessibility here!!

1) 
image (From: https://stackoverflow.com/questions/1832893/python-regex-matching-unicode-properties)

This is why I had to switch over to the new library 

2) Bengali does not seem to be properly recognised: 
image (from: https://stackoverflow.com/questions/68800391/split-a-bengali-string-by-a-regex-without-consuming-the-delimiter-zero-width)

which is why we need the "script=Bengali" part.

hedy.py
```diff
@@ -1310,29 +1301,26 @@ def turn(self, args):
...
+        for a in args:
+            # list access has been already rewritten since it occurs lower in the tree
+            # therefore we should not split it anymore and thread it as 1 variable:
+            if "random.choice" in a or "[" in a:
+                args_new.append(self.process_variable_for_fstring(a))
             else:
-                space = " "
-
-            argument_string += self.process_variable_for_fstring(argument) + space
+                res = regex.findall(r"[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}]+|[^\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}]+", a)
```
boryanagoncharenko, line 1320: This is a nice, creative approach I would not have thought of!

hedy.py
```diff
@@ -1310,29 +1301,26 @@ def turn(self, args):
...
 
-        for argument in args:
-            # final argument and punctuation arguments do not have to be separated with a space, other do
-            if i == len(args)-1 or args[i+1] in self.punctuation_symbols:
-                space = ''
+    def print(self, args):
+        args_new = []
+        for a in args:
+            # list access has been already rewritten since it occurs lower in the tree
+            # therefore we should not split it anymore and thread it as 1 variable:
```
boryanagoncharenko, line 1316: This is a super minor thing, but I am not completely sure what this comment means. Do you mean that any list access would appear as a child node of the current print node, so by the time we reach this piece of code, the list access would already be transpiled, so we could treat it as a variable?

boryanagoncharenko (approved):

grammars/level1.lark
```diff
@@ -58,9 +58,10 @@ textwithoutspaces: /([^\n #]+)/ -> text //anything can be parsed except for a ne
 // FH Sept 2021: More info on this variable format: https://www.unicode.org/reports/tr31/tr31-1.html
 // Exact grammar stolen from: https://lark-parser.readthedocs.io/en/latest/classes.html
 
-NAME: ID_START ID_CONTINUE*
-ID_START: /[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}_]+/
-ID_CONTINUE: ID_START | /[\p{Mn}\p{Mc}\p{Nd}\p{Pc}·]+/
+NAME: LETTER_OR_UNDERSCORE LETTER_OR_NUMERAL*
+
+LETTER_OR_UNDERSCORE: /[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}_]+/
+LETTER_OR_NUMERAL: LETTER_OR_UNDERSCORE | /[\p{Mn}\p{Mc}\p{Nd}\p{Pc}·]+/
```
jpelay, line 64: I like how this is so much clearer now! I was really puzzled the first time I saw this regex and didn't understand the meaning of it!

hedy.py
jpelay: Yeah, this is really smart! And simplifies the grammar!

hedy.py
Felienne: Fair enough! I will make the explanation a bit more clear with an example!

Felienne: (but your interpretation is 100% correct!)

grammars/level1.lark
Felienne: Yeah I just copied the names from the Python definitions but upon revisiting them and being confused 3 times myself I thought: why not make this better for us!

PR #3140, opened 2022-08-09
[BUG FIX] Updating of profile works again

### LINKED ISSUE #3138: [BUG] Unable to update profile due to always invalid birth year
**Describe the bug**
Schermafbeelding 2022-08-09 om 17 39 16

TiBiBa:
**Description**
In this PR we fix an issue where we could no longer signup or update our profile due to a change made in #3114. The casting of the `birth_year` value was removed, but is crucial for it to work as with the front-end clean-up process all data is now sent as strings from the front-end.

**Fixes**
This PR fixes #3138.

**How to test**
Verify that you can signup and update a profile.


Felienne (approved):
Works again!

image

PR #3142, opened 2022-08-09
Allows words with several possibilites to be translated

### LINKED ISSUE #2711: [BUG] repete runs and highlights but can't be translated
**Describe the bug**
Since #2686 we allow multiple forms in the yaml, so now (for example) `repete` in level 8 works and is also highlighted (it used to work but not highlighted) but it can't yet be translated back and forth.

jpelay:
**Description**

Adds the possibility in the translator for words with several options to be translated. For example, in French we allow `repete` and `répète`. We used to be able to just translate `répète` back to other languages, but not repete. Now we can translate both.
For this, in the keyword dictionary I added the possibility of lists, now a keyword can map to a single string or to a list of strings. When we translate back we always take the first option.

**Fixes #2711 **

**How to test**

I added a test with repete to test this.

**Checklist**
  
### DIFF
--- a/hedy_translation.py
+++ b/hedy_translation.py
@@ -24,8 +24,7 @@ def keywords_to_dict(lang="nl"):
     for k, v in command_combinations.items():
         if type(v) == str and "|" in v:
             # when we have several options, pick the first one as default
-            command_combinations[k] = v.split('|')[0]
-
+            command_combinations[k] = v.split('|')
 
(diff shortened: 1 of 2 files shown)


jpelay:
I thought I tested it fully, but I skkiped by accident and some tests are failing :< gonna close it and open it again once its all done.

PR #3143, opened 2022-08-10
Update translations of class customizations to improve readability

### LINKED ISSUE #3095: [UI idea] Make class customization settings consistent
**Idea incl level**
Currently when teachers customize a class they can "hide quiz" or "parsons" which are both contradicting each other in functionality. The "hide quiz" should be selected to hide the quiz while the "parsons" should be selected to actually show the parsons exercises. It would be nice if we made this consistent.

TiBiBa:
**Description**
In this PR we make a minor change to the translations to better explain the class customizations option. Currently the label name for hiding the Parsons puzzels is just "dragging" in English. Making it difficult to understand what the checkbox does. We fix this by renaming it to "Hide puzzle" as we call the Parsons puzzle is all places on the front-end.

**Fixes**
This PR fixes #3095.

**How to test**
Not much to test. Verify that the string is updated (don't forget to compile Babel!).


Felienne (approved):
Better!

PR #3145, opened 2022-08-10
[FEATURE] Enables users to toggle between different keyword languages

### LINKED ISSUE #3128: [UI idea] Better display and explain localized keywords within adventures

TiBiBa:
**Description**
In this PR we introduce a new feature, enabling users to toggle between keywords in English and their profile language if supported. This toggle is implemented in the recently introduced "keyword translation dropdown" containing a table of all English and translated keywords to better help users understand that both English and non-english keywords are valid. On pressing the toggle a new GET request is created with a `?keyword_language=<lang>` parameter, which is caught in the back-end to overwrite the keyword language (if the parameter exists).

**Fixes**
This PR fixes #3128.

**How to test**
Verify both as non-logged and logged in user. Select a non-english language and notice the dropdown menu (which is introduced yesterday). Verify that the toggle switches all adventures keywords between your current keyword language and the other option (always English and your profile language). This should work in any situation for all keyword-supporting languages.

Schermafbeelding 2022-08-10 om 12 28 01
Schermafbeelding 2022-08-10 om 12 28 09


TiBiBa:
This approach seems a bit farfetched, it might be best to use the following idea:
- On the press of the toggle sent a special request to the back-end
- Reload the `/hedy` page with a overrule of the keyword language

Felienne (approved):
Perfect!!! Exactly as I had it in mind!

Issue #1503, opened 2021-12-08
[BUG] unexpected error when printing a variable name before variable definition, level 2 and 3


OnnoEbbens:
**Describe the bug**
When you define a variable in line 2 and you want to print the variable name as text in line 1 you get an error.

**Paste the Hedy code & level**
This code in level 3:
```
print I love animals
animals is dog, cat, mouse
print animals at random
```
raises this error:
```
We can't run your program.

You cannot use the command print with animals because it is a list. Try changing animals to text or a number.
```
This code in level 2 or 3:
```
print I love animals
animals is dog
print animals
```
raises this error:
```
Something went wrong while running the program.

name 'animals' is not defined
```

**Expected behavior**
Variable names that are not defined yet should be printed as text. Although I can imagine that this is hard to implement on the interpretation side. If it to hard an appropriate error for these cases would be nice.

Felienne:
Yeah, we know! This is a duplicate of #165 but it is really hard to fix since it involves program analysis to see where a variable is used. I will try to work on this over Christmas

PR #3146, opened 2022-08-10
Improved unquoted exception

### LINKED ISSUE #832: Reduce the options in the error message generated by printing an unquoted string from two to one
When a kid prints an unquoted string, eg. print naam, there will be an error saying:
"You are trying to print variable name, but you haven't made it yet. OR you forgot the quotation marks."

Reduce these two options to one.

The idea is that we will check before creating the error message if there is a variable that looks a lot like NAAM, and then actually write out "You are trying to print variable name, but you haven't made it yet. " OR "you forgot the quotation marks. " Depending on the situation.

### LINKED ISSUE #1101: Sometimes we do not raise Unquoted?
Level 4:

image

Interestingly enough we do raise it when a question mark is used?

image

Felienne:
Finally fixes #1101 and also fixes #832!


Felienne:
Merging myself 😲 to avoid too many conflicts with #1503

PR #3148, opened 2022-08-10
[FEATURE] Implements report button to enable users to report public programs

### LINKED ISSUE #3147: [UI idea] Create a report button on the /explore page for malicious programs
**Idea incl level**
As discussed in #3044 it would be nice to enable the admin to remove reported programs. The following workflow seems most logical:
- Each program on the /explore page has a "report" button
- When pressed, a mail is sent to the admin with a link
- When the link is clicked: program is removed (if the admin is logged in of course)

TiBiBa:
**Description**
In this PR we add a report flag to the /explore page to enable users to report malicious or undesired programs. The program is then automatically sent to the admin for reviewing. Within this PR we also enable the option for the admin to delete any program by simply visiting it. 

**Fixes**
This PR fixes #3147

**How to test**
This can partly only be tested on Alpha as it requires the server for the mail handling, but some steps can already be tested locally. Make sure there is a public program and you are signed in as an admin. Navigate to the explore page and visit the program, verify that their is a new "delete" button next to the "edit" button. Clicking it should remove the program.

When deployed to Alpha the automatically sent mail on reporting should contain a link to the reported program. From there on you can follow the same flow as above to make sure everything works as expected.


Felienne (approved):
Seems to work (gives me the popup)!

PR #3154, opened 2022-08-11
Changes for-teachers page, give manual a dedicated page

### LINKED ISSUE #3153: [UI idea] Put teachers manual in a separate menu item
I think the teachers manual is not very discoverable as we have it now, and the UI with the buttons is very uncommon so the separate parts are also harder to use than could be necessary, and I also plan to make it a bit more extensive with videos.

So could we add it as a second extra menu item, and replace the green buttons with something else (anchors in the page?)

TiBiBa:
**Description**
In this PR we clean-up the for teachers page a bit by making a dedicated page for the teacher manual. Making it more visible and giving us more freedom on the styling of the manual. For now we merely copied the content and made some minor changes to the `common mistakes` section. But, this should be a good starting point to work from and improve the manual.

**Fixes**
This PR fixes #3153.

**How to test**
Log in and make sure you are a teacher. Verify the new menu option on the dropdown menu. Also verify that the manual is content-wise identical to before, only the location and styling is changed.


Felienne (approved):
Looks good apart from, yes... the conflicts

image

Felienne:
Pffffff that was the last one, will now reset Weblate and hope never to be in this mess again haha

PR #3155, opened 2022-08-11
[LANGUAGE] Adds meta to transpiler to slice, adds new exception for issue #1503

Felienne:
Working on #1503 the phoenix of issues that has resurfaced like 4 times :) 

Almost there, I just need to ask a text for the exception so it will show nicely on the front-end

Current status:

image


Felienne:
Ah forgot about the issue with random, good think I wrote it down in the PR :) Will try to finalize tomorrow!

Felienne:
Now with nice front-end message:

image

PR #3164, opened 2022-08-17
Implements the support for multiple admin users

### LINKED ISSUE #3160: Enable multiple admins
**Idea incl level**
Currently we only enable one admin within our environment as this is set as a string and we verify if the current user is the same as the username set in the environment. It would be nice if we changed this to a list to enable multiple users to be an admin. However, it is up to @Felienne if this is preferable. We can also change it to a list and still keep only one user in the list, but this at least makes it more scalable for the future.

TiBiBa:
**Description**
In this PR we make a small change to how we handle the admin user by implementing support for multiple admins. The `is_admin` function is changed to look into the environment for the `ADMIN_USERS` list and then verify if the current user is in there. If so, the user is an admin. This introduces some complications for our contributors and the following changes to then environment variable should be made to get this to work:
- Change `ADMIN_USER` to `ADMIN_USERS`
- Change the value from `"username"` to `["username"]`

**Fixes**
This PR fixes #3160.

**How to test**
Make sure you make the changes as described above and then verify that all admin functionality is still working. **Important:** we do have to make the same changes in Heroku before merging.


website/auth.py
```diff
@@ -120,8 +121,10 @@ def forget_current_user():
 
 
 def is_admin(user):
-    admin_user = os.getenv('ADMIN_USER')
-    return user.get('username') == admin_user or user.get('email') == admin_user
+    # Get the value from the environment, use literal_eval to convert from string list to an actual list
+    admin_users = os.getenv('ADMIN_USERS')
+    admin_users = ast.literal_eval(admin_users)
```
rix0rrr, line 126: Great job! If I may offer a suggestion, I would try to support both methods of specifying an admin user at the same time.

If you don't, logging in will be impossible during the period that code and config don't match up. Either:

* Deploy code first,  logging in stops working, update config
* Update config first, logging in stops working, deploy code.
 
If we support both methods of specifying the admin user, there's a natural progression:

* Deploy code first, logging in still works
* Update config, loggin in still works

TiBiBa: Agree, will look into this!

rix0rrr:
I changed the PR to make it a bit simpler, using simple Python strings that are split on a `,` instead of using `literal_eval`.

We lose the ability to use `,` in a username or email address, but I don't think that's a big sacrifice (we don't use those anyway), and I feel the usability is better this way (especially in shell scripts where it's harder to add `"` and `'` to values).

Could you make the same change in #3181 ? 

Cheers! 🙏

TiBiBa:
> I changed the PR to make it a bit simpler, using simple Python strings that are split on a `,` instead of using `literal_eval`.
> 
> We lose the ability to use `,` in a username or email address, but I don't think that's a big sacrifice (we don't use those anyway), and I feel the usability is better this way (especially in shell scripts where it's harder to add `"` and `'` to values).
> 
> Could you make the same change in #3181 ?
> 
> Cheers! 🙏

I agree that this simplifies the code but with this change we have to keep two variables in mind, never giving contributors an incentive to update their environment. While this is not a big deal, we might want to enforce users at some point to update these values (we can't always keep support for relic code) to make our code less complex.

Felienne:
> > I changed the PR to make it a bit simpler, using simple Python strings that are split on a `,` instead of using `literal_eval`.
> > We lose the ability to use `,` in a username or email address, but I don't think that's a big sacrifice (we don't use those anyway), and I feel the usability is better this way (especially in shell scripts where it's harder to add `"` and `'` to values).
> > Could you make the same change in #3181 ?
> > Cheers! 🙏
> 
> I agree that this simplifies the code but with this change we have to keep two variables in mind, never giving contributors an incentive to update their environment. While this is not a big deal, we might want to enforce users at some point to update these values (we can't always keep support for relic code) to make our code less complex.

Yes agreed that at one point we want to stop supporting ADMIN_USER (and similarly INVITE_CODE). Maybe make an issue so we don't forget?

TiBiBa:
> > > I changed the PR to make it a bit simpler, using simple Python strings that are split on a `,` instead of using `literal_eval`.
> 
> > > We lose the ability to use `,` in a username or email address, but I don't think that's a big sacrifice (we don't use those anyway), and I feel the usability is better this way (especially in shell scripts where it's harder to add `"` and `'` to values).
> 
> > > Could you make the same change in #3181 ?
> 
> > > Cheers! 🙏
> 
> > 
> 
> > I agree that this simplifies the code but with this change we have to keep two variables in mind, never giving contributors an incentive to update their environment. While this is not a big deal, we might want to enforce users at some point to update these values (we can't always keep support for relic code) to make our code less complex.
> 
> 
> 
> Yes agreed that at one point we want to stop supporting ADMIN_USER (and similarly INVITE_CODE). Maybe make an issue so we don't forget?

Will add this change to the invite code support PR as well. Will also make a separate issue with a to-do for somewhere in the future.

Felienne (approved):
Works!

PR #3166, opened 2022-08-17
Switch Babel dependency to YAML file for tutorials

### LINKED ISSUE #3165: [CONTENT] Change tutorial Babel dependency to YAML files
**Describe the issue**
I would like to propose to change the current Babel dependency for the tutorial into a YAML file one. With #3163 in mind we will have a rapidly growing amount of translations to handle for the tutorials. This is fine, except it will be hard to keep this nice and structured through Babel. By using a YAML file we can use this to approach the tutorials similar to the adventures using a structure somewhat like this:

```
intro:
  steps:
    1:
      title: ...
      step: ...
    2:
      title: ...
      step: ...
teacher:
  steps:
    1:
      title: ...
      step: ...
    2:
      title: ...
      step: ...
levels:
  1:
    steps:
      1:
        title: ...
        step: ...
      2:
        title: ...
        step: ...
```

We have to have a dedicated section for both the intro and teacher tutorials as they are not level-dependent. What do you think @Felienne @Mark-Giesen ?

**Edit**
This change has the added advantage that we can cache all language / keyword language combinations such as we do with the adventures. Speeding up our user experience significantly including the dynamic keywords.

TiBiBa:
**Description**
In this PR we switch the tutorial translations from Babel to a YAML file to better enable us to add new levels and steps without losing the overview. We cache all tutorial language / keyword language combinations on server start to reduce server load and enable dynamic keywords. Currently this is not relevant, but probably will be when adding the tutorials for new levels.

**Fixes**
This PR fixes #3165.

**How to test**
As this is a re-structuring task all should work as before. To verify log in and navigate to `/tutorial`, verify that all steps stil work as expected.


Felienne:
Will merge this once Weblate is up to date with the other changes (probably tomorrow)

Felienne (commented):
The tutorial still works! I did notice that we do not seem to have a step for the Parson's problems, is that correct or is that a mistake? I don't remember if we had that before?

TiBiBa:
> The tutorial still works! I did notice that we do not seem to have a step for the Parson's problems, is that correct or is that a mistake? I don't remember if we had that before?

We never had a step for the Parsons. Will make a seperate issue to add this!

Felienne (approved):

PR #3167, opened 2022-08-17
[FIX] disable test requests on Heroku

rix0rrr:
Test requests are used for e2e tests, which run against local
debug instances. They should not be allowed on Heroku (neither
staging nor the production environment).


Felienne (approved):
Thanks!

PR #3179, opened 2022-08-18
Enables users to request a teacher account after signup

### LINKED ISSUE #3176: Enable users to request teacher account after account creation
**Description**
Currently we only allow users to request a teacher account on account creation. It would be nice if we also allowed this after the creation, somewhere in the profile settings.

TiBiBa:
**Description**
In this PR we implementation the option for users to request a teacher account after signup. Currently this is only possible at the point of signup, which doesn't make any sense. Therefore we add a "request teacher" section and button to the profile page to enable users to request a teacher account later on.

**Fixes**
This PR fixes #3176.

**How to test**
Make sure you are not a teacher and navigate to your profile page. Notice the new section and verify that it shows a button on click which in turn requests a teacher account. Then, as an admin verify that the request is being handled correctly. The section should be invisible is someone is already a teacher or has a pending request.


Felienne (approved):
Works and can be merged when the conflicts are resolved

TiBiBa:
Will fix the conflicts right away!

PR #3183, opened 2022-08-19
Validate username on password reset

JoepdeJong:
**Description**

This PR adds an extra check to the password reset endpoint to validate whether the account that is reset has the same username as the token that is used.


TiBiBa:
Hi @JoepdeJong! Thanks for your first contribution to Hedy, I really like this small validation improvement! Feel free to pick up any open issues if you want to contribute any further.

TiBiBa (approved):
Awesome job!

PR #3185, opened 2022-08-19
Attempt 2 for PR public tags

### LINKED ISSUE #3149: [UI idea] Enrich profile with special tags
Would be cool if public profiles could have a few extra tags that we can assign in the admin interface:

* Teacher (already present in the database, no change needed)
* Certified teacher (new category that we will be making for people that have followed our Hedy course)
* Admin for the username that is admin (maybe it would be even better if we can have multiple admins?)
* And maybe "distinguished user" that we can give out to very special people for various reason?

These tags would then be shown in the user's profile

TiBiBa:
**Description**
In this PR we add a new and exiting feature: profile tags! From now on users will be able to receive tags that will be shown on their public profile page. The "teacher" and "admin" tag are automatically added on updating the public profile, other tags can be added by the admin. At this moment we have the following tags:

- Admin
- Teacher
- Certified teacher
- Distinguished user
- Contributor

**Fixes**
This PR fixes #3149.

**How to test**
Make sure you are an admin and navigate the the `/admin/users` page. Notice the new `tags` column on the table. Verify that clicking the icon displays a modal with all the possible tags enabling the admin to remove / add tags. Also verify that adding and removing tags is correctly stored on the public profile of the user. When no profile exists an error is returned.


Felienne (approved):
Yes, mergen maar!

Felienne:
ok one of those terrible merge conflicts slain @TiBiBa, two to go (I will try to get them all this weekend)

PR #3189, opened 2022-08-21
[BUGFIX] Restores Access Before Assign message 

### LINKED ISSUE #3188: [BUG] Access Before Assign message has disappeared
**Describe the bug**
Somehow, I guess in all the merge troubles, we lost the "Access Before Assign" string in the front end.

f.e. for this level 2 code:

```
print hallo
hallo is waar ben je nu?
```

On Hedy.org it works:

image

On main/alpha, we get:

image

I think it is due to the message being gone form message.pot, will look into this!

Felienne:
Fixes #3188, somehow we lost the message in the merging I think. This seems to fix it!


Felienne:
I also notices a lot of other commented messages in the En and Nl po files, @TiBiBa do you know what it up with those? 

I filled with them a bit and it does not seem to matter if they are there or not, but I do not know enough about Babel to delete them or enable them again, but I think it would be nicer if we clean them up. I guess at least all the tutorial stuff can be removed now we placed that into yamls?

TiBiBa:
> I also notices a lot of other commented messages in the En and Nl po files, @TiBiBa do you know what it up with those? 
> 
> 
> 
> I filled with them a bit and it does not seem to matter if they are there or not, but I do not know enough about Babel to delete them or enable them again, but I think it would be nicer if we clean them up. I guess at least all the tutorial stuff can be removed now we placed that into yamls?

When a string is removed and the .po files are updated Babel keeps the old onces and puts them at the bottom of the file. But they are redundant and can be removed.

Will look into this if we can automatically remove this when using the update command.

Felienne:
I will merge this now because I suspect it will make weblate angry and I don't want to leave you in weblate hell again today :D

PR #3200, opened 2022-08-24
Po cleanup

Felienne:
Removes line numbers for pot and po files to minimize future merge conflicts.

Can you check whether this is as we want it @TiBiBa?


CONTRIBUTING.md
```diff
@@ -123,8 +123,8 @@ When adding new Babel related translation the implementation is a bit more compl
     * When on the front-end (in a .html template) we do this like this: ```{{ _('test') }}```
     * Notice that the ```{{ }}``` characters are Jinja2 template placeholders for variables
     * When on the back-end we do this like this: ```gettext('test')```
-2. Next we run the following command to let Babel search for keys:
-    * ```pybabel extract -F babel.cfg -o messages.pot .```
+2. Next we run the following command to let Babel search for keys. We do not want line numbers since those will lead to lots of Weblate merge conflicts:
+    * ```pybabel extract -F babel.cfg -o messages.pot . --no-location --no-wrap```
```
TiBiBa, line 127: `--no-wrap` should be placed on the `update` command, will fix this!

TiBiBa (changes requested):
Looks good to me! Will fix the minor issue as stated.

CONTRIBUTING.md
Felienne: Ah sorry! I was wondering whether it had had effect but now I know why :D

TiBiBa (approved):
Should all be ready!

PR #3201, opened 2022-08-24
Access before assign message in all langs

Felienne:
Somehow Weblate did not restore this in all languages?


Felienne:
Putting this on draft because I started it before #3200 and I guess we want to do that cleanup first

Felienne:
Closing in favor of #3211

PR #3205, opened 2022-08-25
[CONTENT FIX] Fixes some incorrect translations

### LINKED ISSUE #3202: [CONTENT] Two small textual issues in account page
Two times new password:

image

No string for request_teacher_account:

image

TiBiBa:
**Description**
We fix two smalle missing translations in this PR. However, this is a bigger issue as we lost some translations in merge conflicts. Will have to take a closer look at this (but in a separate issue...). For now the change password strings and request teacher button should work again.

**Fixes**
This PR fixes #3202.

**How to test**
Verify that this is fixed. However, this only works in en/nl until Weblate is up-to-date again!


Felienne (approved):
This will probably upset weblate again but I will deal with that :)

PR #3207, opened 2022-08-25
[FEATURE] Enables admin to filter on username

### LINKED ISSUE #3195: [UI idea] Add username as filter option in /admin/users
Can we add username as a filter in admin/users?

We don't have it now (I am not sure why!) but I often look for users by name. Often (but not always) the email address is similar but it would be nice to explicitly allow usernames too.

image

TiBiBa:
**Description**
In this PR we introduce a new feature to the admin users view: username filtering. Enabling the admin to filter on a username substring in the users overview.

**Fixes**
This PR fixes #3195.

**How to test**
Make sure you are an admin and navigate to `/admin/users`. Verify that the new option is there and everything works as expected.


Felienne (approved):
Works!

image

PR #3209, opened 2022-08-25
Shows profile image in menubar

### LINKED ISSUE #3204: [UI idea] Change the icon next to the username to your avatar
@TiBiBa  I've got one last little issue to fix for you! :) 

image
Can you change the icon next to the usernames into the avatar the users picked?

TiBiBa:
**Description**
In this PR we update the menubar to be better scale-able to new content, but more importantly... we introduce the profile image showing on the menubar! From now on the image chosen for the public profile will always be shown on the menubar, giving this a nice personal touch.

**Fixes**
This PR fixes #3204.

**How to test**
Make sure you are logged in and have a public profile. Your chosen image should be shown next to your username instead of the default icon. Verify that the icon is correctly updated when updating the public profile or when removing our profile.


TiBiBa:
We have to re-style the menubar to prevent the image from screwing up the paddings, so this is a bit more work than expected!

Felienne (approved):
Check! 
image

PR #3211, opened 2022-08-25
Restore Access before assign message in all langs

Felienne:
Another try for #3201 now that we have removed the line numbers


TiBiBa (approved):
Looks good to me!

PR #3213, opened 2022-08-25
[FEATURE] Enable filtering on submitted programs

### LINKED ISSUE #3196: [UI idea] Allow filtering of the programs page on submitted programs
Currently in the programs page of a user you can filter for adventure and level, but for teachers it would be really useful if you could also filter on all programs that are submitted.

TiBiBa:
**Description**
In this PR we add a filtering option to the programs page to enable users (and teachers!) to filter on submitted programs. We already supported filtering on level and adventure, but not yet on wether or not a program is submitted.

**Fixes**
This PR fixes #3196.

**How to test**
Make sure you have several programs, both submitted and not. Verify that the filtering option works as expected.


Felienne (approved):
Works!

image

PR #3217, opened 2022-08-25
[FEATURE] Enables automatic generating of random passwords on multiple account creation

### LINKED ISSUE #3186: [UI idea] Enable generating of random password on multiple account creation
**Idea incl level**
Would save the teacher quite some time when creating student accounts.

TiBiBa:
**Description**
In this PR we add a small feature to the multiple accounts creating process. Enabling teachers to generate random 6 char password for the accounts, making the process of creating multiple accounts easier. In combinatie with the downloading option of the account credentials this makes the process of adding multiple accounts really easy!

**Fixes**
This PR fixes #3186.

**How to test**
Make sure you are a teacher. Navigate to any class and select multiple account creation, verify that the random password generating works and valid passwords are being generated.


TiBiBa:
Schermafbeelding 2022-08-25 om 14 51 48

PR #3219, opened 2022-08-26
[FEATURE] Supports multiple teachers per class

### LINKED ISSUE #3197: [UI idea] Allow multiple teachers for a class?
This year @OnnoEbbens and me will co-teach a course and we were thinking other teachers might also be in this situation, would it be doable to allow 2 teachers to view the progress of a class and maybe also to make modifications? I am not sure how involved that would be?

### LINKED ISSUE #3220: [BUG] Programming experience not correctly stored on /signup
**Describe the bug**
For some reason the programming experience is not stored in the database on signup.

TiBiBa:
**Description**
In this PR we add the functionality for a class to have multiple teachers. As this touches our core code of class functionality a lot of re-writing had to be done. Therefore this PR should be very extensively tested before merging. Shortly the following changes are made:
- Front-end changes are made to support a "Add teacher" button to the class overview
- The `teacher` attribute on the Class table is change to a `teachers` attribute of the set type
- Each teacher user is updated with a new attribute `teacher_classes`
    - This is a set of all the classes of a teacher
    - Necessary to reduce server load and still efficiently retrieve all teacher classes 

Some things that are still open for discussion:
- Currently a user is automatically added as a teacher, is this desirable?
- On account removal all related classes are removed, what if there are multiple teachers?
- Do we want to "lock" customizations when one teacher is working on them?
   - This is really really complex and I propose to not strive to implement this 

**Fixes**
This PR fixes #3197 and fixes #3220.

**How to test**
Make sure you are a teacher and have another teacher account. Verify that all functionality is working as it should (creating, deleting, customizing a class). As well as removing, adding and inviting a student. Add a new teacher and verify that you can only add a user that is actually a teacher. Verify that the other user can log in and also access the class.


TiBiBa:
@Felienne this seems to work (apart from the failing e2e tests). However, we do have to perform some major re-structuring on our live database for this to be correct. Currently each class has a `teacher` value as a string. We change this to a `teachers` value which is a set. We also add a value to the USERS table, `teacher_classes` which is a set of all the classes a teacher has. This is necessary to avoid a big server load for getting all classes of a teacher. For this to work we have to perform the following changes:

- Iterate through all classes
- For all found teachers:
    - Create an empty set `teacher_classes`
    - Append all of their class ids to this set
- For all found classes:
   - Change the `teacher` value to `teachers`
   - Change the string into a set with one item

I'm not sure what would be the best approach to execute this. I would prefer to write a script locally that connects with our production database through our AWS keys and then update the necessary tables. What do you think? Maybe @rix0rrr also has a better idea?

TiBiBa:
I just committed the manually updating script of the database. @Felienne can you take a look? We do have to merge right after or run the script right after merging as all teacher functionality will crash when one of the two isn't up-to-date. This should do the trick to update our database structure:

```
for Class in classes:
    CLASSES.update({'id': Class.get('id')}, {'teacher': None, 'teachers': {Class.get('teacher')}})
    if not teacher_classes.get(Class.get('teacher')):
        teacher_classes[Class.get('teacher')] = set()
    teacher_classes[Class.get('teacher')].add(Class.get('id'))

# Add "teacher_classes" to each user
for teacher, classes in teacher_classes.items():
    USERS.update({'username': teacher}, {'teacher_classes': classes})
```

TiBiBa:
Down to one failing test, but I have not clue (yet) why it fails. All seems to work locally as well as within the tests but for some reason the GET requests after the POST still retrieves some old database while the db should already be updated. Will look into this, but other than that we are ready for merge!

Felienne:
Hi @TiBiBa! I am setting this to draft since I don't want to merge this week (it being the first week of school in many places) since deploying will be a bit tricky (and I might want to make some tiny changes)

Felienne:
Hi @TiBiBa! I will close this for now as it is so much behind. If we want to work more on this feature (not even sure there is really a demand) I think indeed we'd want a primary and a secondary teacher so it is less overhaul.

Felienne:
Sorry for all the work you already put in 😭

PR #3222, opened 2022-08-26
[CLEAN-UP] Removes all unused strings in .po files

### LINKED ISSUE #3221: [CLEAN-UP] Remove unused keys in .po files
**Describe the issue**
With the removal of the line numbers to prevent merging conflicts it might be nice to also remove all redundant commented translations that are no longer used. These are all due to merge conflicts or no longer existing keys.

TiBiBa:
**Description**
In this PR we clean-up all .po files by removing all unused translations and keys.

**Fixes**
This PR fixes #3221.

**How to test**
This is a clean-up, if it passes it works.


Felienne:
Nice work @TiBiBa! I assume you used a command line tool to remove them? Maybe we should document that somewhere so we can run it periodically? This PR is a good place and maybe also in a weblate/babel page of the docs?

Felienne:
Too many conflicts, will just do this anew when all the other weblate PR's are gone

TiBiBa:
> Nice work @TiBiBa! I assume you used a command line tool to remove them? Maybe we should document that somewhere so we can run it periodically? This PR is a good place and maybe also in a weblate/babel page of the docs?

I have to admit I made this PR manually. But as it is closed already I will create a new PR with a script that does this automatically.

In addition to this it might be nice to have a dedicated folder for all our small scripts we create and use to automate some clean-ups etc.

Felienne:
> I have to admit I made this PR manually. 

I did it manually too (with some regex magic)

> But as it is closed already I will create a new PR with a script that does this automatically.

I already fixed this yesterday on a Weblate PR (that was easier)

> In addition to this it might be nice to have a dedicated folder for all our small scripts we create and use to automate some clean-ups etc.

Agreed! But given all the other things you still have on your plate, this is not all too high prio

PR #3230, opened 2022-08-28
[FIX] Startup error in Safari

### LINKED ISSUE #3182: [BUG] Program code not shown when opening from /programs page
**Describe the bug**
When switching the keyword language using the new dropdown menu the code editor remains empty; not showing the default adventure code.

Schermafbeelding 2022-08-19 om 11 10 56

rix0rrr:
In Safari, after loading a program, the following error would occur:

```
TypeError: undefined is not an object (evaluating 'd.innerHTML')
```

After tracing the error, I found it occurred in trying to update line
styles by manipulating the HTML generated by the Ace Editor: finding and
generating `<div>`s with certain styles, in order to achieve a
strike-through effect.

I was not able to figure out why this happened, exactly, and only in
Safari. I suspect a race condition of some sort. However, I did notice that 
the code that was involved in producing this error was making a whole lot
of assumptions on how Ace would render itself to HTML elements, in ways
that seemed dangerous. We would be better off going through a supported
Ace feature.

What I did do in this PR is remove all the code that used to manipulate
the breakpoint styles set by Ace, and replace it with the "markers"
feature that Ace supports. The markers feature allows highlighting of
text, and since this feature is officially supported it's safer than
touching the HTML innards of Ace.

There is one downside to it: because markers render underneath or on top
of the actual text (instead of in line with it) the strikethrough we
generate is a tad hacky: we use `border-top`, and offset the border
down a bit, to render a horizontal white line.

We make the same change (using Markers instead of HTML manipulation)
for the "highlight current line" functionality of the step-by-step debugger.

Fixes #3182.

**How to test**

* Open a program from your saved programs page.
* Click in the gutter to toggle lines on and off (observe that disabled lines have a strikethrough)
* Click the debugger button, and observe that stepping through the program works while the current line is highlighted in blue

**Checklist**
  

Felienne:
Ow @TiBiBa it is ok to approve this one but don't merge it yet because I want to deploy it in isolation (so we can easily roll back if it creates more issues)

Felienne:
Merging this to test it online

Felienne (approved):

PR #3237, opened 2022-08-29
Implement requires admin

### LINKED ISSUE #3236: Implement @requires_admin validation
**Description**
We have a nice way to validate if someone is a logged in user by the checking `@requires_login`. It would be nice if we also write a check like this but for the admin functionalities as currently we copy and paste the check on each function. This is prone for errors and a security risk.

TiBiBa:
**Description**
In this PR we remove a manual check if the user is written and replace it by a `@requires_admin` check on the route. With this check we no longer need a manual if statement on each admin route to check if the user is an admin, making the codebase a bit safer as we can't make mistakes there.

**Fixes**
This PR fixes #3236.

**How to test**
Verify that all admin pages can only be visited as admin, everything should work as before.


Felienne:
Maybe our security expert @JoepdeJong wants to take a peek at this one? 🙏

JoepdeJong (changes requested):
The `requires_admin` decorator looks good to me.

Note that there are more endpoints that are only accessible by admins: e.g. `/admin/getUserTags`, `/admin/changeUserEmail`,`/admin/updateUserTags`, `/programs/set_hedy_choice` and `/program-stats`.

JoepdeJong (approved):

Issue #2777, opened 2022-05-19
[UI idea] Create "intro" tutorial for each level


TiBiBa:
**Idea incl level**
We now have a "UI explain tutorial", it would be nice to have an interactive tutorial for each level as well.

TiBiBa:
To-do:
- Document code
- Write a skelet for level 1 -> generalize

Felienne:
With the new content organization, this seems to be handled!

PR #3242, opened 2022-08-30
Adds 'start tutorial' button if a tutorial exists for the current lang / level combination

TiBiBa:
**Description**
In this PR we built upon the implementation of a tutorial template. We add a start tutorial button if there is a tutorial found for the current language / level combination. When the tutorial is added to the corresponding YAML it should all work automatically. However, when wanting to re-locate the popup, highlight or hide specific elements new functions should be added and called in the corresponding .ts files in the `/static/js/tutorials` folder.

**Fixes**
Not really a related issue, however we are closer to fixing #2777.

**How to test**
Verify that there is a tutorial button showing if the corresponding tutorials YAML contains data for the current level. This is already the case for the English YAML where we added a dummy entry. **This value should be removed after testing**.

**Screenshot**
Schermafbeelding 2022-08-30 om 14 09 05


Felienne:
Hmmm this seems to move around some texts in the po file, not a problem in itself but of course that will lead to merge errors, so I am merging in the current weblate first.

Felienne:
Cool, this seems to work!! But... I don't seem to be able to end the tutorial :) 

image

Do we need a special "end step" in the yaml?

TiBiBa:
> Cool, this seems to work!! But... I don't seem to be able to end the tutorial :) 
> 
> 
> 
> image
> 
> 
> 
> Do we need a special "end step" in the yaml?

We might be able to make this a bit more sophisticated. Maybe we can make a look-ahead on the step retrieval and if the next step returns a "next step not found" error, overwrite the "next step" onclick of the button to hide the tutorial.

Now we have a cumbersome hard-coded way of saying "if step == 14" -> end tutorial. Not really scale-able. Will take a look into this in this PR (tomorrow)!

Felienne:
> > Cool, this seems to work!! But... I don't seem to be able to end the tutorial :)
> > image
> > Do we need a special "end step" in the yaml?
> 
> We might be able to make this a bit more sophisticated. Maybe we can make a look-ahead on the step retrieval and if the next step returns a "next step not found" error, overwrite the "next step" onclick of the button to hide the tutorial.
> 
> Now we have a cumbersome hard-coded way of saying "if step == 14" -> end tutorial. Not really scale-able. Will take a look into this in this PR (tomorrow)!

Ah thanks!! I was hacking around myself a bit but it was hard to make sense of what is what, so if you could polish it a bit more, that would be really helpful!!

TiBiBa:
> > > Cool, this seems to work!! But... I don't seem to be able to end the tutorial :)
> > > image
> > > Do we need a special "end step" in the yaml?
> > 
> > 
> > We might be able to make this a bit more sophisticated. Maybe we can make a look-ahead on the step retrieval and if the next step returns a "next step not found" error, overwrite the "next step" onclick of the button to hide the tutorial.
> > Now we have a cumbersome hard-coded way of saying "if step == 14" -> end tutorial. Not really scale-able. Will take a look into this in this PR (tomorrow)!
> 
> Ah thanks!! I was hacking around myself a bit but it was hard to make sense of what is what, so if you could polish it a bit more, that would be really helpful!!

I looked into this but it requires a bit more of a re-write than I expected. Still possible to change the flow, but I propose to merge this for know and make sure that each tutorial has an `endTutorial()` step that changes the button `onClick` (as we already have for the intro / teacher tutorials).

Felienne:
Note to self: after the merge, this stiil needs:

msgid "tutorial"
msgstr ""

In all languages apart from En.

Felienne:
Let's merge this in, and look at extending it later on!

Felienne (approved):

PR #3243, opened 2022-08-30
Enables admin to view class overview

### LINKED ISSUE #3194: [BUG] Show extra link to class page in /admin
**Describe the bug**
From the classes overview, I can go to the class statistics page, but the button than shows "go back to class":

image

and if I click it, it goes to the class page to which I have no access:

image

(although I am not sure why we do not show a "forbidden" here but an error?)

I think it would be best if the admin could also see the class page (it is cool for me to see customizations people use) so could we:

* Allow that
* Add a column/link to the class page in admin to the class page in addition to the stats page?

TiBiBa:
**Description**
In this PR we enable the admin to view the class overview as well as the class customizations of each class. We change the UI a bit when the admin views these specific pages to make sure it is clear that no alterations can be made. These would still be caught by the server as the admin is not the teacher of the class, but still. 

**Fixes**
This PR fixes #3194.

**How to test**
Make sure you are admin and there are classes for which you are not the teacher. Verify that you can visit these as well as the customizations but the buttons are hidden.

### DIFF
--- a/templates/admin/admin-classes.html
+++ b/templates/admin/admin-classes.html
@@ -9,7 +9,7 @@
 <div class="program w-full mx-auto border-solid border-2 border-orange-400 rounded p-4">
   <table class="table-auto w-full text-sm">
     <thead>
-      <tr class="font-bold text-left">
+      <tr class="font-bold">
           <td id="classname_header">Class name</td>
           <td id="teacher_header">Teacher</td>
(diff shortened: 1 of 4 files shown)


Felienne (approved):
Ah nice, this allows me to more easily snoop on classes :D 

image

PR #3245, opened 2022-08-31
Changes third party structure, store choice in database as well

TiBiBa:
**Description**
In this PR we change the work flow for agreeing with third party offers on signup. Instead of a mail send to the admin the choice is stored in the database. Making it easier to retrieve this later on. We also add the option to the profile page to check/uncheck the option after signup.

**Fixes**
No related issue.

**How to test**
Verify that on signup when checking the third party option this is correctly stored in the database. Also verify that you can uncheck this option in the profile settings.


Felienne (changes requested):
This works, thanks! But... I would also like to see it/filter for it in the /admin page (so I can send all of them an email easily)

Felienne (approved):

PR #3252, opened 2022-08-31
[FIX] Hide submitted programs from /explore and public-profile page

TiBiBa:
**Description**
In this PR we make a minor fix where we no longer show the submitted program on the /explore and public-profile page. The submitted programs are still visible for the admin and teacher of the user but not for anyone else.

**Fixes**
No related issue.

**How to test**
Make sure you have an account that is in your class. Verify that the user has a shared and a shared-submitted program. Now verify that only the shared program can be found through /explore and when visiting the user's their public profile while both programs can be viewed when visiting the user's programs page through the class overview.


Felienne (approved):

Felienne:
Yes this works! We could optimize this though by, when caching public programs on server start, skip all submitted ones so we don't have to check them here again.

PR #3256, opened 2022-09-02
[BUGFIX] Correctly store programming experience

### LINKED ISSUE #3220: [BUG] Programming experience not correctly stored on /signup

Felienne:
Fixes #3220

**How to test**
Create an account and observe that programming experience is stored in the database again

### DIFF
--- a/templates/signup.html
+++ b/templates/signup.html
@@ -45,7 +45,7 @@
                 </select>
             </div>
             <div class="signup-container" style="display: none;" id="keyword_lang_container">
-                 <label class="signup-label" for="keyword_language">{{_('preferred_keyword_language')}}</label>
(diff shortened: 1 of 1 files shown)


TiBiBa:
This issue is also fixed in #3219. Not really nice to store two issues in one PR, but I noticed it when working on the multiple teachers feature.

Felienne:
Yes, I know, I stole the corresponding commit from that PR :) Can you confirm this indeed is the right one (it seems like it?)?

TiBiBa:
This seems to be the correct one! The issue was that the name of the checkboxes was different than the one we checked for in the back-end.

Issue #3259, opened 2022-09-05
[CONTENT] Reorder adventures based on difficulty


Felienne:
**Describe the issue**
We (lightly) follow the convention that adventures left to right are easier to harder but with some new adventures, that is not always the case (f.e. parrot is easier than story) and the blanks (see also #2380) are not always in the later adventures.

**Related but different(might need a separate issue)**
The order of the adventures as it is now is always the same for all levels, while that can potentially break the easier-to-hard order.

Felienne:
Done!

PR #3271, opened 2022-09-10
[CONTENT] Reorder adventures and some other small fixes

Felienne:
Some progress towards #3259. This PR updates the order but not yet smoothens out the use of blanks.


Felienne:
ok I realize this might be a Weblate disaster but we have to do it one day and Weblate is never calm :D

PR #3282, opened 2022-09-15
fix over-colorating of at 

### LINKED ISSUE #3278: [SYNTAX HIGHLIGHTER] Overcoloring in Dutch content
**Describe the bug**
There is some overcoloring in the Dutch content (`op` is highlighted while it should not be)

image

thjazi39:
Solving the problem of over-coloring the at prefix in levels 3 and 4.

The resolution consisted in adding a tag "END_WORD" at the end of the AT transitions in the syntactic highlighting automaton.

image

fix #3278 

This branch also adds a test in English and Dutch for this case

Edit : Same for `to` and `from`


Felienne:
Thanks @thjazi39! This works for `op`!

image

But `uit` remains an issue:

```
uitlaters is papa, mama, Sam, Petra
uitlater_gisteren is ask Wie heeft gisteren de honden uitgelaten?
print uitlater_gisteren hoeft vandaag dan niet nog een keer!
remove uitlater_gisteren from uitlaters
print uitlaters at random is vandaag aan de beurt om de honden uit te laten!
```

Level 3:
image

Could you see if that can be fixed too?

Felienne (changes requested):
Works for `op` but not `uit`

Felienne:
Thanks for still helping out @thjazi39!!! We moved from Slack to Discord, if you still want to chat with the Hedy team you can meet us here: https://discord.gg/fazXQqM6

PR #3296, opened 2022-09-18
[BUG FIX] Admin is able to view own class as teacher again

TiBiBa:
**Description**
Due to an incorrect comparison the the class overview was always viewed as admin when viewing a class with an admin account. While this might not result in an issue the admin was unable to alter their own class, add students and more because of this limitation. This bug is fixed.

**Fixes**
No related issue.

**How to test**
Make sure you are an admin and a teacher. Make a class and verify that you can indeed add students and customize your class again.


Felienne:
Ah thanks for the fix @TiBiBa!

The admin user still lacks the buttons on the customize page:

image

But that is not all too much of an issue for now.

Felienne (approved):

PR #3297, opened 2022-09-18
Makes sure empty code isn't parsed at all

### LINKED ISSUE #3290: [BUG] Run Empty code shouldn't count?
**Describe the bug**
Delete all code from the window and press Run. This results in a succesfull execution. You'll get some encouragement (Bravo!) which is OK, but your succesfull counter also increments. Pressing the run button 50 times, will result in a batch! I think this isn't the intended behaviour?

**Expected behavior**
Running "no code" should result in nothing

TiBiBa:
**Description**
We add a validation to the front-end to make sure there is actual code before parsing and returning some success message. This reduces server-load, achievement spamming and makes the UI less 'flashy' when running an empty program.

**Fixes**
This PR fixes #3290.

**How to test**
Run an empty program. Verify that no POST is made to the server and no success message is returned either.


Mark-Giesen:
I checked out this branch, ran on-deploy to compile ts files (among others) and ran the app.py. I still get the "amazing" and "good job" messages and after 50 times even a batch.

Mark-Giesen:
Ah, thought of something, closing the browser and restarting it, ts might stay in cache. This worked!

Mark-Giesen (approved):
works like a charm.

Issue #3258, opened 2022-09-05
[UI idea] Split the signup workflow into teachers and regular users.


Felienne:
**Idea incl level**
Split the signup workflow into teachers and "regular" users.

We now have one form for all users, but some information like programming experience is mainly useful for students while we might ask teachers other things like "how did you hear about Hedy?" and "Do you want to enrol in our upcoming training?" It would be nice to make two pages.

**Additional context**
This is an excellent issue if you want to get started with Hedy, since it will not require a lot of interaction with the grammars or with the back end, the relevant route is already there and can, for a first implementation, remain the same, before we add specialised questions to each sign-up form.

jdomil:
Hi @Felienne! I'm new to Hedy and open source, but I'd love to contribute to the platform with this issue. I might need some initial guidance though, so I will ask you a couple of questions:

* What's the preferred approach to split the form? 
  1. We give users the option to register as either teachers or students and redirect them to different urls based on that
  2. We have one single sign up page but dynamically prompt certain questions based on whether the user requests to become a teacher
  3. Other approach I haven't considered
  
* With regards to creating the user as a teacher in the database, apart from adding data points about the newly added questions, should we modify anything? I.e. would you like me to create a completely different object such as `teacher`? I assume not, since that would have several downstream implications.

Felienne:
Hi @jdomil! Welcome to Hedy! It would be great if you could also join our Discord: https://discord.gg/fazXQqM6 We use that to talk about all sorts of things, also open issues. 

As it happens to be, yesterday someone else told us they want to pick up this features, but maybe you can do it together? If you join us on Discord we can sync more easily.

Felienne:
> * What's the preferred approach to split the form?
>   
>   1. We give users the option to register as either teachers or students and redirect them to different urls based on that

I would prefer this. Maybe with the default option a student and an extra button/link "click here to register as student"

> * With regards to creating the user as a teacher in the database, apart from adding data points about the newly added questions, should we modify anything? 

I don't think so for now.

PR #3304, opened 2022-09-21
[UI] #3258 Split the signup workflow into teachers and regular users. 

rinkstiekema:
**Description**
The signup process now has an extra step, namely the role selection. This is an extra screen which asks whether the user is a student or a teacher.

I've renamed `signup.html` to `signup-student.html`, and added two extra views (`signup-role-selection.html` and `signup-teacher.html`). 

The role selection page simply links to the respective URL. The signup forms now hide the checkbox for requesting a teacher account. Instead, it uses a hidden input field.

Some extra messages have been added to `messages.po` and two new routes have been introduced (`/signup/student` and `/signup/teacher`). 

Made a minor change to the signup form to place the radio buttons in a flex row. 

#3258

**How to test**
Run the code locally, go to the signup screen and go through the student and teacher flow.


templates/signup-role-selection.html
```diff
@@ -0,0 +1,33 @@
+{% extends "auth.html" %}
+
+{% block main %}
+<div class="w-full mx-auto flex flex-col items-center gap-4">
+    <div class="bg-white shadow-md rounded-lg rounded py-4 px-0 md:px-4 lg:px-8">
+        <h2 class="my-0 text-center">{{_('signup_student_or_teacher')}}</h2>
+
+        <div class="mt-8 mx-auto flex flex-wrap lg:flex-nowrap justify-center gap-8">
+            <a href="/signup/student">
+                <div class="w-full border-2 rounded-lg flex flex-col items-center px-8 bg-white text-center border-green-400 bg-gray-200 hover:bg-green-400 cursor-pointer" onclick="window.location = '/tutorial'">
```
TiBiBa, line 10: We have an `<a>` element as well as an `onclick` on the underlining `<div>`? The second also points to an incorrect location. I assume this is from copying it from the landing page code? Would be best to only use the `onclick` as this is consistent with our code base.

templates/signup-role-selection.html
```diff
@@ -0,0 +1,33 @@
...
+                    <img src="/images/profile_images/1.png" class="h-32 m-8 mb-0" alt="Hedy tutorial icon">
+                    <div class="my-4">
+                        <p class="text-xl">{{_('student')}}</p>
+                    </div>
+                </div>
+            </a>
+            
+            <a href="/signup/teacher">
+                <div class="w-full border-2 rounded-lg flex flex-col items-center px-8 bg-white text-center border-green-400 bg-gray-200 hover:bg-green-400 cursor-pointer" onclick="window.location = '/hedy'">
```
TiBiBa, line 19: Same issue

templates/signup-teacher.html
```diff
@@ -119,9 +121,9 @@ <h4 class="text-blue-500 mt-0 mb-2 text-center">{{_('create_account_explanation'
             </div>
             <!-- The extra check options -->
             <div class="border border-gray-400 rounded-lg mx-4 px-4 md:px-8 flex flex-col gap-1 py-2">
-                <div class="flex flex-row justify-between items-center">
+                <div class="flex flex-row justify-between items-center hidden">
                     <label for="is_teacher">{{_('request_teacher')}}</label>
-                    <input type="checkbox" id="is_teacher" name="is_teacher">
+                    <input type="checkbox" id="is_teacher" name="is_teacher" checked>
```
TiBiBa, line 126: Why not keep the one-page structure, check/uncheck the checkbox based on the flow selection of the user. It can always stay hidden, that's fine.

tests_e2e.py
```diff
@@ -52,13 +54,13 @@ def request(method, path, headers={}, body='', cookies=None):
 
     if response.history and response.history[0]:
         # This code branch will be executed if there is a redirect
-        ret['code']    = response.history[0].status_code
```
TiBiBa, line 55: This is just an example, but please refrain from making these styling changes on code parts unrelated to your PR. This makes it hard to review.

TiBiBa (changes requested):
Hi @rinkstiekema! Thanks for your first contribution to Hedy! First a minor note in general:

- Please refrain from making styling changing on code parts you haven't worked on. This makes your PR hard to review. These changes are still great, but would be better to fix in a separate PR (I assume your IDE fixes this automatically?)

In general I would like to suggest to simplify your implementation. Keep the one-file structure but hide the signup form until an option is chosen (and hide the option selector when chosen). The checkbox for the teacher request can be checked/unchecked on the same onclick. This would greatly simplify your code and keeps the code base more readable.

Let me know if you need any help!

tests_e2e.py
rinkstiekema: Yeh I think VSCode automatically formats the file. I'll go back to those files and revert the changes to make the pull request cleaner 👍

TiBiBa:
Sorry @rinkstiekema, forgot a bit about this PR! Once again thanks for your contribution. I do still think we can simplify this a bit to make the code a bit more readable and understandable by other contributers. For example, at this point we create two separate signup forms, but there are mostly identical; this should be easier to fix!

Maybe we can use a url attribute to contain all the info we need? We can keep the one page approach using the `/signup` route but the buttons re-direct to either `/signup?teacher=false` or `/signup?teacher=true`. We can get this value inside our HTML template using something like this:

```
{% if request.args.get('teacher') %}
# Set the teacher title, explanation and checkbox
{% else %}
# Set the student title, explanation and uncheck the checkbox
{% endif %}
```

Lastly, as the teacher request is already decided by the chosen option it would be nice of we hide the checkbox all together. Simply adding the class `hidden` should fix that! Let me know if you need any help!

rinkstiekema:
I agree. I think the URL param is a good solution. I'll implement it one of these days and push an update.

Coming from React I have to adjust a bit - apologies for the overhead!

TiBiBa:
> I agree. I think the URL param is a good solution. I'll implement it one of these days and push an update.
> 
> Coming from React I have to adjust a bit - apologies for the overhead!

No worries!

Felienne:
Hi @rinkstiekema! Are you planning to work on this soon? If not that is also totally fine of course, but we are sort of eager for this to go live because it will enable other things, so if not, let us know and @jpelay can finish it!

rinkstiekema:
> Hi @rinkstiekema! Are you planning to work on this soon? If not that is also totally fine of course, but we are sort of eager for this to go live because it will enable other things, so if not, let us know and @jpelay can finish it!

Was planning on finishing this today!

Felienne:
Haha, why is the teacher a confused person?

image

Felienne:
Thank @rinkstiekema for making the changes with the URL!!

I have just made some small modifications to the layout (there was a double explanation for teachers)

It looks GREAT, thanks, but creating an account does not work for me, I am being redirected to the choosing screen:

I'm see this:

image

URL bar shows:
http://0.0.0.0:8080/signup?username=felienne&email=feliennehhh%40gmail.com&password=123456&password_repeat=123456&language=eo&keyword_language=en&birth_year=&is_teacher=on&agree_terms=on

Also, I am getting an error in the template regarding `prog_experience`. I know we had issues with that field earlier so maybe that was already an issue:

image

Could you take a look at these issues?

It is also possible i destroyed something while merging main in (in that case: sorry!)

Felienne:
Ah yes the second issue exists in main too, so I guess that is ok (maybe @TiBiBa can confirm?)

image

Felienne:
ok it was a tiny thing so I fixed it! I can confirm this works entirely as we want it! If @TiBiBa is ok with the implementation details we can ship this 🥳

TiBiBa:
> ok it was a tiny thing so I fixed it! I can confirm this works entirely as we want it! If @TiBiBa is ok with the implementation details we can ship this 🥳 

Awesome! Will look into this tomorrow and merge when approved.

templates/signup.html
```diff
@@ -59,69 +94,71 @@ <h4 class="text-blue-500 mt-0 mb-2 text-center">{{_('create_account_explanation'
...
+                    </li>
+                </ul>
             </div>
             <!-- The extra check options -->
             <div class="border border-gray-400 rounded-lg mx-4 px-4 md:px-8 flex flex-col gap-1 py-2">
-                <div class="flex flex-row justify-between items-center">
+                <div class="flex flex-row justify-between items-center hidden">
                     <label for="is_teacher">{{_('request_teacher')}}</label>
-                    <input type="checkbox" id="is_teacher" name="is_teacher">
+                    <input type="checkbox" id="is_teacher" name="is_teacher" {{ "checked" if is_teacher else "" }}>
```
TiBiBa, line 161: This doesn't need to be between quotes, the following should be sufficient:
```
{% if is_teacher %}checked{% endif %}
```

templates/signup.html
```diff
@@ -133,17 +170,23 @@ <h4 class="text-blue-500 mt-0 mb-2 text-center">{{_('create_account_explanation'
                 </div>
             </div>
             <div class="flex flex-row items-center justify-center px-4 md:px-8 gap-4">
+
+            {% if is_teacher %}
                 <input type="checkbox" id="agree_third_party" name="agree_third_party">
                 <label class="w-96 text-center" for="agree_third_party">{{_('agree_third_party')}}</label>
```
TiBiBa, line 176: Why is this only allowed when creating a teacher account?

templates/signup.html
```diff
@@ -1,11 +1,46 @@
...
+{% if request.args.get('teacher') %}
+{% set is_teacher = request.args.get('teacher') == 'true' %}
+{% else %}
+{% set is_teacher = None %}
+{% endif %}
+
 {% block main %}
 <div class="w-full mx-auto flex flex-col items-center gap-4">
     <div class="bg-white shadow-md rounded-lg rounded py-4 px-0 md:px-4 lg:px-8">
+        {% if is_teacher == None %}
```
TiBiBa, line 12: Why don't we simplify this with `{% if not request.args.get('teacher') %}`? This would simplify our code, with the following flow in mind "if there is no url parameter with the name 'teacher' show the selector, otherwise show the signup form".

templates/signup.html
```diff
@@ -1,11 +1,46 @@
...
+                    <div class="my-4">
+                        <p class="text-xl">Teacher</p>
+                    </div>
+                </a>
+
+            </div>
+        </div>
+        {% endif %}
+
+        {% if not is_teacher == None %}
```
TiBiBa, line 34: This could simply be an `{% else %}`.

TiBiBa (changes requested):
Left a few minor comments, for me it is almost good to go!

templates/signup.html
rinkstiekema: I do agree that it's a little more descriptive, but I did not want to include the "complexity" of how this is determined in the template

templates/signup.html
Felienne: This is a change I made: we only send emails from third parties to teachers (this is already current practice) we are never going to try to market stuff to kids :)

templates/signup.html
Felienne, line 12:
```suggestion
        {% if not request.args.get('teacher') %}
```

Felienne: yeah this is a bit of a trade off between readability and maintainability but I agree with @TiBiBa it is not vert descriptive as it is.

templates/signup.html
Felienne, line 34:
```suggestion
        {% else %}
```

Felienne (commented):
Made @TiBiBa's suggestions into commit suggestions (for ease of processing)

templates/signup.html
Felienne, line 161:
```suggestion
                    <input type="checkbox" id="is_teacher" name="is_teacher" {% if is_teacher %}checked{% endif %}>
```

Felienne (commented):
Commit suggestion for the template fix too

TiBiBa (approved):
Made some minor changes to the styling. Once the tests pass this is good to go!

TiBiBa:
Hi @rinkstiekema! Thank you very much for your first PR for Hedy! I just made some minor tweaks and I think this PR is ready for merge. Once merged it will automatically be deployed to [Alpha](https://hedy-alpha.herokuapp.com). Once in a while we make a deploy to production (hedycode.com) where you can see your work in action!

PR #3308, opened 2022-09-23
[BUG FIX] Admin is able to customize own class again

TiBiBa:
**Description**
Same issue as #3296, same fix, other page.

**Fixes**
No related issue.

**How to test**
Make sure you are an admin and the owner of a class. When attempting to customize this class verify that the buttons are shown as expected.


Felienne (approved):
Yup, works!

image

Felienne:
Fun story: I gave a live demo yesterday and I did not use the admin account but my actual teacher account (the one that has my real class) because of this bug 😅

PR #3315, opened 2022-09-24
[REFACTOR] Refactor submodules into blueprints

rix0rrr:
This is a big change, and I understand if you don't necessarily want it, but it fixes some organizational tech debt that's been bothering me.

Organize the different submodules of the website into classes, using the Flask "[Blueprints](https://flask.palletsprojects.com/en/2.2.x/blueprints/)" feature. 

* Blueprints share routes with the same prefix: all routes under--for example--`/admin/*` live together.
* Because we use classes to define Blueprints (not the default mechanism, but I made a `WebsiteModule` class to make it possible) dependencies can now be injected and accessed as class members (`self.db` instead of `DATABASE`).

This PR also adds type hints in a bunch of places to help the editor's autocomplete.


website/website_module.py
```diff
@@ -0,0 +1,41 @@
+import collections
+import functools
+import flask
+
+class WebsiteModule(flask.Blueprint):
+    """A website module is a class with its own routes that can be mounted into an app.
+
+    It's based on Flask Blueprints[1], but instead of defining an object and
+    then registering routes on it, we put everything together in a class so that
+    the routes can also share private variables that are injected from 'app.py'.
```
Felienne, line 10:
```suggestion
    the routes can also share private variables that are injected from 'app.py', for example the database. The alternative (using a global g.db) would be untyped and then the IDE lacks information about the type causing less autocomplete support.
```

Felienne (approved):

PR #3316, opened 2022-09-24
[FIX] Make pylint lint everything

rix0rrr:
Because the `validate-python` script explicitly lists files and directories to lint, parts of the code base were not being linted (leading to possibility of typo mistakes).

Add the missing directories, and update `pylint` to the latest version while we're at it.


Felienne (approved):
Seems not to break things :)

PR #3317, opened 2022-09-24
[FIX] Error marker is not showing

rix0rrr:
There were errors in the TypeScript code because of the use of `@ts-ignore`, which was hiding compilation errors.

Remove the ignore directives and fix the mistake.


Felienne (approved):
Error line works again! 

image

PR #3323, opened 2022-09-25
Setup black formatter on pre-commit for python formatting

rinkstiekema:
**Description**
Added pre-commit configuration to run [Black](https://github.com/psf/black) when committing code. This should make code-reviews much easier and keep styling consistent across the repository.

No linked issue, discussed in thread on Discord

**How to test**
Since black has run on many many files, the changes are too big to code-review. I suggest you reproduce the following steps on the main branch instead, and then merge my pull request into it to verify I haven't changed anything else than the formatting.

- Install pre-commit using pip (`pip install pre-commit`)
- Add pre-commit to `requirements.txt`
- Add `pre-commit-config.yaml` with the same values as in my branch
- Run pre-commit install to install the Black package for pre-commit
- Run `pre-commit run --all-files` to run Black across all python files in the directory, instead of just when changes are made to a file. This is a one-time operation

**Note**
Would be good to include formatting for Jinja templates, Typescript, etc. Didn't want to make the choice on _what_ formatter to use. You can switch out Black for Autopep8 if that's desired. Another good addition would be running Flake8 on pre-commit for python linting. Similar linters can be used for other languages (Typescript, HTML, CSS, etc.) to improve the code consistency and quality.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 22.8.0
+    hooks:
+    - id: black
+      language: python
+      types: [python]
+      args: ["--line-length=120"]
--- a/app.py
+++ b/app.py
@@ -11,19 +11,54 @@ from website import programs
 import utils
 from utils import timems, load_yaml_rt, dump_yaml_rt, version, is_debug_mode
 from website.log_fetcher import log_fetcher
-from website.auth import current_user, login_user_from_token_cookie, requires_login, is_admin, is_teacher, update_is_teacher
+from website.auth import (
+    current_user,
+    login_user_from_token_cookie,
+    requires_login,
+    is_admin,
+    is_teacher,
+    update_is_teacher,
+)
 from website.yaml_file import YamlFile
-from website import querylog, aws_helpers, jsonbin, translating, ab_proxying, cdn, database, achievements
+from website import (
+    querylog,
+    aws_helpers,
+    jsonbin,
+    translating,
+    ab_proxying,
+    cdn,
+    database,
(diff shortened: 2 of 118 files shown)


Felienne:
Thanks @rinkstiekema!! I think this is a great step towards more uniformity and we grow the number of contributors.

I think @rix0rrr might have an opinion on the exact details of the formatting (he is our designated opinion haver 😄) . 

And I also think we want to merge in his big refactoring of Flask (#3315) and then run the formatter again, because otherwise we will be fixing a lot of conflicts.

rinkstiekema:
> And I also think we want to merge in his big refactoring of Flask (#3315) and then run the formatter again, because otherwise we will be fixing a lot of conflicts.

@Felienne I have merged the branch `refactor-blueprints` into this branch and ran the formatting again for the changed files.

Felienne (changes requested):
Can we do this on a PR instead of locally?

rinkstiekema:
> Could we run black on each PR with a GitHub workflow? Or on each commit to main? We already have workflows running on these triggers so if we update these it will go automatically and no one has to remember anything!

Black is now run upon committing your code. It will then apply the formatting on the staged files - I think that should have the same effect as running it on each PR. However, indeed people have to install Black, but as that's part of `requirements.txt` it should happen automatically upon starting with the project?

If you would still like to have it exclusively on PR, you could use [this approach using Github Actions](https://peterevans.dev/posts/github-actions-how-to-automate-code-formatting-in-pull-requests/#further-examples), but I think only owners of the repository can set this up (and then need to maintain it obviously).

Felienne:
> this approach using Github Actions

I will look into how we can do this with an action, thanks a lot for the idea and the link!

PR #3331, opened 2022-09-26
[BUGFIX] fixes printing 5 - 5 

### LINKED ISSUE #3330: [BUG] print 5 - 5 sometimes fails
Thanks to a teacher for pointing this one out:

image

Sometimes (but not always) 5-5 will not be calculated

Felienne:
Fixes #3330

### DIFF
--- a/grammars/level6-Additions.lark
+++ b/grammars/level6-Additions.lark
@@ -18,10 +18,10 @@ assign: var (_IS | _EQUALS) expression | var ( _IS  | _EQUALS) textwithoutspaces
 
 sleep: _SLEEP (INT | list_access | var_access | expression)?
 
-?expression: simple_expression | expression _MULTIPLY atom -> multiplication | expression _DIVIDE atom -> division | expression _PLUS atom -> addition | expression _MINUS atom -> subtraction | _MINUS INT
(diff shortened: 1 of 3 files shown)


Felienne:
I don't love this hackity approach but since of the code snippets is failing I think we should hot patch now and fix this in the grammar later-on (maybe @jpelay or @boryanagoncharenko want to take a deeper look in due time?

Felienne:
> If I understand correctly, there is a grammar ambiguity in level 6-17 that went unnoticed for such a long time. The code `print 'minus is ' 5 - 5` is parsed as a print with its last 2 arguments being (integer, 5), (integer, -5) instead of (subtraction, [(integer, 5), (integer, 5)]). 

Indeed that exactly is the problem and somehow we did not have a matching test!

> If so, would not changing the priority of the expressions in the grammar resolve the matter?
> 
> I changed the first two lines of level6-Additions.lark to be the following and it seems that this works. However, with grammar ambiguities I never know whether I actually solved it or I did not run it enough to trigger the alternative parsing.
> 
> ```
> _print_argument: (_SPACE | quoted_text | list_access | expression | print_expression)*
> ?print_expression: var_access_print | error_unsupported_number | INT
> ```

Maybe! I fiddled with the grammar a little bit but I could not get it to work. I will try your approach too!

PR #3332, opened 2022-09-27
Enable the turtle commands in all levels

### LINKED ISSUE #3299: [BUG] Turtle doesn't work from level 11 onwards
**Describe the bug**
On level 11, the turtle commands are removed from the grammar. This shouldn't happen and the turtle commands should work in higher up levels.

Example from level 11:

imagen

boryanagoncharenko:
**Enable turtle commands for all levels**

_Deletes the explicit removal of turtle commands in level 11._
_Also updates the tests constant for the turtle level so it continues after level 10._

**Fixes #3299**

**How to test**

Go to any level 11 - 18 and verify that the turtle commands work and produce the expected outcome.
Example code:
```
forward 50
color green
turn 45
forward 50
```


Felienne:
Supercool @boryanagoncharenko!! Look what it allowed me to do:

image

image

Two little things:

* The syntax highlighter also needs to be updated (can be done in a separate PR if you want!)
* I can't do calculations in turn (maybe that is not new, but I think in higher levels it is more likely kids want to use it) Compare this to the program above which does the same but does not work if I inline the calc.

image

hedy.py
```diff
@@ -139,14 +139,14 @@ def promote_types(types, rules):
...
-    13 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or'],
-    14 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or'],
-    15 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or', 'while'],
-    16 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or', 'while'],
-    17 :['ask', 'is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or', 'while', 'elif'],
-    18 :['is', 'print', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat', 'and', 'or', 'while', 'elif', 'input'],
+    11 :['ask', 'is', 'print', 'forward', 'turn', 'sleep', 'at', 'random', 'add', 'to', 'remove', 'from', 'in', 'if', 'else', 'for', 'range', 'repeat'],
```
arjank, line 142: We also need to add the color command to all levels.

tests/test_level/test_level_11.py
```diff
@@ -1,5 +1,6 @@
 import hedy
 import textwrap
+from hedy import Command
```
arjank, line 3: This is the only change in this file. Is this import really necessary?

tests/test_level/test_level_12.py
```diff
@@ -282,6 +282,160 @@ def test_print_calc(self):
...
+            forward a""")
+        expected = HedyTester.dedent(
+            "a = 50",
+            HedyTester.forward_transpiled('a'))
+
+        self.multi_level_tester(
+            code=code,
+            expected=expected,
+            extra_check_function=self.is_turtle(),
+            max_level=18,
```
arjank, line 300: I don't think we should specify the max level here, or it should be set to HEDY_MAX_LEVEL. If level 19 is added and it would break things, we need to know.

tests/test_level/test_level_12.py
```diff
@@ -282,6 +282,160 @@ def test_print_calc(self):
...
+        code = textwrap.dedent("""\
+        directions is 10, 100, 360
+        forward directions at random""")
+
+        expected = HedyTester.dedent("""\
+        directions = [10, 100, 360]""",
+        HedyTester.forward_transpiled('random.choice(directions)'))
+
+        self.multi_level_tester(
+            max_level=15,
```
arjank, line 324: If max_level is less than 18 (HEDY_MAX_LEVEL) we should add tests for the higher levels too. So in this case I would expect a similar test at level 16.

arjank:
> Supercool @boryanagoncharenko and of course also @arjank for the first steps!! Look what it allowed me to do:
> 
> image
> 
> Two little things:
> 
>     * The syntax highlighter also needs to be updated (can be done in a separate PR if you want!)
> 
>     * I can't do calculations in turn (maybe that is not new, but I think in higher levels it is more likely kids want to use it) Compare this to the program above which does the same but does not work if I inline the calc.
> 
> 
> image

I don't think calculations in turn (and forward) ever worked. It makes sense to add that though, but I think that should be a different ticket/PR. 
In addition to that I also found that colors, especially those in lists, don't work when you're not using English color names. I don't know how easy/hard it is to solve that, but it should be a different ticket/PR.

hedy.py
arjank: I've added this.

tests/test_level/test_level_12.py
arjank: I've removed both max_level=18 lines in the tests for level 12.

tests/test_level/test_level_12.py
arjank: I've added some tests regarding lists in level 16. Lists are changed there, so new tests are required.

tests/test_level/test_level_11.py
boryanagoncharenko: Thanks for pointing it out! It is unnecessary and I removed it.

boryanagoncharenko:
I created an issue for the adding support for calculations #3345 and one more for fixing the highlighting #3346.

@arjank in you comment you mention that "In addition to that I also found that colors, especially those in lists, don't work when you're not using English color names." I am not sure I understand the issue and I am not even sure what the expected behavior is. Would you be able to create an issue with more details?

Felienne:
> I created an issue for the adding support for calculations #3345 and one more for fixing the highlighting #3346.

Thanks!! Let's merge this now and deal with the rest separately.

> @arjank in you comment you mention that "In addition to that I also found that colors, especially those in lists, don't work when you're not using English color names." I am not sure I understand the issue and I am not even sure what the expected behavior is. Would you be able to create an issue with more details?

I think I understand this, will add a new issue!

Felienne (approved):

PR #3334, opened 2022-09-27
[LANGUAGE] Simpler way to do 3331

Felienne:
I think you were right @boryanagoncharenko (see here: https://github.com/Felienne/hedy/pull/3331#issuecomment-1259330725)


tests/test_level/test_level_06.py
```diff
@@ -535,15 +535,14 @@ def test_print_subtraction_without_text(self):
...
+        expected = "print(f'And the winner is {int(5) - int(5)}')"
         output = 'And the winner is 0'
 
-        for level in range(self.level, 12):
-            print(f"passed for level {level}")
-            result = hedy.transpile(code, level, 'en')
-            self.assertIn(result.code, [expected_1, expected_2])
-            self.assertEqual(output, HedyTester.run_code(result))
+        self.multi_level_tester(
+            max_level=11,
```
arjank, line 542: I think we should also test this for level 12 and up.

Felienne: Good catch!! 

The reason that this runs until 11 only is that the output is different at level 12 (`'''` instead of `'`) This is true for a few more tests and @boryanagoncharenko is still working on a better solution but let's indeed not make it worse by skipping test cases in the name of "we will improve this later" haha.

PR #3351, opened 2022-09-29
Fixed some mistakes in the level 8 quiz

MarleenGilsing:
Fixed some mistakes in the level 8 quiz

@Felienne I've changed both the Dutch and the English file, so this might mess up Weblate.


Felienne:
Haha it will mess with weblate but I will handle it :)

Felienne (approved):

PR #3327, opened 2022-09-26
Only show language switcher on ace when admin

TiBiBa:
First, provide a one line summary of your changes in the **Title of the PR**

**Now fill out the remainder of this template by replacing all _italic_ content**

**Description**

_Changes in detail_

For example: "**Adds** translations of levels 1 to 12 to Polish" (note the use of present tense!)

**Fixes**
This PR fixes #3320

**How to test**

* If this is a UI change: _describe how to run your code to see it in action. See this https://github.com/Felienne/hedy/pull/880#issue-1016304308 for an example_
* If this is a language change: _add a few tests showing the difference_

**Checklist**
Done? Check if you have it all in place using this list:*
  
- [ ] Describes changes in the format above (present tense)
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #3355, opened 2022-09-29
[FEATURE] Allows teachers to hide/show keyword switcher in menu bar

### LINKED ISSUE #3354: [UI idea] Allow teachers to hide new language switcher
**Idea incl level**
In https://github.com/Felienne/hedy/pull/3327 we remove the Ace based switcher:

image

And also mention removing the "hide language"customization (https://github.com/Felienne/hedy/pull/3327#issue-1385516608)

But! We now have another language switcher:

image

Shall we put back the customization feature but use it to disable this switcher (since we also allow hiding of the cheat sheet?)

It is not high prio but would be nice!

TiBiBa:
**Description**
We partly re-implement the code deleted in #3327 where we decided to remove the language switcher from the code editor. We now re-introduce a "hide keyword switcher" but instead aim on the menubar dropdown keyword switcher. The back-end code can remain the same, we only had to make some adjustments to the front-end.

This PR fixes #3354.

**How to test**
Make sure you are in a class and have a keyword-supported-non-english profile language. Verify that you can see and use the keyword switcher dropdown from the menubar. When hidden using the class customizations the button should been gone.

### DIFF
--- a/templates/customize-class.html
+++ b/templates/customize-class.html
@@ -134,6 +134,12 @@
                                 </td>
                             </tr>
                             <tr>
+                                 <td class="text-left border-t border-r border-gray-400">{{_('hide_keyword_switcher')}}</td>
+                                 <td class="border-t border-gray-400">
+                                     <input class="other_settings_checkbox" id="hide_keyword_switcher" type="checkbox" {% if "hide_keyword_switcher" in customizations['other_settings'] %}checked{% endif %}>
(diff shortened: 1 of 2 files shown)


Felienne:
Wow that was quick, thanks!! I will have to again redo a video for the teachers course 🤣 but this is a great addition!

Felienne (approved):

PR #3359, opened 2022-09-30
[TURTLE] Generate dst and png files for machines

### LINKED ISSUE #3357: [UI idea] generate vector in addition to dst for turtle graphics
**Idea incl level**
We now generate a dst file for embroidery machines, but if we want to run the drawing on a pen plotter we have to convert it to a vector. This is not a real problem (I do it now with https://stitchpad.io/ manually) but it would be easier if we generate both files at once so teachers don't have to convert.

Felienne:
Fixes #3357


Felienne:
I am pretty sure this works but a sanity check (maybe by @TiBiBa who worked on this originally?) and a dry run on another machine would be great!

Felienne:
I hobbled this together for a workshop next Tuesday but the workshop is cancelled so we can spend a bit more time figuring out how we want this to work (maybe indeed as Timon suggested with a switcher in the UI (or in the profile?))

Felienne:
Ah I did some renaming in the front-end, I probably messed something up, sorry! Can you take a look at this if you also change the button into a dropdown?

PR #3362, opened 2022-09-30
[CONCEPT] Adds option for teacher to set quiz threshold for unlocking next level

TiBiBa:

**Description**

_Changes in detail_

**How to test**

**Checklist**
  

Felienne:
Ha @TiBiBa!! This seems to work, cool (apart from the e2e tests but I am sure those are test issues and not actual problems)

I do think we can do a tiny bit better in terms of UX.

If I fail the threshold, I see this:

image

Then if I click next level:

image

Could we bring the message up, so that the button is replaced by "you score is not yet the needed <X> to move on"? (potentially we can do this in a separate PR!)

TiBiBa:
> Ha @TiBiBa!! This seems to work, cool (apart from the e2e tests but I am sure those are test issues and not actual problems)
> 
> I do think we can do a tiny bit better in terms of UX.
> 
> If I fail the threshold, I see this:
> 
> image
> 
> Then if I click next level:
> 
> image
> 
> Could we bring the message up, so that the button is replaced by "you score is not yet the needed to move on"? (potentially we can do this in a separate PR!)

Agree that this should be improved. This requires a bit of work so maybe it would be best to fix this in a separate PR indeed. Because we do have to make a few additional db calls when finished the quiz to achieve this:

- Get the class customizations
- Get all user quiz stats -> get highest of level (including the current one)
- Then, if < threshold -> replace button text to restart quiz (?)

Alternatively, we retrieve the threshold once and store this in our session. With the added risk that our session threshold is not up-to-date with the set threshold once the teacher changes this. So we have to decide on either the increase in db calls on one side and the possible issue of incorrect comparisons on the other side.

TiBiBa:
Converted to draft to implemented "greyed out" buttons instead of removing them

Felienne (approved):

Felienne:
This will anger Weblate but I will deal with that :)

PR #3370, opened 2022-10-03
Adding blur button when not in code editor

### LINKED ISSUE #3250: [BUG] When a program is submitted, the blur button does not show
image

jdomil:
**Description**

Displays a blur button in the coding interface for students to be able to hide their code when showing it to others. This button is only displayed when not in edition mode so as not to clutter the UI.

Fixes #3250 

**How to test**

1. Navigate to http://0.0.0.0:8080/hedy#default
2. In this first interface, you shouldn't be able to see any blur button at the bottom right corner of the code editor
3. Follow the level instructions and complete the exercise
4. "Hit Save & share code"
5. Visit the URL in your clipboard and ensure this time there is a blur button on the bottom right corner of the area showing your code
6. Hitting it should blur your code
7. If you click the "Edit code" button again, you should be back to the code editor interface and the blur button should have disappeared.


Felienne (changes requested):
ok I know this sounds very picky, but I do think we want to keep the blur button enabled when in developers mode in non-submitted programs, as we have now.

So: with programmers mode, blur button is there:

image

No programmers mode, no blur:

image

Could you reset that too? I tried to figure out myself how to do it with the new code but I could not explicitly find the place where to do it.

jdomil:
It is not very picky at all! I completely forgot about developer mode 🤦🏻‍♂️. I had to get a bit creative to enable the blur button in developer mode, let me know what you think about the solution. Also, I changed the variable name to `blur_button_available`, but I can name it whatever you think is best as it's an easy change. 

Have a look at the changes and let me know what you think, I don't mind having to go back and rewrite code as I'm learning a lot from this. 😄 .

Felienne (approved):
Works perfectly now and is a lot more readable with the new variable name!

image

Thanks so much for your first Hedy contribution @jdomil!!

PR #3371, opened 2022-10-03
Fixes translation of keywords with several  options

### LINKED ISSUE #2711: [BUG] repete runs and highlights but can't be translated

jpelay:

**Description**

_Changes in detail_

Fixes the translation of keywords that have several options, for example repeat in French can be translated as "répète" and "repete".  This was done by changing the dictionary in the translator from a dictionary that maps strings to strings, to one that maps strings to list of strings. There's one thing I don't feel particularly happy about: while we can translate repete back and forth, we won't get the same result, it means that we can't use repete in multi-level testing because when it's translated from English back to French it will return "répète", meaning that the test witll fail because technically is not the same string. 

**Fixes #2711 **

**How to test**

I added a few tests

**Checklist**
  

Felienne (approved):
Thanks!! I understand the issue the the back and forth but since we don't have the translate button anymore (for kids) that is not too much of a big deal!

PR #3380, opened 2022-10-04
Improves readability on mobile devices

### LINKED ISSUE #3379: [UI idea] Readable text on mobile layouts
**The text should be readable on all devices, including mobile devices**
Currently the text is not really readable on mobile devices, it's too small.

The text on this example image appears bigger than it does in reality. This issue isn't about making things perfect, the main objective is to get a decent starting point that we can use for further improvements.

image

arjank:
Makes sure fonts are displayed properly and also uses full width on smaller devices for adventure texts and example codes. Also keeps the logo nice and round.

**Description**

Makes sure fonts are displayed properly and also uses full width on smaller devices for adventure texts and example codes. Also keeps the logo nice and round.

There's probably a lot more that can be done, but this should be considered a first step.

**Fixes #3379**

**How to test**

* Use the inspector in your browser and enable mobile view (how to do that depends on the browser). Also you could use a mobile device to view the Hedy version on your computer. For that they need to be on the same network.

The changes are template only, so it should be easy to compare current and new situations.

**Checklist**
  
### DIFF
--- a/templates/incl-adventure-tabs.html
+++ b/templates/incl-adventure-tabs.html
@@ -48,12 +48,12 @@
               </div>
             {% for story in adventure.extra_stories %}
                 <div class="lg:flex lg:flex-row border-t-2">
-                    <div class="p-2 ltr:mr-2 rtl:ml-2">
+                    <div class="lg:w-1/2 w-full p-2 ltr:mr-2 rtl:ml-2">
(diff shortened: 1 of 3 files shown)


Felienne:
Ah that improves a bit already, thanks a bunch @arjank!!

image

image

PR #3387, opened 2022-10-07
[FEATURE] Adds admin menubar item if admin

TiBiBa:
**Description**
This PR adds a menubar item for the /admin page in the menubar. Personally I don't like having to manually navigate to /admin each time I log in as an admin: this PR saves us one step.

**Fixes**
No related issue.

**How to test**
Log in as an admin, verify the new menubar item. Also verify that it is not there when not logged in as admin.


Felienne (approved):
ah nice I have thought about something like this MANY TIMES bu I was always to lazy to do something.

image

PR #3392, opened 2022-10-08
[CHORE] Adds editor config

Felienne:
As suggested by @arjank! For more info, see: https://editorconfig.org/


arjank (approved):
Looks good to me :)

We can always extend this for other file types, but having the infrastructure in place is a first requirement.

PR #3393, opened 2022-10-08
Allows sending emails in other languages

### LINKED ISSUE #3289: [BUG] Language of email depends on admin lang not user lang?
**Describe the bug**
I suspect (but have not confirmed 100%) that the email text that users get when their account is changed into a teacher account is dependent on the admin language and not the user's language. It is hard to test since email sending does not work locally but it seems this is the case at least on alpha.

arjank:
**Description**

We usually use the language of the current user when sending emails, but that is not suitable in all cases. Specifically if an admin needs to email a user, this should be done in the language of the user.

This fixes the issue when an admin marks a user as teacher, and also when an admin changes the email address of a user.

**Fixes #3289 **

**How to test**

As stated in the issue, this is hard to test locally because sending emails does not work locally. This would need to be tested on alpha.

**Checklist**
  
### DIFF
--- a/website/auth.py
+++ b/website/auth.py
@@ -8,7 +8,7 @@ from functools import wraps
 import boto3
 from botocore.exceptions import ClientError as email_error, NoCredentialsError
 from flask import g, request, session
-from flask_babel import gettext
+from flask_babel import gettext, force_locale
 import requests
 
 import utils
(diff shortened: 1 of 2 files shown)


website/auth.py
```diff
@@ -314,6 +314,17 @@ def send_email_template(template, email, link=None, username=gettext('user')):
 
     send_email(email, subject, body_plain, body_html)
 
+
+# By default, emails are sent in the locale of the logged-in user.
+# This function is to be used if the email needs to be sent in another locale.
+def send_localized_email_template(locale, template, email, link=None, username=None):
+    with force_locale(locale):
+        if username is None:
+            # We want to use the correct locale for this text
```
Felienne, line 323: Do we mean "current" here? (in stead of correct?)

Felienne:
Ow this is so awesome @arjank! I looked into gettext a bit but I could not find the right setting and you just found it! Great!! I will test it tomorrow!

arjank:
> Ow this is so awesome @arjank! I looked into gettext a bit but I could not find the right setting and you just found it! Great!! I will test it tomorrow!

I just got lucky with Google ;)

website/auth.py
arjank: Could be either, I guess. We want the locale that's used to send the email, so the correct locale. And at this point it's also the current locale. It's not the correct locale when you enter the function, so we can't define it as username=gettext('user') (as is the case in send_email_template() )

Felienne: I think I don't understand in what scenario we do not have a username, which is why the comment it hard to understand (but it is ok, I understand what it does)

Felienne (approved):
Just tested this and indeed, the forced locale works, supercool!! Just added an extra line of comment, so this can now be tested on alpha!

Issue #3378, opened 2022-10-04
[UI idea] tag Dutch teachers in mailchimp api call


Felienne:
We are experimenting with a mailchimp journey for new teachers, currently only existing in Dutch. Once #3304 is ready, we can tag Dutch teachers with #teachertrainingnl and they will then receive a weekly mailing with more info about Hedy features.

This is not a very hard issue to fix (hence the good first issue) it might require a bit of digging into the MailChimp api if you are not familiar with it:

The tag adding itself should be changed here so that is not the hard part.

https://github.com/Felienne/hedy/blob/e70ea2a80f4ffa5b74157d6c49de5b6fc32fdc6b/website/auth.py#L40

PR #3394, opened 2022-10-08
#3378 [UI idea] tag Dutch teachers in mailchimp api call 

jdomil:
**Description**

Includes information on country and user role as tags when creating members in Mailchimp thorugh the Members API. This will allow us to test sending tailored emails to NL teachers weekly about new functionality.

Fixes [3378](https://github.com/Felienne/hedy/issues/3378)

**How to test**

I'm unsure how to test this since it looks like we don't post to the Mailchimp API in test mode on purpose.

**Checklist**
  

jdomil:
Hi @Felienne ! I'm sure this one will require a bit of tweaking. There are two points to raise:

1. The issue states that "we can tag Dutch teachers with #teachertrainingnl". I have opted for a slightly different solution as I think it scales better. Instead of creating logic to add a '#teachertrainingnl' tag for NL teachers, I'm passing both information on role and country to Mailchimp as tags. I thought this would allow us to create  [whatever logic we need to create on Mailchimp](https://mailchimp.com/es/help/send-tags/) to send tailored emails to users with the tags "NL" and "teacher" now, or whatever other combination of tags in the future. That said, let me know if you think there's a better approach for this.\
2. I'm not sure how to test this change as it looks like we don't post to the Mailchimp API in test mode on purpose. I've tested it with my own test Mailchimp account and API and it seems to work just fine, but I'm not sure how you can reproduce it. 

Let me know if you'd like me to make any changes.

jdomil:
Actually, I just realized I introduced a bug with this changes at profile.py. It looks like we make a call to the Members Mailchimp API when a user updates their profile details, so I need to tweak the `mailchimp_subscribe_user` function call there.

Looking into it, I see I can get the country easily from `body['country']`, however, I can't seem to find a reference to the property `is_teacher`. After the changes in [3304](https://github.com/Felienne/hedy/pull/3304), do you know if we are storing in the database whether a user is a teacher or not? It looks like it's not specified in the user's profile at the moment.

Felienne:
Thanks a lot for this @jdomil!!

> 1. The issue states that "we can tag Dutch teachers with #teachertrainingnl". I have opted for a slightly different solution as I think it scales better. Instead of creating logic to add a '#teachertrainingnl' tag for NL teachers, I'm passing both information on role and country to Mailchimp as tags. I thought this would allow us to create  [whatever logic we need to create on Mailchimp](https://mailchimp.com/es/help/send-tags/) to send tailored emails to users with the tags "NL" and "teacher" now, or whatever other combination of tags in the future. That said, let me know if you think there's a better approach for this.\

Ah yes that is better!!! Then we have the flexibility to send different emails, and I am sure I can change the workflow in MailChimp to require 2 tags.

> 2. I'm not sure how to test this change as it looks like we don't post to the Mailchimp API in test mode on purpose. I've tested it with my own test Mailchimp account and API and it seems to work just fine, but I'm not sure how you can reproduce it.

Yeah I don't think we can test this locally. We can merge to main, after which it will be live on hedy-alpha.herokuapp.com and then we can test it there.

jdomil:
Alright @Felienne , this is almost done. There are a couple of considerations:

1. This can't be merged until #3304 is merged since it uses code that the aforementioned PR is introducing, more specifically the parameter `is_teacher` in line 136 of "auth_pages.py". 
2. I modified the control flow logic in line 133 of "auth_pages.py" since after testing, I realized that ` body['subscribe']` never evaluates to True, which is interrupting the Mailchimp flow even for users who decide to subscribe. It's also redundant in my opinion, since we already check if `'subscribe' in body`.
3. I fixed a bug whereby when updating a profile email, we were checking if the new email provided was in Mailchimp instead of the old email on file. This was happening because we are updating the user email with the newly provided one in line 61 of "profile.py", and then using this new user email to check whether it exists in Mailchimp in line 81. I introduced an intermediate variable called `old_user_email` before updating the user email to reference it in the API call to Mailchimp and ensure we are loooking for the old email in Mailchimp to then update it with the new one provided. I have tested with my own Mailchimp account and it now works as expected. 

I'm leaving this as a draft until #3304 is approved, but feel free to suggest any changes before requesting the merge.

jdomil:
All good to go @Felienne. Tested with the new sign up flow and my own Mailchimp API keys and it should work as expected.

Felienne (approved):
Looks good @jdomil! I can't test it offline but I will merge and then test on alpha!

Felienne:
Hi @jdomil!!

This change is now live and indeed it works!! Thanks again 🥳 !

image

PR #3395, opened 2022-10-09
3335 UI idea extend skulpt with pygame4skulpt

ToniSkulj:
Make it possible for Skulpt to execute Pygame (partially), by including PyGame4Skulpt as a external library for Skulpt.

**Description**

Things worthy of mentioning:

1. [PyGame4Skulpt ](https://github.com/Dustyposa/pygame4skulpt) is added to Skulpt
2. variable _hasPygame_ is used analog to the _hasTurtle_ variable, for levels that will need Pygame functionality in the future.
3. the function runPythonProgram is changed to accommodate _hasPygame_ . This way the library is only imported when necessary.
4.  the function runPythonProgram is 'exported', it is accessible via the DOM. This makes it a lot easier to test programs by being able to run real Python, especially when testing Pygame related code.
5. Same canvas is being used for Turtle and Pygame code

**How to test**

1. Open up any level and copy & paste the following Python code into the editor

```python
import math, random

width = 711
heigth = 300

print('Drawing Circles!')

def get_random_color():
    return random.randint(0, 255), random.randint(0, 255), random.randint(0, 255)


def draw_circle():
    r = d / (2 * n)
    kx = width / (2 * n)
    ky = heigth / (2 * n)
    canvas.fill(pygame.Color(247, 250, 252, 255))
    for i in range(n):
        (cx, cy) = ((2 * i + 1) * kx, heigth - (2 * i + 1) * ky)
        pygame.draw.circle(canvas, get_random_color(), (round(cx), round(cy)), round(r), 2)
    pygame.display.update()


d = math.sqrt(width ** 2 + heigth ** 2)

n = 7
draw_circle()

end = False
while not end:
    event = pygame.event.wait()
    if event.type == pygame.QUIT:
        end = True
    elif event.type == pygame.MOUSEBUTTONDOWN:
        if event.button == 1:
            n += 1
            draw_circle()
        elif event.button == 3 and n > 1:
            n -= 1
            draw_circle()
        if event.key == pygame.K_UP:
            n += 1
            draw_circle()
        elif event.key == pygame.K_DOWN and n > 1:
            n -= 1
            draw_circle()

pygame.quit()
```

2. Open up your browser console & execute the following JS
```javascript
let code = hedyApp.get_trimmed_code() 
hedyApp.runPythonProgram(code, false, true, false, false)
```

3. You should now be able to increase the circles with the **up arrow/right mouse click** and decrease with the **down arrow/left mouse click**, you can stop executing Pygame with the **escape button**.
image


TiBiBa:
Hi! Can you set this PR as draft until it is ready for review? Then we know when we can take a look! (As there are still some commits being made after the PR creation).

Felienne (approved):
Check!! 
image

This will be VERY useful in debugging for me in class too (guess I will be the only teacher to use this... but maybe not?!)

PR #3398, opened 2022-10-11
[FEATURE] Changes drawing arrow into turtle

### LINKED ISSUE #3389: [UI idea] Turn the turtle triangle into an actual turtle!
**Idea incl level**

In Python you can change the shape to be a real turtle: 

https://www.geeksforgeeks.org/turtle-shape-function-in-python/

I think Skulpt also supports this, would look nice!

TiBiBa:
**Description**
In this PR we change the turtle arrow into an actual turtle!

**Fixes**
This PR fixes #3389

**How to test**
Run any turtle program and verify that the arrow is now displayed as a turtle.


Felienne (approved):
Works!!

image

PR #3399, opened 2022-10-11
[BUG FIX] 'error or empty' achievement can be achieved again

TiBiBa:
**Description**
This PR fixes an issue where we dependent on an unused variable in the TypeScript code for the 'error or empty' achievement check. Therefore it was never reached; and users where unable to achieve it. Thanks to an observant Hedy user (Kuno) this bug was noted.

**Fixes**
No related issue.

**How to test**
Make sure you have an account that doesn't have the 'error or empty' achievement yet. Create a program without an input but that is an actual program! For example, by using an `ask` without an `echo` in level 1. Verify that the achieved is received.


Felienne:
Hi @TiBiBa 

I have not yet managed to get the achievement again :( 

image

TiBiBa:
> Hi @TiBiBa
> 
> I have not yet managed to get the achievement again :(
> 
> image

Weird! It should be working (tested it on a new account as well), will take another look at this later today.

Felienne (changes requested):
Tried again but I can't seem to get the achievement (Chrome or Safari on mac)

TiBiBa:
Found it! It was broken due to the routing re-structuring. The achievement push from the front-end was made to the `/achievements` route. But due to the re-structuring the route was not found as `/achievements/achievements`. This is fixed and we also fixed a minor issue on a smaller screen where the error/warning box of the code editor would overlay on the achievement pop-up.

Felienne (approved):
Check:

image

PR #3405, opened 2022-10-12
[UI] Make error grey for better readability 

### LINKED ISSUE #3402: [UI idea] error color is unreadable with correct keyword
We were tweaking the error color a while back (https://github.com/Felienne/hedy/pull/3031) and this looks nice for a wrong (and thus not highlighted) keyword:

image

But is the error is not in the keyword, the color overlaps with the keyword color:

image

So I guess we need to change the color to be a bit more red? Or orange or whatever looks nice and doesn't clash.

Felienne:
Fixes #3402.

I readded the red lines from #3031 (they were a great idea after all, I was wrong there!) and now use blue with opacity:

Keyword:
image

Non-keyword:
image


TiBiBa:
If you mean the top and bottom border on the line you can use something like this:

`border-red-500 border-b-2 border-t-2`

Felienne:
> If you mean the top and bottom border on the line you can use something like this:
> 
> `border-red-500 border-b-2 border-t-2`

Yes! I wanted to fetch that from an old commit but this is helpful! Will try it later today to see if the looks nice with the grey.

Issue #3400, opened 2022-10-12
[CONTENT] Remove red text in adventure mode


Felienne:
**Describe the issue**
Remove red sentence in adventure more (for example: https://hedy.org/adventure/parrot)

**Add a screenshot (optional)**
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1003685/195296691-95e060a7-5f64-48ee-a753-f064fc63f104.png">

**Expected text**
Just remove this text, it kids need to go elsewhere, a teacher will tell them.

I am copy-pasting the text here for easy grepping: `You're currently in adventure 'Parrot', click on 'Hedy' to view all adventures.`

Lakshayainani:
@Felienne, may I try resolving this ?

Felienne:
> @Felienne, may I try resolving this ?

Ah sorry @Lakshayainani, in the mean time it has been fixed! Let us know if you want to pick up something else!

PR #3408, opened 2022-10-12
3400 content remove red text in adventure mode

ToniSkulj:
Removes the red text in adventure mode #3400  , also removes the (i guess) unnecessary translations.

image

**How to Test**
Just visit _../adventure/parrot_. The red text should be absent.

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -683,9 +683,6 @@ msgstr ""
 msgid "quiz_tab"
 msgstr ""
 
-msgid "specific_adventure_mode"
-msgstr ""
-
 msgid "example_code_header"
 msgstr ""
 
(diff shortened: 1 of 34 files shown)


Felienne:
Ah this is an interesting question (maybe @Mark-Giesen knows this) If we remove a string, do we manually remove it from the yamls like here or will Weblate handle that?

Felienne:
And thanks for your first (merged) PR @ToniSkulj!!

Felienne (approved):

Mark-Giesen:
> Ah this is an interesting question (maybe @Mark-Giesen knows this) If we remove a string, do we manually remove it from the yamls like here or will Weblate handle that?

If you delete the key from our source language (English), it's deleted in all languages. If you delete it in any other language, it's probably added again some time later.

PR #3416, opened 2022-10-13
(3407) Dockerfile: Install nodejs and npm from NodeSource

### LINKED ISSUE #3407: [CHORE] Add npm to docker container
It would be nice to add [npm](https://www.npmjs.com/) (version 16) to the Docker container. That way, the user can easily start on developing front-end when using Docker. Should be easy if you are familiar with Docker.

pjotrligthart:
**Description**
Adds a line to the Dockerfile that installs nodejs and npm LTS into the Docker image from NodeSource, using the method recommended in the npm and NodeSorce docs ( https://docs.npmjs.com/downloading-and-installing-node-js-and-npm -- https://github.com/nodesource/distributions/blob/master/README.md#debinstall )

I'm a bit hesistant to run code from a shell script piped from a web source using curl. However, this is part of the install method recommended by NodeSource and automates a lot of checks and installation steps, so I still used that method.

Fixes #3407

**How to test**
(Re)build the docker image (possibly with the `--no-cache` option), the image should now contain the `npm` utility


ToniSkulj (approved):
Works as expected!

image

PR #3418, opened 2022-10-14
(3412) No longer copy files to Docker container during docker build, use bind mount instead

### LINKED ISSUE #3412: [DOCS, CHORE] Add mounting folder to docker documentation
Currently the CONTRIBUTING.md documentation states that to start running Docker the following command should be run:
```bash
docker run -it --rm -p 8080:8080 hedy
```
Since the project is making use of file watchers, it's necessary to mount the directory when running Docker. Otherwise, the file watcher won't see any changes! It seems wise to add this to the documentation. 

Following link should help [https://docs.docker.com/storage/bind-mounts/](https://docs.docker.com/storage/bind-mounts/).
Should be easy if you are familiar with Docker.

pjotrligthart:
**Description**
Update the Dockerfile to no longer copy the project files into the container during `docker build`, and update the documentation in CONTRIBUTING.md with instructions to use bind mounts instead. This should eliminate the need to rebuild the docker image after each change to the code!

I tested the updated `docker run` command on my Linux installation and everything seems to work there, but it would be nice if people could test this command on other platforms (Windows+Powershell comes to mind...), so I know whether I need to add additional instructions for those other platforms.

Fixes #3412 

**How to test**
 (Re)build the docker image (possibly with the `--no-cache` option) and use the updated `docker run` command from CONTRIBUTING.md to launch the new version of the container, everything should work as normal!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt install -y curl && curl -fsSL https://deb.nodesource.com/setup_lts.x | b
 
 WORKDIR /app
 
-COPY . .
+#COPY . .
 
 EXPOSE 8080
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt install -y curl && curl -fsSL https://deb.nodesource.com/setup_lts.x | b
 
 WORKDIR /app
 
-#COPY . .
+COPY . .
 
 EXPOSE 8080
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,7 +203,7 @@ docker build -t hedy .
 and then:
 
 ```bash
-docker run -it --rm -p 8080:8080 hedy
+docker run -it --rm -p 8080:8080 --mount type=bind,source="$(pwd)",target=/app hedy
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt install -y curl && curl -fsSL https://deb.nodesource.com/setup_lts.x | b
 
 WORKDIR /app
 
-COPY . .
+#COPY . .
```
ToniSkulj, line 13: @pjotrligthart I suggest leaving in the COPY in the dockerfile. This way if someone forgets (a.k.a does not read docs) to mount, the image still works as expected. Docker will still mount regardless of using COPY or not.

pjotrligthart: Taking a look at this now, thanks for the input!

pjotrligthart: @ToniSkulj I've re-enabled the COPY command, please test/review as needed

ToniSkulj (approved):
Works!

After using the commands in the docs, If I add **test.text** to my Windows hedy directory
image

It shows up in the Docker container without rebuilding the image
image

PR #3423, opened 2022-10-17
[Language] Adds rule for lonely quoted text

### LINKED ISSUE #3375: [Language idea] add extra rule for forgotten command?
**Language idea incl level**

Shall we add an extra gramma rule for text without a command in front (f.e. this in level 4 and up) so we can throw an exception?

```
    print 'Welkom bij de mcHedy!'
    'Bedankt voor uw bestelling.'
```

I am seeing this quite a lot in the logs.

jpelay:
**Description**

Adds a new exception in level 4 to detect lonely quoted text. Like this:

image

**Fixes #3375**

**How to test**

I added a small test for this behavior. 

**Checklist**
  

jpelay:
There was a failing test related to a repeat with only text. I changed that test to throw this new exception, and added a new test with another behavior of that previous exception, since that one works not only for text, but for any print argument.

Felienne:
Sadly this has conflicts now because of the many weblate changes @jpelay. If you fix then we can merge!

Felienne (approved):

jpelay:
Done here as well!

PR #3427, opened 2022-10-17
Enables admin to decline a teacher request

### LINKED ISSUE #3425: Enable admin to also decline a teacher request
Copied from #3421:
_4. Decline teacher request
Right now I get a question mark for a request, if I click it, it get's oked. But sometimes a kid clicks the button and I want to decline. That now is only possible with 2 clicks: accept and then delete again. Not a problem for me to click twice but it also means the kid gets an email that they teacher's account is approved which is annoying._

TiBiBa:
**Description**
In this PR we add the option for the admin to decline a teacher request. A small fix on first thoughts but we didn't have decline callback on the confirm modal (yet). Therefore up until now we where only able to perform an action on a confirmation within the modal but not on a decline. Therefore we add a new callback to the confirm modal to support this action.

**Fixes**
This PR fixes #3425

**How to test**
Make sure you are an admin and have another account with a pending teacher request. Decline it within the admin user overview and the page should reload. Verify in the database that the request is gone or log in to the account with the pending request and verify that a new request can be made.


Felienne (approved):
Thanks voor de snelle service @TiBiBa!!

PR #3442, opened 2022-10-18
Slightly adjust and add to swedish translations

JonathanMurray:
**Description**

Fixes minor typo/formatting in swedish translation, moves some of the levels to make them ordered, and adds translation for level 11.

**How to test**

Visit https://www.hedycode.com/hedy/11 with language=swedish

**Checklist**
  

JonathanMurray:
Hi! Nice project (and presentation on the Strange Loop conference)! 

As I edited the translations manually, it bugged me a bit that they weren't properly ordered, so I moved things around a bit. I want to double-check though: does it make sense to try and maintain ordering in these translation files or is there some bot / procedure that is likely to change things around again (in which case I suppose I'm just making the git history more difficult to read with my change 😄)?

Felienne:
Hi @JonathanMurray!

No it does not make sense to move levels around, because Weblate places them "random" so it will be destroyed when people translate through Weblate (which they mostly do). But it seems no conflicts were created, so all good, and thanks for improving!!

Felienne (approved):

JonathanMurray:
> because Weblate places them "random" so it will be destroyed when people translate through Weblate (which they mostly do)

Ah! As I suspected. I will refrain from re-ordering in the future. Thanks for clarifying!

PR #3461, opened 2022-10-19
[TESTING] Attempt 2 for fixing test optimization

TiBiBa:
Only runs all level and snippet tests on PRs that change the grammar.


TiBiBa:
@Felienne This does seems to do the trick!

Felienne (approved):
ok, let's see if this works 🙏

PR #3462, opened 2022-10-19
Extend list of reserved words

JonathanMurray:
**Description**

Adds 'input' to the list of reserved words, as it's a Python method that our generated code can call.

**Fixes issues/3454**

Fixes https://github.com/Felienne/hedy/issues/3454 (I think and hope)

**How to test**

This program fails on level 11 before the change but should succeed after it:
```
input = "some string"
x = ask "question"
```

If we want, I can try to add a unit-test for this, but I'm not sure where to put it. `test_level_02.py` since the the issue can be reproduced from that level onward? (Or perhaps it's a bit overkill to test every reserved keyword? I'm flexible on the matter.)

**Checklist**
  
### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -33,8 +33,11 @@ local_keywords_enabled = True
 # dictionary to store transpilers
 TRANSPILER_LOOKUP = {}
 
-# Python keywords need hashing when used as var names
-reserved_words = ['and', 'except', 'lambda', 'with', 'as', 'finally', 'nonlocal', 'while', 'assert', 'False', 'None', 'yield', 'break', 'for', 'not', 'class', 'from', 'or', 'continue', 'global', 'pass', 'def', 'if', 'raise', 'del', 'import', 'return', 'elif', 'in', 'True', 'else', 'is', 'try']
+# builtins taken from 3.11.0 docs: https://docs.python.org/3/library/functions.html
+PYTHON_BUILTIN_FUNCTIONS = ['abs', 'aiter', 'all', 'any', 'anext', 'ascii', 'bin', 'bool', 'breakpoint', 'bytearray', 'bytes', 'callable', 'chr', 'classmethod', 'compile', 'complex', 'delattr', 'dict', 'dir', 'divmod', 'enumerate', 'eval', 'exec', 'filter', 'float', 'format', 'frozenset', 'getattr', 'globals', 'hasattr', 'hash', 'help', 'hex', 'id', 'input', 'int', 'isinstance', 'issubclass', 'iter', 'len', 'list', 'locals', 'map', 'max', 'memoryview', 'min', 'next', 'object', 'oct', 'open', 'ord', 'pow', 'print', 'property', 'range', 'repr', 'reversed', 'round', 'set', 'setattr', 'slice', 'sorted', 'staticmethod', 'str', 'sum', 'super', 'tuple', 'type', 'vars', 'zip']
+PYTHON_KEYWORDS = ['and', 'except', 'lambda', 'with', 'as', 'finally', 'nonlocal', 'while', 'assert', 'False', 'None', 'yield', 'break', 'for', 'not', 'class', 'from', 'or', 'continue', 'global', 'pass', 'def', 'if', 'raise', 'del', 'import', 'return', 'elif', 'in', 'True', 'else', 'is', 'try', 'int']
+# Python keywords and function names need hashing when used as var names
+reserved_words = set(PYTHON_BUILTIN_FUNCTIONS + PYTHON_KEYWORDS)
 
 # Let's retrieve all keywords dynamically from the cached KEYWORDS dictionary
 indent_keywords = []


JonathanMurray:
Looking at the list of reserved words, I notice that it seems to cover language keywords but no (or few) function names. Is there some specific reason for not including those, or has it just not been an issue yet?

It seems like errors similar to the one encountered in https://github.com/Felienne/hedy/issues/3454 can be reproduced for many such function names:

(level 11)

('str' object is not callable)
```
int = 1
int = int + 1
print int
```

('str' object has no attribute 'sleep')
```
time = "04:00"
for i in range 0 to 1
    print "good morning"
```

('str' object is not callable)
```
convert_numerals = "yes"
if convert_numerals == "yes"
    print "Let's convert some numerals!"
```
(admittedly a far-fetched example hehe)

Felienne:
> Looking at the list of reserved words, I notice that it seems to cover language keywords but no (or few) function names. Is there some specific reason for not including those, or has it just not been an issue yet?

Yeah, just not something we thought about before! I am not sure we can find a nice list of all buildin functions? And also `t`! I notices a kid use `t` as a variable yesterday in class and that does not work cause we call the Turtle `t`!

Felienne:
Addendum: if we want to do it proper, we should exclude all of these: https://docs.python.org/3/library/functions.html

JonathanMurray:
> Addendum: if we want to do it proper, we should exclude all of these: https://docs.python.org/3/library/functions.html

Makes sense to me! It looks like in some cases we also call functions from modules (like `time.sleep` in the example above)? Not sure if that's a common case or not. There are definitely a lot of modules that _could_ be referenced in that case: https://docs.python.org/3/py-modindex.html

That makes me wonder if it's feasible to track them all beforehand, or if another possibility would be to detect during transpilation which names from the python stdlib that the code will use, and only at that point convert any variables that are clashing. (I haven't looked enough at the transpiler code yet to be able to tell if this is feasible.) A more defensive/lazy approach could be to simply convert all variables in a way that makes conflicts impossible (although I really liked how readable the transpiled code was, when looking under the hood!).

Is this an area where you might be interested in outsider contributions, or are there perhaps already some related plans/efforts?

Felienne:
> That makes me wonder if it's feasible to track them all beforehand, or if another possibility would be to detect during transpilation which names from the python stdlib that the code will use, and only at that point convert any variables that are clashing. 

Great question! I would say, if we exclude the whole list as linked, we already make the code better than we have it now (a gradual approach, haha!)

> (I haven't looked enough at the transpiler code yet to be able to tell if this is feasible.) A more defensive/lazy approach could be to simply convert all variables in a way that makes conflicts impossible (although I really liked how readable the transpiled code was, when looking under the hood!).

Indeed, the reasons we do not do that is to make the Python more readable (and make the variable dropdown nicer without any effort)

> Is this an area where you might be interested in outsider contributions, or are there perhaps already some related plans/efforts?

I don't think we have related efforts, so it would be great if you can pick it up (either in this PR or in a fresh one, whatever you prefer) Are you in Discord? That is typically the place where we discuss these types of things.

Felienne (changes requested):
Let's do this proper and add all built-in functions, or at least the most common ones.

Do you want to pick this up further @JonathanMurray? Otherwise @jpelay can finish it?

JonathanMurray:
@Felienne (cc @jpelay) Hey, sorry for the slow reply. I added the list of function names from the 3.11.0 docs. I figured it might be nice for the future to see easily where different words originate from, so I decided to store them separately from the already existing list of keywords.

I agree a gradual approach seems reasonable here. Feel free to merge the PR if you think it makes sense in its current form.

JonathanMurray:
> And are you in Discord? That is typically the place where we discuss these types of things.

Yes! My username is Venom#1866. I don't normally use Discord much, but I took a peek inside the Hedy discord and noticed there are many channels. I'm guessing #developers would be the one for these sort of topics?

Felienne:
> @Felienne (cc @jpelay) Hey, sorry for the slow reply. I added the list of function names from the 3.11.0 docs. I figured it might be nice for the future to see easily where different words originate from, so I decided to store them separately from the already existing list of keywords.

Ah great idea to keep it separate!!

> I agree a gradual approach seems reasonable here. Feel free to merge the PR if you think it makes sense in its current form.

Will do!

Felienne:
> > And are you in Discord? That is typically the place where we discuss these types of things.
> 
> Yes! My username is Venom#1866. I don't normally use Discord much, but I took a peek inside the Hedy discord and noticed there are many channels. I'm guessing #developers would be the one for these sort of topics?

Ah good to know you are there too to chat! Yes #developers is the channel for discussions about code (although we are still figuring out what works best!)

Felienne (approved):
Yes this works!!

image

Felienne:
Pasting the erroneous code here is case we need it ever again:
```
jump = 0
jumpTime = 1, 2, 3, 4, 5
for i in range 0 to 100
    if jump in jumpTime
        jump = jump - 1
        print 'Its jumping'
    else
        input = ask 'Jump?'
        if input = 'w'
            jump = 5
```

Issue #3413, opened 2022-10-13
Add 'UsesPygame' Transformer 


daandj:
Add a class analogous to the 'UsesTurtle' class which inherits from Transformer and returns true if pygame specific commands are in the tree. (At this moment only the 'pressed' command.)

Also add a UsesTurtle variable in the HTTP-response that is sent to the browser. So the front-end knows whether to enable pygame.

ToniSkulj:
Closed by commit #[Add has_pygame transformer](https://github.com/Felienne/hedy/commit/6108e4bc165476567a27542f7a708d11625424d4) in branch #[Add has_pygame transformer](https://github.com/Felienne/hedy/commit/6108e4bc165476567a27542f7a708d11625424d4)

Issue #3406, opened 2022-10-12
[TESTING] Add unit tests for keymapping level 5


daandj:
❗ https://github.com/Felienne/hedy/issues/3337 has to be completed first.

Issue 3337 adds the keyword 'pressed' to the Hedy language to make key mappings possible. For example, the user could write something like the following:

```python
if pressed is 'a' print 'cool!' else print 'meh'
```

To make sure that this works for levels 5, 6 and 7, unit tests should be added that check if the correct Python code is generated.

Issue #3335, opened 2022-09-27
[UI idea] Extend Skulpt with PyGame4Skulpt


ToniSkulj:
Adding [PyGame4Skulpt](https://github.com/Dustyposa/pygame4skulpt) to Skulpt on the front-end to make it possible of running [PyGame ](https://www.pygame.org/news) relevant code. This would make it possible for the users to make an interactive program and even games. 

:exclamation: Notice that no modifications have to be made to the grammar en transpiler, yet.
:checkered_flag: It should be possible to run the following programs in Skulpt [drawing rectangle](https://www.geeksforgeeks.org/how-to-draw-rectangle-in-pygame/), [drawing circles](https://petlja.github.io/pygame4skulpt/?draw.py), [snake ](https://petlja.github.io/pygame4skulpt/?snake.py) after the completion of this issue. That way we know that at least the functionality of drawing and user input works decently.

PR #3483, opened 2022-10-25
[Language Idea] Add pressed for binding keyboard events

### LINKED ISSUE #3337: [Language idea] Add 'pressed' for binding keyboard events
❗ #3335 has to be completed first, this makes pragmatic testing a lot easier.

It would be nice if the user is able to test if a certain keyboard key has been pressed with the following Hedy syntax:
```python
if pressed is 'a' print 'cool!' else print 'meh'
```
This code has been taken from level 5 (introduction) and adapted, since if statements are necessary the functionality can only be added at level 5 and later. The following Python code would need to be generated (in total):
```python
import pygame as pg
pg.init() 
disp = pg.display.set_mode((711, 300))

while True: 
    for e in pg.event.get():
        if e.type == pg.KEYDOWN:
            if e.key == pg.K_a:
                print('cool!')
            else:
                print('meh')

    pg.display.update()
```
Notice that the generated Python code uses Pygame. 

❗ This is a simplified example, in practice this may result in problems (such as setting the pygame display size)
🏁 the above should give the desired result.

ToniSkulj:
Makes it possible to test if a certain keyboard button is pressed.

**Description**
Fixes #3337, #3413, #3406 and #3546.

- Adds the **ifpressed** grammer rule, for level 5 and upwards
- Adds the **UsesPygame** Transformer
- Uses the PyGame library - #3335 

**How to test**
Open up level 5 and use the following snippet

```python
if x is pressed print 'Ouch!' else print 'Hi!'
```
This will map the keyboard press _x_ to print 'Ouch' and any other to 'Hi!'. You can change _x_ by any Latin letter or one digit number. 
After running the code, if you press _x_ it should output 'Ouch!'. The program should stop executing after the output.

Hedy_—_Mozilla_Firefox_2022-10-25_11-50-32_AdobeExpress(1)

Next, Open up level 8 and use the following snippet
```python
if a is pressed
    print 'Ouch!'
else
    print 'Hi!'
```
This is necessary because in level 8, inline if statements are not allowed anymore. The result should be the same as in level 5.

Try out some turtle commands as well!
```python
if x is pressed forward 50 else turn 90
```
It should get the turtle moving or turning.

**:exclamation: Important Comments**

- The keyword _pressed_ is currently not highlighted. Essentially pressed is a state (for kids: "a special variable") and not a command. I think it would be cool to give these "special variables" a different color (blue maybe?).


RikvanSchaick:
As of now, I found one bug. After running and stopping a snippet with the button under the editor, I can no longer edit my code in the editor. This is most likely the result of the 'Stop program'-button not ending the while loop for pressing a key and therefor not stopping the running program.

ToniSkulj:
> As of now, I found one bug. After running and stopping a snippet with the button under the editor, I can no longer edit my code in the editor. This is most likely the result of the 'Stop program'-button not ending the while loop for pressing a key and therefor not stopping the running program.

Nice catch @RikvanSchaick! commit [Fix stop button not stopping pygame loop](https://github.com/Felienne/hedy/pull/3483/commits/fe86f62556ac4703c9f5e104f3a3f481929d5919) should fix this issue

jpelay:
Hi! Wow, this is really cool and it worked on my machine!
I have one question tho can it work with virtual keyboards? Such as the ones used in tablets. And I understand that this is a first approach, but I think it'd be best if it works with else in the future, at this point we're already teaching kids the concept of `if else` and it kind of breaks that assumption that if can be followed by else.

This also works, which is pretty cool
image

Maybe in the future we'd want to highlight that x differently, to point out that is not a variable but a key. As a first approach I think this is pretty cool, good job!

grammars/level5-Additions.lark
```diff
@@ -17,6 +17,7 @@ assign: var _IS textwithspaces
 error_print_no_quotes: _PRINT (textwithoutspaces | list_access | var_access) (_SPACE (textwithoutspaces | list_access | var_access))*  -> error_print_nq
 
 // new commands for level 5
+ifpressed: _IF SMALL_LETTER_OR_NUMERAL _IS _PRESSED _SPACE* _if_less_command
```
jpelay, line 20: You can add `_SPACE*` to the definition of  `_PRESSED` that way we don't have to use that terminal inside the production rule, keeping grammars cleaner!

jpelay (changes requested):

grammars/level5-Additions.lark
ToniSkulj: Thanks @jpelay for noticing that!! commit [Update level5-Additions.lark - Remove unnecessary space in ifpressed](https://github.com/hedyorg/hedy/pull/3483/commits/11098f2b78e325bdc9df6346684226cfc64f2968) resolves this, This was not supposed to be here, SPACE? is already in _PRESSED.

RikvanSchaick:
While playing with this feature for some time, I stumbled upon another minor bug.

If I have for example the snippet `if x is pressed print 'Ouch!'` in my editor, run the snippet and then stop running the code before pressing key `x`, I get a popup message that says for example '_Well done!_' or '_You did great!_'. 
Doing this with other programs without using this 'pressed'-feature, don't give me those popup messages, though.

ToniSkulj:
> Hi! Wow, this is really cool and it worked on my machine! I have one question tho can it work with virtual keyboards? Such as the ones used in tablets. And I understand that this is a first approach, but I think it'd be best if it works with else in the future, at this point we're already teaching kids the concept of `if else` and it kind of breaks that assumption that if can be followed by else.

Hi @jpelay, **else is also possible now**! Thank you so much for pointing this out! it completely forgot to add 'else'. I updated the branch and the initial PR comment. **Turtle commands are also possible now!**

Felienne:
It seems that the turtle output field always shows when you use `if pressed`:

image

So maybe the transpiler returns `True` for this feature?

Felienne:
And in levels above 8 using the pressed in a loop does not work yet:

image

To paste, wrong code:
```
repeat 10 times
    if x is pressed
        forward 10
```

This does work without issue:
```
if x is pressed
    forward 10
```

ToniSkulj:
> It seems that the turtle output field always shows when you use `if pressed`:
> 
> image
> 
> So maybe the transpiler returns `True` for this feature?

Thank you @Felienne  for noticing that! [Only show canvas when hasTurtle](https://github.com/hedyorg/hedy/pull/3483/commits/40ff58b0a85084b3a4dc1fb69a5490e2b9f9ca0e) fixes this.

grammars/keywords-ar.lark
```diff
@@ -40,3 +40,4 @@ _OR: _SPACE ( "ـ"* "أ" "ـ"* "و" "ـ"*  | "or") _SPACE
 _WHILE: ( "ـ"* "ب" "ـ"* "ي" "ـ"* "ن" "ـ"* "م" "ـ"* "ا" "ـ"*  | "while") _SPACE
 _LENGTH:  "ـ"* "ط" "ـ"* "و" "ـ"* "ل" "ـ"*  | "length"
 _COLOR : ( "ـ"* "ل" "ـ"* "و" "ـ"* "ن" "ـ"*  | "color") _SPACE?
+_PRESSED: ("pressed" | "pressed") _SPACE?
```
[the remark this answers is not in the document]
ToniSkulj: Thanks for the info! I took a closer look at _./content/yaml_to_lark_utils.py_ .I think I understand the process now 😀. 
[Remove non-eng (generated) keyword translations](https://github.com/hedyorg/hedy/pull/3483/commits/224609857b54bd099303bc4b3483f006dd3ec62b) & [Oops, add pressed to keywords template](https://github.com/hedyorg/hedy/pull/3483/commits/f2ac263aac87512f320b0eba69ea888549203f63) should fix this.

ToniSkulj:
> And in levels above 8 using the pressed in a loop does not work yet:
> To copy, wrong code:
> 
> ```
> repeat 10 times
>     if x is pressed
>         forward 10
> ```

Thanks @Felienne! Commit [Fix prefix+indentation issues+pygame event unbinding](https://github.com/hedyorg/hedy/pull/3483/commits/5b53fdf5b70304ddf74b716593fa62a568208935) fixes this issue.

jpelay (changes requested):

jpelay:
For some reason the tests in local are a failing. Am I doing something wrong?

image

tests/test_level/test_level_17.py
```diff
@@ -437,3 +448,147 @@ def test_not_equal_with_diff_types_gives_error(self, left, right):
...
+              except ValueError:
+                raise Exception(f'While running your program the command <span class="command-highlighted">turn</span> received the value <span class="command-highlighted">{trtl}</span> which is not allowed. Try changing the value to a number.')
+              t.right(min(600, trtl) if trtl > 0 else max(-600, trtl))
+              break
+            if event.key == pygame.K_s:
+              trtl = 180.0
+              try:
+                trtl = float(trtl)
+              except ValueError:
+                raise Exception(f'While running your program the command <span class="command-highlighted">turn</span> received the value <span class="command-highlighted">{trtl}</span> which is not allowed. Try changing the value to a number.')
```
jpelay, line 507: To transpile turtle commands you can use `Tester.py::turtle_color_command_transpiled`

jpelay (changes requested):

Felienne:
> For some reason the tests in local are a failing. Am I doing something wrong?
> 
> image

He @jpelay! That is because you did not run the `generate-grammars` script! This PR adds a new keyword and that is not present in the grammars yet, in this PR. But on GitHub the script runs before the tests so it works (If I recall correctly, you added the script on GitHub to run it haha!)

app.py
```diff
@@ -522,6 +527,7 @@ def transpile_add_stats(code, level, lang_):
         result = hedy.transpile(code, level, lang_)
         statistics.add(
             username, lambda id_: DATABASE.add_program_stats(id_, level, number_of_lines,None))
+        print(result)
```
Felienne, line 530:
```suggestion
```

jpelay:
> > For some reason the tests in local are a failing. Am I doing something wrong?
> > image
> 
> He @jpelay! That is because you did not run the `generate-grammars` script! This PR adds a new keyword and that is not present in the grammars yet, in this PR. But on GitHub the script runs before the tests so it works (If I recall correctly, you added the script on GitHub to run it haha!) I tried locally with the grammars generated and then it does work!

Oh yes, I forgot!! Yes I added that part, I forgot!

app.py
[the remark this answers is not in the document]
ToniSkulj: Good catch, my bad! [Fix some mistakes](https://github.com/hedyorg/hedy/pull/3483/commits/eb0fd502308d061a7faf75e792c7a31d32feb8f6) fixes this.

hedy.py
```diff
@@ -830,16 +832,13 @@ def text(self, meta, args):
 
 
 class UsesTurtle(Transformer):
-    # returns true if Forward or Turn are in the tree, false otherwise
     def __default__(self, args, children, meta):
-        if len(children) == 0:  # no children? you are a leaf that is not Turn or Forward, so you are no Turtle command
```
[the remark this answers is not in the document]
ToniSkulj: I did, thanks for noticing this! [Fix some mistakes](https://github.com/hedyorg/hedy/pull/3483/commits/eb0fd502308d061a7faf75e792c7a31d32feb8f6) fixes this.

tests/test_level/test_level_17.py
ToniSkulj: Thank you for point this out, commit [Make use of transpiled turtle commands for tests](https://github.com/hedyorg/hedy/pull/3483/commits/3274ab798a0e892a9289f3593e5146d629729ada) replaces all turtle commands with the appropriate functions and adding necessary indentation.

grammars/level1.lark
```diff
@@ -63,6 +63,7 @@ NAME: LETTER_OR_UNDERSCORE LETTER_OR_NUMERAL*
 
 LETTER_OR_UNDERSCORE: /[\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}_]+/
 LETTER_OR_NUMERAL: LETTER_OR_UNDERSCORE | /[\p{Mn}\p{Mc}\p{Nd}\p{Pc}·]+/
+SMALL_LETTER_OR_NUMERAL: /[a-z]/ | /[0-9]/
```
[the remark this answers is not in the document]
ToniSkulj: Good question! I have not included those because of the following reasons:

1. Pygame will register SHIFT+[the letter] when using this, this also requires the registration of this in the PyGame game loop, which will probably complicate things for now. 
2.  Technically, the letter "A" for example is a combination of SHIFT+a. It would be cool, I think, to make "and" possible with keyboard presses in the near future. This way the users can make there own shortcuts and allowing A breaks this shortcut idea.

Let me know what you think @jpelay!

Felienne (approved):
Great work team, I think we are ready to merge!

I have a small request regarding user feedback, but let's do that in a separate issue!

PR #3484, opened 2022-10-25
[FIX] Grab translation of turn before error

### LINKED ISSUE #3471: [BUG] Wrong error message for the Hebrew `turn` command on invalid arguments
## Description

The error message for the `turn` command in Hebrew (`פנה`) is different than the one used in English.
Specifically, it seems that we're getting the `Text/Parse` error string instead of `Text/Invalid Argument`.

### English

Consider the following Level 1 code:
```
turn bad
```
The error displayed is:
> You cannot use the command turn with bad. Try changing bad to right or left.

image

### Hebrew
When using Hebrew, however, we get a different result:
```
פנה bad
```
(Note that the command is the Hebrew `turn` command, `פנה`, but the input is invalid.)

And we get a different error:
image

The source string for the resulting translation is:
> The code you entered is not valid Hedy code. There is a mistake on line {location[0]}, at position {location[1]}. You typed {character_found}, but that is not allowed.

Felienne:
Fixes #3471


Felienne:
Ah @TiBiBa, this happens to be an interesting case of a change where we would want all tests to run but they don't because of the new workflow but I am pretty sure that is an exception!

Felienne:
Maybe there could be some sort of process (a github label?) we can add to force all tests to run?

Felienne:
For now I guess it would be good it @jpelay runs all tests locally to verify my work?

jpelay:
Hi, Miss Felienne! The tests work for me, but I don't know if I'm doing something wrong, but I'm testing the code in the issue and it still gives the same error, am I doing something wrong?

image

Felienne:
> Hi, Miss Felienne! The tests work for me, but I don't know if I'm doing something wrong, but I'm testing the code in the issue and it still gives the same error, am I doing something wrong?
> 
> image

Ah good thing you checked!! I forgot an argument. Could you try again?

jpelay:
> Ah good thing you checked!! I forgot an argument. Could you try again?

Works perfectly now! All tests pass too!

image

jpelay (approved):

Issue #3487, opened 2022-10-25
[CONTENT] `header.jpg` is 1.2 MB


nycdotnet:
**Describe the issue**
On the main `hedy.org` web site, the header image is 1.2 MB (`static/images/header.jpg`).  This represents a significant portion of the page weight.  This image could be trimmed to a significantly smaller size.

**Expected**
I could PR a web-prepared version in and link to that instead.  Up to you!  Thanks.

Felienne:
What a great catch @nycdotnet! We would very much welcome a PR with a smaller version!

nycdotnet:
OK.  I will do so and may dig into a few other little things.

nycdotnet:
This was fixed by #3491

PR #3491, opened 2022-10-26
Some image trimming (Issue #3487)

nycdotnet:
**Trim some images on homepage**

**Description**

This PR makes a few performance optimizations to the hedy.org home page.  The total bandwidth required on a cold cache to view https://hedy.org/ should be decreased by about 1.2 MB with these changes.

  * It adds three resized image assets.  These images were converted from the existing images using Paint.NET on Windows.  The conversions were done using high quality settings.  There is some cropping that I had to do.  The `achievement.png` was not quite square so I made it square prior to resizing.  I also cropped the `header.jpg` to 500px after resizing (Note this did cut off the bottom of the image - in the original you can see the top of the student's computer).  Each of the images are still at 2x pixels versus the scale they are drawn on the web site to account for high DPI screens.  The `header.jpg` was converted at quality setting 85.  I can't tell the difference visually when switching back and forth, but I am also not an artist/photographer.  The total KB of the three original images is 1392 KB, the new images total 190 KB - about 14 % of the original image bytes.
  * I switched over the img tags on the home page *only* to use the new scaled images.  I think it would likely be fine to do the other pages, but I don't know enough about the app and I didn't want to mess with it more haha.  If you're happy with this PR and it gets merged, it would be easy for me to do the others.
  * I added height and width html attributes to the logo that appears in the top corner to ensure that layout can occur prior to the .png downloading fully.  I got this suggestion from webpagetest.org.  Here is [their analysis](https://www.webpagetest.org/result/221026_BiDc8D_YH/) for hedy.org.
  * I added a lazy loading attribute on the achievement image.  This image is downloaded even though it is not displayed for (I assume) the majority of users hitting the home page (it's hidden by CSS) so I feel like there's no downside.

**Fixes 3487**

**How to test**
Run the web server and hit the home page.  In your browser dev tools on the network tab, you should see the new images downloading and they should be smaller.  There should be no human-visible difference to the web site other than it loads using less bandwidth.  (Again this is subjective and if I've over-compressed the images or screwed something else up, please let me know and I can address it).

**Checklist**
  

Felienne:
And congrats on your first Hedy contribution!!! 🥳

nycdotnet:
Thank you!!

nycdotnet:
@Felienne the reason I didn't was because I wanted to ensure the original high res version was there in case it was needed for something else.  I also did not remove the original high res version of the Hedy logo or achievement image.

PR #3494, opened 2022-10-26
[BUG FIX] Incorrect inputs correctly highlighted again multiple accounts creation

### LINKED ISSUE #3473: [BUG] Creating students fails if 1 already exists
**Describe the bug**

As reported on Discord: 
_"I'm creating a batch of student accounts all at once, entered some 30 names. Only to get an error saying "One or more account already exists"..._

_But it doesn't tell me which ones 😭"_

A quick fix of course is to highlight the one that exists in red and I think we at one point even had that working but it was somehow removed?

TiBiBa:
**Description**
We make a but of an ugly work-around that requires some explanation. For styling of input attributes we have some global classes set. In this case the `input:required` and `input:required:valid` classes are relevant. When an input is required but not valid (yet), the `input:required` styling is set: setting a red border on any input element. When the input is `:valid` we overwrite this styling with a green border using `!important`.

This is what is causing the issue. On the moment of receiving the faulty feedback from our POST request we style each input element matching the input with the class `border-red-500`. But as the element is already styling with a green border through the `!important` suffix the overwrite doesn't work! In this PR we make a bit of dirty workaround where we remove the `required` attribute from the inputs at the moment of making the POST request. This doesn't have any functional implications as all verification is handled by the server. However, the border will no longer turn green after the first attempt.

**Fixes**
This PR fixes #3473.

**How to test**
Validate that the invalid field border turns red when an error is returned on multiple accounts creation.


Felienne (approved):

Felienne:
The red border is back and my extra fix also revives the reset button :)

PR #3495, opened 2022-10-26
[BUG FIX] Admin is able to set user tags again

TiBiBa:
**Description**
Fix an issue where Dynamo would be mad when sending a dictionary that also contained the index key to the database for altering an entry. Locally this was not an issue so this was hard to detect.

**Fixes**
No related issue.

**How to test**
Deploy and verify that you, as an admin, can set user tags again.


Felienne (approved):

website/admin.py
```diff
@@ -246,8 +246,11 @@ def update_user_tags(self, user):
         if body.get("contributor"):
             tags.append("contributor")
 
-        user['tags'] = tags
-        self.db.update_public_profile(user['username'], user)
+        # We have to pop the username otherwise Dynamo gets mad -> we can't update a value that is also the index key
+        db_user.pop('username')
+        db_user['tags'] = tags
+
+        self.db.update_public_profile(user['username'], db_user)
```
Felienne, line 253: @TiBiBa I guess this user should be body?

PR #3496, opened 2022-10-26
[BUG FIX] Another fix to really fix the tags setting by the admin

TiBiBa:
Fix issue in tags setting by admin


Felienne (approved):
Works locally, let's see on the server!

PR #3497, opened 2022-10-26
Normalize paths and replace instead of rename to make atomic files work in Windows.

Mark-Giesen:
Paths are now normalized by os. Some forward/backward slashes sometimes cause errors in finding files on windows. 
os.rename in utils changed to os.replace to make sure it works on Windows aswell (atomic when succesfull)


Felienne (approved):
Ah thanks @Mark-Giesen!! This does not seem to break anything and (I assume!) enables Windows users test locally again.

PR #3503, opened 2022-10-27
[IMPROVEMENT] Changes highest level reached from highest saved program to highest level quiz score

### LINKED ISSUE #3501: [UI idea] Update "Highest level reached" to represent Quiz Completion
**Update "Highest level reached" to represent completed quizzes in the class interface**
In the Class interface, when displaying the list of students, there is a column called "Highest level reached". Based on a discussion on Discord, this value is an artifact of an earlier time, and actually reflects the highest level of a student's saved programs. I would suggest that this value be changed so that it instead shows **the level of the highest level quiz completed** by the students. This would be a helpful change for classes using a score threshold to unlock subsequent levels. I would also suggest that the column in the class interface be renamed to "Highest level completed", for a more accurate name.

For example, the value should start at 0 for new accounts, as it does now. When Quiz Level 1 is completed, the value should change to 1. Completing the Quiz for Level 2 should change the value to 2, and so on.

**Additional context**
Alternatively, the value could start at 1, showing that students are working on Level 1, and then change to 2 after they complete the Level 1 Quiz. This would better match the current name. I've been having my students complete one or two levels each class, with the Quiz being the culminating task for the day. So, for my class, I care more about what levels are _complete_, rather than what students are presently working on, but other teachers may be using Hedy differently. If anyone else has different thoughts about how this should be handled, please share your perspective.

TiBiBa:
**Description**
As pointed out by @JimJamShazam our current implementation of "Highest level reached" is a bit outdated and still depends on relic code from when we didn't do any user logging (and didn't have any more stats than "saved programs"). In PR we improve this by iterating the user logs and getting the highest level for which a quiz is finished.

Note: We don't look into the actual quiz score, only that it is finished.

**Fixes**
This PR fixes #3501.

**How to test**
Make sure you have a class with students in it. Verify that the highest level reached column now depends on the highest level for which a quiz is finished instead of the highest level a program is saved in.

### DIFF
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -3,7 +3,8 @@ from flask_babel import gettext
 import hedy
 import hedyweb
 from .achievements import Achievements
-from website.auth import requires_login, is_teacher, is_admin, current_user, validate_student_signup_data, store_new_student_account
+from website.auth import requires_login, is_teacher, is_admin, current_user, validate_student_signup_data, \
+    store_new_student_account
 import utils
 import uuid
 from flask import g, request, jsonify, session
@@ -12,6 +13,7 @@ import os
 import hedy_content
 from .database import Database
 from .website_module import WebsiteModule, route
+
 
 class ForTeachersModule(WebsiteModule):
     def __init__(self, db: Database, achievements: Achievements):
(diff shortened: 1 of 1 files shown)


jpelay:
Hi! Timon! This works perfectly!

After signing up the students:
image

After running some quizzes:
image

PR #3505, opened 2022-10-27
Fix language handling when loading adventure

babaMar:
**Description**

_The request handler for single adventure was not handling the `keyword_language` parameter. In turn the language to be passed to `load_adventures_per_level` function was always the language in app context `g`. At the same time, in `load_adventures_per_level` function the language argument was not to use as key for `ADVENTURES`.

Fixes for code style._

**Fixes https://github.com/Felienne/hedy/issues/3485 **

It probably has some impact on https://github.com/Felienne/hedy/issues/3433 as well. In the top-right corner I don't see the language switcher though, only my username, so not able to test that.

**How to test**
Run locally, and browse to single adventure. Use the translation button to switch language

### DIFF
--- a/app.py
+++ b/app.py
@@ -94,7 +94,7 @@ def load_adventures_per_level(level, keyword_lang):
                 loaded_programs[program_key] = program
 
     all_adventures = []
-    adventures = ADVENTURES[g.lang].get_adventures(keyword_lang)
+    adventures = ADVENTURES[keyword_lang].get_adventures(keyword_lang)
 
     for short_name, adventure in adventures.items():
         if level not in adventure['levels']:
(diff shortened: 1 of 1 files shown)


jpelay:
Hello! Thank you for resubmitting your PR, this one is a lot cleaner than the last one! We still are figuring out the fixes related to the styling and how's best to enforce it. The logger is a good idea, perhaps you want to talk about it in the Discord where we'll all discuss it and then open and issue and a PR!

jpelay (approved):

app.py
```diff
@@ -94,7 +94,7 @@ def load_adventures_per_level(level, keyword_lang):
                 loaded_programs[program_key] = program
 
     all_adventures = []
-    adventures = ADVENTURES[g.lang].get_adventures(keyword_lang)
+    adventures = ADVENTURES[keyword_lang].get_adventures(keyword_lang)
```
TiBiBa, line 97: This is an incorrect fix and should not have been merged. When retrieving the adventures we want to get the adventures of the user language, not the user keyword language.

We do however want to parse all commands with the keyword language. But this already happens correctly.

So: What is the purpose of this change?

TiBiBa (commented):
Please reverse this PR.

TiBiBa:
I don't understand why this PR is merged. Did it fix the related issue? The issue is the not correct showing of the switcher when in a specific adventure mode. 

If I understand correct this code change introduce a new issue by retrieving the incorrect adventure file.

PR #3509, opened 2022-10-28
Refactor Logging

babaMar:
**Description**

- centralized loggers.py module where module loggers are defined and imported from
- consistent use of logging throughout the application via module loggers (removed print statements in `hedy_content.py`)
- logging essential (debug) information if `NO_DEBUG_MODE` is **_not_** set, set the logging level to INFO otherwise

### DIFF
--- a/logger.py
+++ b/logger.py
@@ -1,0 +1,11 @@
+import logging
+
+LOG_FORMAT = '[%(asctime)s] %(filename)s %(funcName)s %(lineno)d <%(levelname)s>: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
+
+app_logger = logging.getLogger('app')
+log_queue_logger = logging.getLogger('querylog')
+jsonbin_logger = logging.getLogger('jsonbin')
+
+for _logger in (app_logger, log_queue_logger, jsonbin_logger):
+    _logger.setLevel(logging.DEBUG)
--- a/logger.py
+++ b/logger.py
@@ -1,11 +1,14 @@
+import os
 import logging
 
+LOG_LEVEL = logging.DEBUG if not os.getenv('NO_DEBUG_MODE') else logging.INFO
 LOG_FORMAT = '[%(asctime)s] %(filename)s %(funcName)s %(lineno)d <%(levelname)s>: %(message)s'
 logging.basicConfig(format=LOG_FORMAT)
 
 app_logger = logging.getLogger('app')
 log_queue_logger = logging.getLogger('querylog')
 jsonbin_logger = logging.getLogger('jsonbin')
+hedy_content_logger = logging.getLogger('hedy_content')
 
 for _logger in (app_logger, log_queue_logger, jsonbin_logger):
-    _logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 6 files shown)


logger.py
```diff
@@ -0,0 +1,14 @@
+import os
+import logging
+
+LOG_LEVEL = logging.DEBUG if not os.getenv('NO_DEBUG_MODE') else logging.INFO
+LOG_FORMAT = '[%(asctime)s] %(filename)s %(funcName)s %(lineno)d <%(levelname)s>: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
```
rix0rrr, line 6: I would feel more comfortable if configuring this and setting the log levels would be done in a method call that is explicitly invoked at application startup, rather than during module loading.

logger.py
```diff
@@ -0,0 +1,14 @@
...
+
+app_logger = logging.getLogger('app')
+log_queue_logger = logging.getLogger('querylog')
+jsonbin_logger = logging.getLogger('jsonbin')
+hedy_content_logger = logging.getLogger('hedy_content')
+
+for _logger in (app_logger, log_queue_logger, jsonbin_logger):
+    _logger.setLevel(LOG_LEVEL)
```
rix0rrr, line 14: I believe that you can also configure a level as part of calling `logging.basicConfig()`. Does that remove the need for configuring the log level on each individual logger?

------

And if that removes the need for configuring individual loggers, then do we still need to centralize all loggers here in this one file? I was under the impression it's pretty common to initialize a logger in every file, using the current file name.

I think you're even supposed to do this:

```py
logger = logging.getLogger(__name__)
```

In every file. At least, that's what [this page](https://docs.python.org/3/library/logging.html#logger-objects) tells me. 🤔

rix0rrr (commented):
Hey! Thanks for consolidating the `print` statements to log lines, that's very useful.

The only thing I'm not sure about is centralizing the loggers into a single file. I believe the old style was more or less straight from the Python docs as recommended style, and I find it hard to judge the benefits of one vs the other

Felienne (dismissed):
I think @rix0rrr is right about this proposal not being exactly what the intention of the loggers is. Is there a good reason to do it this way that we are overlooking here @babaMar?

logger.py
babaMar: This was my initial approach, but surprisingly it didn't produce the expected results. I had to add the explicit `setLevel` call at the logger level to see the logs being issued. There seem to be a side effect when having `import logging` in `website/log_queue.py`, the configuration gets overwritten.

babaMar:
> I think @rix0rrr is right about this proposal not being exactly what the intention of the loggers is. Is there a good reason to do it this way that we are overlooking here @babaMar?

I reworked the changes based on @rix0rrr comments. Initially the reason to do it that way was to have it working, I stumbled on some side effects and ended up with the previous implementation.

logging_config.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+
+LOGGING_CONFIG = {
```
rix0rrr, line 4: Noice!

rix0rrr (approved):

PR #3514, opened 2022-10-28
Updates to level 9 calculator example (bn,de,en,fr,hi,it) - issue 3489

LeslieMurphy:
**Description**

Enhance the sample calculator to provide pass/fail on each response.  Provide the correct answer if the response is wrong.  Make the number of questions configurable via a variable.

**NOTE**: I changed several languages, but _not all of them_.  For some reason a lot of the languages have all the code on a single source code line using newline characters.

If it is safe to edit all the other languages and make this code match en I can do so.  I'm not quite sure how to test the other languages to be honest.

Also, I noted a minor difference in the original code regarding syntax highlighting.  I made the 6 languages bn,de,en,fr,hi,it all consistent.

For example en and fr had `{repeat} 10 {times}` yet de had `repeat 10 times`

**NOTE2** - the Hindi language version within my browser shows unicode characters in the program which I don't understand.  The program still runs fine.  Example:
दोहराना question_count बार

**Fixes** [3489](https://github.com/Felienne/hedy/issues/3489)

**How to test**

View level 9 in user interface, then run sample code.  Test with both correct and wrong answers.  Test in each language.

**Checklist**
  
- [X ] Describes changes in the format above (present tense)

### DIFF
--- a/content/adventures/bn.yaml
+++ b/content/adventures/bn.yaml
@@ -1383,17 +1383,20 @@ adventures:
                 example_code: |
                     ```
                     score = 0
-                    {repeat} 10 {times}
-                        numbers = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
+                    question_count = 10
+                    numbers = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
+                    {repeat} question_count {times}
                         number1 = numbers {at} {random}
                         number2 = numbers {at} {random}
                         correct_answer = number1 * number2
                         {print} 'What is ' number1 ' times ' number2 '?'
                         answer = {ask} 'Type your answer here...'
-                        {print} 'Your answer is' answer
                         {if} answer {is} correct_answer
                             score = score + 1
-                    {print} 'Great job! Your score is... ' score ' out of 10!'
+                            {print} 'Your answer ' answer ' is correct!'
+                        else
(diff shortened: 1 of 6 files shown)


Felienne:
Thaks for the suggestion @LeslieMurphy! I am not sure we want this separate change, we are working on more holistic content developments with increasing complexity at the moment, we might favor this change in too in a while but not now.

Felienne:
To answer some questions for future reference:

>  NOTE: I changed several languages, but not all of them. For some reason a lot of the languages have all the code on a single source code line using newline characters.

That is Weblate flattening the strings

> If it is safe to edit all the other languages and make this code match en I can do so. I'm not quite sure how to test the other languages to be honest.

They will all be tested automatically on each PR!

> Also, I noted a minor difference in the original code regarding syntax highlighting. I made the 6 languages bn,de,en,fr,hi,it all consistent. For example en and fr had {repeat} 10 {times} yet de had repeat 10 times

Good catch!! We recently allowed translation of keywords and not all translators have picked this up yet. Maybe I should make a separate tissue for it!

> the Hindi language version within my browser shows unicode characters in the program which I don't understand. The program still runs fine. Example:
दोहराना question_count बार

Is this a browser issue? For me it looks like this with that line of code pasted in:
image

PR #3526, opened 2022-10-29
[FIX] Make Mergify regular-merge Weblate PRs

rix0rrr:
With the old config, all PRs are squash-merged, which is great for history but confuses Weblate.

Except Weblate PRs to make them be regular-merged instead.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,5 @@
 pull_request_rules:
-  - name: Automatic squash merge on approval
+  - name: Automatic squash merge on approval for non-Weblate PRs
     conditions:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -1,15 +1,29 @@
 pull_request_rules:
-  - name: Automatic merge on approval
+  - name: Automatic squash merge on approval
```
Felienne, line 2:
```suggestion
  - name: Automatic squash merge on approval for non-Weblate PRs
```

Felienne (approved):

Issue #3531, opened 2022-10-31
[TESTING] Create front-end tests for log-in features


JimmyOei:
**Description**
Create front-end tests in Cypress for all the features in the `login` page.
- ‘Log in’ button
- ‘Create account’ button
- ‘Forgot your password?’ button

Felienne:
Thanks for the issue but maybe we can split it up into a few more issues (Testing all teacher features, all logging features etc.) That will be useful to track progress and to see what we still have to do when you all leave.

PR #3532, opened 2022-10-31
Login page and recover page Cypress tests

JimmyOei:
**Description**
This PR consists of the first Cypress tests for the front-end testing, namely tests for the login page and the recover page. Also including some tools that helps with commonly used implementations and site navigations for all the Cypress tests.

_Changes in detail_
- **Adds** a directory for the login page tests, consisting of tests for all the buttons, except the login button, and fields on this page and the modal alert popup when trying logging in.
- **Adds** a directory for the recover page tests, consisting of tests for the button and field on this page. Eccluding the test for the modal alert popup when trying to send the recovery link.
- **Changes** to the cypress config file for the environment variables of pages and the base url.
- **Changes** to the tsconfig.json file. Here I added the baseUrl, because this gave an error when running the cypress scripts on the command line, saying that the baseUrl was missing.

**Running the tests**
Run the tests on the command line with the following: `npx cypress run --spec "[path to test(s)]"`
An example of running cypress: while being in the hedy/tests directory execute the following:  `npx cypress run --spec "cypress/e2e/login_page/*"`
This runs all the login_page tests.

Works on the issue #3531.


Felienne:
ok I managed to run the tests on the command line: 

image

Some are failing though and I am not sure that is expected? If you add instructions on how we should run the tests so we can verify it works as proposed that would be great!

JimmyOei:
I removed the two tests that didn't work and will further work on this in a new branch. The problem is that it involves loggin in a user that should be registered, but the user registered locally on one of our laptops is, ofcourse, not registered in the official hedy database or any other machine you are running the test on. Current tests should work and I added a description on how to run the tests.

PR #3540, opened 2022-11-01
[UI] Allows ordering of adventures per level

### LINKED ISSUE #3535: [UI idea] Allow reordering of adventures per level
This is something I have been wanting to do for a while but it is annoyingly cumbersome so I have not yet done it, but it must be done :) 

We now order the adventures themselves, here:
https://github.com/hedyorg/hedy/blob/4ce87a3d8298e4f927cf3f6e867bf2e225580ec8/hedy_content.py#L41

But we would like to be able to reorder per level, like so:

```
ADVENTURE_ORDER=
{1: 'default',
    'parrot',
    'fortune',
    'haunted',
    'restaurant';
 2: 'default',
     'fortune',
     'haunted',
     'parrot',
     'restaurant'...}
```

This is not hard per see, but a lot of work because of the caching code. 

But it would allow us to work more on #3259 and also help the ideas outlined in #3533 (specifically point 2)

jpelay:
**Description**

Introduces a more granular way to order adventures in each one of the levels. This is still hard coded in `hedy_content.py`, in the future we want this to be retrieved from the database so every teacher can customize this.

For now, to change the ordering of the adventures you can modify them in `ADVENTURE_ORDER_PER_LEVEL`
If someone adds a new adventure in a level, they should add the adventure to this dictionary too.

**Fixes #3535**

**How to test**

I left the ordering the same as it already is, to change it and see if the test really works you can change the position of one of the adventures in one of the levels and then see if it's changed in the site.

**Checklist**
  

Felienne (approved):
Cool thanks, this works and will help us make the adventures even better!

image

Felienne:
Wil greatly ease further work on https://github.com/hedyorg/hedy/issues/3259 (cc @MarleenGilsing)

PR #3542, opened 2022-11-02
[CHORE] Update PULL_REQUEST_TEMPLATE.md

Felienne:
Suggests new template including categories


jpelay:
We can also add UI for new user interface changes! Maybe one that says TESTING, when we're only updating tests? And at last, what about changes to the logic, but that don't affect the interface BACKEND?

Felienne:
> We can also add UI for new user interface changes! Maybe one that says TESTING, when we're only updating tests? And at last, what about changes to the logic, but that don't affect the interface BACKEND?

Good idea! Feel free to update the template!

jpelay:
@TiBiBa You sometimes use [UI IMPROVEMENT] and [IMPROVEMENT] would you like us to include those categories, how do you feel about the ones we're including?

PR #3545, opened 2022-11-02
[BUG FIX] Fixes single adventure language switcher

### LINKED ISSUE #3485: [BUG] Language switcher does not work in single adventure
If we use a single adventure (f.e. https://hedy.org/adventure/parrot/1) the language switcher does not seem to show well:

image

Also, using the link directly does not seem to work to start with the language in the link:

https://hedy.org/adventure/parrot/1?keyword_language=nl

(Maybe these are two separate issues, feel free to split out)

TiBiBa:
**Description**
In this PR we fix the keyword language switcher on the specific adventure mode. As we didn't have the `commands` list yet on this route the switcher was empty. We also add the correct adventure GET dependent on the keyword language to make sure this changes as well when switching.

**Fixes**
This PR fixes #3485.

**How to test**
Navigate to any specific adventure such, such as `/adventure/parrot/1`. Verify that the language switcher works and the adventure is also correctly parsed in the chosen keyword language.


Felienne (approved):
Thanks!!

image

PR #3552, opened 2022-11-03
[CONTENT] Quiz level 12 and 13

MarleenGilsing:
Quiz level 12 and 13


Felienne (approved):
Thanks @MarleenGilsing!

Note that if we have a code block in a question, the ``` are not needed (they will then be printed as part of the code which is not what we want)

I removed them all, so for now all good this is just a tip for next time!

PR #3571, opened 2022-11-03
[CHORE] Add database with data for local testing

### LINKED ISSUE #3448: [CONTENT] Create a filled, local, test database
If you want to test more complex features (f.e. the teacher features) you will need to manually create accounts. It would be nice it we create a version of the test database that we can share in the repo to get started. 

This will entail:
* Making the data once
* Updating the docs to explain local data
* Making sure the file exists in the repo but is not used online (Maybe this would already work if we just stop ignoring `dev_database.json`, I am not sure)

jpelay:
**Description**

Adds a database with data to feed the local database when needed. This file does not replace the local database, instead I added a shell script that copies the content from one to the other, avoiding merge conflicts and so. Also updates the CONTRIBUTING.md file to reflect the contents of this database.

**Fixes #3448**


Felienne (approved):
Ah thanks this will make it even easier to people to run Hedy locally!

PR #3576, opened 2022-11-04
[LOCALIZATION] Implements localized 'search' string for ace search box

### LINKED ISSUE #3536: [LOCALIZATION] Localize Ace search box
The way we place "search..." in the Ace search editor now means that it cannot be localized. Would be nice if that is also possible.

_Originally posted by @Felienne in https://github.com/hedyorg/hedy/pull/3504#discussion_r1010375528_

TiBiBa:
**Description**
In this PR we make a bit of an ugly work-around to localize the 'search' string in the ace search box. As we can only inject Jinja variables inside our HTML templates we create an empty div that contains our string. Then we replace the placeholder of our search box with the innerHTML of this div: making it localized. However, there is also another solution and I don't really have a preference. Two options:

- Create hidden div with localized string, replace placeholder in `ext-searchbox.js`
- Create minor script in `incl-editor-and-output.html` and overwrite placeholder

Note that in this PR we fix this issue with the first workaround. However, this does leave a hidden div in our template.

**Fixes**
This PR fixes #3536.


jpelay:
Hi! It works for me, and albeit is a bit of a hack, it's not bad. The empty div is understandable if one look at it and see its `id`.

image

Felienne (approved):
Yes, works, great!

image

And I agree with @jpelay a comment for such "magic" solutions is a good idea, added one!

PR #3586, opened 2022-11-05
[GRAMMAR] Allow empty comments

Donaim:
This small grammar change makes Hedy ignore also those comments that have zero length.
Such comments are allowed in Python, but yield errors in Hedy, until this patch is applied.


Felienne (approved):
Ah thanks that is a nice addition!!

Felienne:
And I think this is your first PR in Hedy (apart from translations?)?

PR #3598, opened 2022-11-08
[FIX, WIP] Trying to get Babel to properly understand Punjabi, take 2

Felienne:
I added a separate Weblate language now and dynamically load between options. Still not sure whether the bash script will work like this with Weblate but let's first try if the rest works.


Felienne:
Closed in favor of #3599

PR #3410, opened 2022-10-12
Introduce automated checks for PEP8-compliance using flake8

mennoliefstingh:
**Description**

This pull request serves as a first step towards making Hedy a PEP8-compliant project by introducing automatic linting of code using flake8. Flake8 is configured in setup.cfg, with a set max line length of 100 and an exclusion of rule F541 (f-strings without placeholders) for the tests folder. 

Two checks are added to the repository. First of all, it introduces a Git pre-commit hook that developers can install to automatically lint modified code on each commit. Secondly, it introduces a Github Actions workflow that lints the code on pushes and pull requests. It is possible to configure this Workflow such that failing the linting denies the pull request, but as that is outside the scope of issue 3377 I will make a new issue about making the entire codebase PEP8-compliant. 

Finally, CONTRIBUTING.md is updated with instructions for using the Git pre-commit hook, instructions for using autopep8 to automatically fix some issues and a friendly request to make sure all new or modified code is PEP8-compliant. 

**Fixes issue**
#3377 

**How to test**

To test the Git pre-commit hook, follow these commands:
``` 
pip install pre-commit
pre-commit install  # installs the pre-commit configuration from pre-commit-config.yml
```
After this, stage a commit with Python code and your commit will be prevented if your modifications fail the linting. You can force a commit by adding a `--no-verify` flag to your git commit. 

To test the Github action, simply push to the repository and visit the 'Actions' tab in Github: you should see the results of the linting in the workflow.

PR #3602, opened 2022-11-08
[CLEANUP] Make Hedy PEP8 compliant

jpelay:
**Description**

Since #3410 we introduced a git pre-commit hook to lint modified code. However, much of the Hedy codebase is still not PEP8 compliant, in this PR we'll change that by applying autopep8 to the whole codebase with the following parameters:

```
autopep8 --in-place --max-line-length 120  -aaa --jobs 0 [files]
```
A few changes have been made to the pre-commit hook:
* The max line length now is 120
* I excluded the max line check from utils.py  because the number dictionary is too long.

**Checklist**
  

jpelay:
I'm already done with all of the files!

mennoliefstingh:
Great work! I did a scan of a few files and didn't see any weird things. Tests pass too so I'm inclined to approve this PR, but would like to get @TiBiBa in on it if possible first.

TiBiBa (changes requested):
All looks good to me! One minor related change: can we add `pre-commit` to `requirements.txt`?

jpelay:
> All looks good to me! One minor related change: can we add `pre-commit` to `requirements.txt`? On the long run I would also like to add some bash script that handles all the setup for the contributor as it is getting a bit complex for a new contributor to get started with Hedy. But, this seems best for a separate issue to solve.

I agree with you! I did add it to requirements.txt, however the user still has to manually run: `pre-commit install` due to the various reason listed by pre-commit author here: https://stackoverflow.com/questions/71086922/install-pre-commit-on-package-install-in-setup-py-file

TiBiBa:
> > All looks good to me! One minor related change: can we add `pre-commit` to `requirements.txt`? On the long run I would also like to add some bash script that handles all the setup for the contributor as it is getting a bit complex for a new contributor to get started with Hedy. But, this seems best for a separate issue to solve.
> 
> I agree with you! I did add it to requirements.txt, however the user still has to manually run: `pre-commit install` due to the various reason listed by pre-commit author here: https://stackoverflow.com/questions/71086922/install-pre-commit-on-package-install-in-setup-py-file

I see! But we would still be able to run this using a bash script, I can look into this.

TiBiBa (approved):
This is fine by me once the merge conflicts are fixed!

PR #3599, opened 2022-11-08
Fixes support for pa_Arab_pk

TiBiBa:
**Description**

_Changes in detail_

For example: "**Adds** translations of levels 1 to 12 to Polish" (note the use of present tense!)

**Fixes _issue or discussion number_**

Always link the number of the issue or of the discussion that your pr concerns.

**How to test**

* If this is a UI change: _describe how to run your code to see it in action. See this https://github.com/Felienne/hedy/pull/880#issue-1016304308 for an example_
* If this is a language change: _add a few tests showing the difference_

**Checklist**
Done? Check if you have it all in place using this list:*
  
- [ ] Describes changes in the format above (present tense)
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #3603, opened 2022-11-08
[FIX] Manually adds support for Tswana

### LINKED ISSUE #3600: Adds support for Tswana in Babel
**Description**
Following the implementation of the work-around for `pa_PK` in #3599 we can use the same logic to (finally) implement support for Tswana. Should be to hard but it would also be nice to generalize the implementation a bit so we can easily add languages later on.

TiBiBa:
**Description**
In this PR we fix the support for Tswana that is not natively supported by Babel. We use the same work-around as we used in #3599 for adding support for `pa_PK`. Simply re-direct Babel to another language when the non-supported language is called but **do not** overwrite the locale parser -> keep this to the original "unsupported" language.

**Fixes**
This PR fixes #3600.

**How to test**
Verify that Tswana is listed in the language dropdown selector. Also verify that the translations are wrapped correctly by both changing some stuff in the `tn.yaml` files in `/content` and the .po files in `/translations`. In the second case don't forget to compile the translations.

### DIFF
--- a/app.py
+++ b/app.py
@@ -13,7 +13,7 @@ from website.auth import current_user, login_user_from_token_cookie, requires_lo
 from website.yaml_file import YamlFile
 from website import querylog, aws_helpers, jsonbin, translating, ab_proxying, cdn, database, achievements
 import hedy_translation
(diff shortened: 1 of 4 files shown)


Felienne (approved):
Works!!

image

PR #3604, opened 2022-11-08
[UI IMPROVEMENT] Improves language dropdown UI

### LINKED ISSUE #3601: [UI idea] Improve language dropdown selector
**Idea incl level**
Hedy is getting a bit too popular in terms of supported languages and the current language dropdown doesn't handle the large amount very well. We should re-think this part of the UI to make sure it looks nice again:
Schermafbeelding 2022-11-08 om 17 59 26

TiBiBa:
**Description**
We improve the language dropdown UI by splitting the column into two columns, making sure we don't need to scroll to select our relevant language:
Schermafbeelding 2022-11-08 om 20 04 42

**Fixes**
This PR fixes #3601

**How to test**
Verify that the new UI is in place and no languages are missing.

### DIFF
--- a/templates/incl-menubar.html
+++ b/templates/incl-menubar.html
@@ -59,9 +59,18 @@
                       <div class="font-medium border-b-2 border-gray-700 mx-4 pb-2 text-center">
                         {{_('keyword_support')}}
                       </div>
-                      {% for lang in other_keyword_languages() %}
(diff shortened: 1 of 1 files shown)


jpelay:
Works perfectly!

jpelay (approved):

PR #3618, opened 2022-11-10
[UI] Implements button to hide / show editor and expand the output panel

### LINKED ISSUE #3608: [UI idea]Make editor "foldable"
**Idea incl level**
Kids really want to have a larger output field, esp. when using the turtle. Could we add a button to fold the code "in", like the files panel in repl.it:

Folded out:

image

Folded in:

image

Since this is just a small change is ts, it could be a good first issue!

This could also replace the blur button that is also used to "hide" code.

TiBiBa:
**Description**
In this PR we implement a button as an overlay on the code window to enable the user to hide/show the editor and thereby expand the output panel. This makes the blur option redundant and enables the user to get more space for their output.

**Fixes**
This PR fixes #3608

**How to test**
Notice the new button as an overlay on the code editor. Verify that it can be used to hide/show the code editor and thereby shrink / enlarge the output window. Also verify that the button isn't shown in a screen smaller than Tailwind `lg`.


Felienne:
Ah this is so coo @TiBiBa!!! Kids will love this.

Just one small thing I noticed, if you fold the code in and out with the turtle, in a submitted program (only in that scenario!) the turtle hops into the grey borders:

image

TiBiBa:
> Ah this is so coo @TiBiBa!!! Kids will love this.
> 
> Just one small thing I noticed, if you fold the code in and out with the turtle, in a submitted program (only in that scenario!) the turtle hops into the grey borders:
> 
> image

Whoops, that is a very specific bug. Will look into this if I'm able to fix this, otherwise might be acceptable for merge? Once we complete re-code the template this should be handled anyway.

TiBiBa:
I'm unable to re-produce this bug:
Schermafbeelding 2022-11-11 om 12 59 58

Felienne:
> I'm unable to re-produce this bug: Schermafbeelding 2022-11-11 om 12 59 58

Neiether am I:

Felienne (approved):

PR #3622, opened 2022-11-10
[TESTING, CYPRESS] Tests Teacher page

### LINKED ISSUE #3560: [TESTING] Create front-end tests for the teacher features in the for-teachers page
**Description**
Create front-end tests in Cypress for all the features in the `for-teachers` page.

*My Classes:*
- Create a new class button
- Duplicate class button
- Remove class button
- View Class button

*My adventures:*
- Create adventure button
- View adventure button
- edit adventure button
- remove adventure button

martinj2001:
**Description**

The first tests for the teacher page that test creating classes, duplicating classes and logging in as a teacher.

Fixes #3560 

**How to test**

Run 
`export ADMIN_USER=admin_user; python app.py` in your virtual environment. This ensures that there is an active admin_user and Hedy is running.

Then you can go to the /tests folder and run: 

`npx cypress run --spec cypress/e2e/teacher_page`

All specs should have passed!


Felienne (changes requested):
Great work @martinj2001! 

This style of testing is really good where you verify options are enabled for example. Maybe you can help to improve the other Cypress PRs a bit along the same lines (like #3626 and #3627)?

However I did find a (small) issue in running the tests, I do not think the script in `tools/login/login.js` creates the user teacher_user. The tests as they are now failed for me locally, and they only started to work when I manually added that user. Can you check the login scripts?

Felienne:
> Can you maybe supply the error? The reason why I would not work properly, is maybe because the ADMIN_USER is not set to admin_user? That's the only thing I can think of right now

Ah yes you are right, I could have been more clear! 

You can reproduce the issue by:

1. Removing teacher_user from the local database of users
2. Running the tests. They will then fail because the user cannot login (since it does not exist)

Hope that helps, if not I can also provide screenshots later (am now in a train).

martinj2001:
Ah, now I understand! We have all used our local environment for a while now, and we all once executed the function: setupDatabase(). This changes the dev_database.json file to have all accounts ready that we use. How do you suggest we use this in the login functions? Because you don't want to write the database file every time right?

Felienne:
> Ah, now I understand! We have all used our local environment for a while now, and we all once executed the function: setupDatabase(). This changes the dev_database.json file to have all accounts ready that we use. 

I suspected something like that yes! Where does this `setupDatabase()` live, in your own local Python code?

> How do you suggest we use this in the login functions? Because you don't want to write the database file every time right?

I would be totally ok with you running the `feed_dev_database.sh` file (or something similar) before the Cypress tests to ensure the DB is what you expect. That is proper test setup as far as I am concerned. But maybe you can do it with Cypress too (this sounds like a problem other apps might have too).

But I am very ok with merging this PR and opening a separate issue for the DB setup (if we can avoid it I'd rather not have PRs live for more than a few days)

martinj2001:
The function now lives inside the Cypress tests tooling, we can use it from
there. The only problem is: once you write the database file, you have to
restart flask... I am going to look into it, if you could merge this one, I
will open a new PR when I have a fix for the DB issues!

Felienne (approved):
Let's merge this one for now, I made a separate issue (#3691) for the db startup

PR #3623, opened 2022-11-10
[TESTING] Expands on partly testing to include hedy.app and /tests folder

### LINKED ISSUE #3607: [CHORE, TESTING] Run tests with label and with changes to hedy.py
In #3461 we optimize the test running, which is nice so you do not have to wait for all grammar tests in a front-end PR, but changes to hedy.py do not trigger a run (causing issues in #3587 cause I thought it was complete but it was not).

I propose 2 things:
* Also run the tests if hedy.py changes
* Make a label like `force-tests` that we can put on a PR so all tests are ran. We can do that with mergify I think.

TiBiBa:
**Description**
In this PR we further improve the testing flow to make sure GitHub performs some smart testing to decrease the time before a PR is validated. We add `hedy.app` and the `tests` folder to the changes that will trigger all tests and we also check if the `force_tests` label is active on the PR, if so: run all tests as well.

**Fixes**
This PR fixes #3607

**How to test**
This should be fine as long as the tests run. Once the label is implemented, verify this as well.


jpelay:
Hi, Timon! This is great! I also added hedy.py to all tests, because sometimes you don't change the grammar, but change hedy, so it principle that should trigger the tests too!

.github/workflows/unittests.yml
```diff
@@ -41,12 +41,12 @@ jobs:
...
       run: |
         $diff = git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }}
-        $SourceDiff = $diff | Where-Object { $_ -match '^content/' -or $_ -match '^grammars/' -or $_ -match '^highlighting/' }
+        $SourceDiff = $diff | Where-Object { $_ -match '^content/' -or $_ -match '^grammars/' -or $_ -match '^highlighting/' -or $_ -match '^tests/' -or $_ -match '^app.py' -or $_ -match '^hedy.py'}
         $HasDiff = $SourceDiff.Length -gt 0
         Write-Host "::set-output name=docs_changed::$HasDiff"
     - name: Run all tests
       shell: pwsh
-      if: steps.check_file_changed.outputs.docs_changed == 'True'
+      if: steps.check_file_changed.outputs.docs_changed == 'True' -or ${{ github.event.label.name == 'force_tests' }}
```
Felienne, line 49: Ah cool, I did not know this was possible!

Felienne (approved):

TiBiBa:
> Hi, Timon! This is great! I also added hedy.py to all tests, because sometimes you don't change the grammar, but change hedy, so it principle that should trigger the tests too!

My bad! I actually had in mind to add that one as well but added `app.py` instead. Maybe we should remove that one?

.github/workflows/unittests.yml
TiBiBa: Not really sure that it works, but according to Stackoverflow it should...

jpelay:
> > Hi, Timon! This is great! I also added hedy.py to all tests, because sometimes you don't change the grammar, but change hedy, so it principle that should trigger the tests too!
> 
> My bad! I actually had in mind to add that one as well but added `app.py` instead. Maybe we should remove that one?

Yes, I feel the same way! I think we should remove hedy.py.

PR #3626, opened 2022-11-10
[TESTING, CYPRESS] tests login page and recover page final two tests

JimmyOei:
**Description**
This PR consists of the last two Cypress tests for the login page and the recover page. Also including some tools that can be used for login in an account using the test database (see https://github.com/hedyorg/hedy/pull/3571). This PR finishes all the tests for the login page and recover page.

**Changes in detail**

**Adds** the test for the modal alert popup in recover page
**Adds** the test for the login button in the login page
**Adds** tool functions for loggin in a user, student and teacher, which exists in the test database

**Running the tests**

**Note** that these tests uses the accounts in the test database, to use this the database has to be fed first by running `bash feed_dev_database.sh` 
An example of running cypress: while being in the hedy/tests directory execute the following: `npx cypress run --spec "cypress/e2e/login_page/*"`
This runs all the login_page tests.

Closes issue https://github.com/hedyorg/hedy/issues/3531.


tests/cypress/e2e/login_page/login_button.cy.js
```diff
@@ -0,0 +1,19 @@
+import {goToLogin} from '../tools/navigation/nav.js'
+import {login, loginForUser} from '../tools/login/login.js'
+
+describe('Login button test', () => {
+  it('passes', () => {
+    goToLogin();
+
+    // Tests login button type and visibility
+    cy.get('button[class*="green-btn mt-2"]')
```
TiBiBa, line 9: Not sure why we get the button on the classes. What if we decide to re-style the login button? The e2e tests would fail in this case! However, the functionality would still be the same.

I assume we can also get the button by the id or value? I know that both are not set at this moment, but feel free to modify the templates as well to remove ambiguity for testing. It would make sense to simply give this button an id to clarify the template as well as the tests (and remove the dependency on classes):

`<button type="submit" class="green-btn mt-2" id="login_button">{{_('login')}}</button>`

tests/cypress/e2e/login_page/login_button.cy.js
```diff
@@ -0,0 +1,19 @@
...
+      .should('be.visible')
+      .should('have.attr', 'type', 'submit')
+      .should('have.text', 'Log in');
```
TiBiBa, line 12: This only works for English, do we know in which language the tests would be run? The `_('login')` is a placeholder that is replaced on render with the user language. Using this as a testing validation is a bit tricky.

tests/cypress/e2e/recover_page/modal_alert_popup.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+    // Tests hidden modal alert text
+    cy.get('#modal_alert_text')
+      .should('not.be.visible')
+      .should('be.empty');
+
+    // Tests visibility of error modal alert text/ recover non existing account
+    cy.get('#username')
+      .type('anonexistingaccount123@#$%^!')
+    cy.get('button[class*="green-btn mt-2"]')
```
TiBiBa, line 15: Same here as explained before, can be improve this?

tests/cypress/e2e/recover_page/modal_alert_popup.cy.js
```diff
@@ -0,0 +1,34 @@
...
+      .clear();
+
+    // Tests visibility of succes modal alert text/ recover existing account
+    cy.get('#username')
+      .type('user1')
+    cy.get('button[class*="green-btn mt-2"]')
+       .click()
+    cy.get('#modal_alert_text')
+      .should('be.visible')
+      .contains('You should soon receive an email with instructions on how to reset your password.');
```
TiBiBa, line 32: This is also a tricky one because of the language of the user or what if we decide to change the string a bit? I understand that you need values and attributes to validate, but try to fallback on "robust" ones that are actually determining the functionality as much as possible.

TiBiBa (changes requested):
Left a few comments, feel free to contact met through Discord if you have any questions (or don't agree of course)!

tests/cypress/e2e/login_page/login_button.cy.js
Felienne: Yes I ran into this! I started the server with language set to non-English and then a test failed. Could we force the tests to run in a certain language?

tests/cypress/e2e/login_page/login_button.cy.js
JimmyOei: Ah, I didn't know we would be allowed to change anything outside of our test, but this helps a lot!

tests/cypress/e2e/login_page/login_button.cy.js
JimmyOei: Hm, I think, since you highlighted the tests should be flexible, checking the texts should not be done, because texts are very likely to change.

TiBiBa: I would suggest to focus on the `id` attribute of most elements. If it is not there, feel free to add it to the templates. The id should be unique by design (if we have duplicate ones this should be fixed). This is unlikely to change and should directly re-present the functionality of the element. For example, the login button could have an `id` attribute `loginButton`.

JimmyOei:
Made the tests more flexible, by creating ID's in the html code for some buttons. I also removed all tests that check on certain texts, because these are likely to change and have nothing to do with the identity of a feature.

tests/cypress/e2e/login_page/login_button.cy.js
JimmyOei: I did exactly this is my rerequest. Maybe I am doing something wrong? Are you still seeing the old files?

TiBiBa: You did not, but my comment is older than your latest changes. Will review your PR again later today!

Felienne:
Hi @JimmyOei! I noticed you merged manually, but that is not needed. After an approving review, the PR will be automerged.

JimmyOei:
What do you mean? I think I merged automatically with the bot. Should I not do that?

Felienne:
> What do you mean? I think I merged automatically with the bot. Should I not do that?

In the logs it looks like you also merged (clicked the merge button)? If the bot did it: all good!

JimmyOei:
Ah it's probably the bot, cause I haven't worked in hedy today :)

PR #3628, opened 2022-11-10
[CONTENT] #3259 reorder adventures based on difficulty

MarleenGilsing:

**Description**
Work towards #3259 
All adventures in level 1 - 13 are now reordered so they go from easiest to hardest. I also added a lot of blanks in the example codes and added challenges to each adventure, so the students know what to do with it and not just play around with the example. 

I still need to do this for level 14 -18, but that's a job for next week


Felienne (approved):
Nice work @MarleenGilsing! A lot more clear for students what to do in the higher levels!

PR #3633, opened 2022-11-11
[CLEAN-UP] Improves server responses

### LINKED ISSUE #3616: [CLEAN-UP] Improve response dictionaries by remove duplicate code
**Description**
Currently we make a distinction between returns with an achievement and without in some places. This makes sense, as we only want to return an achievement when one is actually achieved. However, the rest of the response is duplicate coding in a lot of places. Take the following example:

```
if self.achievements.verify_save_achievements(user['username'],'adventure_name' in body and len(body['adventure_name']) > 2):
      return jsonify({'message': gettext('save_success_detail'), 'name': body['name'], 'id': program_id, "achievements": self.achievements.get_earned_achievements()})
return jsonify({'message': gettext('save_success_detail'), 'name': body['name'], 'id': program_id})
```

The complete return is 90% the same coding, which is a starting point for unfindable bugs. I would propose to re-structure all response using a dictionary that we dynamically fill with key/value pairs dependent on the request flow. For example:

```
response = {}
if achievement:
    response['achievement'] = 'bla bla bla';
response['message'] = 'This all went well';
return jsonify(response)
```

This happens in a lot of places so takes quite some work to clean-up, but I think this is worth it in the end.

TiBiBa:
**Description**
This is a big work-in-progress clean-up task to improve the server responses we return on (mostly) POST requests made from the front-end. To remove some ambiguity and better align our codebase we work on keeping the following styling guidelines:

- Don't add a response code when using `jsonify` when success -> 200 is default
- When returning an error use 400, **don't** `jsonify` the response
    - This is due to the front-end using the `response.responseText` for displaying getting the error
- When return an empty succes, use `return '', 204`

**Note**
It would be even better to also use the correct 4xx code for each error return but for now this should work. There is already quite an improvement by splitting the 2xx response codes and cleaning up the codebase itself.

**Fixes**
This PR fixes #3616.

**How to test**

**Checklist**
  

TiBiBa:
This (of course) will have a lot of test issues as we validate against a 200 response code in the e2e-tests. Should be easy to fix but takes some time!

TiBiBa:
Closed as with the new PEP8 styling the merge conflicts are too complex to fix. Will pick up again in a new PR.

PR #3636, opened 2022-11-11
[CLEAN-UP] Adds teacher authentication wrapper

### LINKED ISSUE #3635: [CLEAN-UP] Add @requires_teacher wrapper to remove duplicate code
**Description**
We currently support to authentication wrappers `@requires_login` and `@requires_admin`. Would be nice to add a wrapper for teacher authorization as well as we have a lot of manual check if the user is a teacher or not.

TiBiBa:
**Description**
In this PR we perform a clean-up to remove some duplicate code and replace this by a wrapper. We already have a wrapper in place to easily check if the current user is logged in or an admin. In this PR we add one for validating if someone is a teacher. This makes our code much more read-able, see the example below:

**Before**
```
@route('/create_adventure', methods=['POST'])
    @requires_login
    def create_adventure(self, user):
        if not is_teacher(user):
            return utils.error_page(error=403, ui_message=gettext('create_adventure'))
```

**After**
```
@route('/create_adventure', methods=['POST'])
    @requires_teacher
    def create_adventure(self, user):
```

**Fixes**
This PR fixes #3635

**How to test**
Nothing to test as this is a clean-up. If all tests pass we are ready to go!


Felienne (approved):
Seems not to break anything!

PR #3640, opened 2022-11-11
[FEATURE] Enables users to view highscores with all class members

### LINKED ISSUE #3638: Show achievements of all class members when filtering highscores on class
**Describe the bug**
_The title is self-explanatory. For some reason the classes highscores remain empty even when multiple students have achievements._

Actually not a bug but a design choice. However, it would be nice if users would be able to view all highscores when filtering on their own class. Currently they are only able to view users with a public profile.

TiBiBa:
**Description**
In this PR we change the design of the highscores slightly. Currently users are only able to view the highscores of users with a public profile. However, in the case of a class some users might not have a public profile while it is still fun to compare the amount of achievements.  We solve this by retrieving achievements of all users when a user visits the class highscores.

**Fixes**
This PR fixes #3638

**How to test**
Make sure you are in a class with users that all have achievements but some might not have a public profile. Verify that these users are still visible when filtering the highscores on class.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -44,13 +44,20 @@ jobs:
         $SourceDiff = $diff | Where-Object { $_ -match '^content/' -or $_ -match '^grammars/' -or $_ -match '^highlighting/' -or $_ -match '^tests/' -or $_ -match '^app.py' -or $_ -match '^hedy.py'}
         $HasDiff = $SourceDiff.Length -gt 0
         Write-Host "::set-output name=docs_changed::$HasDiff"
+    - name: Run weblate tests
+      shell: pwsh
+      id: weblate_tests
+      if: ${{ github.event.pull_request.user.login == 'weblate' }}
+      run: |
+        Write-Host "::set-output name=run::True"
+        build-tools/github/validate --weblate
     - name: Run all tests
       shell: pwsh
-      if: steps.check_file_changed.outputs.docs_changed == 'True' -or ${{ github.event.label.name == 'force_tests' }}
+      if: ${{(steps.check_file_changed.outputs.docs_changed == 'True' || github.event.label.name == 'force_tests') && steps.check_file_changed.weblate_tests.run != 'True'}}
       run: |
         build-tools/github/validate --all
     - name: Run non-code related tests
       shell: pwsh
-      if: steps.check_file_changed.outputs.docs_changed != 'True'
+      if: ${{steps.check_file_changed.outputs.docs_changed != 'True' && steps.check_file_changed.weblate_tests.run != 'True'}}
       run: |
         build-tools/github/validate
--- a/build-tools/github/validate
+++ b/build-tools/github/validate
@@ -10,6 +10,8 @@ $scriptdir/validate-typescript
 $scriptdir/validate-yaml
 if [[ "${1:-}" == "--all" ]]; then
   $scriptdir/validate-tests --all
+elif [[ "${1:-}" == "--weblate" ]]; then
+  $scriptdir/validate-tests --weblate
 else
   $scriptdir/validate-tests
 fi
--- a/build-tools/github/validate-tests
+++ b/build-tools/github/validate-tests
@@ -12,6 +12,8 @@ echo "------> Python unit tests"
 # This is expected to run from the repo root
 if [[ "${1:-}" == "--all" ]]; then
   python -m pytest
+elif [[ "${1:-}" == "--weblate" ]]; then
+  python -m pytest tests/test_snippets
 else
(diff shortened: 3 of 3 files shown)


Mark-Giesen:
I'd like to have this and agree, but does @Felienne?. I think being in a class together is enough reason to share this, but privacy could be an issue for some?

Felienne:
> I'd like to have this and agree, but does @Felienne?. I think being in a class together is enough reason to share this, but privacy could be an issue for some?

Yes I agree, this should be fine from a privacy perspective if it is only visible in the class, but I would want teachers were able to enable/disable this, because you might not want to encourage this type of competition, it is not fun for all students (i.e, not fun for "slower" students). 

So if a teacher does not enable it, students can still compare themselves but only to other students that choose to have a public profile.

Could we add an option for that in the customization page?

Mark-Giesen:
Good point, but that might need a student by student option, since there might always be a student that doesn't like it? Maybe to much trouble and more important things to do? The ones who like to compete can create a public profile.

Felienne:
> Good point, but that might need a student by student option, since there might always be a student that doesn't like it? Maybe to much trouble and more important things to do? The ones who like to compete can create a public profile.

No, my proposal is that the teacher en/disables it for the whole class, so not too much work. If students want to make their profiles public they can, and then they can compare only to other public profiles in the class.

Mark-Giesen:
Yes, I got that, but there might allways be one student that doesn't want to compete, so teachers might never be able to turn it on. That's why just leaving it like it is, might be good enough. But If we talk a little longer, Tibiba might have finished it :-)

Felienne:
> Yes, I got that, but there might allways be one student that doesn't want to compete, so teachers might never be able to turn it on. That's why just leaving it like it is, might be good enough. But If we talk a little longer, Tibiba might have finished it :-)

Yes, but I propose that turning it on or off would not require the student's permission, only the teachers. Bcause as a teacher you need some control over the classroom atmosphere and you do not want "he harry has - achievements hahahaha" without having turned that view on.

And yeah I am sure @TiBiBa can easily add that :D

Felienne (dismissed):
Love the idea! however I would like teachers to be able to turn it off (and let's turn it off by default)

TiBiBa:
> Love the idea! however I would like teachers to be able to turn it off (and let's turn it off by default)

All right! Will fix this coming sunday.

TiBiBa:
Should be good to go!

Felienne:
> It would be a nice solution to transfer the "last achieved" value from the public profile table to the achievements table. 

Ah yeas I agree that makes a lot more sense!

> Write a script to manual transfer them from public profile -> achievements table

Let's migrate the values indeed because I see how students can be attached to them! I would however not advise a script on the db (it is always scary to run scripts in the whole database) but a bit of code that, when a student logs in or when a students get a new achievement (or both) migrates the date then. What do you think?

TiBiBa:
> > It would be a nice solution to transfer the "last achieved" value from the public profile table to the achievements table.
> 
> Ah yeas I agree that makes a lot more sense!
> 
> > Write a script to manual transfer them from public profile -> achievements table
> 
> Let's migrate the values indeed because I see how students can be attached to them! I would however not advise a script on the db (it is always scary to run scripts in the whole database) but a bit of code that, when a student logs in or when a students get a new achievement (or both) migrates the date then. What do you think?

I'm afraid this makes us a bit too dependent on relic code, always having to keep a check for the value in the public profile table. My preference would be to write a script and migrate this at once (after testing on Alpha of course). As we only touch the achievements and public profile tables it wouldn't have any app breaking consequences but would really simplify our implementation in some places.

Mark-Giesen:
It's the achievement issue you allready addressed. I haven't logged in all students, so if the migration has been put there, that might be the issue?

TiBiBa:
> It's the achievement issue you allready addressed. I haven't logged in all students, so if the migration has been put there, that might be the issue?

Whoops this issue occurs when a student doesn't have any achievements. Will implement a check!

TiBiBa:
Fixed!

Mark-Giesen:
Yep, only the name of the fix is incomplete, the error occurs with students with public profile aswell. But is is fixed indeed.

Mark-Giesen (approved):
Nice new feature.

TiBiBa:
> Yep, only the name of the fix is incomplete, the error occurs with students with public profile aswell. But is is fixed indeed.

That should be possible: everyone with a public profile should have at least one achievement. (As you get one for creating the profile)

Mark-Giesen:
OK, you're talking about students in class, I think. I was talking about the logged in user.

Felienne:
> I'm afraid this makes us a bit too dependent on relic code, always having to keep a check for the value in the public profile table. My preference would be to write a script and migrate this at once (after testing on Alpha of course). 

That is true, it is a trade off! What is we copy, rather than migrate the data? I am afraid moving might mean we break stuff if we still use the old data somewhere accidentally.

PR #3642, opened 2022-11-12
[CHORE] Fix typos in various .md files

0WN463:
**Description**

Fix various typos in markdown files.
Since I wasn't sure if the document should be in British English or American English, I opted to retain those words as it is.

I see that in the `CONTRIBUTING.md` that first PR's should ideally be targeting open issues, but I don't seem to see any issues surrounding this, and this seems reasonable enough of a pull request.

Used a built-in dictionary on Linux, so may have false positive/false negatives.

**How to test**

No functional code changes, so no testing needed

**Checklist**
  
### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 Helping build Hedy
 ------------
 
-We would be grateful if you help make Hedy better! First you will want to follow the instructions below to run the code locally and configuring your manchine as explained below. After that, you want to look at these things:
+We would be grateful if you help make Hedy better! First you will want to follow the instructions below to run the code locally and configuring your machine as explained below. After that, you want to look at these things:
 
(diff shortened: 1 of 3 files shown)


Felienne:
Hi @0WN463! 

I totally understand you are trying to be helpful, but working directly against our guidelines is not a great strategy. They do not say "ideally", they say: please no PR without an issue. We have these guidelines for a reason (and the reason is to not overwhelm the team with tiny PRs that are not adding value for users. Each PR, even such a small one, add to our workload).

Precisely "this seems reasonable enough of a pull request" is the problem here, what seems reasonable to an outsider might not feel like that to us, who, as you can see, are working on 20+ PR this week only. 

If you want to join the Hedy community, hop in our Discord and ask us how you can contribute in a meaningful way!

PR #3647, opened 2022-11-12
[CHORE] Fix prod deploy workflow

rix0rrr:
I copy-pasted something from the internet, which turned out not to work.

According to https://docs.github.com/en/actions/learn-github-actions/contexts#github-context, `actor` is a string and not an object.

Just check against individual usernames instead of roles.

**How to test**

Run the workflow and see if it works.

### DIFF
--- a/.github/workflows/deploy-to-prod.yml
+++ b/.github/workflows/deploy-to-prod.yml
@@ -9,7 +9,7 @@ jobs:
       # Restrict who can run this workflow
       - if: github.actor != 'Felienne'
         run: |
-          echo "You must be a Maintainer or Admin to run this workflow." >&2
+          echo "You must be Felienne to run this workflow :D" >&2
           exit 1
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-prod.yml
```diff
@@ -6,9 +6,8 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      # Require one of these roles, configured ON THE REPOSITORY itself
-      # (On the Organization is not enough)
-      - if: github.actor.role != 'Maintainer' && github.actor.role != 'Admin'
+      # Restrict who can run this workflow
+      - if: github.actor != 'Felienne'
         run: |
           echo "You must be a Maintainer or Admin to run this workflow." >&2
```
Felienne, line 12:
```suggestion
          echo "You must be Felienne to run this workflow :D" >&2
```

Felienne (approved):

PR #3654, opened 2022-11-14
[UI] Enables multi-selecting debug lines by pressing shift key

### LINKED ISSUE #3644: [UI idea] Allow disabling of multiple lines at once
Disabling a line of code with the "sleeping guy" is very helpful when debugging but in a long program it is a lot of clicking. Could we change the Ui to make it easier to select more lines at once? Not sure how is easiest to build, maybe by

1. Allowing dragging by keeping the mouse button pressed and/or...
2. Selecting multiple lines by clicking two lines when pressing something like shift and then disabling all lines in between?

TiBiBa:
**Description**
In this PR we make a small QoL improvement by adding an option to multi-select lines for debugging when pressing the shift key. When the shift key is pressed when clicking the line number all lines between the nearest (above) selected debug line and the current one are set as debug lines. Removing the need for selecting all these lines manually.

**Fixes**
This PR fixes #3644.

**How to test**
Verify that you can multi-select debug lines by keeping the shift key pressed when clicking the line number. De-selecting multiple lines is not possible.


Felienne (approved):
Nice UX fix!

PR #3656, opened 2022-11-14
[TESTING] Only test snippets when PR comes from Weblate

### LINKED ISSUE #3652: [TESTS] Run only /test_snippets on Weblate PRs
I just realized we could optimize the tests a bit more by running only the test_snippets tests on Weblate PRs, since they can only change content! Would be nice for me to wait less long on Weblate PRs!

TiBiBa:
**Description**
In this PR we expand our Github actions file to only run the snippet tests when a PR comes from Weblate. As we know this PR can only contain content changes we should be fine by only validating the code snippets. This highly speeds up the process of testing. We do have to keep in mind that, if for some reason, the Weblate translations break the front-end we don't test this any longer.

**Fixes**
This PR fixes #3652.

**How to test**
Merge this PR, re-run the jobs for a weblate PR and verify that only the snippets are tested.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -48,7 +48,7 @@ jobs:
       shell: pwsh
       id: weblate_tests
       if: ${{ github.event.pull_request.user.login == 'weblate' }}
-      run: |F
+      run: |
         Write-Host "::set-output name=run::True"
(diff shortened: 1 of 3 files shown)


Felienne (approved):
Let's see if this works :)

PR #3657, opened 2022-11-15
[FIX] Gutter no longer overlays code in RTL main editor

TiBiBa:
Work towards https://github.com/hedyorg/hedy/issues/3662


Felienne (approved):
This margin indeed does the trick! Good find @TiBiBa

PR #3659, opened 2022-11-15
[BACKEND] ICU MessageFormat support

### LINKED ISSUE #3658: Support for ICU MessageFormat interpolation
This change would enable more powerful features for translations.

Here is an illustration:

> You used an {echo} before an {ask}, or an {echo} without an {ask}. Place an {ask} before the {echo}.

might turn into

> You used an echo before an question, or an echo without an question. Place an question before the echo.

because (in this example) users can switch between echo<->echo and ask<->question keyword sets.

This refactoring will open the door to translators to use this string as a translation:

> You used {echo, select, echo {an} other {a}} {echo} before {ask, select, ask {an} other {a}} {ask}, or {echo, select, echo {an} other {a}} {echo} without {ask, select, ask {an} other {a}} {ask}. Place {ask, select, ask {an} other {a}} {ask} before the {echo}.

With arguments `{ask: "ask", echo: "echo"}` this would be interpolated in the same, correct way, but with `{ask: "question", echo: "echo"}` it would become

> You used an echo before a question, or an echo without a question. Place a question before the echo.

This is an English example, which makes it a bit nonsensical, but especially in languages with grammatical gender, this issue is easier to run into and harder to circumvent by careful phrasing.

I'll discuss my implementation under a PR.

KovacsGG:
### Description

**Adds** requirement for PyICU-binary package
**Adds** icu_i18n.py which exports a gettext wrapper and ICU formatter method

**Removes** import of babel's gettext from app.py, utils.py and hedyweb.py
**Adds** import of gettext from icu_i18n.py to app.py and utils.py
**Refactors** translate_error and translate_list in app.py

**Removes** str.format calls from hedy_content.py
**Adds** calls to icu formater to hedy_content.py

**Fixes #3658**

### How to test

* This PR should introduce minimal or no changes in current behavior
* It should correctly process strings with ICU parameters, similar to "Place {ask, select, ask {an} other {a}} {ask} before the {echo}."
* It should do this regardless if the string is in a .mo file or a .yaml

### Checklist
  
### DIFF
--- a/app.py
+++ b/app.py
@@ -581,7 +581,7 @@ def translate_error(code, arguments, keyword_lang):
 
 def translate_list(args):
     # Deduplication is needed because diff values could be translated to the same value, e.g. int and float => a number
-    translated_args = {gettext(a) for a in args}
+    translated_args = list({gettext(a) for a in args})
 
     if len(translated_args) > 1:
         return f"{', '.join(translated_args[0:-1])}" \
--- a/app.py
+++ b/app.py
@@ -575,8 +575,6 @@ def translate_error(code, arguments, keyword_lang):
                 # Arguments ("tip") may themsleves be patterns, making formatting necessary here
                 arguments[k] = gettext(v, **arguments)
 
-    print(arguments)
-
     error_template = gettext(code, **arguments)
     return error_template
 
--- a/app.py
+++ b/app.py
@@ -579,9 +579,6 @@ def translate_error(code, arguments, keyword_lang):
     return error_template
 
 
-
-
-
 def translate_list(args):
     # Deduplication is needed because diff values could be translated to the same value, e.g. int and float => a number
     translated_args = {gettext(a) for a in args}
--- a/app.py
+++ b/app.py
@@ -17,9 +17,10 @@ import hedy_translation
 from hedy_content import ADVENTURE_ORDER_PER_LEVEL, COUNTRIES, ALL_LANGUAGES, ALL_KEYWORD_LANGUAGES, NON_LATIN_LANGUAGES
 import hedyweb
 import hedy_content
-from flask_babel import gettext, Babel
+from flask_babel import Babel
 from flask_compress import Compress
 from flask_helpers import render_template
+from icu_i18n import gettext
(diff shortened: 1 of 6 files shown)


KovacsGG:
There might be issues with the PyICU package on deployment. PyICU-binary provides manylinux wheels for python 3.9, so Heroku should be able to install it.

Because it's a C extension, static analysis tools can't verify that icu.Formattable and other members exist. This could be fixed by providing a .pyi stub file, ignored, or having the package updated with it. I've filed an issue with the developer (https://gitlab.pyicu.org/main/pyicu/-/issues/156), until then I think it's fine to ignore.

KovacsGG:
This PR will not solve issues related to calls like `gettext('print')`. messages.po doesn't have keys for keywords, those are in .yaml files, where afaik babel doesn't look. This means that keywords in error massages still can't be translated. This doesn't affect strings constructed by hedy_content.py, because that parses the necessary .yaml files, and doesn't rely on babel.

Eventually, we should introduce functions in icu_i18n.py that can look for arguments in the given pattern, and perform the gettext or yaml fetching as necessary. Currently exception message behavior is a bit unruly, the `needs_translation` `needs_formatting` arrays are incomplete, it's a bit arbitrary what they name their arguments. They should also probably take care of formatting themselves, instead of doing it after the fact. (Or they should provide a list of argument names to apply formatting to.)

Please imagine my whole comment is between "I think( ... )". I just wrote it this way for brevity.

Felienne:
Thanks for the PR @KovacsGG and for the detailed explanation! I am not really sure how to move this PR forward (I am not a huge babel/weblate expert) Do you need to do more work or do you need help or input from us?

Mark-Giesen:
Thanks @KovacsGG for this effort. I think this might be a good start for the next level in i18n support. This year we focused on ease of translations for non programmers and achieved this by introducing Weblate and are still in the process of tweaking and tuning the way of working there. I've looked at this code and think I understand, but like to discuss some of the features before introducing ICU in our codebase. I think I reached out to you on Discord, but we can also discuss this in the issue that is attached to this PR. 
I'll close this PR for now, but leave the issue open so it remains in focus and can be reviewed and used as soon as we move to the next level in translations support.

PR #3664, opened 2022-11-16
[CONTENT] Allow non-latin letters for pressed commands

### LINKED ISSUE #3663: [CONTENT] Allow non-Latin letters for pressed commands
currently, only Latin letters and numerals can be used for mapping pressed commands:
```python
if a is pressed
    print "Ouch!"
```
This should also be made possible for non-Latin letters and numerals such as in Arabic, Russian and Hebrew:
```python
if ش is pressed
    print "Ouch!"

if ф is pressed
    print "Ouch!"

if ש is pressed
    print "Ouch!"
```

ToniSkulj:
Fixes #3663, makes it possible to use non-Latin letters for the pressed commands.

image

Also, this implementation makes it possible to use 'special' buttons such as:

image

**Test**

1. There is an extra simple test _test_if_pressed_non_latin_ in _test_level_05.py_

2.  Try out your non-Latin language!


Felienne (approved):
Works!!

image

Felienne:
Love this, thanks a lot! It is a bit out of scope for this PR/issue but the special keys are super cool!! However, must Shift be case-sensitive? I'd prefer it if shift would also work (Can you make a separate issue for that if you think it is feasible)

ToniSkulj:
> Love this, thanks a lot! It is a bit out of scope for this PR/issue but the special keys are super cool!! However, must Shift be case-sensitive? I'd prefer it if shift would also work (Can you make a separate issue for that if you think it is feasible)

My pleasure! This is definitely doable. I opened [#3701](https://github.com/hedyorg/hedy/issues/3701) for this.

PR #3674, opened 2022-11-18
[FIX] Fixes issue with french front-end error when running program

### LINKED ISSUE #3661: [CONTENT/BUG] quote in French error message breaks Javascript
**Describe the issue**
French elif example in level 17 gives a front-end error:

image

Because the text of the error message is: `msgstr "Il y a déjà un programme en cours, finissez d'abord celui-ci."` and that quote in "d'abord" upsets Javascript when the message is loaded on the front-end!

I can fix it for this error (just reqrite the sentence without the quote), but I want to explicitly make an issue because this is a fascinating problem needs a bit of work to be fixed  in general. 

More details to be added later (no help needed at this point)

TiBiBa:
**Description**
In this PR we fix an issue with the `runit()` function that was broken in French. This was due to the default error message `already_programming_running` containing an `'` character. As the placeholder was also surrounded by `' '` to clarify as a string for Typescript the function broke. We fix this by using the Jinja2 function `|tojson` to always safely escape html special characters.

**Fixes**
This PR fixes #3661

**How to test**
Verify that Hedy works again in French.

### DIFF
--- a/templates/incl-editor-and-output.html
+++ b/templates/incl-editor-and-output.html
@@ -139,7 +139,7 @@
     <div class="order-2 lg:order-4" id="code_related_buttons">
       <div class="flex justify-between gap-2 items-start h-12 overflow-visible">
         <div id="runButtonContainer">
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Ah that was a small fix for a large issue! 

image

Issue #3293, opened 2022-09-16
[BUG] Handling of indexed data not always  consistent.


eremmel:
**Describe the bug**
One can use index data from a list. In a number of situations you can use that index-expression (e.g. mylist[i]) directly as expression in a more complex situation. But not at all logical places.

The most minimal implementation is to use an index-expression as the only r-value in a assignment expression. This approach will always work (line 6).
Some more complex and working examples are seen below in the screen shot.
line 7: can be in boolean expression.
line 8: can be in print statement

Accepted with questionable results:
line 9: treated as unquoted literal string (any way wrong at level 18)
line 10: treated as unquoted literal string (?) at least value is not removed.
An not excepted expression is 'if mylist[i] in mylist3:'
 
**Paste the Hedy code & level**
Run this only on lvl 18:
![image](https://user-images.githubusercontent.com/1874108/190668103-64136f3d-53c0-4aba-ac53-47b574aefbc5.png)

**Add a screenshot (optional)**

**Expected behavior**
Allow index expression on all places where a  'input'  type is allowed or force only the most simple approach: indexed data needs to be  copied to variable first and that variable can be used in expressions.

**What machine and browser you were using (optional)**
N/A Server issue.

Felienne:
Hi @boryanagoncharenko!

Here is another language related issue. If might be fixed by #3685, but I am not sure (I did not check) if it does, feel free to close it!

boryanagoncharenko:
PR #3685 addresses having list access of the left-hand-side of adding and removing from/to lists, so I am closing the issue.

PR #3685, opened 2022-11-21
[FIX] Adds support for adding/removing numbers and list accesses to/from lists (#3293)

jakeklaris:
[FIX] added support for adding/removing numbers and list accesses to/from lists (#3293)

**Description**

Adds support to levels 3+ for adding integers or floats to lists, and removing integers of floats from lists. Also fixes the issue where accessing a list item with square bracket indexing gets changed into a literal string when adding or removing from a list (ex. add list1[2] to list2, appends 'list1[2]' as a string to list2)

**Fixes Issue #3293**

**How to test**

Adds a test to test_level_03.py for checking that a number can be added to a list
Adds tests to test_level_16.py for checking that square bracket list indexing correctly gets added and removed from a list


Felienne:
Welcome to Hedy @jakeklaris and well done on picking up a language bug as your first PR! We can simplify the code a bit but it does seem to work (leaving a few details comments now)

hedy.py
```diff
@@ -1469,12 +1469,27 @@ def list_access(self, meta, args):
...
+            value = before_index + '-1' + ']' + after_index # account for 1-based indexing
+        else:
+            value = self.process_variable(args[0], meta.line) 
+
         list_var = args[1]
         return f"{list_var}.append({value})"
 
     def remove(self, meta, args):
-        value = self.process_variable(args[0], meta.line)
+        if args[0].isnumeric(): # is int/float
```
Felienne, line 1485: Would be good to repeat the comment from add here (future readers will not always think to read add when looking for more info on remove)
```suggestion
      # only process arg if arg is a string, else keep as is (ie. don't change 5 into '5', my_list[1] into 'my_list[1]')
        if args[0].isnumeric(): # is int/float
```

hedy.py
```diff
@@ -1469,12 +1469,27 @@ def list_access(self, meta, args):
...
+            value = args[0]
+        elif ('[' in args[0] and ']' in args[0]): # is list indexing
+            before_index, after_index = args[0].split(']', 1)
+            value = before_index + '-1' + ']' + after_index 
+        else:
+            value = self.process_variable(args[0], meta.line)
```
Felienne, line 1491: These lines are exactly the same as for add, can we make a function for this and call it from both places (a bit like arithmetic operators do in level 6 by all calling `process_calculation()`)

Felienne (changes requested):
These are all supersmall style tips, the mechanic seems to work just fine!

jakeklaris:
> These are all supersmall style tips, the mechanic seems to work just fine!

Great! Just updated the code to account for the style tips. Let me know what you think.

Felienne (approved):
Thanks for the fixes, and congrats on getting your first PR merged @jakeklaris!

PR #3690, opened 2022-11-21
[FIX] Fix highlighting to use unicode categories

### LINKED ISSUE #3476: [BUG] Erroneous syntax highlighting in Hebrew. CHARACTER regex without Hebrew characters
**Describe the bug**
In Hebrew, there are two separate keywords: "ב" (at) and "בתוך" (in). When typing the keyword "בתוך", the editor recognizes "ב" as a complete word, and doesn't highlight the rest of the word.
I believe this is down to the fact that the constant `CHARACTER` in the file [highlighting/definition.py](https://github.com/Felienne/hedy/blob/main/highlighting/definition.py), doesn't include Hebrew characters. It should include the range א-ת (Aleph א through Tav ת).

Observed in level 10
image

**Expected behavior**
The entire word "בתוך" should be highlighted as a keyword, not just the first letter "ב".

jpelay:
**Description**

Changes the regex used to detect a character and the end of a word in the highliter, preventing problems with it not detecting end of word correctly in languages that don't use the Latin alphabet.

**Fixes #3476**

**How to test**

Change language to hebrew and try this code in level 5:
```
לכל מספר בתוך טווח 1 עד 5
    הדפס מספר
```

**Checklist**
  

Felienne (approved):
Yes all good and the regex issues should be fixed by #3688

PR #3696, opened 2022-11-23
[CHORE] JavaScript flow: use push token on checkout

rix0rrr:
We need to check out the source with a push-enabled token, otherwise the commit action later on will fail. Never mind that we pass that the correct token, it will not overwrite the token in the git checkout.

**How to test**
Run more workflows and see if they fail.


Felienne (approved):
Here goes nothing!

PR #3697, opened 2022-11-23
[REFACTORING] Change Parsons YAML structure for code lines

### LINKED ISSUE #3677: [CONTENT] Rethink Parsons yaml structure
I am seeing a lot of translation mistakes in the Parsons problems, which is annoying cause I usually end up fixing them. Often the issues are in between lines (different variable names in different lines) and that make me take a look at Weblate and the yamls, and it seems each line is saved and shown separately, f.e.:

https://hosted.weblate.org/translate/hedy/parsons/es/?checksum=f07cac93dbaca3e4&q=comida+es+pizza%2C+helado%2C+ensalada%2C+sandwich&sort_by=-priority%2Cposition

image

Of course this makes sense because the yaml structure is:

image

Can we make the code block into 1 block so it will be shown as one in Weblate? Or somehow make these lines show as one with some weblate magic?

TiBiBa:
**Description**
In this PR we change the structure of the Parsons code_lines in the YAML to make translating easier. Instead of a dict structure we change this to a whole snippet that is splitted on the front-end on runtime. We keep support for both `code_lines` and `code` as a yaml attribute for now to support relic YAML files. This should be cleaned-up once all translations and weblate are happy.

**Fixes**
This PR fixes #3677.

**How to test**
All should work as before as this is only a refactoring.


TiBiBa:
Propose: When merging create a new issue to remove the relic code once Weblate and all translations are up-to-date.

Felienne:
Looks Good @TiBiBa, thanks for the cleanup. We will need to merge the last Weblate PR first (which still has a few issues) I will try to update that over the weekend so we can merge this one in after.

Felienne (approved):

PR #3715, opened 2022-11-25
[WIP, TESTING] Update adventure testing

Felienne:
Turns out we were not testing all snippets in adventures! This is a fix but also requires some updates so weblate locking for now until Weblate is done


Felienne:
We did it @jpelay!!!

image

Issue #3564, opened 2022-11-03
[TESTING] Create front-end tests for the teacher features in the customize adventure page


JimmyOei:
**Description**
Create front-end tests in Cypress for all the features in the `/for-teachers/customize-adventure/ [adventure id]` page.
- change name
- change level
- checkbox to agree to make adventure publicly available
- add adventure to class checkboxes
- preview button
- save button
- remove button
- go back to teachers page button

PR #3721, opened 2022-11-26
3564 customize adventure tests

JimmyOei:
**Description**
This PR consists of all the front-end tests for the ``customize-adventure`` page, which are 9 different cypress tests in total for all the buttons/fields/etc.. on this page. This PR also includes an added adventure to the test database, since there wasn't one before and it is needed to test the ``customize-adventure`` page. There are also some new cypress tools implemented to go to this page and to make a new adventure. Some HTML files have been edited to incude id's for some buttons, this makes the tests easier and more flexible.

**Changes in detail**

**Adds** the 9 tests for the  ``customize-adventure`` page in the ``for-teacher/customize-adventure_page`` directory
**Adds** a function tool to create a new adventure in a new tool directory, namely ``tools/adventures``
**Changes** an added adventure to the test database in ``data-for-testing.json``
**Changes** added id's to the HTML files of ``customize-adventure`` page and ``for-teacher`` page

**Running the tests**

An example of running cypress: while being in the hedy/tests directory execute the following: `npx cypress run --spec "cypress/e2e/for-teacher_page/customize-adventure_page/*"`

Closes issue https://github.com/hedyorg/hedy/issues/3564.


martinj2001 (changes requested):
Looks great to me! I think as a group we should all follow this testing mechanism. 
One thing however, maybe you can add something to the actual buttons to test if they are enabled and thus clickable. This ensures that buttons work properly without executing the logic behind the buttons.

martinj2001 (approved):
Nice!

Issue #3728, opened 2022-11-28
[TESTING] For the login page and recover page tests add testing of enabled buttons


JimmyOei:
For the login page and recover page front-end tests in cypress, add the testing of whether the buttons are enabled or disabled.

PR #3730, opened 2022-11-28
[TESTING, CYPRESS] improved some tests and added the checking of disabled/enabled buttons

JimmyOei:
**Description**
Adds testing whether the buttons are enabled or disabled. Along the way I also improved some tests.

**Changes in detail**

**Changes** the tests for the  ``login`` page
**Changes** the tests for the  ``recover`` page 

**Running the tests**

An example of running cypress: while being in the hedy/tests directory execute the following: `npx cypress run --spec "cypress/e2e/login_page/*"`
This tests all the login page tests.

Closes issue https://github.com/hedyorg/hedy/issues/3728.


Felienne (approved):
Works! (and a lot nicer with the id's instead of the class names!)

image

PR #3738, opened 2022-11-29
[FIX] if pressed not working eventtype object has no attribute unicode

### LINKED ISSUE #3737: [BUG] If pressed not working - 'EventType' object has no attribute 'unicode'
For some reason, _if pressed_ does not work anymore.
The error produces the following error message: **'EventType' object has no attribute 'unicode'**

Example:
image

ToniSkulj:
Fixes #3737. _if pressed_ should work after this PR.


Felienne (approved):
Works again, thanks for the quick action (and nothing was broken on prod, so all good!)

image

PR #3742, opened 2022-11-30
[FIX] Don't lock level on quiz threshold when there is no quiz

### LINKED ISSUE #3741: [BUG] Level 14 quiz is missing
**Describe the bug**
Level 14 doesn't have a quiz. The students can't advance if I demand a certain score for the quiz.

Level 14

TiBiBa:
**Description**
In this PR we fix a bug where level 15 would be blocked because there is no quiz in level 14. When a threshold is set a user would never be able to get it and unlock the next level. We fix this by checking if there is a quiz, otherwise unlock the level anyway.

**Fixes**
This PR fixes #3741

**How to test**
A bit of a hassle but... set a quiz threshold and make sure you have quiz score getting this threshold for all levels. Then verify that you can still go to level 15 even without achieving a score in level 14 (as there is no quiz there).


Felienne (changes requested):
Seems the linter has complaints :) 

image

TiBiBa:
> Seems the linter has complaints :)
> 
> image

Big oopsie on my side, fixed it!

PR #3743, opened 2022-11-30
[CHORE, TESTING] Improves test output

Felienne:
As discussed in the HCH, this improves the test output a bit more so it is easier for translators to see their errors.


jpelay:
With: 
 > We still need an extra issue to be opened to translate the local keywords back into {print}
 
 You mean that the output of the code snippet should be in English and not in the local keyword language?

Felienne:
> It works flawlessly! Gonna merge this in the other branch :3
> 
> image

Great! I just merged it is to make fixing easier for myself!!
> 
> With:
> 
> > We still need an extra issue to be opened to translate the local keywords back into {print}
> 
> You mean that the output of the code snippet should be in English and not in the local keyword language?

Yes! I made an issue now here: #3747

Felienne:
I have merged these changes in with #3715

PR #3748, opened 2022-12-01
[LANGUAGE] Adds buttons to Hedy

### LINKED ISSUE #3678: [LANGUAGE idea] Add buttons to the Hedy language
**Language idea incl level** _(linked with [#3680](https://github.com/hedyorg/hedy/issues/3680))_
A description of your idea, f.e. I would like to be able to delete things from lists in level 5.
The idea is to add a new keyword and syntax to enable the Hedy users to dynamically add buttons to the screen. The functionality of these buttons can then be programmed by the Hedy users themselves using the `if a is pressed` syntax that is already available for normal keypresses on the keyboard from level 5 onwards.

The new syntax will look like this:
```python
foo is button
if foo is pressed print "Hello"
```

This will create an onscreen button and a variable in Hedy in which it is stored. Upon pressing it, the program will print "Hello" to the screen. Only one new keyword is necessary, namely 'button' and the syntax for variables and if statement will be reused to keep it intuitive.

This new keyword will be available from level 5, at the same time as the `if a is pressed` for keyboard keys.

daandj:
**Description**

Adds a new rule to the grammar to make it possible to define a button, with the syntax:
```python
hello is button
```

Changes the grammar rule ```ifpressed``` so that instead of a keyboard character the name of a button can be used, with the expected outcome.

Changes the front-end HTML templates to add a div containing the user created buttons just below the output screen.

**Fixes**

This PR fixes #3678.

Tip, if you use the words fixes before the issue number in this description, the related issue will automatically close then the PR is merged! 

**How to test**

In level 5 run the following Hedy code:
```python
hello is button

if hello is pressed print "Hello world!"
```
A button should appear, and pressing the button should result in the string `Hello world!` appearing in the output and the program terminating.

**Checklist**
  

Felienne:
Looking good @daandj and team!!

Can you let us know what still needs to be done before we can review? It seems to work and there tests are there, so maybe it is ready?

daandj:
As far as I know everything works right now, but we wanted to complete a few more tests to make the test suite more comprehensive. This should be done in a about a day and then the PR can be switched to ready.

RikvanSchaick:
I've found a small bug while making unit tests for this feature. If I make a button in the turtle adventure, the button in the will be pushed down by the turtle canvas, which makes the button not very visible. 

Try running the following code in level 5 #turtle:
```python
x is button
forward 50
```

jonathanvdheuvel:
When you create a button with a name of a lowercase letter e.g. 
```python
x is button
if x is pressed print 'joe'
```
It will create the button on-screen, but when you click the button nothing happens, because the pressed feature is only mapped to key 'x' instead of the on-screen button 'x'.

This may seem logical for us, but I think that a kid would prefer to see the on-screen button work as well.

RikvanSchaick:
Another bug found: 
The button syntax does not work yet from level 12 and up. It gives a `var undefined` error, what possibly indicates a grammar issue.

Felienne:
I can also imagine we merge that we have now and open issues for the remaining problems, so we minimize conflicts

daandj:
All three bugs have now been fixed, and I manually fixed the merge conflicts with main. So this PR should be completely ready...

daandj:
I fixed the three errors and manually fixed all the merge conflicts. The checks still seem to fail, I think this is because of a type in the unit tests, but I'm not sure

Felienne:
Hi team (@daandj!) 

I think this can be reviewed (will take a look either way!)

daandj:
Hi @Felienne, yes I think so too!

Felienne:
I found a few small things still (we can make them into issues and merge this for now)

1. There is a weird white bar under the button:

image

2. Now that we also have real buttons, would it not be clearer if keyboard pressed read:
"waiting for a key press"

image

3. When waiting for a button press, I also see the key pressed pop up. That is a bit confusing I think?

Felienne:
Hi @daandj,

One tests now fails in level 17, maybe due to changes in other PRs:

image

Felienne:
Ah @daandj, issues keep popping up, now that the test is fixed, we now have conflicts in the grammars (I will take a look too)

Felienne (approved):

Issue #3557, opened 2022-11-03
[TESTING] Create front-end tests for the home page


JimmyOei:
**Description**
Create front-end tests in Cypress for all the features in the hedy `home` page.
- Selecting languages
- Clicking headers
- `Try it' button
- Youtube video

KavyaManocha:
Hi, I would love to help work on this feature!

JimmyOei:
Hey! This issue is unfortunately not meant to be done by other developers. The issue belongs to a project I and other Leiden University students are doing for the course Software Engineering, where we develop front-end tests for Hedy. I'm sorry for the inconvenience and will assign all the issues of the project, such that there is no confusion. But I'm sure there are lots of other issues you can help with.

PR #3750, opened 2022-12-01
[TESTING] Front-end tests for the home page

martinj2001:
**Description**

- Added tests for the home page that test the navigation buttons, buttons on the home page as well as the language selection.

Fixes issue #3557 

**How to test**

Run Hedy locally, then navigate to the /tests folder and run:
`npx cypress run --spec cypress/e2e/home_page`


JimmyOei:
Maybe add an id to the 'Try it' button to make the test more modular.
Also, it is better to do the testing of the visibility of the buttons in the individual tests of the buttons. So, instead of one test for all the buttons visibility (nav_visible), include these checks in the individual test fo the button (e.g. hedy_nav_works).

martinj2001:
Fixed that!

JimmyOei (approved):
Great!

PR #3754, opened 2022-12-01
[TESTING, CYPRESS] 3704 testing creating front end test for coding on the hedy page

### LINKED ISSUE #3704: [TESTING] Creating front-end test for coding on the hedy page
Create front-end tests for the following features on the /hedy page:

- Typing in code box
- Read aloud -> selecting a language
- 'Go to level x' button
- 'Go back to level x' button
- Emoji button that shows a list of things to try out -> 'Try' buttons
- Clicking headers within a level to different parts of that level

hanna1231:
[TESTING]

**Description**
These front-end tests tests many features on the `/hedy` page. I was unable to test the read aloud button because it did not show in Cypress and thus I was unable to select it. This pull requests tests the editor box, the 'try' button, back to level button, next level button and switching between adventures in a level.

Fixes #3704 

**How to test**


Felienne (approved):
Works!

image

Felienne:
also, this is great but only tests English, while other languages are more likely to break. Could we test multiple languages (see https://github.com/hedyorg/hedy/issues/3760)

templates/incl-adventure-tabs.html
```diff
@@ -3,7 +3,7 @@
         <div class="flex flex-row overflow-x-scroll overflow-y-hidden mb-2 px-2 pt-2" id="adventures-buttons">
             {% for adventure in adventures %}
               {% if 'adventures' not in customizations or (customizations['adventures'][adventure.short_name] and level|int in customizations['adventures'][adventure.short_name]) %}
-                <div class="tab {% if loop.index == 1 %}tab-selected{% endif %} z-10 whitespace-nowrap flex items-center justify-center" tabindex="0" data-tab="{{adventure.short_name}}">
+                <div class="tab {% if loop.index == 1 %}tab-selected{% else %}adventure{{ loop.index0 }}{% endif %} z-10 whitespace-nowrap flex items-center justify-center" tabindex="0" data-tab="{{adventure.short_name}}">
```
TiBiBa, line 6: I don't agree with this approach, as we use a unique identifier as a class. This should be used as the `id` attribute.
Can we create a new issue and fix this? I also think there is a typo in `{{ loop.index0 }}` and this doesn't work as expected at this point.

Felienne: I made an issue so we don't forget: #3765 (cc @hanna1231)

PR #3758, opened 2022-12-02
[DOCS] Update CONTRIBUTING.md

Felienne:
New contributors find it hard to see what to work on, and sometimes pick up stuff we are still making our minds up about, so let's be a bit more deliberate about this!


jpelay:
I like it!

PR #3761, opened 2022-12-02
[BUG] Improve error message level 5

### LINKED ISSUE #3329: [BUG] error message unclear in level 5 when forgetting the last quote
When you forget the last quote in a `print` or `ask` statement the error is unclear (although most of the times not completely off). When you run the same code in level 4 the error is very clear.

```
print 'Hoi ik ben Hedy de Waarzegger
print 'Ik kan voorspellen wie morgen de loterij wint!'
naam is ask 'Wie ben jij?'
```

Error in level 5:
image

Error in level 4:
image

In this example the error in level 5 is still quite clear. The error type is correct only pointing to the wrong line of code. In some cases, that I cannot reproduce right now, the error was completely wrong when the last quote was missing.

Felienne:
**Fixes #3329**

Current main:
image

New:
image

Tip, if you use the words fixes before the issue number in this description, the related issue will automatically close then the PR is merged! 

**How to test**

Try the code snippet above and see the right text is shown or run the added test.

**Checklist**
  

jpelay (approved):
You make it look easy! :3 Looks great

PR #3763, opened 2022-12-02
[TESTING] Fixes GitHub actions to only test snippets when Weblate PR

TiBiBa:
**Description**
We fix an ongoing issue where the Weblate check in the GitHub Actions was incorrect. This should be fixed now.

**Fixes**

**How to test**
Merge, update Weblate PR with main and verify that the flow works as expected.


Felienne (approved):
Let's see what happens on the next Weblate PR (should be incoming)

PR #3775, opened 2022-12-05
Don't show achievements for the first x programs ran

### LINKED ISSUE #3734: [UI idea] disable achievements in level 1
When students start with Hedy they see a lot of new stuff. The first few programs they run return a pop-up with an achievement. For some students the achievements are more confusing than rewarding. Some think the pop-up is the result of a program rather than the text in the output window. I propose to start with the achievements in level 2 when the students are more used to the interface.

TiBiBa:
**Description**
As suggested by @OnnoEbbens we throw a lot of achievements to the users when they just start using Hedy. In this PR we prevent this by setting an `achievements threshold`, currently set to `5`. When the run count is lower than the threshold no run-related achievements are shown. This should fix the issue.

**Note**
I personally would like to remove the "run 1 program" achievement as this no longer makes sense getting after running x programs. However, this introduces quite some complications to fix, might not be worth the effort.

**Fixes**
This PR fixes #3734.

**How to test**
Create a new account, verify that you don't get any run-related achievements for the first 5 programs ran.


Felienne:
> **Note**
> I personally would like to remove the "run 1 program" achievement as this no longer makes sense getting after running x programs. However, this introduces quite some complications to fix, might not be worth the effort.

I agree, and while it might be a lot of effort, it might be worth it to keep the code base clean?

Felienne (changes requested):
This works, thanks, but it is a bit weird I get the 1 program ran after 5 runs? So maybe removing it is better? Or at least rename it to "run 5 programs"?

TiBiBa:
> This works, thanks, but it is a bit weird I get the 1 program ran after 5 runs? So maybe removing it is better? Or at least rename it to "run 5 programs"?

Fixed in the back-end and in English and Dutch. I hope Weblate takes care of the rest?

Felienne (approved):
Top!

image

PR #3780, opened 2022-12-05
[TESTING, CYPRESS] 3703 tests for saving code on /hedy page

### LINKED ISSUE #3703: [TESTING] Create front-end tests for saving code on hedy page
Create front-end tests for the following buttons on the `/hedy` page:

- 'Save code' button
- 'Save & share code' button

sharandaa:
**Description**

This pull request adds tests for the save button and save & share button on the `/hedy` page when trying to save a level.

Fixes #3703

**How to test**


tests/cypress/e2e/hedy_page/save_button.cy.js
```diff
@@ -0,0 +1,40 @@
...
+        cy.get('#modal-yes-button').click();
+      }
+      
+      goToHome();
+      goToHedyPage();
+
+      // checks if it is correctly saved:
+      cy.get('#editor > .ace_scroller > .ace_content .ace_line').each((element, index) => {
+        if(index == 0) {
+          cy.get(element).should('have.text', 'print hello world!');
```
TiBiBa, line 32: What if we change the order or have a non-english language? Then this would fail! I would propose to first empty the editor, then add the line `print test` and then save the program. This way we are always sure the program contains one line that is the line set in this file.

tests/cypress/e2e/hedy_page/save_share_button.cy.js
```diff
@@ -0,0 +1,59 @@
...
+      // clicking on button for copying the share link
+      cy.get('#modal-copy-button')
+      .should('be.visible')
+      .click();
+
+      // checking if correct link is copied to clipboard:
+      //http://localhost:8080/hedy/d89964c395f647bfb07c0ef4bcce8f31/view (this is the correct link)
+      cy.window().then((win) => {
+        win.navigator.clipboard.readText().then((text) => {
+          expect(text).include('/hedy/d89964c395f647bfb07c0ef4bcce8f31/view');
```
TiBiBa, line 39: I think (correct me if I'm wrong) that the program string is created dynamically one save/share and is different each time. You can find the code that creates this code at `website/programs.py#105` where the program id is set by getting a random hex: `program_id = uuid.uuid4().hex`.

tests/cypress/e2e/hedy_page/save_share_button.cy.js
```diff
@@ -0,0 +1,59 @@
...
+      goToHome();
+      goToHedyPage();
+
+      // checks if it is correctly saved:
+      cy.get('#editor > .ace_scroller > .ace_content .ace_line').each((element, index) => {
+        if(index == 0) {
+          cy.get(element).should('have.text', 'print hello world!');
+        }
+        if(index == 1) {
+          cy.get(element).should('have.text', 'print test');
```
TiBiBa, line 52: Same comment as before, can we make this more dynamically?

TiBiBa (changes requested):

tests/cypress/e2e/hedy_page/save_share_button.cy.js
sharandaa: this cypress test is a javascript file, but the programs is a python file. Is it possible to import a python variable into a javascript file?

TiBiBa: I think that would be quite cumbersome. The easiest solution to completely verify this flow would be:
- Save/share the program
- Navigate to `/programs`
- Get the copy link of the last program (as we just saved it it is the last one)
- Logout
- Navigate to the link, verify that it works

sharandaa: okay noted! I will implement this.

sharandaa: I cannot seem to fetch the clipboard content (which contains the link) in order to navigate to this link. I will try to fix the other comments, but I don't think I can resolve this copied link part before our project deadline (which is Monday).

tests/cypress/e2e/hedy_page/save_share_button.cy.js
sharandaa: I resolved this.

tests/cypress/e2e/hedy_page/save_button.cy.js
sharandaa: I resolved this.

tests/cypress/e2e/hedy_page/save_share_button.cy.js
TiBiBa: It is stored as an `onclick` attribute on the `modal-copy-button`. You can find the code at `app.ts#637`:

```
if (shared) {
      $('#modal-copy-button').attr('onclick', "hedyApp.copy_to_clipboard('" + viewProgramLink(response.id) + "', '" + response.share_message + "')");
```

TiBiBa (dismissed):
I think this is ready to go once the clipboard issue is fixed! We can also decide to remove that and leave the copy button out-of-scope but that is up to you as a project decision.

sharandaa:
We have indeed decided to leave the clipboard link out of scope for this project so I removed this part. The test still checks whether this copy link button is visible and not disabled. I just pushed the final version to the branch so it should be ready to go.

Felienne (approved):

PR #3781, opened 2022-12-05
[TESTING, CYPRESS] 3765 added an id for the adventure classes and use this to select the buttons

### LINKED ISSUE #3765: [TESTING, CYPRESS] Improve on loop from 3754
I don't agree with this approach, as we use a unique identifier as a class. This should be used as the `id` attribute. Can we create a new issue and fix this? I also think there is a typo in `{{ loop.index0 }}` and this doesn't work as expected at this point.

_Originally posted by @TiBiBa in https://github.com/hedyorg/hedy/pull/3754#discussion_r1038221303_

hanna1231:
[TESTING]

**Description**
There is now id for every adventure tab in the `/hedy` which is used to select the tabs. To check if the tab is selected I use the class of the element.

Fixes #3765

**How to test**


TiBiBa (approved):
Looks good, thanks for the changes!

PR #3782, opened 2022-12-05
[CHORE] Removes view columns on classes overview

### LINKED ISSUE #3776: [UI] Make name of class link (instead of the eye symbol)
We now have an extra column to visit a class with the eye at 'Bekijk' (sorry for the Dutch img)

image

Could we remove that column, and make the name a link? I think that is more conventional.

And we might have the eye in more places, let's do this everywhere.

TiBiBa:
**Description**
In this PR we remove the view column on the classes overview page and make the name clickable instead.

**Fixes**
This PR fixes #3776

**How to test**
Verify that you can navigate to a class by clicking the name now that the eye symbol is gone.


Felienne (approved):
Works thanks!

image

PR #3787, opened 2022-12-05
[UI] Make newsletter sign up only for teachers & add new question to sign-up form only for teachers

### LINKED ISSUE #3769: [UI IDEA] Change signup forms a bit
Two small changes I would like to see in the sign-up page:

1) Remove the field "sign up to newsletter" for student users. When we send email it is always adult facing (the fact that it is there is legacy from when we had just 1 signup)
2) Add a 'How have you heard about Hedy?' option for teacher users, with the options:
* From another teacher
* Social media
* From a video
* From a magazine or website
* Other

And store these results in the database so we get a better understanding of what leads teachers to Hedy.

sfayyad:
**Description**
In this PR, we make some changes to the sign-up form:
- The "subscribe to the newsletter" option is only available for teachers signing up and NOT students signing up.
- The "how did you hear about Hedy?" question and its options is only available for teachers signing up and NOT students signing up.

**Fixes**
This PR fixes #3769 

**How to test**

- Try signing up as a student to verify that they can't see the newsletter option and new section.
- Try signing up as a teacher a verify that they can see the newsletter option and new section.
- Run tests_e2e.py (specifically test_invalid_signups() and test_signup()) to test back-end functionality.


Felienne:
Thanks a lot [sfayyad](https://github.com/sfayyad)!! Well done! I have updated a front-end test that tested whether the signup was there for students.

templates/signup.html
```diff
@@ -104,6 +104,45 @@ <h4 class="text-blue-500 mt-0 mb-0 text-center">{{_('create_student_account_expl
                         {% endfor %}
                     </select>
                 </div>
+                <!-- START OF NEW QUESTION -->
+                {% if is_teacher %}
+                <div class="signup-container">
+                    <label class="signup-label" for="heard_about">{{_('heard_about_hedy')}}</label>
+                    <ul class="signup-input appearance-none">
+                        <li class="flex" style="justify-content: end;">
```
TiBiBa, line 112: Please style these using Tailwind instead of hard-coded CSS. `class="flex justify-end"` should do the trick!

templates/signup.html
```diff
@@ -104,6 +104,45 @@ <h4 class="text-blue-500 mt-0 mb-0 text-center">{{_('create_student_account_expl
...
+                            <div class="flex flex-row items-center gap-2">
+                                <input type="checkbox" name="heard_about" value="from_another_teacher">
+                                <label style="width: 201px;">{{_('from_another_teacher')}}</label>
```
TiBiBa, line 115: Same here, please style using Tailwind. It might take few tries, but something like `class="w-64"` should work. I personally would like to not style the label at all as it should automatically take the width of it's content. As there is already a `gap-2` on the container `div` this should all work fine.

templates/signup.html
```diff
@@ -104,6 +104,45 @@ <h4 class="text-blue-500 mt-0 mb-0 text-center">{{_('create_student_account_expl
...
+                                <input type="checkbox" name="heard_about" value="from_magazine_website">
+                                <label style="width: 201px;">{{_('from_magazine_website')}}</label>
+                            </div>
+                        </li>
+                        <li class="flex" style="justify-content: end;">
+                            <div class="flex flex-row items-center gap-2">
+                                <input type="checkbox" name="heard_about" value="other_source">
+                                <label style="width: 201px;">{{_('other_source')}}</label>
+                            </div>
+                        </li>
```
TiBiBa, line 141: Do we want to enable users to select multiple options? Otherwise a radio button or a select dropdown would make more sense (and removes a lot of duplicate code).

website/auth_pages.py
```diff
@@ -146,6 +146,15 @@ def signup(self):
         if "country" in body:
             if not body["country"] in COUNTRIES:
                 return gettext("country_invalid"), 400
+        if "heard_about" in body:
+            if isinstance(body["heard_about"], str):
+                body["heard_about"] = [body["heard_about"]]
+            if not isinstance(body["heard_about"], list):
+                return gettext("heard_about_invalid"), 400
+            for language in body["heard_about"]:
+                if language not in ["from_another_teacher", "social_media", "from_video", "from_magazine_website", \
```
TiBiBa, line 155: Please use a more explanatory variable name than `language`, maybe `option`?

website/auth_pages.py
```diff
@@ -168,8 +177,7 @@ def signup(self):
         if not is_testing_request(request) and "subscribe" in body:
             # If we have a Mailchimp API key, we use it to add the subscriber through the API
             if MAILCHIMP_API_URL:
-                role = "teacher" if "is_teacher" in body else "student"
-                mailchimp_subscribe_user(user["email"], body["country"], role)
+                mailchimp_subscribe_user(user["email"], body["country"], "teacher")
```
TiBiBa, line 180: As we are providing a hard-coded string here to the function, maybe remove it all together and make sure the function always handles the request as a teacher?

TiBiBa (changes requested):
Hi @sfayyad! Thanks for your contribution, awesome! I left some comments but otherwise it looks great. Please let me know if you need any help solving them. I'm also a bit struggling with the current UI, could we align the options with the other input fields?
Schermafbeelding 2022-12-06 om 11 24 46

templates/signup.html
sfayyad: Applied changes! I had these in-line styles only for my current design; however, I don't need them if we are aligning the content to the left instead of the right, so I removed all in-line styles.

templates/signup.html
sfayyad: Applied changes! (removed all, same reason as comment above)

website/auth_pages.py
sfayyad: Completely agree! Don't know why I didn't change it before the PR. Anyways, applied changes!

website/auth_pages.py
sfayyad: Applied changes!

TiBiBa:
Awesome thanks for the quick fixes, will take a look again!

TiBiBa:
Whoops! Had some issues with the keys. When I extract and update we lose quite some translations for some reason. Let's keep it this way with only the English translations. Weblate should take care of the rest.

PR #3794, opened 2022-12-06
[BUG] GitHub Actions Error "ALSA lib ..."

### LINKED ISSUE #3793: [BUG] GitHub Actions Error "ALSA lib ..."
After [Fix run code with pygame](https://github.com/hedyorg/hedy/pull/3698/commits/0b299b90e1e2b2bfa8675c3bae8414320a3856e0) (whoops my mistake) the test runner on Github actions gives the following errors/output:
image

It seems that the ALSA (the audio driver) is missing.

ToniSkulj:
Fixes #3793. it sets the audio drivers for PyGame to _disk_ (essentially a dummy driver). 

Found on: https://stackoverflow.com/questions/72311056/alsa-not-found-github-actions-unit-testing-setting-env-envvars-correctly


Felienne (approved):
Thanks for the quick fix!

Hard to know if this will work in Actions, but it looks legit so let's merge and see!

PR #3804, opened 2022-12-07
[FIX] i escaping and modified tests accordingly - bug #3676

### LINKED ISSUE #3676: [BUG] Variable i should be escaped
**Describe the bug**
The label in the right top corner of the output window shows variable in the program that ran. But if you use ```repeat``` an extra variable i is introduced. This variable should be escaped. 

image

KavyaManocha:
** Fixes #3676 **

I changed the variable "i" to be __i __. This change causes i to not show up on the variable drop down. I also had to change test suite 7 by replacing all of the "i's" with __i __ so that they would work correctly. 

Note: I'm not 100% sure why changing i to __i __ fixes this bug, I was messing around with changing the variable and it just worked. So, this could be an incorrect approach. 

**How to test**
On level 7: Story, you can see that without the fix, the variable "i" shows up in the variable drop down. With the fixed code, the "i" variable doesn't show up. 

Link: https://hedy.org/hedy/7#story

Screen Shot 2022-12-07 at 6 55 03 PM
Screen Shot 2022-12-07 at 6 54 24 PM


Felienne:
Thanks for this PR @KavyaManocha! I am also not sure why it works like this, but I guess we escape variables starting with _ in the frontend... I will dive in.

The issue with `trtl` also remains: 

image

But we can also fix that in a separate PR!

Felienne:
Hi @KavyaManocha!

I found why it works, we escape varis with `__` in them (because they are often Python built-in vars), see here:

https://github.com/hedyorg/hedy/blob/648d8409dc640ac9b0a098fb352d56edba57ee4e/static/js/app.ts#L1522

How awesome you stumbled on this! I have also found fixed the turtle issue so we should be good to go as soon as I update the tests

Felienne (approved):

Felienne:
Congrats on getting your first PR merged @KavyaManocha!! This is nice fix for a bug that is confusing for kids!

KavyaManocha:
Thank you Felienne I'm glad I could contribute!

Issue #3708, opened 2022-11-24
[TESTING] Create front-end tests for the features on the admin statistics page


hanna1231:
Create front-end tests for the following features on the `/admin/stats` page:

- 'Return to admin page' button 
- Loading date for this week/past 2 weeks/past 4 weeks/past year

PR #3811, opened 2022-12-08
[TESTING] 3708 stats page front-end tests

JimmyOei:
**Description**
This PR consists of all the front-end tests for the ``admin/stats`` page, which are 5 different cypress tests in total: 1 for the "go back to admin page" button and 4 for the different period toggles. This PR also includes  some changes to the HTML file of the stats page, namely for id's for the buttons and toggles.

**Changes in detail**

**Adds** the 5 tests for the  ``customize-adventure`` page in the ``for-teacher/customize-adventure_page`` directory
**Adds** the pages in the cypress config file ``cypress.config.js``
**Adds** id's to ``admin-stats.html``

**Running the tests**

The tests are for the admin features, therefore the user we are expecting to be the admin, which is *"user5"* in the test database (with password *"123456"*), should be be made an admin user before testing. We do this by running the following command before running the application:  `export ADMIN_USER=user5`. So, you can do for example: `export ADMIN_USER=user5; python app.py`. After doing this the cypress tests can be run as usual.
An example of running cypress: while being in the hedy/tests directory execute the following: `npx cypress run --spec "cypress/e2e/for-teacher_page/customize-adventure_page/*"`

Closes issue https://github.com/hedyorg/hedy/issues/3708.


Felienne (approved):
Check and well done with the id's!

image

Issue #3548, opened 2022-11-02
[CONTENT] Adding example Hedy code with the new keyword 'pressed' in the interface 


RikvanSchaick:
❗ https://github.com/Felienne/hedy/issues/3337 has to be completed first.

Issue 3337 adds the keyword 'pressed' to the Hedy language to make key mappings possible. For example, the user could write something like the following:

```python
if pressed is 'a' print 'cool!' else print 'meh'
```

In levels where the new keyword 'pressed' can be used, it is necessary for the user of hedy to have some example programs in the interface of hedy with this keyword. Think of something like:

 ```python
name is ask 'what is your name?'
if a is pressed print name
else print 'Hedy!' 
```

PR #3814, opened 2022-12-08
added pressed intro to level 5 and 8

maximilionis:
[Fix] Now there are short intros for pressed in level 5 and 8. No other big changes are introduced to pressed.

Fixes issue #3548


maximilionis:
@Felienne if you'd prefer me to explain the option of pressed more in detail in the indroduction too, and not only in the Pressit! adventure, please inform me here. The reason I only did level 5 and 8 was because there were significant changes to pressed here.

Felienne (approved):
Nice, thanks!! (hope this does not upset Weblate...)

PR #3815, opened 2022-12-08
[TESTING] Fixed some problems with admin page test, made them more neat

martinj2001:
**Description**

Added cypress.env variables to make it the code base more maintainable
Added id's to buttons
Added goToAdminUsersPage

**Fixes _issue or discussion number_**
Fixes #3705 

**How to test**
Run in the /tests folder:
```
npx cypress run --spec cypress/e2e/admin_page
```

PR #3816, opened 2022-12-08
Revert "[TESTING] Fixed some problems with admin page test, made them more neat"

Felienne:
Reverts hedyorg/hedy#3815

I am not able to run these tests locally successfully, and I can't seem to use local admin anymore at all. Reverting to check.


Felienne:
Never mind, all seems in order!

PR #3824, opened 2022-12-09
[CHORE] Update cypress installation instructions

rix0rrr:
`cypress` was being installed independently, without a locked version, by every individual tester.

Put the version of `cypress` we are expecting in `package.json`, so that at least everyone has the same version, and our use of the `cypress` dependency is encoded in a machine-readable location (for GitHub Actions etc).

**How to test**

Follow the instructions, observe that the tests still run.


Felienne (approved):
Yup this still seems to work!

Issue #3707, opened 2022-11-24
[TESTING] Create front-end tests for the features on the admin adventures overview page


hanna1231:
Create front-end tests for the following features on the `/admin/adventures` page:

- Return to admin page button 
- Link to view adventure

PR #3826, opened 2022-12-10
[TESTING] 3707 tests for adventures overview page

JimmyOei:
**Description**
This PR consists of all the front-end tests for the ``admin/adventures`` page, which are 2 cypress tests in total: 1 for the "go back to admin page" button and 1 for button to go to an adventure. This PR also includes  some changes to the HTML file of the adventures overview page, namely for id's for the buttons and toggles. And very small improvements to the tests for the ``admin/stats`` page (just some semicolons I forgot, which do not conflict with code, because cypress can still run without them).

**Changes in detail**

**Adds** the 2 tests in the ``admin_page/adventures_page`` directory
**Adds** the variable for the ``admin/adventures`` page in the cypress config file ``cypress.config.js``
**Adds** id's to ``admin-adventures.html``
**Adds** some semicolons I forgot for the tests for the ``admin/stats`` page in the ``admin_page/stats_page`` directory

**Running the tests**

The tests are for the admin features, therefore the user we are expecting to be the admin, which is *"admin"* in the test database (with password *"123456"*), should be made an admin user before testing. We do this by running the following command before running the application:  `export ADMIN_USER=admin`. So, you can do for example: `export ADMIN_USER=admin; python app.py`. After doing this the cypress tests can be run as usual.
An example of running cypress: while being in the hedy/tests directory execute the following: `npx cypress run --spec "cypress/e2e/admin_page/adventures_page/*"`

Closes issue https://github.com/hedyorg/hedy/issues/3707.


Felienne:
Leaving is open in case you forgot to push more commits, but feel free to close if the changes indeed have been merged!

Issue #3800, opened 2022-12-07
[LANGUAGE idea] Add new exception for `at random` used on level 16


jpelay:
**Language idea incl level**

I really don't know if this error is at all common, but I've noticed that when a user tries to access a list with `at random` on level 16, it throws an ugly parse error. Maybe we can add that rule as an error on level 16 and throw a nice error message instead?

**Additional context**

This should throw a nice error message:
![image](https://user-images.githubusercontent.com/45865185/206235973-ac822dcb-fc81-475f-88d7-ab95b3a17173.png)

heemlanden-bch:
This would be nice indeed, I also have some students that are running into this.

Perhaps a simpler fix could be to add a line to the introduction of Level 16 that states something like: Also, from now on you can write `fruit[random]` instead of `fruit at random` to get a random item from the `fruit` list?

boryanagoncharenko:
It seems that both the better error message and the content change are already there:
<img width="748" alt="Screenshot 2024-04-04 at 16 07 38" src="https://github.com/hedyorg/hedy/assets/3010723/812ef0d6-1431-4de4-8597-6f2935b8270c">

PR #3805, opened 2022-12-07
[LANGUAGE] Adds new exception to catch 'at random' instead of 'list[random]'

reiades:
**Description**
Fixes issue #3800 so now if a student attempts to use 'at random' in either a print statement or an expression or just by itself (fruit at random), an informative message will tell them to use 'list_name[random]' instead. 

**How to test**

Added some test cases to test for "print fruits at random," "x = fruits at random," and "x is fruits at random"

**Checklist**
Done? Check if you have it all in place using this list:*
  
- [ ] Contains one of the PR categories in the name
- [ ] Describes changes in the format above
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #3767, opened 2022-12-03
[GRAMMAR] Simplify grammar list access

Felienne:
Work towards #1619 (OMG an issue number starting with 1!!)

PR #3831, opened 2022-12-11
[LANGUAGE] Adds new exception to catch the use of `at` in level 16+

reiades:
**Description**
Fixes issue #3800 so now if a student attempts to use 'at random' in either a print statement or an expression or just by itself (fruit at random), an informative message will tell them to use 'list_name[random]' instead. 

**How to test**

Added some test cases to test for "print fruits at random," "x = fruits at random," and "x is fruits at random"

**Checklist**
  

reiades:
By accidentally, my old fork was deleted so I had to grab my commits from my other PR and make a new one. I closed PR #3805

Felienne:
> By accidentally, my old fork was deleted so I had to grab my commits from my other PR and make a new one. I closed PR #3805

NP, thanks for this! @jpelay can make take a look today, if not I will look at it tomorrow.

grammars/level16-Additions.lark
```diff
@@ -1,9 +1,12 @@
 // adds list access with numbers and brackets
```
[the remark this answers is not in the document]
Felienne: Hi @reiades!

I think we should mark this as draft since you are still aiming to make changes. If you need any help, let us know!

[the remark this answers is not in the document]
reiades: I think this has all the correct functionality for this issue. I just was pointing out that I couldn't find a command that would parse "print fruit at random" in the lark file and had to do it when traversing the AST @Felienne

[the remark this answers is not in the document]
Felienne: ok, I will take a look!

hedy.py
```diff
@@ -1687,6 +1697,16 @@ def __init__(self, lookup, numerals_language):
         super().__init__(lookup, numerals_language)
         self.ifpressed_prefix_added = False
 
+    def list_access_var(self, meta, args):
+        var = escape_var(args[0])
+         # if we tried to use 'at random' in level 16 raise an exception
```
Felienne, line 1702: I don't really understand this command because I don't see what exception you are referring to here?

Felienne:
Hi @reiades!

Your changes seem to work, but I am not 100% sure about that new method and the comment I refer to above. Can you explain in a bit more depth?

Felienne:
Ah, this was my bad! I missed a few instances of list_access_var in #3767 that can be removed.

hedy.py
```diff
@@ -823,6 +824,12 @@ def get_type(self, tree):
             if in_lookup:
                 return type_in_lookup
             else:
+
+                # if the variable name is 'at', it's because they tried to print "at random" 
+                # in a level that wasn't allowed
+                if var_name == 'at' and self.level >= 16:
```
Felienne, line 830: Do we need this? We already have the error rule right (and it will only work in En if we hard code 'at' like this)

reiades: I was having some trouble with adding a rule for print in the lark file so did it this way instead. I looked at how "print fruit at random" was handled in previous levels, but it appears that when "list_access" was updated to now be "fruit[x]" opposed to "fruit at x," I couldn't correctly get the rule to work anymore. I agree that "At" will only work in English, so need to think of a different solution. 

I tried something like this 

```error_random_print: _PRINT (_print_argument_error)? -> error_random_keyword
_print_argument_error:  (_SPACE | quoted_text | list_access | expression | print_expression | at_random)``` but it didn't parse correctly. I feel like im a pretty close though?

Felienne: Maybe you could make a draft PR with the grammar attempt so I can take a look more easily?

Felienne (changes requested):
Let's see if we can make it less dependent on English

hedy.py
reiades: hi! sorry school ended and I went on vacation and then it was Christmas. I will take a look again soon- but for now converted it to draft!

jpelay:
Hello @Felienne! As discussed in the meeting I fixed the problems that were present in this PR and got it working!

Felienne (approved):
How nice to get this out @jpelay!!

PR #3833, opened 2022-12-11
[LANGUAGE] Allowing to change items by index in a list

### LINKED ISSUE #3832: [BUG] Change list by index doesn't work
**Describe the bug**
I noticed than in level 16 there is a rule `change_list_item` that allows to change the item inside a list using its index. But when you try this in the parser it throws an ugly parse error.

Level 16
```
l = [1, 2]
m is 2
l[m] = 3
```

jpelay:
**Description**

Adds the rule `change_list_item` to the `command` rule in level 16. This rule had its own funtion in hedy.py, it just was orphan

**Fixes #3832**

**How to test**

I added a couple of tests


Felienne:
> Adds the rule `change_list_item` to the `command` rule in level 16. This rule had its own funtion in hedy.py, it just was orphan

Probably a refactor gone wrong somewhere? Thanks for the fix!

Felienne (approved):

Issue #3634, opened 2022-11-11
[CLEAN-UP] Replace 403 error (forbidden) by 401 (unauthorized)


TiBiBa:
**Description**
We currently throw a 403 error code in a lot of places. For example in the `@requires_login` wrapper. However, the `403` code is reserved for responses where the server refused to respond. This is not the case: The user isn't allowed. In this case we should throw a `401` code, specifically for an unauthorized request. It would be a nice clean-up to fix this in (all) places.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

**Edit**
To clarify a bit on the subject: throwing a 403 error is fine when something is forbidden. However, at the point of authentication we should throw a 401. This is, for example, the case when logging in.

PR #3836, opened 2022-12-11
[REFACTORING] Clean-up the server responses and response codes

### LINKED ISSUE #3616: [CLEAN-UP] Improve response dictionaries by remove duplicate code

TiBiBa:
**Description**
This is a big clean-up task to improve the server responses we return on (mostly) POST requests made from the front-end. To remove some ambiguity and better align our codebase we work on keeping the following styling guidelines:

- Replace all `jsonify()` returns to `make_response()` with the corresponding response code
- Remove all hard-coded response messages, even when not likely to be returned to user
- When return an empty succes, use 204 instead of 200
- Improving the other error codes is handled in a separate issue #3634 

We have quite some hard-coded English return statements that should never occur as long as the user follows the "normal" front-end flow. We already enforce the right behavior on the front-end and the flow is only entered when a user is hacking a bit. Therefore I propose to make the following change:
```
if not isinstance(body, dict):
    return make_response("body must be an object", 400)
if not isinstance(body.get("name"), str):
    return make_response("name must be a string", 400)
```
into
```
if not isinstance(body, dict) or not isinstance(body.get("name"), str):
    return make_response(gettext("request_invalid"), 400)
```

**Note**
It would be even better to also use the correct 4xx code for each error return but for now this should work. There is already quite an improvement by splitting the 2xx response codes and cleaning up the codebase itself.

**Catch all in a `make_response()` function!**

**Fixes**
This PR fixes #3616.

**How to test**
<Todo>


TiBiBa:
@Felienne I'm struggling a bit with cleaning up the hard-coded responses. We only return hard-coded once when a flow occurs that should be possible. For example when a user makes a direct POST with invalid fields instead of using the normal front-end (where we already check the fields). Therefore I would like to propose the following:
```
if not isinstance(body, dict):
    return make_response("body must be an object", 400)
if not isinstance(body.get("name"), str):
    return make_response("name must be a string", 400)
```
into
```
if not isinstance(body, dict) or not isinstance(body.get("name"), str):
    return make_response(gettext("request_invalid"), 400)
```

What do you think? This cleans-up a lot! And we can use the `gettext()` to return so generic error message.

Felienne:
> @Felienne I'm struggling a bit with cleaning up the hard-coded responses. We only return hard-coded once when a flow occurs that should be possible. For example when a user makes a direct POST with invalid fields instead of using the normal front-end (where we already check the fields). Therefore I would like to propose the following:
> 
> ```
> if not isinstance(body, dict):
>     return make_response("body must be an object", 400)
> if not isinstance(body.get("name"), str):
>     return make_response("name must be a string", 400)
> ```
> 
> into
> 
> ```
> if not isinstance(body, dict) or not isinstance(body.get("name"), str):
>     return make_response(gettext("request_invalid"), 400)
> ```
> 
> What do you think? This cleans-up a lot! And we can use the `gettext()` to return so generic error message.

That does seem a lot cleaner! I will double check with @rix0rrr whether there are some best practices to follow here

Felienne:
> > @Felienne I'm struggling a bit with cleaning up the hard-coded responses. We only return hard-coded once when a flow occurs that should be possible. For example when a user makes a direct POST with invalid fields instead of using the normal front-end (where we already check the fields). Therefore I would like to propose the following:
> > ```
> > if not isinstance(body, dict):
> >     return make_response("body must be an object", 400)
> > if not isinstance(body.get("name"), str):
> >     return make_response("name must be a string", 400)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > into
> > ```
> > if not isinstance(body, dict) or not isinstance(body.get("name"), str):
> >     return make_response(gettext("request_invalid"), 400)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think? This cleans-up a lot! And we can use the `gettext()` to return so generic error message.
> 
> That does seem a lot cleaner! I will double check with @rix0rrr whether there are some best practices to follow here

@rix0rrr had no notes, so this seems like a fine plan forward!

TiBiBa:
> > > @Felienne I'm struggling a bit with cleaning up the hard-coded responses. We only return hard-coded once when a flow occurs that should be possible. For example when a user makes a direct POST with invalid fields instead of using the normal front-end (where we already check the fields). Therefore I would like to propose the following:
> > > ```
> > > if not isinstance(body, dict):
> > >     return make_response("body must be an object", 400)
> > > if not isinstance(body.get("name"), str):
> > >     return make_response("name must be a string", 400)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > into
> > > ```
> > > if not isinstance(body, dict) or not isinstance(body.get("name"), str):
> > >     return make_response(gettext("request_invalid"), 400)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What do you think? This cleans-up a lot! And we can use the `gettext()` to return so generic error message.
> > 
> > 
> > That does seem a lot cleaner! I will double check with @rix0rrr whether there are some best practices to follow here
> 
> @rix0rrr had no notes, so this seems like a fine plan forward!

On second thoughts this still can get quite messy when we have a lot of values. I would like to implement some sort of validation scheme that we use to verify our POST dictionaries. I've been searching and something like this might be useful:

https://pypi.org/project/jsonschema/

It enables us to store our validation schemes separately and call them on an object. Which makes our routing code a lot cleaner and better to read whereas we can update our validation in one place.

TiBiBa:
Note that this is only for validating attributes that should be there in the first place within a normal UX flow. Therefore we can fallback to a generic error message at the user should never see this unless they are making direct requests / POSTs to our server.

jpelay:
Hello @TiBiBa! I really like what you're doing in this PR, but it's drifting away from main more and more. Therefore I'm going to close it and keep this as an issue (#4029) for now!

PR #3842, opened 2022-12-11
[TESTING] 3830 testing create front end tests for the admin page

### LINKED ISSUE #3830: [TESTING] Create front-end tests for the admin page
Create front-end tests for the following features on the ``/admin`` page:

- "User overview" button
- "Classes overview" button
- "Adventure overview" button
- "Achievements overview" button
- "Statistics" button
- "Logs" button

CasProgram:
**Description**

This includes the front-end tests for the admin overview page (`/admin`)

Fixes #3830 

**How to test**

Run 
`export ADMIN_USER=admin; python app.py` in your virtual environment. This ensures that there is an active admin and Hedy is running.

Then you can go to the /tests folder and run: 

`npx cypress run --spec cypress/e2e/admin_page/admin_overview_page`

The specs should have passed!


CasProgram:
Since there is no standard admin user in the dev_database, I have added the following command: `export ADMIN_USER=teacher1`. If this could be done easier I would gladly hear it!

JimmyOei:
There should be a user named "admin" in the test database, which can be used as admin (i.e. ``export ADMIN_USER=admin``)

CasProgram:
> There should be a user named "admin" in the test database, which can be used as admin (i.e. `export ADMIN_USER=admin`)


Thanks Jimmy, just updated this!

TiBiBa (changes requested):
The tests all seem to fail for me with a similar error:
```
 AssertionError: Timed out retrying after 4000ms: Expected to find element: `#achievements_button`, but never found it.
```

Your code itself is fine, but the separate files contain a lot of duplicate code. Resulting in an exploding amount of source files that are hard to keep track of and maintain by other contributors. Every file only tests if one button can be clicked, could you look into a way to make this more generic?

f.e. Make a dictionary of button names and expected urls, iterate this and verify if all are as expected. This way we can reduce a lot of files with duplicate functionality and keep the codebase read-able. Let me know if you need any help!

CasProgram:
Thanks for the feedback @TiBiBa. I have just combined all seperated files into one file to test all buttons. Could you take a look if it is OK now?

JimmyOei:
> The tests all seem to fail for me with a similar error:
> 
> ```
>  AssertionError: Timed out retrying after 4000ms: Expected to find element: `#achievements_button`, but never found it.
> ```
> 
> Your code itself is fine, but the separate files contain a lot of duplicate code. Resulting in an exploding amount of source files that are hard to keep track of and maintain by other contributors. Every file only tests if one button can be clicked, could you look into a way to make this more generic?
> 
> f.e. Make a dictionary of button names and expected urls, iterate this and verify if all are as expected. This way we can reduce a lot of files with duplicate functionality and keep the codebase read-able. Let me know if you need any help!

I totally agree that this will result in a lot of small test files, however it's important to mention that we agreed on this structure from the beginning (a single test per button) to test individual buttons seperately, so now the file structure deviates for these tests from the rest. This might be useful to know for if you want to change some things in the future (this will be the last couple of tests before we finish the project).

CasProgram:
> The tests work correctly for me. I would suggest changing the the folder name "admin_overview_page" to "general", such as we did with the "for-teacher_page/general", because these are the tests for the buttons of the actual "admin_page".

Good point!

Felienne:
I have not been able to make this one work

Issue #1431, opened 2021-11-30
[UI] When running the same code twice, both the green "good job" modal and confetti button should no longer appear


johannadekker:
When you run code, you get a nice popup with a confetti icon:

<img width="662" alt="image" src="https://user-images.githubusercontent.com/1003685/205442083-a8fc17f5-84f8-499e-ba98-9683325117fb.png">

But if you run the exact same code again, it also shows the bar and the icon:

<img width="640" alt="image" src="https://user-images.githubusercontent.com/1003685/205442098-23fcf76d-dced-45df-84e5-74de25648080.png">

We want to prevent this second showing. This would require a variable at the front end called something like `last_code` and compare that with the current code. Should not be super hard to fix I think!

PR #3853, opened 2022-12-12
[UI] when running the same code twice, green "good job" modal and confetti button no longer appear #1431

jl200589:
**Description**

* **Closes** [UI] When running the same code twice, both the green "good job" modal and confetti button should no longer appear #1431
* Only shows the modal and button on first run/ if the code has changed

**How to test**

* Go to http://localhost:5000/hedy/2
* In the introduction exercise, click 'Run code' button. There should be good job modal and confetti
* Click 'Run code' button again with no changes. There should be no good job modal and confetti
* Make a change to hello world code and rerun. The good job modal and confetti should be back 🎉


static/js/app.ts
```diff
@@ -1127,7 +1129,10 @@ export function runPythonProgram(this: any, code: string, hasTurtle: boolean, ha
     }
     if (!hasWarnings) {
       if (debug == null) {
-        showSuccesMessage();
+        if (code !== last_code) {
+          showSuccesMessage();
+          last_code = code;
+        }
```
TiBiBa, line 1135: Maybe we can combine these into one statement?
```
if (!hasWarnings && code !== last_code) {
    showSuccesMessage();
}
```

TiBiBa (changes requested):
One code suggestion, other than that it works like a charm!

TiBiBa (approved):

PR #3854, opened 2022-12-12
[UI] Allow teachers to customize adventure ordering

### LINKED ISSUE #3541: [UI idea] Allow teachers to customize adventures ordering
**Idea incl level**
In #3540 we start allowing to customize the ordering of adventures in each on the levels, but this is hard coded and we should allow teachers to reorder this, and also reorder their own adventures, and choose which one appears in which level.

I was thinking about something like this (in the customize class menu):
adventures

You could drag and drop the adventure around and that would reorder it, a button to delete it, a menu to add new adventures, etc.

jpelay:
**Description**

This is the WIP towards #3541

Things to do:
* [x] Fix the customization page to allow teachers to reorder the adventures.
* [x] Change the Database design to save the change
 

**Fixes #3541**

**How to test**

**Checklist**
  

jpelay:
Very early stage prototype, but I think this is the hardest part after changing the DB.

Peek 2022-12-12 11-48

jpelay:
Advances :D 
https://user-images.githubusercontent.com/45865185/210259027-f2778c52-58c1-4392-b0b2-928806455ccd.mp4

jpelay:
So this is done, but the Cypress tests will fail. I'm not sure how to change them, but @Felienne can begin testing!

Felienne:
Ah well done @jpelay!! This is a lot better!!

Some small things to be fixed (can also be done in a separate issue/PR since this is already so big)

* Now that we have the new layout, I think we can also move quiz and puzzle into the "dragger" instead of a tickbox below?
* We need to change the text for the enable/disable buttons (I can do that)
* If there are no adventures available, can we remove the drowdown? And empty one is a bit weird. We can even consider a button there with "add you own"

jpelay:
> We now have 2 places for all levels (in addition to the dragger)

I will make separate issues for this! Also, there was a button that is necessary still, but haven't been able to make it to work the "Reset adventures" button, it will reset the adventures to the defaults in "hedy_content"

Felienne (approved):

Felienne:
> I will make separate issues for this! Also, there was a button that is necessary still, but haven't been able to make it to work the "Reset adventures" button, it will reset the adventures to the defaults in "hedy_content"

Hi @jpelay! Did we ever make this issue? I can't seem to find it (I was thinking about it because I just opened an issue for some other small improvements to the customization page, see: #4084 and #4085

> * We can even consider a button there with "add you own"

Hahahaha, this, more or less, is #4084.

PR #3861, opened 2022-12-13
[BACKEND] Hash parsed programs to prevent duplicate parsing programs

TiBiBa:
**Description**
In this PR we hash programs after parsing and use a look-up table to check if the program is already parsed previously. This reduces a lot of duplicate server load by parsing identical programs. We would have to keep track of the server to make sure the memory doesn't overflow by this caching, otherwise a maximum size might be nice.

**Fixes**
No related issue.

**How to test**
Run several programs, add some debug prints to `PARSED_PROGRAMS` to verify that only new programs are added to the dictionary.


Felienne:
> We might be able to improve this by adding a frequency value to each hashed program. This way we can limit the amount of cached programs to the x most ran ones (since server start it is). Preventing the memory from getting overflown with an unknown amount of programs, but popping a random least frequent program if a new one is parsed and storing the new one. Giving it a structure like this:
> 
> ```
> PARSED_PROGRAMS = 
> {
>     '123456': {'result': program, 'frequency': 3},
>     '654321': {'result': program, 'frequency': 5}
> }
> ```

Yes that does sounds like a plan, because I fear otherwise the memory will be full very soon.

TiBiBa:
> > We might be able to improve this by adding a frequency value to each hashed program. This way we can limit the amount of cached programs to the x most ran ones (since server start it is). Preventing the memory from getting overflown with an unknown amount of programs, but popping a random least frequent program if a new one is parsed and storing the new one. Giving it a structure like this:
> > ```
> > PARSED_PROGRAMS = 
> > {
> >     '123456': {'result': program, 'frequency': 3},
> >     '654321': {'result': program, 'frequency': 5}
> > }
> > ```
> 
> Yes that does sounds like a plan, because I fear otherwise the memory will be full very soon.

I just improved the implementation to only keep the 500 most frequent programs cached. I still think can be a huge improvement on the server load, maybe even upping the threshold to 1000 as the memory won't be the biggest issue.

app.py
```diff
@@ -541,11 +544,31 @@ def remove_file(response):
...
+    global PARSED_PROGRAMS
+    PARSED_PROGRAMS[uniq_hash]['frequency'] += 1
+    return PARSED_PROGRAMS.get(uniq_hash).get('result')
+
+
+def add_hash_parse(uniq_hash, result):
+    global PARSED_PROGRAMS
+    if len(PARSED_PROGRAMS) >= 500:
+        min_freq = min(PARSED_PROGRAMS, key=PARSED_PROGRAMS.get('frequency'))
+        min_keys = [key for key in PARSED_PROGRAMS if PARSED_PROGRAMS['frequency'] == min_freq]
```
Felienne, line 557: I fear this too will consume a lot of time (to loop over the whole list), maybe we can consider a second dictionary from freq -> hashes?

TiBiBa:
@rix0rrr will look into your comments this week! There should indeed already be some existing LFU cache Python library.

Felienne:
Let's close this one. It is far behind and lacking the analysis described above, we are not even sure it will help and with all the other improvements we have made (mainly to /explore) performance is now nicely in check even with a lot of traffic!

TiBiBa:
Sorry completely forgot about this one! I agree that it will not influence performance much. Merely from a server load and "this can be improved" perspective it was a nice attempt, but not something for the end-user.

Felienne:
No problemo! We are cleaning some PRs where there is not a clear path forward to keep a better overview of what needs to be looked at.

Issue #3377, opened 2022-10-04
[CHORE] Run code formatter on PRs


Felienne:
We can probably do that this way: 

https://github.com/marketplace/actions/autopep8 (earlier version of this issue suggested black but I think PEP8 will do just fine!)

Felienne:
In principle, I think this is a great first issue since it does not require knowledge of our code base and will be easy, given that you have done such a thing before!

mennoliefstingh:
All right never mind, I've done some further tests and even after formatting the code base with both autopep8 and black, my linter (flake8) still reports a lot of warnings and non pep-compliancy.  Autopep8 is a lot less strict and e.g. doesn't split long lists over multiple lines like Black does, meaning some lines remain extremely long. Black makes the code look a lot better in my opinion, but still doesn't solve a lot of issues like the following:

- Unused imports
- Unused local variables
- Lines that are too long and cannot be split by black/autopep8, usually because they're comments
- 'Bare' exceptions (except: instead of except ValueError: for example)

Specifically in the tests:
- Lambda expressions that should be regular functions (def:)
- f-strings without placeholders in the tests (so f"blablabla" instead of "blablabla",  maybe there's a reason for this)

Don't shoot the messenger as these are flake8's words and not mine, but making the code base pep-compliant and subsequently enforcing style through git hooks would require a lot more work than just running a formatter and writing a yaml file. This basically leaves two options: either reformatting the repo with Black, run flake8 and manually fix non-compliant code or just saying 'never mind' and stick with the codebase as it is now.

bspans:
Was wondering if this issue is still a problem, or if I can help with this. In the afore mentioned issue #3410 It is said that there should be two pull requests, but i'm not sure on what needs to be done at this moment.

craigmulligan:
Hey :wave: 

I have a little experience using both black & flake8. While black is PEP8 compliant it is conservative in what it formats and won't change anything that could break the program, which means it won't automatically fix things like unused imports. So you'll likely have to use both a linter and a formatter to cover the issues mentioned in this thread. 

The two tools work fairly well together, black has some documentation on that [here](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8).

* I had a quick look at the repo a first pass of flake8 there are *4479* PEP8 violations.
* After running black there were *825* PEP8 violations left. 
[...]

PR #3862, opened 2022-12-13
[CHORE] Fix setup.cfg by extending it with specific autopep8 config

mennoliefstingh:
**Description**

This edits the setup.cfg to prevent autopep8 from trying to inherit per-file-ignores from the flake8 config as autopep8 does not support this (yet, anyone up for a coding challenge?) 

It also updates a ton of files to be pep8-compliant again as these slipped through since the last time I ran a formatter. 

Flake8 also found a test that didn't do anything in test_level_07.py; please verify if I added the assert correctly. The unit tests are still running but I'm assuming it should be all right. 

Related to #3377

**How to test**

run autopep8 on the entire repo with this setup.cfg, you'll see that it skips the tests repo


mennoliefstingh:
actually it makes more sense to include the commits relating to pep8 compliance here; we added a few files that weren't compliant since I last ran the formatters. There's also a broken test that doesn't do anything in tests/test_levels/test_level_07.py, which is missing an assert. The next commit will be pep8 compliance, the commit after that adds the assert. Just a minute!

Felienne (approved):

Issue #3863, opened 2022-12-14
[BUG] Unit test with no assertion.


RikvanSchaick:
**Description of the bug**
The last unit test of level 7 does not have an assertion, so will not be tested.

**Regarding unit test**
`test_if_pressed_repeat()`

PR #3864, opened 2022-12-14
[FIX] Fixing unit test (level 7)

RikvanSchaick:
Menno noted that the last unit test of level 7 did not have an assert, which let to a test that will not be used. 

To fix this, I added the missing assertion in unit test 'test_if_pressed_repeat' and made sure the unit test does not fail. This PR therefor fixes issue #3863.


mennoliefstingh (approved):
Looks good! Thanks for fixing this on such short notice :)

mennoliefstingh:
pre-commit.ci fails, but that is because a portion of the code is still not pep8-compliant. PR #3862 fixes this, so we can merge this regardless

PR #3883, opened 2022-12-20
[FIX] Remove startup translation of keywords in YAMLs

rix0rrr:
The eager keyword translation upon startup adds about ~1 minute to the startup time of the server, and complicates the code.

Replace with on-demand translation. After timing, this seems to add 10-15ms to the very worst page load: loading all adventures. That seems like an acceptable price to pay to simplify the code a lot, improve startup time, and improve memory usage as well since we don't have to keep multiple copies of the YAML data in memory. Plus, reading afresh from the `YamlFile` object on every page load makes the hot reloading feature work.

Reduction of memory usage may lead to us being able to run more processes in parallel on the server, which will be good for responsiveness.

**How to test**

Everything should behave exactly as before.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
   - repo: https://github.com/pre-commit/mirrors-autopep8
-    rev: v1.4.4
+    rev: v2.0.1
     hooks:
     -   id: autopep8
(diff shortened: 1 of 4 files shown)


Felienne (approved):
Seems to not break things!

PR #3886, opened 2022-12-21
[REFACTOR] Reduce memory footprint by not caching YAML data in memory

rix0rrr:
`YamlFile` used to cache all YAML data for all languages in memory. All YAML files together are 15MB on disk, and when loaded into memory take about 38MB (as reported by the Python `tracemalloc` module). Taking into account we run 5 server processes, this adds 5 * ~40MB = ~200MB of memory usage to the server.

In this PR, turn off caching to free up more memory on the machine, which we can use to run more server processes in parallel.

I've made sure to not negatively impact runtime performance:

- We still do per-request caching (just not eternal global caching anymore).
- We still do pickling, and in fact store the pickled files in the user's TEMP dir so that we can turn feature on even on developer machines.
- Add logging to measure how much time impact this all has.

After this change, loading all YAMLs for a page load adds a second to a page load if the pickles don't exist, and a millisecond if they do, which seems acceptable.

Hot reloading still works: if the YAML file is changed, the original file is re-read instead of the pickled one (as determined by their relative time stamps).

**How to test**

Everything works exactly as before.


content/adventures/nl.yaml
```diff
@@ -7,7 +7,7 @@ adventures:
             1:
                 story_text: |
                     ## Welkom bij Hedy!
-                    In Level 1 kun je de commando's `{print}`, `{ask}` en `{echo}` gebruiken.
+                    In Level poep kun je de commando's `{print}`, `{ask}` en `{echo}` gebruiken.
```
Felienne, line 10:
```suggestion
                    In Level 1 kun je de commando's `{print}`, `{ask}` en `{echo}` gebruiken.
```

Felienne (approved):

PR #3888, opened 2022-12-21
[BUG FIX] Quiz and parsons now use the correct keyword language when toggled

### LINKED ISSUE #3884: [BUG] Keyword language in quiz does not does not when not logged in
**Describe the bug**
Just using the switcher does not make the quiz keywords Dutch (I am not even sure this ever worked...):
image

Being logged in does work so it is probably something small:

image

TiBiBa:
**Description**
We fix a bug where the keyword language wouldn't be correct for both the parsons and the quiz when using the UI toggle and we "force" the keyword language. This is due to the structure were we GET each question or exercise separately and depend on the `g.keyword_lang` value. However, this is incorrect when forcing the language! Therefore we now first check if there is a forced parameter in the url, if so: change the url and get the correct question / exercise.

**Fixes**
This PR fixes #3884.

**How to test**
Make sure that both the parsons and quiz work when not logged in and using the keyword language toggle.


Felienne (approved):
Check! 
image

PR #3889, opened 2022-12-21
[REFACTOR] Reduce memory usage by not querying Babel

rix0rrr:
The Python library `babel` wraps `gettext` and a locale database.

However, when the information about a locale is accessed, they choose to cache that information in memory indefinitely. We load information about all locales at startup, which fills the cache with data we'll never read again.

This accounts for at least 40-70MB of memory usage per server process.

Instead, we will read the bits of information we need from the Babel locale database in a build tool, and write it to a file; we will then read that file in the server process.

In the future, the `Locale` features from babel should not be used anymore.

**How to test**

Everything should behave exactly as before


Felienne:
> In the future, the `Locale` features from babel should not be used anymore.

Just want to clarify that using `force_locale` is still ok, like here:

https://github.com/hedyorg/hedy/blob/d1eb428bada03a115b058cd1bb90caa4b0f3b1ee/website/auth.py#L363

Felienne (approved):

PR #3912, opened 2023-01-04
[CHORE] Design: move teacher nav to top bar

rix0rrr:
Move the teacher navigation out of the profile dropdown, into the top bar, where it will be more prominently visible.

Also make the teacher's manual publicly visible, not only to teachers, so that teachers (finding the link via the colophon) can have a preview at the learning material that's available to them, hopefully enticing them to register.

## After

image


Felienne (approved):
Check, works on Dutch now too!

image

PR #3917, opened 2023-01-04
[CONTENT] Replace copied strings with newer content

Felienne:
#3844 replaced some content in English, this PR does the same to other (untranslated) languages


Felienne:
Superceded by #3918

PR #3920, opened 2023-01-06
[feat] Create page to easily find commands of a program

Felienne:
I often give my students the assignment to create a program with commands such and such and this will help me grade (I will make it nicer looking, one day!)


Felienne:
I am merging this since it will ease my grading a lot, I will make an issue to improve it later!

PR #3926, opened 2023-01-11
[FIX] Design: tabs are not scrolling horizontally

rix0rrr:
When the browser window was narrow, the adventure taps were keeping the scrollable area wide for some reason, causing content (and buttons) to fall off the side of the screen.

This was very hard to diagnose, but ultimately was due to having `class="flex"` (vertical flexbox) on parents `<div>`s in the level page.

Flexbox behaves weirdly with sizes on the cross-axis.

I don't quite understand it, but removing the `flex` on the parents seems to fix it: the tab bar now understands its max width properly and will add a scroll bar when necessary.

**How to test**

Go to the level 5 page, make the browser window narrower.


Felienne (approved):
Seems to fix the problem!

image

PR #3940, opened 2023-01-16
[FIX] Design: next level block overlay hides button

rix0rrr:
If the "next level" button is disabled because a teacher chooses to do so, the popup informing children was hiding the "Reply" button.

Move it down over the button itself so the blocks don't interfere.

Also make the Login screen a bit wider, the word "Gebruikersnaam" is too long for the current box.

## Before

image

image

## After

image

image


Felienne (approved):
Thanks, works!

image

PR #3943, opened 2023-01-16
[UI] Enable alternative error captions for a testing group

hbycht:
_PR DRAFT of the Köln students :)_

**Description**

_At every `/parse` request that leads to a `HedyExeption`, the users receive an alternative error caption, if they are in the test group **AND** the used language is a supported test language.
If the user have no `test_group` field in their DB entry, it will be created. (1/3 users with supported languages will be selected.)_

**Fixes _#3191_**

**How to test**

* Login with a user of your choice in a local env.
* Change language to _En_ or _De_ (if it isn't already).
* Go to Hedy application and Click on "Run code".
* You should now find a field `test_group` within your user entry with `True` or `False` in `./dev_database.json`.
* If `test_group` is `False`, set it to `True`.
* In the code section of the Hedy app just type `echo` without anything to provoke a error.
* You now should see the alternative error caption `B_TEST: We can't run your program.` (Still a template.)
* If it doesn't appears, clear your browser cache and Hedy cookies. 

_We will find some better alternatives for the transpile error caption in the next steps and check how to archive and analyze the test data._


Felienne:
Thanks for your first PR @hbycht! As you can see, stuff moves fast here and you already have some conflicts on this PR. Let me know if you need help resolving them!

Felienne:
Ow for static/js/appbundle.js and static/js/appbundle.js.map you can just take either version, because these are generated files (see docs)

app.py
```diff
@@ -368,6 +369,16 @@ def parse():
     username = current_user()['username'] or None
     exception = None
 
+    user = DATABASE.user_by_username(username)
```
TiBiBa, line 372: This way we make an additional database call on each `/parse` request. It would be nice to store the `test_group` value on the user (in the `session` for example). If it is already there, we don't have to check it each time. As we parse quite a lot this will introduce a lot of db calls.

app.py
```diff
@@ -368,6 +369,16 @@ def parse():
...
+
+    # If user has no  test_group field, create and save it into database
+    if "test_group" not in user.keys():
+        test_group = alternative_error_messages.select_users_test_group()
+        DATABASE.update_user(username, {"test_group": test_group})
```
TiBiBa, line 377: I'm a bit hesitant to update the user table. What if we decide to no longer use this option, or want to change the group distribution? Start over? This user table is the core of our app and is hard to clean-up or alter later on. It might be nice to use an additional table that stores this data. Simply like a `username - test group` pair. This also allows us to increase the amount of groups (for example) later on.

website/auth.py
```diff
@@ -94,7 +94,7 @@ def password_hash(password, salt):
 # You can remove the current user from the Flask session with the `forget_current_user`.
 def remember_current_user(db_user):
     session["user-ttl"] = times() + 5 * 60
-    session["user"] = pick(db_user, "username", "email", "is_teacher")
+    session["user"] = pick(db_user, "username", "email", "is_teacher", "test_group")
```
TiBiBa, line 97: We already store it in the session! This means we only would have to remove the db call (if the value exists).

hbycht:
> Ow for static/js/appbundle.js and static/js/appbundle.js.map you can just take either version, because these are generated files (see docs)

@Felienne So we can just add them both to our gitignore file and untrack them from our repo, right?

app.py
hbycht: First of all @TiBiBa, thanks a lot for your helpful suggestions!

We have discussed a few ways to fix the problem: 
1. we can link the test_group field to the session user only, so no real database query is needed.
2. we move the check if the test_group field exists to the /login endpoint (login via login page as well as login via session cookie).

If we decide to persist the test_group field in the database in its own table, can you give us some guidance here on how to create our own table and attach it to the user? A link to a doc or tutorial would be perfect as we are not that familiar with this yet.

Felienne:
> > Ow for static/js/appbundle.js and static/js/appbundle.js.map you can just take either version, because these are generated files (see docs)
> 
> @Felienne So we can just add them both to our gitignore file and untrack them from our repo, right?

Yes that does sound reasonable doesn't it? So reasonable that maybe there is a reason that is not the case :)? 

Our contributing guise explains why these files still live in the repo despite being generated:

https://github.com/hedyorg/hedy/blob/main/CONTRIBUTING.md#working-on-the-web-front-end-in-typescriptjavascript

app.py
TiBiBa: I would like this to be checked at the `/login` endpoint. Also automatically assigning a user to a group if it isn't in one yet. I would be happy to help! Please send me a message in Discord (if your already in there?) to get in touch, I can give you some pointers on adding a table to our database.

hbycht:
We just have to change the "alternative error heading" to an actual good alternative. 
What do you think in which languages we should translate it?
EN, DE, ES, NL?

Felienne:
> We just have to change the "alternative error heading" to an actual good alternative. What do you think in which languages we should translate it? EN, DE, ES, NL?

The team usually maintains Dutch and English only, but of course if you know German, adding that is really nice too!

hbycht:
We now added real alternative error headings in english, german and dutch.

Maybe you can check the dutch translation by your own (path is: `/content/client-messages/nl_test.yaml`).
In english the wording is: "Something went wrong! Try again 🙂"

We mark the draft as "Ready for review"!

alternative_error_messages.py
```diff
@@ -0,0 +1,44 @@
+from random import random as random
+
+from flask import session
+from hedy_content import ALTERNATIVE_ERROR_LANGUAGES
+from website.auth import (current_user, update_test_group_in_session)
```
TiBiBa, line 5: Not sure the `()` are needed, could you remove them to be consisted with our code base?

alternative_error_messages.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def is_supported_lang(lang):
+    return lang in ALTERNATIVE_ERROR_LANGUAGES
+
+
+def select_users_test_group():
+    lang = session.get('lang')
+    if is_supported_lang(lang):
+        if random() > 0.66:
```
TiBiBa, line 25: You we replace the `0.66` with some constant and an explanation? This value is unclear and requires some comment to explain.

alternative_error_messages.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Creation of test_group entry if it doesn't exist in session or db
+
+
+def update_users_test_group(db):
+    user = current_user()
+    # Check if there is a logged-in user
+    if user['username']:
+        # Check if test_group field exists or is filled properly in session
+        if 'test_group' not in session.keys() or session.get('test_group') is None:
```
TiBiBa, line 37: You can just use `if not session.get('test_group'):`

app.py
```diff
@@ -197,6 +198,9 @@ def initialize_session():
     utils.session_id()
     login_user_from_token_cookie()
 
+    # Creation of test_group field if it doesn't exist in session or db user
+    alternative_error_messages.update_users_test_group(g.db)
```
TiBiBa, line 202: This is called before each request. Can't we just call this function once in the `/login` route?

app.py
```diff
@@ -437,6 +444,7 @@ def parse():
         'code': code,
         'server_error': response.get('Error'),
         'exception': get_class_name(exception),
+        'alternative_error_message': alt_err_msgs,
```
TiBiBa, line 447: What happens if `alt_err_msgs` is empty? It is unclear to me what value this variable can contain.

app.py
```diff
@@ -1658,12 +1667,23 @@ def client_messages():
...
     response = make_response(render_template(
         "client_messages.js", error_messages=json.dumps(d)))
 
-    if not is_debug_mode():
+    if is_debug_mode():
+        response.cache_control.max_age = 0
+    else:
         # Cache for longer when not developing
         response.cache_control.max_age = 60 * 60  # Seconds
+
```
TiBiBa, line 1686: I think @rix0rrr is also working on some re-written of the `client-messages` structure. Might have to check with him if this is still suitable with the new approach.

static/js/app.ts
```diff
@@ -1709,7 +1710,7 @@ $("#output").animate({ scrollTop: $(document).height() }, "slow");
 export function modalStepOne(level: number){
   createModal(level);
   let modal_editor = $('#modal-editor');
-  initializeModalEditor(modal_editor);
+  initializeModalEditor(modal_editor);
```
TiBiBa, line 1713: Please remove this empty space again.

utils.py
```diff
@@ -172,6 +190,20 @@ def set_debug_mode(debug_mode):
...
+def debug_only(f):
+    @wraps(f)
+    def wrapped(**kwargs):
+        if not current_app.debug:
+            abort(404)
+
+        return f(**kwargs)
+
+    return wrapped
+
```
TiBiBa, line 206: I really like this approach, but is this related to the issue of your PR?

website/database.py
```diff
@@ -730,3 +732,17 @@ def to_year_week(self, d):
     def get_username_role(self, username):
         role = "teacher" if USERS.get({"username": username}).get("teacher_request") is True else "student"
         return role
+
+    def test_group_by_username(self, username):
+        test_group = TEST_GROUP.get({"username": username})
+        if test_group:
+            return test_group["test_group"]
+        return None
```
TiBiBa, line 740: This can be simplified with `return test_group.get('test_group')`.

website/database.py
```diff
@@ -730,3 +732,17 @@ def to_year_week(self, d):
...
+
+    def add_test_group_to_username(self, username, test_group: bool):
+        user_test_group = self.test_group_by_username(username)
+        if not user_test_group:
+            user_test_group = {"username": username, "test_group": test_group}
+            TEST_GROUP.create(user_test_group)
+            return True
+        return False
```
TiBiBa, line 748: Do we need this return statement?

TiBiBa (changes requested):
I left quite some comments, don't worry they are mostly really small issues. It might be a good idea to schedule a meeting through Discord? It is still a bit unclear to be what the main goal is of this PR. Would be nice to be brought up-to-date! Let me know!

alternative_error_messages.py
hbycht: done!

alternative_error_messages.py
hbycht: done. 
(It's the percentage of users that are in the test group.)

alternative_error_messages.py
hbycht: done.

app.py
hbycht: We want to maximize the percentage of users which are used for the test group selection. So if we place it in the /login route, all the users with active session don't be available.

app.py
hbycht: As you can see in line 380, `app.py`:
`# Check if alternative error messages are available in this parse`
`alt_err_msgs = alternative_error_messages.is_available()`

Always a boolean value is returned and we have to log it to analyze it later.

static/js/app.ts
hbycht: done.

utils.py
hbycht: deleted.

website/database.py
hbycht: done.

website/database.py
hbycht: deleted.

app.py
rix0rrr: Good catch, thank @TiBiBa! It would be most helpful to me if we can put the new error messages in the same file, but give them different keys.

So, instead of:

```
client-messages/nl.yaml
   Transpile_error: "We konden je code niet goed lezen."

client-messages/nl.yaml
   Transpile_error: "Dat ging fout. Probeer het opnieuw 🙂"
```

I'd rather see:

```
client-messages/nl.yaml
   Transpile_error: "We konden je code niet goed lezen."
   NEW_Transpile_error: "Dat ging fout. Probeer het opnieuw 🙂"
```

Is there a reason not to do it that way?

(Is the reason Weblate?)

app.py
aspargillus: Doing that only on `/login` will not cover all the sessions handled by `login_user_from_token_cookie()`. I think this really does the minimum required to reliably get the test group flag.

Felienne:
Let's close this for now, we sadly don't have someone to push it forwards

Issue #3865, opened 2022-12-14
[BUG] Pygame multiple keyboard input checks


sjensen19:
**Describe the bug**
`if x is pressed` only works once. Mulitple keyboard input checks result in an error

**Paste the Hedy code & level**
```
repeat 3 times if x is pressed forward 15
repeat 3 times if y is pressed forward 15
repeat 3 times if z is pressed forward 15
```
Level 7

**Add a screenshot (optional)**
![image](https://user-images.githubusercontent.com/55716630/207572245-ff867333-cf0c-4cc1-8543-b069bbbf2516.png)

PR #3944, opened 2023-01-16
[BUG] Correctly handle multiple key presses

Felienne:
Work towards #3865


Felienne:
@ToniSkulj can pick this up quicker and better :)

PR #3948, opened 2023-01-17
[FIX] - Comparing int and float gives wrong answer

### LINKED ISSUE #3726: [BUG] Comparing int and float gives wrong answer
**Describe the bug**
Hedy gives the wrong answer when comparing an integer and a float

Level 14:
var = ask 'What is your number?'
if var > 15.0
    print 'Above 15.0'
else
    print 'Below'

Try this code with var = 16  (=> Below) and var = 16.0 (=> Above 15.0)

noelgolding:
## Description

* `convert_numerals` function now returns an int, float or string accordingly
* no longer necessary to pad with leading zeros 

Fixes #3726

### Automated Unit test:

```
python -m pytest tests/test_level/test_level_14.py -k 'test_greater_than_with_int_and_float or test_not_greater_than_with_int_and_float'
```

### Manual UI test:

1. reproduce bad behavior
2. apply fix
3. confirm corrected behavior

#### To reproduce the broken behavior, before this fix is applied:

Navigate to Level 14, enter and run the following code:

```
age = 16
if age > 15.0
    print 'You are older than me!'
else
    print 'You are younger than me!'

age = 7.0
if age > 15
    print 'You are older than me!'
else
    print 'You are younger than me!'
```

expected output

```
You are older than me!
You are younger than me!
```

actual output &#9785;

```
You are younger than me!
You are older than me!
```

#### To confirm fix after it has been applied:

Execute the reproduction steps above, and confirm that now the expected output matches the actual output 😃


semantics/SEMANTICS.md
```diff
@@ -1,4051 +1,4051 @@
...
-    <td>(if | for | while) (boolean) (newline without ':' first) </td>
-    <td>Invalid Exception</td>
-  </tr>
-  <tr>
-    <td>(any lines with code without `if`) elif (boolean):</td>
-    <td>Invalid Exception</td>
-  </tr>
-</tbody>
-</table>
+# Semantics
```
noelgolding, line 1: not sure why it shows as the entire file has changed.  The change here should just be the removal of `.zfill(100)`

Felienne: This probably has to do with line endings, but it is not such a relevant file to keep the git blame history of, don't worry!

Felienne:
Thanks for the fix Neil and removing the error prone hack that I though up!! Somehow, an Cypress test now fails, I am not entirely sure why. Do you want to take a look or shall I?

Felienne:
> Thanks for the fix Neil and removing the error prone hack that I though up!! Somehow, an Cypress test now fails, I am not entirely sure why. Do you want to take a look or shall I?

Never mind, I think I fixed it by rerunning the ts compiler!

Felienne (approved):

Felienne:
Thanks for the fix @noelgolding!!

PR #3952, opened 2023-01-18
[FIX] Allow for key pressed using variables

### LINKED ISSUE #3957: [LANGUAGE idea] Allow variables in pressed
This code currently does not work, since variables are not allowed in `if` `pressed` (names with more than 1 letter are assumed to be buttons without checking)

Level 5:
```
letters is a, b, c, d, e, f
letter is letters at random
print 'Press ' letter
if letter is pressed print 'Yes' else print 'No'
```

Result:
image

Felienne:
Allow for key pressed using variables.

Fixes #3957


Felienne:
Works at the back end, but 2 issues remain:

* Two if's in a row do not work (this is a known and separate issue)
* The front-end still shows "button press" I guess that somewhere also depends on the length of the variable!

Felienne:
Hi @ToniSkulj!

This fixes a missing feature (using variables for letters) but also removes some weird duplication that I think can be removed, but I am not 100% sure.

Also, I think the lines

```
            if event.unicode == 'x':
```

could be merged into one with and?

```
          if event.type == pygame.KEYDOWN and event.unicode == 'x':
```

simplifying the code even more but I wanted to hear your opinion before starting that large operation!

Felienne:
Merging this to enable the new adventure in level 9, will update the rest of the PyGame code/issues later!

ToniSkulj:
> Hi @ToniSkulj!
> 
> This fixes a missing feature (using variables for letters) but also removes some weird duplication that I think can be removed, but I am not 100% sure.
> 
> Also, I think the lines
> 
> ```
>           if event.type == pygame.KEYDOWN:
>             if event.unicode == 'x':
> ```
> 
> could be merged into one with and?
> 
> ```
>           if event.type == pygame.KEYDOWN and event.unicode == 'x':
> ```
> 
> simplifying the code even more but I wanted to hear your opinion before starting that large operation!

Hi @Felienne! Sounds like a great idea, the only downside that I can come up with is that _if event.type == pygame.KEYDOWN and_ will be tested multiple times instead of just once, but this will all be generated code so I don't think it's problematic. It will make the code that generates this more simplified like you say.

let me know if you want any help!

Felienne:
> > Hi @ToniSkulj!
> > This fixes a missing feature (using variables for letters) but also removes some weird duplication that I think can be removed, but I am not 100% sure.
> > Also, I think the lines
> > ```
> >           if event.type == pygame.KEYDOWN:
> >             if event.unicode == 'x':
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > could be merged into one with and?
> > ```
> >           if event.type == pygame.KEYDOWN and event.unicode == 'x':
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > simplifying the code even more but I wanted to hear your opinion before starting that large operation!
> 
> Hi @Felienne! Sounds like a great idea, the only downside that I can come up with is that _if event.type == pygame.KEYDOWN and_ will be tested multiple times instead of just once, but this will all be generated code so I don't think it's problematic. It will make the code that generates this more simplified like you say.
> 
> let me know if you want any help!

I will make an issue, if you have time, feel free to pick it up (I can do it too so I don't necessarily need help, but I'd appreciate it!)

PR #3956, opened 2023-01-19
[CONTENT] Fixing broken slides

jpelay:
**Description**

The slides are broken because of an error in the .yaml. As I side note, we should test this! 
**How to test**

Slides should work :D


Felienne:
Ah thanks for the fix! We validate the yamls with schema, but this does not throw a hard error indeed! I think somewhere we do check placeholders but not in this was. Let's make a note to add a test!

Felienne (approved):

PR #3961, opened 2023-01-20
[CHORE] Remove `window.State` and a lot of other cleanup

rix0rrr:
This PR changes the following:

BACKEND:

- Remove the `adventure_name` parameter, it seems to always be the adventure of the `loaded_program`, which we have already.
- Organize the `templates/` directory a little with subdirectories.
- Remove dynamic loading of `/client_messages.js`. Instead, we use Python to generate a TypeScript source file at build time that contains all string translations we might need (`generate-client-messages.py`). This removes the need for the `level_title` parameter from HTML to JavaScript.
- Put the code of `generate-rules-highlighting.py` into a `main()` function... just because.
- Remove the code for the (deprecated) built-in translator interface.
- Load all scripts in `layout.html`, instead of duplicating the script loading across all different pages.

FRONTEND:

- Remove the global `window.State`. We now have `APP_STATE`, which is still a global but it's read-only, and it contains drastically fewer variables. Most variables should be local to a module (`tabs.ts`/`app.ts`/`teacher.ts`).
- The application doesn't initialize itself automatically on JavaScript bundle load anymore. Instead, there's an `initialize()` function that gets called explicitly, which calls `initializeApp()`/`initializeTabs()` etc in turn.
- Instead of having a global with `unsaved_changes`, have functions for it that attach/remove the `beforeunload` listener.
- Rename `disable_run` -> `askPromptOpen`.
- Rename `ErrorMessages` -> `ClientMessages`.
- Make the showing/hiding of divs in `tabs.ts` calculated off of booleans.
- Remove the code for the (deprecated) built-in translator interface.
- Remove unused fonts.
- Rewrite jQuery handlers for `submit`/`click`/`change` from legacy format to recommended format (`$(...).submit(...)` -> `$(...).on('submit', ...)`.

**How to test**

Everything should work exactly the same. "Should" being the operative word!


static/js/auth.ts
```diff
@@ -82,123 +82,141 @@ export function request_teacher_account() {
...
-        data: convertFormJSON($(this)),
-        contentType: 'application/json; charset=utf-8'
-      }).done (function () {
-        afterLogin({"first_time": true});
-      }).fail (function (response) {
-        modal.alert(response.responseText, 3000, true);
-      });
-});
+export function initializeFormSubmits() {
+  $('form#signup').submit(function(e) {
```
jpelay, line 86: Since we already are changing so much of the code, can we change the `submit` function and instead call `on`? `submit` and similar shorthand event handlers were deprecated since JQuery 3.3.

static/js/state.ts
```diff
@@ -0,0 +1,143 @@
...
+   * to show a debugger,  to load parsons exercises, to initialize a default save name.
+   */
+  level: number;
+
+  /**
+   * Current keyword language
+   *
+   * Written: by every page, on page load.
+   *
+   * Used: set on the Ace editor, and then is used to do some magic that I don't
```
jpelay, line 35: Maybe @TiBiBa can help us here?

TiBiBa: If I recall correctly we did use to set a keyword language on each ace editor. This is from when we supported the dynamic keyword switcher within the editor. Enabling us to support a separate keyword language on the main editor and the adventure editors at the same time. 

The last time I worked on this code we decided to keep the switcher available to the admin. So if that is still the case we should keep it, it might still be used to determine the highlighter language?

tests/cypress/e2e/for-teacher_page/customize_class_page/customize_class.cy.js
```diff
@@ -1,22 +1,28 @@
...
     beforeEach(() => {
       loginForTeacher();
-      createClass();
+      className = ensureClass();
       cy.getBySel('view_class_link').first().click(); // Press on view class button
       cy.getBySel('customize_class_button').click(); // Press customize class button
+
+      // Remove any customizations that already exist to get the class into a predictable state
+      // This always throws up a modal dialog
+      cy.getBySel('remove_customizations_button').click({ force: true });
```
jpelay, line 14: This is actually pretty good! I think we can use this one instead of create class every single time?

jpelay:
Hello! This is a big PR so I'll run through the changes!

1. I like the change with the statically generated client messages, it makes a lot more sense and is quite elegant!
2. Removing the use of `window.State` is a pretty good move. That global variable was used everywhere.
3. The other minor improvements are very helpful too.

Do you plan to further remove APP_STATE? Or leave it as it is?

I'd be inclined to approve this as it is, but I'd like @TiBiBa to chip in and also say his opinions since this is such a big PR!

tests/cypress/e2e/for-teacher_page/customize_class_page/customize_class.cy.js
rix0rrr: Yes, I was trying to save the database a little by not creating something new for every test run.

I've seen that kind of thing done at work and sooner or later the database always gets in trouble after you've run the test 1000s of times 🥸

rix0rrr:
Hey!

> Do you plan to further remove APP_STATE? Or leave it as it is?

My thoughts exactly! I'm already working on more changes to get rid of most of `APP_STATE`, but this change is already growing larger and larger and I want to land some of it to make merge conflicts less likely in the future.

(And to think I only started because I wanted to remove the need for the modal "unsaved changes" popup 😅...)

tests/cypress/e2e/for-teacher_page/customize_class_page/customize_class.cy.js
jpelay: Cool! I'm refactoring the cypress tests and will take this into account!

tests/cypress/e2e/tools/classes/class.js
```diff
@@ -3,22 +3,40 @@ import { goToTeachersPage } from "../navigation/nav";
...
+ * Make sure that at least one class exists
+ *
+ * Create a class if one doesn't exist already.
+ */
+export function ensureClass()
+{
+    const classname = `test class ${Math.random()}`;
+    goToTeachersPage();
+
+    const viewClassLink = cy.getBySel('view_class_link');
```
jpelay, line 30: Hello @rix0rrr this is something I missed yesterday, and just realized now trying to incorporate this on another refactor in #4032 . You can't assign the return value of a Cypress command to variable because they are ran asynchronously. There's more info here: https://docs.cypress.io/guides/core-concepts/variables-and-aliases

jpelay (changes requested):

tests/cypress/e2e/tools/classes/class.js
jpelay: I propose we change it to this, since we only need the value in the function `createClassAndAddStudents`, which is used in create_accounts.cy,js and I think we can just put that code in the `beforeEach` of that particular file as suggested by the cypress documentation:

```
export function ensureClass()
{
    goToTeachersPage();
    if (cy.getBySel('view_class_link').length === 0) {
        createClass();
    }
}
```

rix0rrr: Good call. For now, I turned this into a function that returns a `Promise`, that can be `await`ed in the caller:

```ts
function ensureClass() {
  return new Promise(ok => {
    cy.get('#whatever').then(x => ok(x));
  });
}

beforeEach(async () => {
   await ensureClass();
});
```

Should be good enough for now, while you do the refactor?

static/js/state.ts
Felienne: Yes, we support the feature to translate programs from the front-end, but it is only available to me, to show in demos, kids don't really need it.

tests/cypress/e2e/tools/classes/class.js
jpelay: Yes this is good enough for now!

jpelay (approved):
Hello @rix0rrr I think this is great and good to go now :D Waiting for those further changes you're planning!

PR #3962, opened 2023-01-20
[FIX] Fixing broken customization of adventures

jpelay:

**Description**

There was a bug where class without customizations, adventures wouldn't load correctly. Also there were other edge cases fixed:

* Need to update the adventure creation page.

**Checklist**
  

templates/customize-adventure.html
```diff
@@ -47,7 +47,7 @@ <h3 class="text-center mt-0 mb-4">{{ _('adventure') }}</h3>
                         <label for="agree_public" class="text-sm italic w-full">{{_('adventure_terms')}}</label>
                     </div>
                     {% if class_data %}
-                        <div class="flex flex-col items-center">
+                        <!-- <div class="flex flex-col items-center">
```
rix0rrr, line 50: This is commented out. Is this left in for debugging purposes?

If this was intentional, then wouldn't we better delete the code?

(Approving this PR regardless since it is a bugfix and it is important to get out).

rix0rrr (approved):

templates/customize-adventure.html
jpelay: It's a feature that I plan to reenable today!

PR #3964, opened 2023-01-21
[UI] Fix Overflow For Pygame Button

### LINKED ISSUE #3963: [UI idea] Fix Overflow For Pygame Buttons
Currently when using the buttons feature the **x scroll bar** will always be visible:
image

The scroll bar should only be visible when the buttons do not fit inside the container anymore.

ToniSkulj:
Fixes #3963,
Now the scroll bar will only be visible when the buttons do not fit anymore.

when it does fit:
image
when it does not fit:
image


Felienne (approved):
Thanks for the fix @ToniSkulj!!

PR #3965, opened 2023-01-22
[FIX] Customize class: back button doesn't work

rix0rrr:
A variable named `isDirty` was being referenced that was not defined.

Also in this PR:

* move the page initialization code of the Customize Class page from the HTML template into the TypeScript code, so all the client side code is together.
* Remove the `id="remove"` attribute from the close buttons. Not all HTML elements need to have an `id` attribute, but if it is used it should be unique (so we cannot use a static id in a loop). It did not appear to be used however, so this shouldn't make a difference.
* Found a couple of duplicates of the HTML that defines the draggable tabs occurs 3 times in the code base, so add a comment that indicates where the duplicates are so they are easier to find.


jpelay:
Thank you for the improvements! The code is much clearer and better now!

The remove customization button now is broken! I can fix that and upload the fix if you agree!

jpelay:
@rix0rrr Hi Rico! I made a few changes to the code, and I want to hear your input is possible!

* I removed the object you created and instead added those variables to window.State, because those variables are used in the `remove_customizations` function and will be used in the `reset_customizations` function.
* To do this I also added a few interfaces and modify the function you created.

Now, both the Go back and Remove customization buttons work!

Felienne (approved):

rix0rrr:
> I removed the object you created and instead added those variables to window.State, because those variables are used in the remove_customizations function and will be used in the reset_customizations function.

Please don't add anything to `window.State` anymore. Global variables are a bad idea, and I'm specifically preparing a PR to remove the global state altogether (https://github.com/hedyorg/hedy/pull/3961).

The only reason to move variables to a "global" place, which can be read and modified from anywhere, is because we apparently don't know which code needs access to those variables -- which, if true, is a bad reason! We would be better off refactoring our code so that the state can be local (to a module, class, or function).

It's unfortunate that this got merged in the mean time... but I will take it into account for my refactor when I get back to working on #3961. (If we could get some Cypress tests on the functionality in the mean time, I would know if I accidentally broke something!)

-------

One of the reasons for storing state in the global object is probably so that we can access it both from JavaScript and in event handlers that are written directly on HTML tags (`<button onclick="callFn(globalVar)">`). We shouldn't be doing that anymore, and if you see it currently being done it was a historical mistake.

The only things we should be doing are:

* `<button onclick="hedyApp.someHandler()">`, without variables. The `someHandler()` function can look up the variables purely inside the JavaScript.
* Or, preferred: `<button id="my_button">` and then in the JavaScript `$('#my_button').on('click', () => { ... })`, to define the event handler fully in JavaScript. We also don't need to export any functions if we do that.

PR #3972, opened 2023-01-25
[TESTING] Adding new tests to customize class page

jpelay:
**Description**

Recently, the customize class page has gone through a lot of changes, and ever since a handful of bugs emerged. In order to prevent this from happening in the future, this PR adds a handful of tests for the following behaviors:

* Clicking the 'go back' button.
* Clicking the 'save' and 'remove customizations' buttons
* Reordering the adventures through drags and drops.
* Deleting an adventure and adding it again.
* Switching through levels.

Also, minor changes:

* A debug print was removed.
* Added ids to some buttons in order to query them more easily in the tests.


tests/cypress/e2e/for-teacher_page/customize_class_page/go_back_to_class_button.cy.js
```diff
@@ -0,0 +1,25 @@
+import {loginForTeacher} from '../../tools/login/login.js'
+import { createClass } from "../../tools/classes/class";
+
+describe('Back to class button', () => {
+  it('passes', () => {
```
rix0rrr, line 5: These strings are the test name. It will help the reader/debugger if they describe the behavior the test is looking for.

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
```diff
@@ -0,0 +1,77 @@
+import {loginForTeacher} from '../../tools/login/login.js'
+import { createClass } from "../../tools/classes/class";
+
+describe('Back to class button', () => {
+  it('passes', () => {
+    loginForTeacher();
+    cy.wait(500);
```
rix0rrr, line 7: Do we needs this `wait` ? This always looks a bit scary to me, because if we're ever on a bit slower computer in which this takes 600ms, the test fails. Is it necessary?

Cypress best practices have something to say about this: https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
```diff
@@ -0,0 +1,77 @@
...
+      .as('first_adventure');
+
+    cy.get('#level-1')
+      .children()
+      .eq(1)
+      .invoke('attr', 'adventure')
+      .as('second_adventure');
+    
+    // Getting their values first, and then moving them around
+    cy.get('@first_adventure').then(first_adventure => {
```
rix0rrr, line 44: Ooo I didn't know you could do this! That's nice!

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
```diff
@@ -0,0 +1,77 @@
...
+      cy.get('@second_adventure').then(second_adventure => {
+       
+        // Move the second adventure to the first place
+        cy.get('#level-1')
+          .children()
+          .eq(1)
+          .trigger('dragstart')
```
rix0rrr, line 51: Woah! Well found!

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
rix0rrr: Could we use promises and `await` here? It would look nicer...

rix0rrr: After some Googling, looks like not...

tests/cypress/e2e/for-teacher_page/customize_class_page/remove_adventure_button.cy.js
```diff
@@ -0,0 +1,63 @@
...
+    cy.get("#available")
+      .children()
+      .should('have.length', 1)
+    
+    // store the name of the adventure we're going to delete
+    cy.get('#level-2 div:first')
+      .invoke('attr', 'adventure')
+      .as('adventure');
+    
+    cy.get("@adventure").then(adventure => {
```
rix0rrr, line 37: From reading the docs, I don't think this `as() + get()` is necessary. You can probably do:

```ts
cy.get(...)
   .invoke(...)
   .then(adventure => {
    });
```

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
```diff
@@ -0,0 +1,77 @@
...
+    // Now it should be visible
+    cy.get('#level-1').should('be.visible');
+
+    cy.wait(500)
+
+    // Get the first and second adventure
+    cy.get('#level-1')
+      .children()
+      .eq(0)
+      .invoke('attr', 'adventure')
```
rix0rrr, line 34: What is `invoke` doing here? Is this reading an attribute value using jQuery?

rix0rrr (commented):
I love that you're adding these tests, and you taught me a bunch of things about Cypress here, so thanks for that! 

There are two mini-things I'd like to request you do slightly differently. I understand that you based the code off of the examples that were already there, but those could have been better. That doesn't mean all existing code has to be changed, but we should be mindful of these things so that we will make the code easier to work with in the future.

## Test organization

Even though the students who originally wrote these test suites didn't know it, it's possible (and I want to say desirable) to put more than one test in the same file. That has a couple of advantages:

* It's easier to navigate (at least in my mind)
* We can use that file to define a couple of helper functions to share logic between tests, so we don't need to copy/paste code 
* We can use `beforeEach()` to share common setup code between tests.

A test suite can look like this for example:

```ts
describe('teacher page', () => {
  beforeEach(() => {
    loginForTeacher();
    createClass();
  });

  describe('back to class button', () => {
    it('sends us back to the previous page', () => { 
     // ...
    });
  });

  it('contains a list of all classes', () => {
    // .. 
  });
});
```

We don't need to repeat code that way. 

A `describe` with a single `it` in it looks a bit wonky... it makes sense because you are being taught that you're building a sentence that must start with the word `it`... but you don't actually have to. It just describes a test, so you could also do the following and that'd be fine:

```ts
  it('back to class button sends us back to the previous page', () => {
    // ...
  });
```

Just reads weird, but executes fine.

## Selecting HTML elements by ID

I see you added a couple of `id="xyz"` to elements, in order to target them with Cypress.

That works, but Cypress has a best practices section where they themselves recommend using a `data-` attribute:

image

Source: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

I would feel more comfortable if `id`s were reserved for making the actual functional web app work, and we'd use the `data-cy` attribute for Cypress functionality. It'd then be very obvious what is what.

------

Other than that, I love everything about this PR!

jpelay:
Hi @rix0rrr! Thank you so much for your thoughtful feedback, I really appreciate it :D 

With respect to test organization I'll be sure to do that. As I understand it you want me to put the tests of, for example, the customize class inside a single file? I can make that for the tests in this PR, and then we can make an issue for all the other tests what do you think?

I didn't know about that feature, It seems very useful and I'll be using it!

tests/cypress/e2e/for-teacher_page/customize_class_page/remove_adventure_button.cy.js
jpelay: Yes, good idea!

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
jpelay: Thank you!

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
jpelay: Yes, exactly. The invoke function calls the jQuery function to get the attribute. I got that info from here: https://stackoverflow.com/questions/57585785/cypress-get-attribute-value-and-store-in-variable

tests/cypress/e2e/for-teacher_page/customize_class_page/drag_and_drop.cy.js
jpelay: You are right, I upload them as an error, because they make it easier see the test in action, but in these cases there's no reason to keep the wait.

jpelay:
Hi @rix0rrr! I refactored the tests following your advises, I did the following:

1. Put all of the tests in a single file file with a `describe` and multiple `it`, the `beforeEach` function does make this a lot better.
2. I stopped using the `id` for grabbing elements and instead added a `data-cy` to the elements instead.
3. I removed all of the `wait`s

jpelay:
Why does removing the waits fixes the problems with the tests?

Felienne:
> Why does removing the waits fixes the problems with the tests?

Could be because (as @rix0rrr said above) "This always looks a bit scary to me, because if we're ever on a bit slower computer in which this takes 600ms, the test fails."

rix0rrr (approved):
❤️

jpelay:
> > Why does removing the waits fixes the problems with the tests?
> 
> Could be because (as @rix0rrr said above) "This always looks a bit scary to me, because if we're ever on a bit slower computer in which this takes 600ms, the test fails."

Sorry I worded it incorrectly, putting the wait back again makes them work!

Issue #3982, opened 2023-01-26
[UI, BUG] Programmer's mode now has a weird gap


Felienne:
Since a while (I suspect since the new design?) the programmer mode has a weird gap on Mac

<img width="1213" alt="image" src="https://user-images.githubusercontent.com/1003685/214883124-42eea98e-988e-4de1-a774-68389c63450c.png">

MarleenGilsing:
Same goes for windows

PR #3984, opened 2023-01-26
[FIX] Design: output height

rix0rrr:
A different fix for the code editor scrollbar, one that is compatible with the *programmer's mode* toggle.

Right now, it works by putting fixed heights (`22rem` and `36rem`) on the elements.

**How to test**

http://localhost:8080/hedy/7

```
repeat 30 times print 'Hedy is fun!'
```

See a scrollbar. Toggle programmmer's mode.

Fixes #https://github.com/hedyorg/hedy/issues/3982.


Felienne (approved):
Thanks for the quick fix!

PR #3988, opened 2023-01-29
[UI] Highlight variables after back-end response

ToniSkulj:
Implements the solution discussed in #3929.
Essentially, after running the program all variables are highlighted in yellow.

Before Running:
image

After Running:
image


ToniSkulj:
Hi @Felienne !
[Fix bug](https://github.com/hedyorg/hedy/pull/3988/commits/3c663cacafe6fa23ecf5721e79635f938c97f121) fixes part of the issue that you mentioned during the Hedy's contributing hangout.

A couple of examples:

Level 2:
image
image

Level 10:
image

The problem that still remains is that I can not find a (simple) way to make the distinction between text and variable
image
image

Let me know what you think!

Felienne:
> Hi @Felienne ! [Fix bug](https://github.com/hedyorg/hedy/pull/3988/commits/3c663cacafe6fa23ecf5721e79635f938c97f121) fixes part of the issue that you mentioned during the Hedy's contributing hangout.
> 
> A couple of examples:
> 
> Level 2: image image
> 
> Level 10: image

Ah that is more like it would be in my mind! 
 
> The problem that still remains is that I can not find a (simple) way to make the distinction between text and variable image image
> 
> Let me know what you think!

The issue of course is line numbers here, we only want to highlight after a certain line. We could send the line number to the front-end (I don't think we currently do because it is not needed) but can Ace do something with that info?

Felienne:
As discussed in the Hedy meetup, we are not really highlighting the variables in string values in levels before 4:

We want to make a difference here between the text and the variable names, now they are the same:

image

But maybe this is hard?

ToniSkulj:
@Felienne, I might have found a way with [Make distinction between text and variables](https://github.com/hedyorg/hedy/pull/3988/commits/36c8bd1394036dbe92a1a7211d4ccf004b55af04)! (Just using pure old JS)

Examples:
_The default/easy case_:
image

_Notice that all uses of answer are yellow except from the text!_:
image
Some magic is used here, essentially there is a check on every line if it starts with "{variable} is". if so, we only highlight the first. 

_From level 4 onwards only text between ' & " is allowed, this makes it easier to filter text_:
image
This was quite simple, just a regex with negative lookback for ' & "

There is a small bug left, when changing code from level 1 to 4 and running the code again the variables are not highlighted again. Should be easily fixable though.

`And... if we go deeper down this road, can we make use different colors in the future for singulars and lists...? You would also be cool!`
Yes we could! It's possible to figure out if a variable is a list or not from the front-end.

Felienne:
Hi @ToniSkulj!

Since we are going to replace our editor (see #4479) this is now a bit outdated :( But maybe @jpelay can take this idea into account in the new system, I still love the idea!

Felienne:
I made an issue so we don't forget (#4599)

PR #3989, opened 2023-01-29
[LANGUAGE] Add clear screen feature

### LINKED ISSUE #3645: [LANGUAGE idea] Add "clear screen" feature
**Language idea incl level**
As suggested by kids in my class, could we add a language element like clear screen for textual output? It is not all too hard for kids and seems doable to build, I think (although I am not sure how whether Skulpt supports it)

**Additional context**
I am not sure what a good level would be. Maybe 4 because it does not have many "cool" features?

ToniSkulj:
Fixes #3645.
Adds the _clear_ keyword that erases all content in the output from level 4.

Fun example in level 9:
```python
repeat 10 times
    if a is pressed
        print "A is pressed!"
    if x is pressed
        clear
```

**Make sure to generate the grammar locally first! Use the command:**
```bash
./build-tools/heroku/generate-grammars-and-js
```


Felienne:
Ah @ToniSkulj what a lovely addition!! I am going to test right away!

Felienne (approved):
Love it! We do need 2 changes (content and turtle) but I will make separate issues for that.

PR #3992, opened 2023-01-30
[LANGUAGE] Add clear to turtle

### LINKED ISSUE #3991: [LANGUAGE idea] Add clear to turtle
**Language idea incl level**
In https://github.com/hedyorg/hedy/pull/3989 we have added clear for clear the screen, which works for text but not for the turtle screen, can we make that work too?

ToniSkulj:
Fixes #3991.
Fun example to test in level 9:
```python
print "Lets Draw!"

repeat 30 times
    if w is pressed
        forward 20
    if a is pressed
        turn -90
    if d is pressed
        turn 90
    if x is pressed
        clear
        print "Cleared!"
```

When using the _clear_ keyword it should now also clear the turtle canvas.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -1709,7 +1709,15 @@ class ConvertToPython_4(ConvertToPython_3):
         return ConvertToPython_2.print(self, meta, args)
 
     def clear(self, meta, args):
-        return 'extensions.clear()'
+        return f"""extensions.clear()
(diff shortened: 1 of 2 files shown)


Felienne (approved):
Works, thanks!!

Even more fun like this!

image

PR #4006, opened 2023-02-02
[CONTENT] Turtle exercises - old

MarleenGilsing:
For level 1 and 2

Let's see if i'm allowed to add images. It seems like I'm not allowed to make a folder for the turtle images, but a seperate folder would be recommended.


Felienne:
I think you should be make a folder but I made one and moved the images! I also added a link so you can see how to do it!

Felienne:
Also, can we change the spaces in the filesnames into underscores? Spaces don't work on all computers (I think they do now on almost all systems, but they used to not work at all, so they give old programmers like me a bit of a tummy ache!)

Felienne:
Hi @MarleenGilsing! I have set this to draft since it is not entirely completed. You can pull my changes locally and then continue (if you need help, let me know)

SabinaChita:
I will help with this.

PR #4015, opened 2023-02-05
fix discrepancy in linelength text vs argument

Mark-Giesen:
the text states linelength should be 120, but the autopep8 argument states 100. Guess those should be the same.


Felienne (approved):
Ah yes, thanks for the catch!

PR #3993, opened 2023-01-31
[UI] Upload of a CSV file to create student accounts for a class

noelgolding:
Work In Progress
This hopefully should give an idea of the user experience.  

Items outstanding
* better UI styling (would love help from someone more visually inclined than me)
* error handling
* content (need to understand proper way to add english and request translations)
* testing

Fixes #3936 
resolves discussion #3933

**How to test**
TBD, ideally cypress, got to figure out how to use it
For now you can try it out manually
* For Teachers > Select a class > Add Students 
* press the new upload_csv button
* drag and drop or select a csv file to upload (csv file must only contain two columns with headers "username" and "password")
* confirm you want to upload the file
* review the changes on the create accounts form
* from here it works as if the accounts were typed in manually.
do all the things you would do if you were going to add them manually, e.g. generate passwords, append classname etc.

01 new upload csv button
02 drag and drop or choose csv file
03 confirm you want to upload csv
04 preview accounts imported from csv

PR #4018, opened 2023-02-06
[REFACTORING] create accounts form

noelgolding:
**Description**

* fix typo in intro
* add class name to the page header
* hide CLASS input field and column header
* ~~use template for cloning an account row~~
    * ~~client side reset~~

in preparation for https://github.com/hedyorg/hedy/pull/3993

**How to test**

```
cd tests
npx cypress run --spec "cypress/e2e/for-teacher_page/class_page/create_accounts_page/*"
```


noelgolding:
these cypress tests are passing for me locally.  Is there a way to view the screenshots of the failed tests?
```
(Screenshots)

  -  /home/runner/work/hedy/hedy/tests/cypress/screenshots/for-teacher_page/class_pag     (1280x720)
     e/create_accounts_page/create_accounts.cy.js/Passes -- before each hook (failed)               
     .png 
```

Felienne:
> these cypress tests are passing for me locally. image
> 
> Is there a way to view the screenshots of the CI failed tests?

Not unless we pay for a premium Cypress Cloud account. I will try them locally and see if I can reproduce the issue!

translations/en/LC_MESSAGES/messages.po
```diff
@@ -456,7 +456,7 @@ msgid "create_multiple_accounts"
 msgstr "Create multiple accounts"
 
 msgid "accounts_intro"
-msgstr "On this page you can create accounts for multiple students at the same time. There are autoamtically added to the current class, make sure the selected class is correct!. By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross. Make sure no rows are empty when you press \"Create accounts\". Please keep in mind that every username needs to be unique and the password needs to be <b>at least</b> 6 characters."
+msgstr "On this page you can create accounts for multiple students at the same time. They are autoamtically added to the current class, make sure the selected class is correct!. By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross. Make sure no rows are empty when you press \"Create accounts\". Please keep in mind that every username needs to be unique and the password needs to be <b>at least</b> 6 characters."
```
Felienne, line 459:
```suggestion
msgstr "On this page you can create accounts for multiple students at the same time. These are automatically added to the current class, so make sure the selected class is correct! By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross. Make sure no rows are empty when you press \"Create accounts\". Please keep in mind that every username needs to be unique and the password needs to be <b>at least</b> 6 characters."
```

Felienne:
Ok the "good" news is that for me the test also fails locally :)

Felienne:
Here is the screenshot:

image

Felienne:
image

templates/create-accounts.html
```diff
@@ -2,7 +2,7 @@
 
 {% block regular_content %}
 <div>
-    <h2>{{_('create_multiple_accounts')}}</h2>
+    <h2>{{_('create_multiple_accounts')}}: {{current_class.name}}</h2>
```
jpelay, line 5: This looks nicer now! Thanks a lot!

translations/en/LC_MESSAGES/messages.po
```diff
@@ -456,7 +456,7 @@ msgid "create_multiple_accounts"
 msgstr "Create multiple accounts"
 
 msgid "accounts_intro"
-msgstr "On this page you can create accounts for multiple students at the same time. There are autoamtically added to the current class, make sure the selected class is correct!. By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross. Make sure no rows are empty when you press \"Create accounts\". Please keep in mind that every username needs to be unique and the password needs to be <b>at least</b> 6 characters."
+msgstr "On this page you can create accounts for multiple students at the same time. These are automatically added to the current class, so make sure the selected class is correct! By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross. Make sure no rows are empty when you press \"Create accounts\". Please keep in mind that every username needs to be unique and the password needs to be <b>at least</b> 6 characters."
```
jpelay, line 459: I don't think we need such a contrived explanation for what is a fairly simple page!

"On this page you can create accounts for multiple students at the same time." => This is pretty obvious from the page itself, since you're prompted with several empty fields!

 These are automatically added to the current class, make sure the selected class is correct!. => This needs to be removed, because is redundant now that they're added to only the current class.
 
 "By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross." => I think this is clear from the buttons themselves? What do you think?
 
"Make sure no rows are empty when you press "Create accounts". Please keep in mind that every username needs to be unique and the password needs to be at least 6 characters. => In my opinion this is the only part worth keeping, but still maybe is better to enforce this from the backend itself instead of asking this to our users.

jpelay, line 459: I don't think we need such a contrived explanation for what is a fairly simple page!

"On this page you can create accounts for multiple students at the same time." => This is pretty obvious from the page itself, since you're prompted with several empty fields!

These are automatically added to the current class, make sure the selected class is correct!. => This needs to be removed, because is redundant now that they're added to only the current class.

"By pressing the green + on the bottom right of the page you can add extra rows. You can delete a row by pressing the corresponding red cross." => I think this is clear from the buttons themselves? What do you think?

"Make sure no rows are empty when you press "Create accounts". Please keep in mind that every username needs to be unique and the password needs to be at least 6 characters. => In my opinion this is the only part worth keeping, but still maybe is better to enforce this from the backend itself instead of asking this to our users.

jpelay (changes requested):

jpelay:
Thank you for this contribution @noelgolding! I like the change and cleans the UI a lot, I just think we can clean it further by removing the explanation at the top of the page!

translations/en/LC_MESSAGES/messages.po
noelgolding: I agree, I think this text can be simplified.  I'll defer to @Felienne on the verbiage.

Felienne:
HI both! I removed a bit of the explanation about the x and + buttons, I think that speaks for itself (or fits betters in a teacher manual) If you agree with the text, feel free to approve @jpelay (I feel a bit weird merging my own text now)

jpelay (approved):
Hi! Now that @Felienne added the text I feel everything's perfect good to go now :smile: Thank you again for this amazing contribution @noelgolding!

PR #4020, opened 2023-02-07
[CHORE] Add a tool to check the length of arrays in YAMLs

rix0rrr:
Compares the lengths of arrays in YAML files between `en.yaml` and the other languages, expecting them to be the same.

The new check is run as part of the `validate-yaml` script, which runs as part of regular GitHub PR validation.


jpelay:
Hello @rix0rrr! I don't know how hard this is, but can we be a little more precise as to where this is happening? I'm thinking about showing the specific key where the mismatch happens! What do you think?

Felienne:
> Hello @rix0rrr! I don't know how hard this is, but can we be a little more precise as to where this is happening? I'm thinking about showing the specific key where the mismatch happens! What do you think?

Hi @jpelay!

I just deliberately deleted a few elements in the Dutch yamls and I think the output is pretty helpful:

image

Is there anything more you'd like to see? (if not, please approve since I would feel better if we can catch these type of things earlier, I have already lost 2 afternoon fixing them late 😭 )

jpelay:
> > Hello @rix0rrr! I don't know how hard this is, but can we be a little more precise as to where this is happening? I'm thinking about showing the specific key where the mismatch happens! What do you think?
> 
> Hi @jpelay!
> 
> I just deliberately deleted a few elements in the Dutch yamls and I think the output is pretty helpful:
> 
> image
> 
> Is there anything more you'd like to see? (if not, please approve since I would feel better if we can catch these type of things earlier, I have already lost 2 afternoon fixing them late 😭 )

I thought the exact key where the arrays diverte might be helpful, but if this is good for you then I have no qualms approving this!

jpelay (approved):

Felienne:
> I thought the exact key where the arrays diverte might be helpful, but if this is good for you then I have no qualms approving this!

Ah yeah I see that could add a little bit of info yes but this is already so helpful!

rix0rrr:
> I thought the exact key where the arrays diverte might be helpful

It's a good idea, but how would you know? 

Let's say you have these two arrays:

```
- Banana
- Strawberry
- Plum
```

And

```
- банан
- слива
```

Now which is the index that I removed? Is it 0, 1 or 2?

PR #4022, opened 2023-02-08
[FIX] Design: indent lists

### LINKED ISSUE #4000: [UI] Numbered lists render to the left
This is tiny but numbered lists are rendered more "to the left" than looks nice (f.e. level 11 Haunted House (on alpha, not yet released!)

image

rix0rrr:
Indent lists a little so that the bullets don't stick outside the left hand margin, fixes #4000.

This also increases the line distance for the front page a bit, to give the text a bit more breathing room.

## Before

image

## After

image
In the adventure tab:

image


build-tools/heroku/tailwind/styles.css
```diff
@@ -353,11 +353,16 @@ code {
...
 }
 
 ol {
-    list-style: decimal outside none;
+  list-style: decimal outside none;
+}
+
+.text-display-copy ul, .text-display-copy ol,
+.text-business-copy ul, .text-business-copy ol {
+  margin-left: 2em;
```
TiBiBa, line 365: I think we can achieve the same with `list-style-position` and when using Tailwind the `list-inside` class:

https://tailwindcss.com/docs/list-style-position

I would prefer this approach as it is less code and better aligns with our CSS already being as much Tailwind as possible. Of course, I the goal is to also give it a bit of an indent on the text itself, I agree with your solution.

TiBiBa (changes requested):

build-tools/heroku/tailwind/styles.css
rix0rrr: `list-style-position` gives a different effect for dangling lines though, and arguably one that's less scannable (or at least, in my opinion). Examples:

```
list-style-position: inside
```

image

```
list-style-position: outside
```

image


We can't really control the bullet style on the lists we're interested in, as most of them are rendered from Markdown, and so we don't write the HTML directly where we would add the Tailwind `class="list-inside"` class.

Felienne:
This PR does not fix the thing it needed to fix (it works on the home page but not on the adventures) so let's put it as draft for now.

rix0rrr:
Reviving this old PR

Felienne (approved):
Nice! Looks a lot more clean:

image

Felienne (approved):

PR #4031, opened 2023-02-14
 [UI] Removes thin styling from cheatsheets

### LINKED ISSUE #4030: [BUG] Text in cheatsheets is being shown very thin
**Describe the bug**

The font in the cheatsheets is being styled as thin, and therefore it's a little hard to read. I think they should be styled with a normal weight. 

image

jpelay:
**Description**

As stated in the title, removes the `font-thin` class from the cheatsheets making them easier to read!

image

**Fixes #4030 **

### DIFF
--- a/templates/level-page.html
+++ b/templates/level-page.html
@@ -25,7 +25,7 @@
                 <h2 class="px-2 py-6">{{_('cheatsheet_title')}}</h2>
                 {% for example in cheatsheet %}
                   <div class="flex justify-between items-center gap-4 px-4 py-8 border-t border-dashed border-green-500">
-                    <div class="font-thin flex-grow h-min min-w-80" tabindex=0>
+                    <div class="flex-grow h-min min-w-80" tabindex=0>
                       {{ example.explanation|commonmark }}
                     </div>
                     <div class="flex-none">


Felienne:
Yes that does look nicer! What I was referring to when I mentioned the yamls, si that we have 2 fields in the files:

image

Name and description. I think that is where the thin comes from, we used to also show the name but in bold. I was thinking that we can maybe remove the names? They only add content (to be translated) and we don't show them here, but I am not sure they are not used elsewhere (in the printable sheets?) and if removing them leads to errors. Maybe worth an issue since we are cleaning up?

Felienne (approved):

jpelay:
> Yes that does look nicer! What I was referring to when I mentioned the yamls, si that we have 2 fields in the files:
> 
> image
> 
> Name and description. I think that is where the thin comes from, we used to also show the name but in bold. I was thinking that we can maybe remove the names? They only add content (to be translated) and we don't show them here, but I am not sure they are not used elsewhere (in the printable sheets?) and if removing them leads to errors. Maybe worth an issue since we are cleaning up?

Ahhhh, I understand what you mean now, miss Felienne. I dug a bit and I think you're right that it has to do with the cheatsheet name, but these are indeed used in the printable cheatsheets: https://github.com/hedyorg/hedy/blob/e73e5e1bcf3964c0db666eead2ccc089a8dffa19/templates/printable/cheatsheet.html#L22 although there the explanation still doesn't have the `font-thin` class.

PR #4045, opened 2023-02-16
fix: dropdown-selected keyword language is not passed to JavaScript

rix0rrr:
The keyword language selected in the keyword switching dropdown caused the `?keyword_language=` parameter to be added to the URL, but this parameter did not ultimately make its way into `g.keyword_lang`, which is the variable that gets passed to the JavaScript. So according to the JS, the keyword language was always English. Respect the keyword argument language if passed in the URL.

Additionally, when a keyword language was selected, it was committed to the `session`. If the user then subsequently changes language, the keyword language is not re-negotiated, because there is already one in the session! Remove the keyword language from the session when the user switches, so that renegotiation happens again.

Fix both issues, and add a test to validate.


tests/cypress/e2e/hedy_page/keyword_language.cy.js
```diff
@@ -0,0 +1,23 @@
+import {goToHedyPage} from "../tools/navigation/nav";
```
Felienne, line 1: Thanks for the added test, great to prevent this from happening in the future. This indeed fails on the current main so correctly catches the issue!!

Felienne (approved):

PR #4047, opened 2023-02-17
[DRAFT, LANGUAGE] allow list access with a variable

Felienne:
Work in progress (not sure it is going anywhere)


Felienne:
Note to self: have to check whether add and remove work with vars too (probably not)

jpelay:
> Note to self: have to check whether add and remove work with vars too (probably not)

Don't know if you checked this, but it does work! 

image

jpelay:
I think this is ready to go, but maybe is important to take this case into account in the future!

image

jpelay (approved):

Issue #4041, opened 2023-02-16
[TESTING] Create accounts Cypress test is flaky 


Felienne:
**Describe the bug**
This test keeps passing and failing and passing and failing. Sometimes it works in a PR and then fails on main, or the other way around, I can't really find a pattern. Can you take a look @noelgolding or maybe @jpelay? I can also disable the test for now (because it interferes with merging if it fails on a PR) but it is of course better to fix it.

<img width="733" alt="image" src="https://user-images.githubusercontent.com/1003685/219321901-77a38aea-c67e-4266-84ba-f0f1fe024cf6.png">

noelgolding:
@jpelay the only way I can reproduce this locally is to comment https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/for-teacher_page/class_page/create_accounts_page/create_accounts.cy.js#L21

Do you think a `cy.wait(500)` is required before looking for `.username-cell`?

```
    cy.get('#back_to_class_button').click();
    cy.wait(500);
    cy.get('.username_cell').should(($div) => {
```

noelgolding:
actually, I think this is a cleaner approach, the original test was taking advantage of a side effect.
```
    navigateToClass(classname);
    cy.get('.username_cell').should(($div) => {
```

PR #4048, opened 2023-02-17
[Fix] Flaky Create accounts Cypress test

noelgolding:
Fixes  issue #4041


jpelay:
Hello @noelgolding! Were you able to reproduce the issue? I'm testing both this branch and main and they never return me an error when I run the test!

noelgolding:
@jpelay I was not able to reproduce the flakiness locally. I think it is a timeout issue waiting for the page to load after clicking the back to class button in ci pipeline. That being said none of the tests that first navigateToClass seem to exhibit this flaky behavior, I think this update should address it.

Felienne (approved):
Let's see if this helps (it seems mostly to be flaky on GitHub?)

PR #4049, opened 2023-02-18
[BUG] Consequitive If pressed with other command in between

### LINKED ISSUE #4042: [BUG] Consequitive if pressed with print in between
**Describe the bug**
Ok, @ToniSkulj we are getting these, but this remains (level 8)

```
als s is ingedrukt
    print 'leuk'
anders
    print 'wat ben jij stom doei'
print 'maak jij de kinderen wakker'
als a is ingedrukt
    print 'ja ik maak ze wakker'
anders
    print 'wat vervelend ben jij dan doe ik het wel.'
```

The second if creates an if without the while true around it:

```
while not pygame_end:
  pygame.display.update()
  event = pygame.event.wait()
  if event.type == pygame.QUIT:
    pygame_end = True
    pygame.quit()
    break
    if event.unicode == 's':
      print(f'leuk')
      break    
    else:
      print(f'wat ben jij stom doei')
      break
    
print(f'maak jij de kinderen wakker')
    if event.unicode == 'a':
      print(f'ja ik maak ze wakker')
      break    
    else:
      print(f'wat vervelend ben jij dan doe ik het wel.')
      break
```

image

I think the fix is to change the semantics so that each `if pressed` get its own `while`, not very hard to fix I think! Let me know if you have the bandwidth to look at it!

ToniSkulj:
Fixes #4042,

Makes it possible to have statements in between _if_pressed_ commands without breaking the code.
For example:
```python
als s is ingedrukt
    print 'leuk'
anders
    print 'wat ben jij stom doei'

print 'maak jij de kinderen wakker'

als a is ingedrukt
    print 'ja ik maak ze wakker'
anders
    print 'wat vervelend ben jij dan doe ik het wel.'
``` 

To-Do
- [x] Rewrite Pygame related tests


ToniSkulj:
Hi @Felienne, [Add pygame prefix to all ifpressed commands](https://github.com/hedyorg/hedy/pull/4049/commits/fba52bba35b3d7457dc0005fb90021df86e1a1e3) changes the semantics like we discussed in https://github.com/hedyorg/hedy/issues/4042. It will encapsulate all _if_pressed_ commands with its own while loop.

But I was quite sad to see that the (I believe) most fun/popular program would not work anymore in level 9 (drawing with the turtle):
```python
repeat 10 times
    if w is pressed
        forward 20
    if a is pressed
        turn -90
    if d is pressed
        turn 90
```

So I added [Add post processing](https://github.com/hedyorg/hedy/pull/4049/commits/256bfd1f253a9eedf915c0082ff153c05cbbd199), it removes the while loops only if multiple if_pressed commands are followed by eachother. It's definitely not the best solution but this way the problem is fixed but the fun program is still fun. 

Also changing the semantics to [Add pygame prefix to all ifpressed commands](https://github.com/hedyorg/hedy/pull/4049/commits/fba52bba35b3d7457dc0005fb90021df86e1a1e3) will mostly defy the purpose of the buttons for example:
```python
hey is button
bye is button

if hey is pressed print "hey!"
if bye is pressed print "bye!"
```

It would not be possible to click on bye and expect anything to happen.
Let me know what you think!

Felienne:
Thanks for the fix @ToniSkulj!! I see you still have to fix the tests but some observations already here.

> 
> ```python
> als s is ingedrukt
>     print 'leuk'
> anders
>     print 'wat ben jij stom doei'
> 
> print 'maak jij de kinderen wakker'
> 
> als a is ingedrukt
>     print 'ja ik maak ze wakker'
> anders
>     print 'wat vervelend ben jij dan doe ik het wel.'
> ```

Yes, this is perfect, because the two if's are separated by a non-if (the print)!

Felienne:
> But I was quite sad to see that the (I believe) most fun/popular program would not work anymore in level 9 (drawing with the turtle):
> 
> ```python
> repeat 10 times
>     if w is pressed
>         forward 20
>     if a is pressed
>         turn -90
>     if d is pressed
>         turn 90
> ```

> So I added [Add post processing](https://github.com/hedyorg/hedy/pull/4049/commits/256bfd1f253a9eedf915c0082ff153c05cbbd199), it removes the while loops only if multiple if_pressed commands are followed by eachother. It's definitely not the best solution but this way the problem is fixed but the fun program is still fun.

Agreed this is a different case since the if's are living "together" (without an intermediate something else) so they don't need their own while! I am not sure that is possible to generate correctly form the inception, rather than doing (somewhat complex) post-processing? But at least I agree with the semantics you suggest here!!

> Also changing the semantics to [Add pygame prefix to all ifpressed commands](https://github.com/hedyorg/hedy/pull/4049/commits/fba52bba35b3d7457dc0005fb90021df86e1a1e3) will mostly defy the purpose of the buttons for example:
> 
> ```python
> hey is button
> bye is button
> 
> if hey is pressed print "hey!"
> if bye is pressed print "bye!"
> ```

Wouldn't this also fall under the "if's together" rule so be in one while? I am not sure why it would be different here?

In any case, good work! Let mw know if it needs discussing more.

ToniSkulj:
Thanks for the feedback @Felienne! 

> I am not sure that is possible to generate correctly form the inception, rather than doing (somewhat complex) post-processing?

Me neither, I think it is a decent approach to fix it like this for now, in the future I can always try to figure out a way (if there is any). Let me know if you think otherwise.

> Wouldn't this also fall under the "if's together" rule so be in one while? I am not sure why it would be different here?

Yess it would, if we use the  "if's together" rule. I was just not sure if you would agree on that.

I'm going to think about this a little more, maybe I can find a more elegant solution for combining the ifs in one while loop. In any case, I will open up a PR for review before the end of Friday so that this we will (at least) have a fix.

ToniSkulj:
I unfortunately did not find another solution using the limited time that I had. 

This solution may bring some problems in the future for projects such as 
https://github.com/hedyorg/hedy/discussions/3397 and therefor also https://github.com/hedyorg/hedy/discussions/3192. But for now at least, the problem is fixed.

hedy.py
```diff
@@ -981,6 +981,28 @@ def NEGATIVE_NUMBER(self, args):
...
+while not pygame_end:
+  pygame.display.update()
+  event = pygame.event.wait()
+  if event.type == pygame.QUIT:
+    pygame_end = True
+    pygame.quit()
+    break""")
+
+    regex_string_subsequent_use = (
+        r"\s*# End of PyGame Event Handler"
```
jpelay, line 995: This is a pretty smart move, for us to be able both if pressed consecutive commands and with other commands in between!

tests/test_level/test_level_08.py
```diff
@@ -826,21 +830,27 @@ def test_double_if_pressed(self):
...
+            if event.unicode == 'x':
+              print(f'first key')
               break
-            if event.type == pygame.KEYDOWN:
-              if event.unicode == 'x':
-                print(f'first key')
-                break
-            if event.type == pygame.KEYDOWN:
-              if event.unicode == 'y':
-                print(f'second key')
```
jpelay, line 842: I see that you changed the tests but didn't add one with command in between them?

jpelay (approved):

tests/test_level/test_level_08.py
ToniSkulj: I completely forgot to add that test (I guess I was tired after changing all tests😅), thanks for adding the test @jpelay !!

PR #4051, opened 2023-02-19
[CHORE] Automation of the texts translate with DeepL

emartynov:
**Description**

This PR adds a script to automate draft translations with DeepL. 

Unfortunately, I'm not fluent in python, so add the script in Kotlin. 
The script is not a product of precise specification but a weekend hack of things I need to know. So script code could be refactored to a more readable and maintainable structure.
I'm also new to DeepL API. I can probably format text to XML, JSON, or DeepL has another solution to preserve keys.

The script has two bugs:
1. It Doesn't run for big files since DeepL has a limit on the request size, and I didn't come up with an easy translation split that will not break the `key` to `value` structure
2. The format of the string 

**Related to** #2031

Manual translating is slow and exousting. Feeding content to this script leaves only tiny corrections left to the translator.

**How to test**

* Read README in the script folder to setup the environment
* Run script 
```
./tools/kotlin/translate-content.main.kts --content content --language <language to translate> --deepl-key <deepl-key>
```

**Checklist**
  
- [ ] Runs well on the files that content for translation is bigger than the DeepL limit
- [ ] Fix a bug when DeepL changes int to a different int in translation
- [x] The Ukrainian translation is reviewed by a native speaker (me)

### DIFF
--- a/content/slides/uk.yaml
+++ b/content/slides/uk.yaml
@@ -1,66 +1,115 @@
 levels:
-    0:
-        1:
-            text: "Hedy is a programming language.\n"
-            header: Welcome to Hedy!
-        2:
-            header: Programming
-            text: "Programming is giving instructions to a computer, using the language of the computer.\n\nSuch a language is called a \"programming language\".\n"
-        3:
-            header: Programming Language
-            text: "Maybe you already know some programming languages?\nSome examples are:\n* Scratch\n* Python\n* HTML\n"
-        4:
-            header: Programming
-            text: "If you can program, you can do many cool things:\n* Build interactive stories\n* Create computer drawings and draw these on paper or fabric\n* Build your own apps to use on your phone\n"
-        5:
-            header: Programming in Hedy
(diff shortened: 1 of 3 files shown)


Felienne:
HI @emartynov!

Thanks for this PR, but we can't accept it. There are a few reasons:

* We doubt whether this is useful. Automated translations are not always good, esp. for small languages, and are unlikely to suggest kid-facing formulations. I encourage you to try to yourself while translating and see if it indeed saves time.
* We can't take on maintenance of Kotlin code

I really understand you are trying to be helpful, but if you want to help we have so many open issue to work on. What you are now doing--even though it is surely coming from all the best intentions!!--is adding to our workload and distracting us from what we have decided is now our main focus (teacher support). 

Feel free to drop into a meeting if you want to discuss the road map more!

emartynov:
@Felienne would you accept just the translation part? I will remove commits related to the script and I will just use in for personal purposes.

PR #4052, opened 2023-02-19
[CHORE] Explicit JS initialization everywhere

rix0rrr:
Remove `APP_STATE` and `passFromHtml` features. All JavaScript initialization is now done by calling `initialize()` (in `initialize.ts`), and optionally by passing page-specific options (from Python to JavaScript in `javascript_page_options`, which get dispatched in `initialize.ts`).

Each module now has their own copy of some globals they might need... this is still not the ideal end state, but at least each individual module can now be refactored independently. 

Also making an attempt to reduce the size of `app.ts`, which is starting to grow quite large, by moving functionality out into their own files where it  makes sense.

In this PR:

- Remove `APP_STATE`
- Introduce page-specific initialization options
- Move the app-specific logic (around loading programs, hiding and showing various page elements based on the current tab) from `tabs.ts` to `app.ts`, by using an event mechanism (`tabs.on('afterSwitch', () => { ... })`).
- Move the functionality that has to do with debugging from `app.ts` into `debugging.ts`.
- Move the `onElementBecomesVisible` functionality out of `app.ts` into its own file
- Move some `<script>` tags out from HTML into JavaScript


rix0rrr:
Note: currently untested, this is for preview

rix0rrr:
Ready for review!

static/js/event-emitter.ts
```diff
@@ -0,0 +1,66 @@
...
+ * An event emitter with the ability to cancel and defer events
+ */
+export class EventEmitter<Events extends object> {
+  private events: {[K in keyof Events]: EventHandlers<Events[K]>};
+
+  constructor(events: {[k in keyof Events]: true}) {
+    this.events = Object.fromEntries(Object.keys(events).map(k => [k, []])) as any;
+  }
+
+  public on<E extends keyof Events>(name: E, handler: EventHandler<Events[E]>) {
```
jpelay, line 11: I really like this!

static/js/tabs.ts
```diff
@@ -24,140 +31,82 @@ let _currentTab: string | undefined;
...
-  else if (adventures[tabName]?.loaded_program) {
-    $ ('#program_name').val (adventures[tabName].loaded_program!.name);
-    theGlobalEditor?.setValue (adventures[tabName].loaded_program!.code);
-  }
-  else {
-    if (tab.hasClass('teacher_tab')) {
-      $ ('#program_name').val (tabName);
-      theGlobalEditor?.setValue ("");
+    // We don't do an even for the very first tab switch
+    if (this._currentTab != '') {
```
jpelay, line 96: I like how this is a lot cleaner now

jpelay (approved):
Love this @rix0rrr!

jpelay:
Ok for some reason the highlighting is not working in the adventures tab!

image

PR #4056, opened 2023-02-22
[FEAT] UI: focus language input on opening dropdown

rix0rrr:
If you are looking for a language, you previously needed to click the language dropdown, then the input, and only then could you start typing.

Focus the language input immediately on opening the dropdown so you can start typing right away.

**How to test**

Click the Language dropdown, start typing.


jpelay (approved):
Hi! Love this, approving right away!

PR #4060, opened 2023-02-23
[TESTING] Add coverage to front-end testing

jpelay:
As discussed in the last Hedy meeting this PR adds coverage support for the cypress testing.


jpelay:
Marking as ready for review! 

This PR adds a tool and a script to check for front end coverage, and this is how the report looks once you execute it:

image

I couldn't figure out how to exclude the node_modules files from the coverage, since the file we use to execute the Javascript is `appbundle.js` that has everything in it. And this is why I think the coverage won't be 100% accurate but still offers a good guidance.

**Why not instrument the typescript code?** 

I tried doing this, but since we link appbundle.js directly in our file, and that's the file that gets executed, instrumenting the typescript code does nothing. Or at least I didn't figure out how, if someone does know how to do this it'd be greatly appreciated!!

**How to get the report**

I added a script that makes it easier to do all the steps necessary to generate the report, that is:

1. Delete folder with instrumented code (this prevents the .ts files to be instrumented).
2. Compile the typescript code.
3. Instrument code.
4. Replace the appbundle.js in static/js with the instrumented one.
5. Run the tests. This is why we need the instrumented appbundle, since this is the one that can count which lines are ran.
6. If the tests run succesfully it'll restore the original appbundle

There's something that I don't love and is that if the cypress tests fail the last step, which is restoring the original appbundle won't be executed.

Felienne:
It seems I am getting an error. I tried form the project root and in the tests folder, some error shows up:

image

Felienne (changes requested):
See above (maybe I am missing a config file?)

Felienne (approved):
Yes, works after an npm update!!

Great work @jpelay!!

image

jpelay:
> Yes, works after an npm update!!
> 
> Great work @jpelay!!
> 
> image

Thank you Miss Felienne!!!

reedspool:
I ran into the above error as well. This happens because the test script uses `npx`. `npx` will ask to install a script if it doesn't find it in the local `node_modules`. But unfortunately when it installs that script, it installs it to a global place (kind of like `npm install -g` but a different location). And further problems, the `nyc` library doesn't work when global, since it looks for a particular config file which `node_modules` installs. 

This can all be prevented by simply disallowing `nyc` from installing in that global place. Where you have `npx nyc`, put the `--no` option in. This will stop npx from asking or attempting to do its global install.

`npx --no nyc`

This will print out a useless error if the person hasn't run `npm install` yet. So either catch that error or just always print something above it like ```echo "👀 If you get 'npm ERR! canceled' try `npm install`. 👀"```

Felienne:
> I ran into the above error as well. This happens because the test script uses `npx`. `npx` will ask to install a script if it doesn't find it in the local `node_modules`. But unfortunately when it installs that script, it installs it to a global place (kind of like `npm install -g` but a different location). And further problems, the `nyc` library doesn't work when global, since it looks for a particular config file which `npm` installs in the project-local `node_modules`.
> 
> This can all be prevented by simply disallowing `nyc` from installing in that global place. Where you have `npx nyc`, put the `--no` option in. This will stop npx from asking or attempting to do its global install.
> 
> `npx --no nyc`
> 
> This will print out a useless error if the person hasn't run `npm install` yet. So either catch that error or just always print something above it like `` echo "👀 If you get 'npm ERR! canceled' try `npm install`. 👀" ``

This sounds good, but the details are too much for me to fully grasp. Is this something people need to do locally or something we can add to our scripts?

reedspool:
Looks like @rix0rrr solved this in `build-tools/heroku/generate-typescript`

```sh
if [[ ! -f node_modules/.bin/tsc ]]; then
    echo "👉 You need to run the following command ONCE before running 'generate-typescript':" >&2
    echo "" >&2
    echo "    npm ci" >&2
    echo "" >&2
    exit 1
fi
```

Drop it into this coverage script and this won't happen again :) Probably replace `tsc` with `nyc` in the `if` conditional.

PR #4062, opened 2023-02-24
[FIX] "Next question" button on quiz sometimes loads wrong question

rix0rrr:
Every time a quiz answer comes back, the frontend does the following:

```ts
$('#next_button').on('click', () => loadLevel(level + 1));
```

The problem is that none of the old event handlers ever get cleared, so by the time you click the button for the 5th time, we run event handlers that try to load level 2, 3, 4, 5 and 6 all at once.

Whichever one happens to load slowest wil be the one that gets displayed.

Fix this by clearing old onclick event handlers before attaching a new one.

**How to test**

Click around in the quiz (with the Network tab open) and observe that every time you click the **Next question** button, only one quiz question gets loaded.


Mark-Giesen (approved):
Works fine now. The off().on() part looks a bit ugly? Maybe that's normal in FE or TS, but gut feeling tells me this should be handles an other way. It's a bugfix, so make users happy first.

PR #4069, opened 2023-02-28
[UI IMPROVEMENT] Smoothen dropdown menus

### LINKED ISSUE #4066: [UI idea] Hide dropdown instead of re-open when pressing button again
**Idea incl level**
Title should be quite explanatory. When you press a button twice that handles a dropdown it shouldn't open again, but instead close it. This makes the UI a lot less crowded when pressing the button multiple times.

TiBiBa:
**Description**
Slide toggle hide a dropdown when the button is pressed again instead of the hacky re-slide open we currently have.

**Fixes**
This PR fixes #4066

**How to test**
Verify that the language dropdown, user profile dropdown and both the cheatsheet and language switcher dropdown all close when the button is pressed again. This makes the animation a lot more smooth than what currently happens when we press the button twice.

### DIFF
--- a/templates/incl/menubar.html
+++ b/templates/incl/menubar.html
@@ -22,7 +22,7 @@
             </a>
         </li>
         <li class="flex-initial menubar-item dropdown relative z-20{% if current_page == 'my-profile' %} active{% endif %}">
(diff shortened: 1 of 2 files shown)


jpelay:
Hi @TiBiBa! This indeed looks a lot better! Thank you!

jpelay (approved):
Hi @TiBiBa! This indeed looks a lot nicer and is more aligned to what an user expects.

PR #4071, opened 2023-02-28
[FIX] test 'generate-client-messages'

rix0rrr:
This fixes the version of pybabel to a version number so that any babel behavior will be consistent between different contributors and adds a test to see if we are not missing babel messages.

**How to test**

No testing involved.


Felienne:
Thanks for the fix @rix0rrr! I am marking this draft for now because this will totally mess up Weblate and cause conflicts so I don't want to merge it accidentally. Let us know if this is needed for you to progress.

Felienne (approved):

PR #4065, opened 2023-02-26
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Hedy/Adventures](https://hosted.weblate.org/projects/hedy/adventures/).

It also includes following components:

* [Hedy/Achievements](https://hosted.weblate.org/projects/hedy/achievements/)

* [Hedy/slides](https://hosted.weblate.org/projects/hedy/slides/)

* [Hedy/Parsons](https://hosted.weblate.org/projects/hedy/parsons/)

* [Hedy/Webpages](https://hosted.weblate.org/projects/hedy/webpages/)

* [Hedy/Quizzes](https://hosted.weblate.org/projects/hedy/quizzes/)

* [Hedy/Cheatsheets](https://hosted.weblate.org/projects/hedy/commands/)

* [Hedy/Texts](https://hosted.weblate.org/projects/hedy/web-texts/)

* [Hedy/client-messages](https://hosted.weblate.org/projects/hedy/client-messages/)

* [Hedy/Keywords](https://hosted.weblate.org/projects/hedy/keywords/)

* [Hedy/Tutorials](https://hosted.weblate.org/projects/hedy/tutorials/)

Current translation status:

Weblate translation status

PR #4077, opened 2023-03-01
[LANGUAGE] Chinese punctuation

### LINKED ISSUE #2724: [Language idea] Translation of punctuation in code
The code (keyword) translation needs to be extended to translate other punctuation than western. 

In Chinese for instance the colon is not ':' but the '：' (unicode #ff1a). Same thing for the ';', this would be '；' (#ff1b). Comma's and periods might be the same. Apparently it's not hard for Chinese to use western punctuation, but it would be a lot friendlier if we supported both.

### LINKED ISSUE #4012: [LANGUAGE idea] Localization: Chinese quote marks and other characters
**Language idea incl level**
I would like to propose to use Chinese quote marks in the Chinese version of Hedy language. Quote marks are introduced in Level 4. They are "" and '' in English while they are “” and ‘’ in Chinese. 

**Additional context**
The usage and semantics of Chinese and English quote marks are virtually the same, so we safely replace them in the Hedy language. 

The table of unicode for quotation marks is the below

| Quotation mark    | Unicode |
| ---- | ---- |
| “ (Chinese Left Quotation Mark, A.K.A Left Double Quotation Mark in Unicode) | U+201C |
| ” (Chinese Right Quotation Mark, A.K.A Right Double Quotation Mark in Unicode) | U+201D |
| ‘ (Chinese Left Single Quotation Mark, A.K.A Left Single Quotation Mark in Unicode) | U+2018 |
| ’ (Chinese Right Single Quotation Mark, A.K.A Right Single  Quotation Mark in Unicode) | U+2019 |
| “ (English Quotation Mark) | U+0022 |
| ‘ (Apostrophe) | U+0027 |

Fun fact BTW, this header file has become kind of meme in Chinese C programmer community, which replaces frequently used characters in Chinese version with English ones:
```c
#define ； ;
#define ， ,
#define “ "
#define ” "
```
So I also suggest replace `,` and `;` with their Chinese counterparts `，` and `；` so to complete the localization.

Felienne:
Fixes #2724 (Chinese brackets and colon are now supported) and fixes #4012 (Chinese quotes).

This will hopefully fix some of the issues currently going on in #4065!


Felienne:
Note to self: Don't forget to also update the error translation for other characters as in https://github.com/hedyorg/hedy/pull/4077/commits/9ed20b9f37eec4555e24996dfc1456209c42ae04

Felienne:
Guess this was not so long-lived after all!

PR #4081, opened 2023-03-01
[WEBSITE] Implements dynamic multi-langual logo

### LINKED ISSUE #4080: [UI idea] Make logo dynamically multi-lingual
**Idea incl level**

**Additional context**

TiBiBa:
**Description**
Implements a dynamic text placeholder for the logo.

**Fixes**
This PR fixes #4080.

**How to test**
Try different languages, verify that is always works. Especially in the case of LTR and RTL.


messages.pot
```diff
@@ -941,12 +827,21 @@ msgstr ""
 msgid "try_it"
 msgstr ""
 
+msgid "print"
```
Felienne, line 830: Do we need a new key here? We already have the keyword print traslated in content/keywords?

Felienne: We could even fetch print hello from the first adventure, then we don't have to translate anything exta:

https://github.com/hedyorg/hedy/blob/411a9b3e9da3f8749b9c4982bb3a59b1c12ccb66/content/adventures/en.yaml#L18

Felienne:
Ow and RTL needs a bit of love I think?

image

messages.pot
TiBiBa: We could, but this would make the code a bit more complex as we have to split on the whitespace as well for styling. New key might not be ideal, but I think the most straight-forward solution.

TiBiBa:
> Ow and RTL needs a bit of love I think?
> 
> image

Just fixed this, should be ready for merge if the keyword approach is agreed upon.

Felienne:
Thanks for the updates @TiBiBa!! The broken unit tests were my bad (should be fixed now) but there are also issues in the Cypress tests, can you take a look?

TiBiBa:
> Thanks for the updates @TiBiBa!! The broken unit tests were my bad (should be fixed now) but there are also issues in the Cypress tests, can you take a look?

Whoops sorry for the delay! This should be fixed including a check for a custom file, if not: set the title with css.

Felienne (changes requested):
Sorry I kept this lying around so long @TiBiBa!

I have merged into main, renaming the keys so it is a bit more clear what they are meant for.

We are getting there, but there is still a small issue with font size. Can we make that dynamic based on the string's length. Otherwise we get this:

image

I committed the Bulgarian translation so you can play around with it!

TiBiBa:
> Sorry I kept this lying around so long @TiBiBa!
> 
> I have merged into main, renaming the keys so it is a bit more clear what they are meant for.
> 
> We are getting there, but there is still a small issue with font size. Can we make that dynamic based on the string's length. Otherwise we get this:
> 
> image
> 
> I committed the Bulgarian translation so you can play around with it!

Fixed! It is still a bit "houtje-touwtje", but should still be a nice addition until more logo's are added manually.
Schermafbeelding 2023-03-28 om 20 12 00

Felienne (approved):
Good to go, thanks for the final fixes @TiBiBa!

Issue #3823, opened 2022-12-09
[BUG] Arabic code editor incorrectly editing text when using text selection with backspace and/or delete 


alaaeddin-swidan:
**Describe the bug**
When editing code in the Arabic language: specifically trying to change some text using simple text selection with the backspace and delete buttons, it is then impossible to determine where the cursor is going to take the action and therefore the editing of the textual code is hindered. 

**Paste the Hedy code & level**
Applies to all levels. But let's take the following as an example: Level 2 -> introduction -> first example code: 
الاسم هو هيدي
العمر هو ١٥
قول الاسم وعمرها الان العمر سنة

**Video screen comparing Arabic and English code 

https://user-images.githubusercontent.com/7933089/206694714-1fcc8e7a-5a43-4261-87f8-fc6a28e38709.mp4

The actions above show that a shift is happening to the left whenever a selection is done on a letter or a word, the first operations were a backspace: what they do is they remove a character ahead (left of the selection by 2 spaces (I think)) rather than going backward to the right while deleting the character.. Another one is when selecting the whole word الان followed by a backspace: the word itself stays as is (not deleted) but the deletion happens also somewhere to the left (ahead of the word).

https://user-images.githubusercontent.com/7933089/206694752-c3d5e114-9515-4db5-84d1-68a045de744e.mp4

In the English editor the actions are as expected. 

**What machine and browser you were using (optional)**
Windows. Firefox browser.

Felienne:
Hi @alaaeddin-swidan!

Is this a new issue (in my experience this has always been the case) Just trying to understand if this is something we broke or that is most likely in Ace.

vzsky:
I'm not sure whether the arabic one is the same as this case. But with Thai, AceAjax edit seems not to fully support the language. 

for example, try the word "คำที่มีปัญหา" in the official(?) ace editor [here](https://ace.c9.io). When you move the cursor to the back of the word, it extends beyond the actual position of the end of the visual word as there are 0 length characters.

I suspect something with this issue on ace : https://github.com/ajaxorg/ace/issues/4512 
Or this one in the case of arabic? : https://github.com/ajaxorg/ace/issues/460

In the past, I had used [codemirror](https://codemirror.net), where there were no problem with Thai. But still have no up-to-date information.

alaaeddin-swidan:
Hi @Felienne no this is not a new issue. 
Thanks for the input @vzsky. I tried the official Ace edi


[...]

PR #4090, opened 2023-03-07
[WIP, UI] Force Ace font

Felienne:
This might be a fix for #3823?

Can you take a peek @alaaeddin-swidan? 

It does make the editor look ugly in Latin because the font is too small, but if this addresses the issue I am sure we can patch that too. If this does not work, in any case it can be a branch for us to try stuff out in (other options mentioned in #3823) because you now know where to fiddle with the css?

### DIFF
--- a/static/css/additional.css
+++ b/static/css/additional.css
@@ -24,7 +24,7 @@
 }
 
 .ace_editor * {
-    font-family: monospace !important;
+    font-family: Courier, monospace !important;
     font-size: 12pt;
 }
 
--- a/static/css/additional.css
+++ b/static/css/additional.css
@@ -23,6 +23,11 @@
   display: none;
 }
 
+.ace_editor * {
+    font-family: monospace !important;
+    font-size: 12pt;
+}
+
 input:checked ~ .dot {
   transform: translateX(100%);
   background-color: #48bb78;


vzsky:
Does not work for Thai. I did change a few css but still no luck. I think it might be more or less related to https://github.com/ajaxorg/ace/issues/3141 and this open issue https://github.com/ajaxorg/ace/issues/460

Felienne:
> Does not work for Thai. I did change a few css but still no luck. 

:( 

> I think it might be more or less related to [ajaxorg/ace#3141](https://github.com/ajaxorg/ace/issues/3141) and this open issue [ajaxorg/ace#460](https://github.com/ajaxorg/ace/issues/460)

Yes, I think that too! I took this idea from [ajaxorg/ace#460](https://github.com/ajaxorg/ace/issues/460

reedspool:
In one of these threads I saw the suggestion that the default monospace font in chrome "Monako" is insufficient, but that Courier works better. So something to try is

```
.ace_editor * {
    font-family: Courier, monospace !important;
    font-size: 12pt;
}
```

Felienne:
> 



> In one of these threads I saw the suggestion that the default monospace font in chrome "Monako" is insufficient, but that Courier works better. So something to try is
> 
> ```
> .ace_editor * {
>     font-family: Courier, monospace !important;
>     font-size: 12pt;
> }
> ```
> 
> I tried this myself and found it worked a bit better

Thanks for diving in deeper! Somehow, I cannot seem to reproduce the issue as shown in the video of @alaaeddin-swidan above, both in Chrome for macOS and Safari backspace does go the right way in Arabic (towards the gutter) and I am also able to select a word and delete the whole word with backspace too. How have you reproduced the issue?

> , but didn't fix the issue completely on long lines of text. The issue seems to be worse on longer lines.

Should that be the case we can maybe patch that a bit by giving a warning for long lines.

> There's a chance that different fonts work best for different languages. But I fear that, at best, this kind of solution will only fix Ace half-way, leaving more issues to be patched individually as they pop up. 

This is probably true but giving we are not likely to move over to CodeMirror soon, half fixed is better than nothing?

> I hope my fears are wrong though!!

I have pushed your suggestion, maybe @alaaeddin-swidan can take a look, or @vzsky for Thai (assuming this has the same root cause)?

Felienne:
we are going to slowly move away from Ace, so this can be closed

PR #4094, opened 2023-03-07
[FEATURE] [LANGUAGE] Source-mapper

ToniSkulj:
PR for adding a source-mapper to Hedy


ToniSkulj:
Hi @jpelay! I added [Concept 2 ?](https://github.com/hedyorg/hedy/pull/4094/commits/46d4b9e767409b74eb1782ea9f8408d7b012d4ea), this might give us some more functionalty when using the sourcemapper. Essentialy, I added some types so that we could make a dictionary that is dedicated to the source mapping.

My idea here was to store as much info as possible (which might be useful), after that we could use the _get_response_object_ method to strip the object of only the essential things we need, to keep performance good.

This might be waay to complicated than what is necessary😅. Feel free to revert all of this!

ToniSkulj:
It's easy to go from Hedy "from this line number  to this line number to Python" but going the other way around might be more tricky. As in, which line number in Python is it?

ToniSkulj:
I added [Concept 3? (I like this one the most)](https://github.com/hedyorg/hedy/pull/4094/commits/78c616a3c705c3729b14eaf7ac0199e5538eb274) this makes the mapping more from code to code instead of Hedy source range to python code. 

Now the above problem is cleary visible in
_SourceRange(0, 0, 0, 0),  # this is tricky_
in hedy_sourcemap.py

hedy_sourcemap.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __str__(self):
+        return str(self.map)
+
+
+def source_map_rule(source_map: SourceMap):
+    def decorator(function):
+        def wrapper(*args, **kwargs):
+            meta = args[1]
+            generated_python = function(*args, **kwargs)
```
jpelay, line 71: I have a doubt, here we are executing the function twice? Once for compiling and one for the source mapping?

jpelay:
This looks great, Toni, OMG! I do like the added complexity, let's keep it for now unless we see that it won't serve us any purpose in the future. 

For the line numbers in Python I just added a counter for the current line number, and count the number of new lines so we now the length of the snippet. This is how it looks with the turtle:

```
{1/0-1/10 --- forward 50: 1/0-0/7 --- __trtl = 50
try:
  __trtl = int(__trtl)
except ValueError:
  raise Exception(f'While running your program the command <span class="command-highlighted">forward</span> received the value <span class="command-highlighted">{__trtl}</span> which is not allowed. Try changing the value to a number.')
t.forward(min(600, __trtl) if __trtl > 0 else max(-600, __trtl))
time.sleep(0.1), 2/11-2/20 --- turn left: 8/0-0/8 --- t.left(90)}
```

In this case, I don't know how to assign the start position and end position, maybe that would be all of it? From beginning to end? This really only matters for ifs!

jpelay:
Also, what about prefixes? We add a bunch of prefix code sometimes, shall we just ignore it for our purposes? Or maybe do a bit of post processing and shift the line numbers?

hedy_sourcemap.py
ToniSkulj: Yeah, sort of, i'm wrapping the function twice to be able to pass the _source_map_ parameter to the decorator, this is the only reason I did it like this. I believe there is also another way to achieve this.

ToniSkulj:
> This looks great, Toni, OMG! I do like the added complexity, let's keep it for now unless we see that it won't serve us any purpose in the future.

Yesss, good to hear you like it Jesús 😃!

> For the line numbers in Python I just added a counter for the current line number, and count the number of new lines so we now the length of the snippet. This is how it looks with the turtle:

Perfect!! This is an awesome, but stil simple, solution.

> In this case, I don't know how to assign the start position and end position, maybe that would be all of it? From beginning to end? This really only matters for ifs!

I think we can use something like a character count. Pretty much the same solution you have made for the lines but counting characters instead of lines.

> Also, what about prefixes? We add a bunch of prefix code sometimes, shall we just ignore it for our purposes? Or maybe do a bit of post processing and shift the line numbers?

Hmm, this is a good one. Maybe we could make it possible to generate the prefixes from the syntax tree? Right now we check for Pygame, Turtle and more at the end. If we could do that at the start maybe we can add custom prefix rules at the top of the syntax tree? This way we would also not need the prefixes in JS. Let me know what you think of this.

jpelay:
> I think we can use something like a character count. Pretty much the same solution you have made for the lines but counting characters instead of lines.

Yes, I like this Idea. I was thinking along the lines. This isn't pretty, but we can maybe do a special case, something like this:
```
if line has 'if' and 'else' do
    count characters between 'if' and 'else
    count characters after 'else    
 ```
 > Hmm, this is a good one. Maybe we could make it possible to generate the prefixes from the syntax tree? Right now we check for Pygame, Turtle and more at the end. If we could do that at the start maybe we can add custom prefix rules at the top of the syntax tree? This way we would also not need the prefixes in JS. Let me know what you think of this
 
 If we can find a way to do manipulate the AST then let's do it! Otherwise I think shifting the lines ain't so difficult? Less elegant for sure!

**Comment by ToniSkulj** [comment]
I changed some things. Firstly I added the source-mapper to all levels using a decorator to decorate the level classes. Instead of decorating all methods.

Secondly, I made the sourcemapper map the Python code after all code is generated, this may not be an efficient way but solves a lot of problems! For instance, it immediately works for code blocks (which we introduce in level 8). An example:

```
1/0-1/32 --- name is ask 'What is your name?':
        1/0-1/35 --- name = input(f'What is your name?')

3/53-3/73 --- print 'Welcome Hedy':
        3/111-3/133 --- print(f'Welcome Hedy')

4/78-4/116 --- print 'You can play on your computer!':
        4/136-4/176 --- print(f'You can play on your computer!')

6/135-6/152 --- print 'INTRUDER!':
        6/185-6/204 --- print(f'INTRUDER!')

7/157-7/194 --- print 'You cannot use this computer!':
        7/207-7/246 --- print(f'You cannot use this computer!')

2/33-8/204 --- if name is Hedy
    print 'Welcome Hedy'
    print 'You can play on your computer!'#ENDBLOCK
else
    print 'INTRUDER!'
    print 'You cannot use this computer!'
#ENDBLOCK:
        2/36-7/246 --- if convert_numerals('Latin', name) == convert_numerals('Latin', 'Hedy'):
          print(f'Welcome Hedy')
          print(f'You can play on your computer!')
        else:
          print(f'INTRUDER!')
          print(f'You cannot use this computer!')

```

This is the sourcemap for level 8, using the second code example. There is one problem with this solution, that is currently the code does not take into account the code that appears twice (or more). This is fixable, by 'marking' the code we already have mapped.

If this solution is something that we want to go further with there's only a couple of things we need to do (famous last words):
- [ ] Fix code that appears multiple times
- [ ] Remove postprocessing done for Pygame code
- [ ] Give the mapping back to client (front-end)
- [ ] Currently, only commands are mapped, I see no reason why we could not extend this to all rules.

After that, we should have a full blown source-mapper !! unless I'm forgetting something..

**Comment by ToniSkulj** [comment]
Hi @jpelay! I Added [Add basic response to client](https://github.com/hedyorg/hedy/pull/4094/commits/43f88fe4f98c24ae6e227aca7b56bca86c13590f) as a simple response object for the front-end. Let me know if you think we should compress it even more (I am not sure how) or if we should make it more informative.

**Comment by ToniSkulj** [comment]
A couple of updates:

- fixed an edge case problem ([Fix find returning -1](https://github.com/hedyorg/hedy/pull/4094/commits/bcff24c5ce997ff58799738c59eae7ee4ab38fe6))
- Added tests ([Add tests](https://github.com/hedyorg/hedy/pull/4094/commits/ab37a74b7779a49ab0c268156e30f62e99f2a170))
- Removed some unnecessary prints and added some informing comments ([Remove debug prints](https://github.com/hedyorg/hedy/pull/4094/commit/b0efba614ba78d87f32571bd294e400a6f47015f) & [Add comments & docstrings for clarification](https://github.com/hedyorg/hedy/pull/4094/commits/8412750a7d35200440bdadc83b4e251ec660abd8))

I think we are pretty much there, we now have a fully functional (basic) source mapper!!

@jpelay, I was thinking about the extra information that we wanted to add where every entry in the source map would indicate if it would be a full line or not. Maybe we should implement that later? I think the basic source-mapper that we have in this PR should be enough for both you and me to continue for both our goals. That way the added functionality would be in the appropriate PR. Let me know!

Anyway, the basic functionality is there, that's why I'm opening this PR for review.
@jpelay, @Felienne hope you can check it out when you have the bandwidth for it!

File: tests/test_level/test_level_01.py [hunk 2]
```diff
@@ -642,3 +642,23 @@ def test_non_keyword_with_argument_gives_invalid(self):
             extra_check_function=lambda c: c.exception.arguments['invalid_command'] == 'aks',
             max_level=17,
         )
+
+    def test_source_map(self):
+        code = textwrap.dedent("""\
+        print Hallo welkom bij Hedy!
+        forward 50
+        ask Wat is je lievelingskleur
+        echo je lievelingskleur is""")
```
jpelay, line 651: Can we add Python code in this tests? I think that is better if the Python code is next to the Hedy code and we can compare. Therefore, this will also document the Source Mapper for people who don't know how it works, which is also one of the purposes of the tests

jpelay:
Also, as I commented on Discord, in the future we need to purge the `#ENDBLOCK`s from being added to the sourcemap, as they're not really part of the code, but rather a hacky way of parsing indentation.

tests/test_level/test_level_01.py
ToniSkulj: Great idea! [Add testing Python code to the source-mapper tests](https://github.com/hedyorg/hedy/pull/4094/commits/b8cc9d624eee1c829a71844145b96968d75ee406) fixes this.

ToniSkulj:
> Also, as I commented on Discord, in the future we need to purge the `#ENDBLOCK`s from being added to the source map, as they're not really part of the code, but rather a hacky way of parsing indentation.

Hi @jpelay! [Remove #ENDBLOCK from Hedy code](https://github.com/hedyorg/hedy/pull/4094/commits/1e3647fcae4693208d30188bf1d6bf08beba3794) fixes this.
Hopefully, we can merge the source-mapper now to test it in production!

Felienne:
> Also, as I commented on Discord, in the future we need to purge the `#ENDBLOCK`s from being added to the sourcemap, as they're not really part of the code, but rather a hacky way of parsing indentation.

Are you calling my work hacky 🤣?!!

Felienne:
Hi @ToniSkulj and @jpelay! I just deployed main to hedy.org, so feel free to merge this so it can be tested on alpha!

jpelay (approved):

Issue #4002, opened 2023-02-02
[CONTENT] Improve the quizzes


MarleenGilsing:
The quizzes need to be improved, so I'll take care of that. 

Note to self: 
En de preview is /quiz/preview-question/<level>/<question>
Dus bijvoorbeeld: http://0.0.0.0:8080/quiz/preview-question/6/1

Issue #2380, opened 2022-03-30
[Content] make sure all blanks use the same symbol (and explain it!)


thjazi39:
Blanks are not recognized on valid lines at levels 1-3:
![image](https://user-images.githubusercontent.com/48509466/160877692-c55c0405-3c0e-4c9a-b637-ed3120d6662d.png)

This is the opposite of bug #2359, because here the syntax highlighting is not applied, but the interpreter considers that YES, it should be colored (while in the case of bug #2359, it considers that the # should not be colored)
![image](https://user-images.githubusercontent.com/48509466/160877897-2e03f835-8834-495a-9827-7401ab445694.png)

As a result, the previous code, which is not excutable, cannot give the following result
![image](https://user-images.githubusercontent.com/48509466/160877987-aa2b40ac-2b7b-4ebf-91b7-1cb52e43b8b4.png)

jpelay:
The current implementation doesn't allow to print blanks, not even on further levels. We preprocess blanks before they're fed to the transpiler (the program that makes Hedy code Python code). Maybe a solution to this is to not detect blanks that are inside a print? But this still leaves the issue that at levels 1-3 we can't distinguish string and vars, so maybe in this case this is actually what we want: you can't print blanks because blanks mean that you need to fill in some code there.

Felienne:
Hi both!

I would say the highlighter is ok as it is, I consider this a content issue, we are content creators can make sure we don't use blank in that way, kids will never type blanks anyway (of they do, they don't mean them as blanks!)

Felienne:
> Ok, so in that case, I have to correct the syntax highlighting so that the blanks are highlighted? (so that the kids understand where the error comes from)

No, leave this one for now. As I said I consider it a content issue.

The other changes to the highlighter (less overcoloring) are a lot more important, focus on those first

Felienne:
Just a note to self, currently, both _ and _?_ are highlighted as blanks:

<img width="67" alt="image" src="https://user-images.githubusercontent.com/1003685/188414869-d62b1de9-182a-4670-a7f3-d53bc09dab01.png">

So the issue that remains here is to make sure: 
1. All locations use the same blank and
2. Explain what it means to users

jpelay:
I feel the same way, this is not a bug it's a feauture!

thjazi39:
> Just a note to self, currently, both _ and _?_ are highlighted as blanks:

[...]

PR #4098, opened 2023-03-09
[CONTENT] Quizzes update

### LINKED ISSUE #3277: [CONTENT] Quiz question 1 level 4 (Dutch) multiple correct answers, only one is allowed
Question 1 level 4 (Dutcht) answer 2 "In level 4 werkt ask anders dan in level 2" is marked as wrong and answer 1 "In level 4 werkt print anders" is marked correct. If 1 is correct 2 should be correct as well.

image

MarleenGilsing:
The quizzes desperately needed to be looked over again. I'm done upto level 7, level 7 and higher are for another time. They are on my to do list as issue #4002 

Also fixes #3277 and #2380

### DIFF
--- a/content/quizzes/en.yaml
+++ b/content/quizzes/en.yaml
@@ -1318,7 +1318,7 @@ levels:
                 feedback: "No you can repeat a line."
             -   option: "1"
                 feedback: "Correct, one line at a time"
-            -   option: "2"
+            -   option: "3"
                 feedback: "In this level only one line at a time"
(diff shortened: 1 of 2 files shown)


MarleenGilsing:
The problem is that I've changed question 10 of level 1. It used to only have 2 answer options, but now it has 4. The other languages don;t have that (except for dutch now).

Felienne (approved):
Let's hope this does not annoy Weblate too much!

PR #4120, opened 2023-03-12
Fix kurdish translations

### LINKED ISSUE #4119: [CONTENT?] Kurdish shows placeholders!
This is an interesting WTF:

image

This is broken on the site and locally too. Of course I ran

` hedyorg % pybabel compile -f -d translations`

image

It even tells me it has compiled kmr but places holders are still shown.

Not sure who can dive into this, maybe @jpelay or maybe @TiBiBa?

We could temporarily remove kmr as a language so that this ugly view is not shown?

And also, we must add a test to detect place holders on the site, even though we do `pybabel` on the server so "this should never happen"... apparently is does happen!

TiBiBa:
**Description**
Fixing the language gap between weblate and babel by manually overwriting the Kurdish language key.

**Fixes**
This PR fixes #4119

**How to test**
Check that no placeholders are shown for Kurdish.


Felienne (approved):
Thanks for the quick fix @TiBiBa!!!

image

Issue #3388, opened 2022-10-07
[Language idea] Use meta in all exceptions


Felienne:
**Language idea incl level**
Since #3155 we have access to meta information (aka location of a parse error) everywhere in the transpiler so we can improve upon some error messages, such as missing quote (which now only shows the problematic string in the message and then the user has to go on a hunt for where that string is!)

Felienne:
Here is a concrete issue to improve with the use of meta, Spanish, level 3:

```
comida es pizza, helado, ensalada, sandwich
toppings es chocolate, queso, atún, chispas
imprimir Tu comida aleatoria para hoy es...   <-- the error is here but no line is indicated
dormir 2
imprimir comida en aleatorio con toppings en aleatorio 
```

PR #4128, opened 2023-03-15
[LANGUAGE] Add line number to InvalidArgumentType

Felienne:
Work towards #3388

### DIFF
--- a/exceptions.py
+++ b/exceptions.py
@@ -99,12 +99,13 @@ class CyclicVariableDefinitionException(HedyException):
 
 
 class InvalidArgumentTypeException(HedyException):
-    def __init__(self, command, invalid_type, allowed_types, invalid_argument):
+    def __init__(self, command, invalid_type, allowed_types, invalid_argument, line_number):
         super().__init__('Invalid Argument Type',
                          command=command,
                          invalid_type=invalid_type,
                          allowed_types=allowed_types,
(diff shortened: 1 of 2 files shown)


Felienne:
Ok we are still missing a few but I think we can merge this because I am out of time now and this already adds a lot of value!!

jpelay:
Hi Miss Felienne! This is great! The only thing needed was to add the meta.line after all?

Felienne:
> Hi Miss Felienne! This is great! The only thing needed was to add the meta.line after all?

Yes, since we already had meta available everywhere thanks to previous efforts from @boryanagoncharenko! Can't believe this took me 6 months to pick up! There are a few rarer exceptions (LonelyEcho, for example, and blank in som cases) that need a bit more work since they rely on a hand-build function rather than a parsing but I will take a look at that also.

jpelay:
> > Hi Miss Felienne! This is great! The only thing needed was to add the meta.line after all?
> 
> Yes, since we already had meta available everywhere thanks to previous efforts from @boryanagoncharenko! Can't believe this took me 6 months to pick up! There are a few rarer exceptions (LonelyEcho, for example, and blank in som cases) that need a bit more work since they rely on a hand-build function rather than a parsing but I will take a look at that also.

I see! Still this improves the error so MUCH with respect to our previous efforts! Specially the unquoted text!

PR #4129, opened 2023-03-15
[UI] Start customizations at minimum level

jpelay:
Deletes the default option "Select a level" from the level dropdown in customize class page


jpelay:
This PR changes the past behavior that the customization levels had, showing an empty box when entering the page or disabling a level. Now, that empty box is never shown, instead if the class have customizations the minimum level enabled is shown by default, otherwise the level 1 is shown. 

I also changed that when you disable the current level you're editing a message is displayed, letting you know that you can't edit this level since it's been disabled:

image

Felienne (approved):
Ahhh the grey is so much better!

Speaking about colors? Shall we give a different one to levels that open later on (see #4170)

PR #4131, opened 2023-03-16
Quizzes update level 8 - 13

MarleenGilsing:
Level 8 - 13 quizzes update in NL en EN


Felienne:
> Oops I see I continued this update on the same branch as last week. Hope that doesn't cause any problems.

It is better to avoid it, but GitHub will be able to understand the changes in most cases.

content/quizzes/en.yaml
```diff
@@ -2593,10 +2593,10 @@ levels:
             code: |-
                 courses = appetizer, main course, dessert
                 names = Timon, Onno
-                {for} name {in} names
-                	{for} course {in} courses
-                        	food = {ask} name ', what would you like to eat as your ' course '?'
-                        	{print} name ' orders ' food ' as their ' course
+                {for} name in names
```
Felienne, line 2596: Ah I think here merge issues have caused some good changes to also be gone, I will take a further look!

Felienne:
Superseded by #4143

PR #4146, opened 2023-03-18
[LANGUAGE] Improve pressed command

### LINKED ISSUE #4138: [BUG?] Trying to figure out pressit
Not sure if it's a bug or my lack of programming skills. I'm trying to figure out how the pressed command works in level 5 and I'm confused: 

In this code, I can't type a and then b. Because it stops after pressing a. 
```
if a is pressed print 'a'
else print 'Hallo'
if b is pressed print 'b'
else print 'doei'
```

In this code, I can press a and then b. 
```
if a is pressed print 'a'
else print 'Hallo'
print 'test'
if b is pressed print 'b'
else print 'doei'
```

Is this a bug? 

And is there a reason that in the higher levels the pressit adventures is not online yet? It's not in the hedy_content.py , is that intentional, or should I add it.

ToniSkulj:
Fixes #4138,

This PR changes how the _ifpressed_ command works. 
**From level 5 till level 15: it is required to use an else**

```
if x is pressed print "x!" else print "other one"
```

**From level 15: using it without else is allowed**
```
if x is pressed 
  print "x!"
```

**From level 17: using it with elif is possible (The drawing game !!)**
```
repeat 20 times
    if w is pressed:
        forward 20
    elif a is pressed:
        turn -90
    elif d is pressed:
        turn 90
    elif s is pressed:
        forward -20
```

I added an error when the user forgets to add the else in level 5 till 15, still need to figure out how to make translatable messages
afbeelding

**Still To-Do:**
- [x] Add/Fix tests


ToniSkulj:
Hi @Felienne, I'm not sure if I did [Add translatable error message](https://github.com/hedyorg/hedy/pull/4146/commits/78ab3eccf96d2a1c0a82731d9c2ad672646b1e13) correctly. Hope you have the bandwidth to check it out.

Felienne:
Thanks for this @ToniSkulj!! I think we are on the right track, the levels up to 15 now totally make sense! I only made a few cosmetic changes to the error messages.

For level 15, I am not sure what we have as semantics: Should one press be able to stop a loop? I would think so, but the current implementation does not seem to agree:
 
image

ToniSkulj:
> Thanks for this @ToniSkulj!! I think we are on the right track, the levels up to 15 now totally make sense!

Yeahh, good to hear @Felienne !!

> Maybe it just sees the first line as body (I did not read the generated Python)

This was exactly the problem, good catch! I fixed it in [Fix level 15 missing multi lines](https://github.com/hedyorg/hedy/pull/4146/commits/f9a6bd345a167600fb6914448a333bd00f7a1112)

Felienne:
Thanks for the progress @ToniSkulj! I will fiddle a bit more with the adventures.

I did notice one thing, in a turtle program, I do not get the 'waiting for key press' bar. Is that to be expected?

image

``` (level 5)
if w is pressed forward 100
else turn 90
```

Felienne:
It also seems that in some cases, we do not get the missing else error message?

image

I have not dived into why that is the case!

Felienne:
I have found the cause, our old enemy the newline!

With:

image

Without:
image

ToniSkulj:
> I have found the cause, our old enemy the newline!

Oh no my archnemesis!! Luckily, defeated with [Remove ifpressed preprocessing](https://github.com/hedyorg/hedy/pull/4146/commits/3eab3501e7c373c4b2b50f5eae9e20f59ce3896f). We really do not want to add an _else_ when forgotten.

> I did notice one thing, in a turtle program, I do not get the 'waiting for key press' bar. Is that to be expected? 

I believe I did this on purpose so that the canvas size (for the turtle) would not be decreased. Let me know if we want to change this @Felienne, it's easy to do that.

Felienne:
> Oh no my archnemesis!! Luckily, defeated with [Remove ifpressed preprocessing](https://github.com/hedyorg/hedy/pull/4146/commits/3eab3501e7c373c4b2b50f5eae9e20f59ce3896f). We really do not want to add an _else_ when forgotten.

Great thanks!!

> > I did notice one thing, in a turtle program, I do not get the 'waiting for key press' bar. Is that to be expected?
> 
> I believe I did this on purpose so that the canvas size (for the turtle) would not be decreased. Let me know if we want to change this @Felienne, it's easy to do that.

Yeah, let's do that. We do it also for ask after all with the turtle, shrinking the field too. I think it would be more clear what is going in.

I will play a bit more with it and add a few simple adventure steps!

ToniSkulj:
> Yeah, let's do that. We do it also for ask after all with the turtle, shrinking the field too. I think it would be more clear what is going in.

Hi @Felienne , [Add PygameModal when Turtle is used](https://github.com/hedyorg/hedy/pull/4146/commits/9ef7d4c12f414b19fa8d311db1946123f2ff59c2) fixes this!

Felienne:
@jpelay: Can you take a look at the Spanish version of pressit? Thanks!

Felienne:
Otherwise, we are good to go!

jpelay:
> @jpelay: Can you take a look at the Spanish version of pressit? Thanks!

Hello @felienne. I fixed the translations that were already there and added the translation for level 9!

Felienne:
Great, I will merge Weblate, fix the conflicts that arise and then we are finally good to go!!

@ToniSkulj I totally owe you a coffee and a piece of pie!

PR #4147, opened 2023-03-18
[TESTING] Typescript instrumentation for code coverage

reedspool:
**Description**

Fixes typescript implementation in code coverage. The trick, @jpelay, was to add the instrumentation to the TypeScript code before compiling it. I also only compile with esbuild, but I believe you can interject `tsc` as you had it before, as long as you are running `tsc` on the instrumented TypeScript. I would test all of this but I ran out of time :)

I fear this does NOT properly manage appbundle.js so someone needs to take these changes and fix the script.

Discussion in PR https://github.com/hedyorg/hedy/pull/4060

**How to test**

1. Run the code coverage script as normal
2. See that you are able to inspect source of TypeScript files in `static/js` on code coverage webpage (see screenshot below)
3. See if there is now an `appbundle_orig.js` untracked in your repo. If so, then this script is messing with that incorrectly.

Screenshot from 2023-03-18 00-52-45


reedspool:
I'm excited to learn something because I am clueless about what the exclamation point does in a shell script like that. Can you explain or point to some documentation about that syntax? I believe you that it lets the script continue, but I'd like to know how.

I think you missed that there's no `__tmp__` directory created anymore :)

jpelay:
> I'm excited to learn something because I am clueless about what the exclamation point does in a shell script like that. Can you explain or point to some documentation about that syntax? I believe you that it lets the script continue, but I'd like to know how.

It basically bypasses the -e flag in the shell script, allowing the script to continue even if that line fails. I think it's better this way since even if the tests fails appbundle will allways return to its original state and we will never have a pesky appbundle_orig. Didn't notice that! Will remove that line.

jpelay:
> https://unix.stackexchange.com/a/437824 this is the only doc I've been able to find on that usage. It says it negates the exit status. Wouldn't that mean that if that line exits without error, that the script would halt there?

They explain it in the Stack Overflow answer I linked!

> What you are doing is to by-pass the set -e flag in the command1. Why?
> * if command1 runs properly, it will return a zero status. ! will negate it, but set -e won't trigger an exit by the because it comes from a return status inverted with !, as described above.
> * if command1 fails, it will return a non-zero status. ! will negate it, so the line will end up returning a zero status and the script will continue normally.

jpelay:
> Darn I thought I could delete my silliness before you saw it :) you nailed it. I read your comment before coffee and missed your informative link entirely. I apologize :)

Don't worry!

jpelay (approved):

Issue #4102, opened 2023-03-10
Deprecate/remove `generate-development-css` script and update `CONTRIBUTING.md`


reedspool:
The [Contributing Doc](https://github.com/hedyorg/hedy/blob/main/CONTRIBUTING.md) suggests one should use `./build-tools/heroku/tailwind/generate-development-css` during development, and then (in the same dir) `generate-css` before merging.

But in almost a year, no one has thought to add the really-very-useful `--watch` mode to `generate-development-css`, after it [was added](https://github.com/hedyorg/hedy/pull/2842) to the `generate-css` script. That signals to me that no one is actually using the script, making it useless cruft and misleading to newbies like me as that's what the docs say to use.

Please let me know if I'm wrong and this script is in use or does something useful that the other script doesn't do. If my hunch is correct, I'll gladly make a PR to remove the unused script and update the contributing doc.

Felienne:
Good catch @reedspool, I think there might be a use for this but I am not sure. @rix0rrr will know I think!

rix0rrr:
Hey @reedspool, thanks for the interest in improving the developer experience!

> That signals to me that no one is actually using the script

Well, I use it all the time 😅 but I agree that it is confusing to have two scripts, especially since their functions are very similar and a computer could figure out when it needs to do one thing and when the other.

Here is the the reason for this script existing (if this rationale is not in the contributing guide, we should probably add it):

Our goal currently(*) is that a developer should only need git and Python, and only need to check out the repository and run `python app.py` to get up and running with Hedy. This is to encourage contributions with a minimum of hurdles. That means that we can't require any additional build steps to be run before starting local development. *That* in turn means that generated artifacts should be committed to the repository in a "good to go" state.

Tailwind has "strip unused CSS classes" feature to minimize the size of the CSS file, which is great. But if we commit a stripped version of the CSS to git, then any time you want to use a new class that hasn't been used before, your HTML will not work and you need to regenerate the CSS file. So we have decided to commit the multi-megabyte "full" CSS File to git, so that, while developing, all classes are available. The minimized file will be generated on deployment to the server.

It's `generate-development-css`'s job to generate the full file, and `generate-css`'s job to generate the stripped file. Why someone has added `--watch` to a script that should only be run on a build server, I don't know 😅, but it kinda shows that having these two scripts was probably a bad idea on my part, since the distinction is subtle and hard to explain.

I've been thinking about solving this already, but never got around to doing it. What do you think of the following:

[...]

PR #4148, opened 2023-03-18
[REFACTOR] Clearly distinguish prod and development css

rix0rrr:
After the discussion in #4102 and #4130, I will try to clean up the mess I made!

Here is the new situation:

* We now have two different scripts, `generate-prod-css` and `generate-development-css`.
* These generate into two different files, `generated.full.css` and `generated.css`. The `generated.css` file is `gitignore`d.
* Depending on whether we are in debug mode or not, the right file is referenced in the HTML.

Hopefully, this will get rid of all the confusion surrounding the purpose of these files!

@TiBiBa, @reedspool, what do you think?


rix0rrr:
Here's another thought: we can potentially NOT minify the files that get committed to version control, to reduce the chances of merge conflicts in them!

TiBiBa:
> Here's another thought: we can potentially NOT minify the files that get committed to version control, to reduce the chances of merge conflicts in them!

Would love this! Not minifying the development version makes it easier to understand for new contributors as well, with the addition that CSS changes in a PR can be clearly distinguished in the file.

Felienne (changes requested):
Setting this to changes requires for the not minifying (otherwise it is too hard for me to keep track of the PRs)

reedspool:
Nice! Could consider `.gitignore` for the deployment CSS now.

This definitely feels like a great step especially and much more informative docs!

CONTRIBUTING.md
```diff
@@ -204,29 +204,47 @@ Make sure to reload your browser (and work in incognito mode) to see the changes
...
+library](https://tailwindcss.com/docs/installation).  This library has generated
+classes for styling which we can apply to HTML elements.
+
+You normally do not need to think about this. During development, a multi-megabyte
+CSS file will be served that contains most classes. During deployment, a minimized CSS
+file is automatically produced.
+
+You may need to regenerate the development CSS file if you want to do one of the following things:
+
+* Want to use a *conditional* Tailwind class (for example, a class that starts with `hover:`).
```
rix0rrr, line 218:
```suggestion
* Use a *conditional* Tailwind class (for example, a class that starts with `hover:`).
```

Felienne:
> LGTM!

Do you have enough rights to officially approve the PR? You have my blessing to do so!

reedspool:
I do not see a button to do so

PR #4152, opened 2023-03-20
Created the initial page to host the live dashboard for the New Media Lab Course

HcKide:
created an html file which will house all of the dashboard content
edited the existing class overview page for teachers, to include a button to the live stats
added a route to the statistics.py file for the live stats page.

hedy_Rad

### DIFF
--- a/templates/class-live-stats.html
+++ b/templates/class-live-stats.html
@@ -5,7 +5,6 @@
 {% block regular_content %}
 
 <div>
-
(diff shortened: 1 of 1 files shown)


templates/class-live-stats.html
```diff
@@ -0,0 +1,22 @@
+{% extends "auth.html" %}
+
+{% from "macros/stats-shared.html" import insert_accordion_tab %}
+
+{% block regular_content %}
+
+<div>
+
```
karlijnbok, line 8:
```suggestion
```

PR #4156, opened 2023-03-20
[UI] Hide submit button on code page when not in class

### LINKED ISSUE #4154: [UI idea] Hide submit option on code page when not in a class

TiBiBa:
**Description**
The title should be self-explanatory, when the current user is not in a class, hide the submit button. Similarly we might want to hide the bar all together when a user is not logged in. But I'm not sure if that is desirable. I did find two new bugs when working on this, still have to make issues on them:

- When logged in and navigating to a join link, you are logged out (this should not be the case!)
- When joining a class the value on `current_user()` is not updated, the user is required to log out / log in again to set this right.

**Fixes**
This PR fixes #4154.


Felienne:
> **Description** The title should be self-explanatory, when the current user is not in a class, hide the submit button. 

Yes it is a great idea to hide the submit icon when not in a class.

> Similarly we might want to hide the bar all together when a user is not logged in. But I'm not sure if that is desirable.

I think I am ok with how it is now, so the used sees the same UI all the time, that might helps understandability. Kids that are not logged in might otherwise panic that buttons are gone.

> I did find two new bugs when working on this, still have to make issues on them:
> * When logged in and navigating to a join link, you are logged out (this should not be the case!)
> * When joining a class the value on `current_user()` is not updated, the user is required to log out / log in again to set this right.

Ah thanks for the test drive @TiBiBa!!!

rix0rrr:
> When logged in and navigating to a join link, you are logged out (this should not be the case!)

I'm not sure how to reproduce this. We tried the obvious ways but that doesn't seem to do it. While trying to repro I did find some other issues that I fixed in #4157.

rix0rrr (approved):

PR #4157, opened 2023-03-20
[FIX] Improve logout experience

rix0rrr:
Previously, when you log out, you would be shown a JSON error message (because the browser will send `Accept: */*`, which includes JSON, and that's what we feature tested for). The goal was to respond with JSON to the Ajax requests. Fix by feature testing for yes JSON, but not HTML.

Also, the previous PR made it so that if you log out, you stay on the same page. But if this page is now a "my profile" page or "programs" page, it shows you a very big unfriendly "you are not logged in message".

Add a new `@requires_login_redirect` decorator, which redirects you back to the front page if you end up not logged in on certain pages.

**How to test**

Log in, go to your profile page, then log out. You end up back at the front page.

### DIFF
--- a/app.py
+++ b/app.py
@@ -40,7 +40,7 @@ from website import (ab_proxying, achievements, admin, auth_pages, aws_helpers,
                      profile, programs, querylog, quiz, statistics,
                      translating)
 from website.auth import (current_user, is_admin, is_teacher,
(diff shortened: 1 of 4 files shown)


Felienne (approved):
Thanks for the fix, this indeed solves the ugly error message on logout

PR #4159, opened 2023-03-20
[FIX] Student needs to wait 5m or logout after joining class

rix0rrr:
The classes that a student is part of is cached on their session (so we don't have to query the database on every page load).

However, we failed to refresh the cache when the user joins a class. The result was that they had to wait 5 minutes for their session data to be refreshed as usual, or log out and back in to force a refresh.

Instead, do the refresh on joining and leaving a class.

Thanks for catching this @TiBiBa!

**How to test**

Invite a new student that's not in any classes. Join as that student, observe that you can submit programs immediately (instead of after 5 minutes)

(Fixing a bug mentioned in #4156)

### DIFF
--- a/website/auth.py
+++ b/website/auth.py
@@ -94,9 +94,10 @@ def password_hash(password, salt):
 # You can remove the current user from the Flask session with the `forget_current_user`.
 def remember_current_user(db_user):
     session["user-ttl"] = times() + 5 * 60
-    session["user"] = pick(db_user, "username", "email", "is_teacher", "classes")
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -1153,6 +1153,8 @@ def index(level, program_id):
     max_level = hedy.HEDY_MAX_LEVEL
     level_number = int(level)
 
+    print(current_user())
```
Felienne, line 1156:
```suggestion

```

I think we can remove this debug?

Felienne (approved):
Yes, works now! Thanks!!

PR #4160, opened 2023-03-20
[UI] Translate tooltips for new buttons

rix0rrr:
The tooltips for the new "share" and "hand in" buttons were accidentally not translated. Translate them as well.

**How to test**

Pick Dutch. Hover over the new buttons next to the program name.


Felienne:
Thanks!!

The English user facing text calls it "exercise" so I think it is more readable if we use that term in the code too:

image

I fixed that everywhere in the last commit

Felienne (approved):
Let's hope this does not annoy Weblate too much!

PR #4158, opened 2023-03-20
[FIX] Handle incomplete database rows

rix0rrr:
Not all database rows have all columns they should have. The current code that was dealing with database rows was assuming they did.

Remove that assumption, change `r['name']` into `r.get('name', '')` so that the code doesn't throw.

Fixes #4153.

**How to test**

It's not clear how to reproduce this, it depends heavily on what's in the database.

PR #4161, opened 2023-03-21
[FIX] Survive another missing program name

rix0rrr:
Similar to #4158, there was another place where we were assuming that all programs have a `name` field, which is not true in the actual database.

A couple more fixes around error reporting:

- The `Accept` header may not contain charsets in the MIME type, so remove them. This caused the content negotation to not work, and the error page to still render HTML.
- When we are rendering the error as JSON, include the exception message. This will be good for technical debugging, and make it so we may not have to go into the server logs to find exceptions.
- Get an appropriate error message out of the server response.

**How to test**

Log in to alpha, switch between private and public.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -11,6 +11,7 @@ import string
 import random
 import uuid
 import unicodedata
+import traceback
 
 from flask_babel import gettext, format_date, format_datetime, format_timedelta
 from ruamel import yaml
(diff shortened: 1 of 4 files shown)


utils.py
```diff
@@ -277,7 +278,9 @@ def error_page(error=404, page_error=None, ui_message=None, menu=True, iframe=No
 
     if request.accept_mimetypes.accept_json and not request.accept_mimetypes.accept_html:
         # Produce a JSON response instead of an HTML response
-        return jsonify({"code": error, "error": default}), error
+        return jsonify({"code": error,
+                        "error": default,
+                        "exception": traceback.format_exception(exception, None, None) if exception else None}), error
```
rix0rrr, line 283:
```suggestion
                        "exception": traceback.format_exception(type(exception), exception, exception.__traceback__) if exception else None}), error
```

Felienne (approved):

PR #4162, opened 2023-03-21
[FIX] Return newest state of record on update

rix0rrr:
The new program storing logic was assuming that `update` would return the updated state of the record.

This is an option in DynamoDB, but it is not the default.

Switch it on.

**How to test**

(can only be tested on alpha) Sharing a program will no longer throw an error.


Felienne (approved):
Let's see if this helps!

PR #4166, opened 2023-03-22
[UI] adds Hedy dashboard & common error detection

karlijnbok:
**Description**

* Adds a button to the `for-teachers/class/{{class_info.id}}` page to open the live statistics dashboard.
* Adds a live statistics dashboards page at `/live_stats/class/{{class_info.id}}`.


templates/class-overview.html
```diff
@@ -49,6 +49,7 @@ <h1>{{ class_info.name }}</h1><span class="cursor-pointer" onclick='hedyApp.rena
                 <button class="green-btn" id="add-student" onclick=$('#add_students_options').toggle();$(this).toggleClass('green-btn');$(this).toggleClass('blue-btn');>{{_('add_students')}}</button>
                 <button class="green-btn" id="customize-class-button" data-cy="customize_class_button" onclick="window.location.href = '/for-teachers/customize-class/{{class_info.id}}'">{{_('customize_class')}}</button>
             {% endif %}
+            <button class="green-btn" id="live_stats_button" onclick="window.location.href = '/live_stats/class/{{class_info.id}}'">{{_('class_live')}}</button>
```
Felienne, line 52: class_live I think is a new babel string? Be sure to run the pybabel extraction command (see docs) so it will be added to the pot files and can be translated.

Felienne:
Just took a peek at your progress and this is going amazing team!!!

image

website/database.py
```diff
@@ -775,12 +779,21 @@ def get_quiz_stats(self, ids, start=None, end=None):
 
     def add_program_stats(self, id, level, number_of_lines, exception):
         key = {"id#level": f"{id}#{level}", "week": self.to_year_week(date.today())}
-
         add_attributes = {"id": id, "level": level, "number_of_lines": number_of_lines}
+        p_stats = PROGRAM_STATS.get_many({"id": id, "week": self.to_year_week(date.today())})
+
+        chart_history = []
+        if p_stats.records and 'chart_history' in p_stats.records[0].keys():
+            chart_history = p_stats.records[0]['chart_history']
```
Felienne, line 787: This is not very defensive (if there is no 'chart_history' it will fail hard) can we change it into a get?

website/database.py
```diff
@@ -775,12 +779,21 @@ def get_quiz_stats(self, ids, start=None, end=None):
...
+        slice = MAX_CHART_HISTORY_SIZE if len(chart_history) > MAX_CHART_HISTORY_SIZE else 0
+
         if exception:
             add_attributes[exception] = dynamo.DynamoIncrement()
+            new_chart_history = chart_history + [0]
         else:
             add_attributes["successful_runs"] = dynamo.DynamoIncrement()
+            new_chart_history = chart_history + [1]
```
Felienne, line 795: same here, if chart_history would be a string it would fail. First cast to a list maybe?

website/database.py
karlijnbok: I didn't write the code, but doesn't the if statement above check for the key existing though? Or perhaps I don't understand what you mean exactly. But I will change it into a get, that's nicer anyway :)

static/images/new_media_lab/progress.png
Felienne, line 1: I think these images are no longer needed?

translations/en/LC_MESSAGES/messages.po
```diff
@@ -19,88 +19,183 @@ msgstr ""
 "Generated-By: Babel 2.11.0\n"
 
 msgid "Access Before Assign"
-msgstr "You tried to use the variable {name} on line {access_line_number}, but you set it on line {definition_line_number}. Set a variable before using it."
+msgstr ""
+"You tried to use the variable {name} on line {access_line_number}, but "
+"you set it on line {definition_line_number}. Set a variable before using "
+"it."
+
+msgid "Adventure Progress"
```
Felienne, line 27: Messages are added here but do not seem to be added in the general pot file. I suspect that is going to lead to issues in translation. Can you add them to the pot file too? That should just work with the extraction commands explained in contributing.md

Felienne (changes requested):

Felienne:
I find this icon a bit confusing for folding:

image

Could we use a plus and minus? Or > as in the teacher's guide:

image

Felienne:
I have made a few errors as a student, but I am not yet able to see them in the dashboard:

image

Can you help me understand what steps to take to see something?

HcKide:
> I have made a few errors as a student, but I am not yet able to see them in the dashboard:
> image
> 
> Can you help me understand what steps to take to see something?

We could not replicate exactly what you have I am afraid. In theory (upon reload) one should see the last program run of the student. So logging in as a student in an incognito tab while the dashboard is open, and then running a few programs (both correct and purposefully wrong) should show up in the dashboard upon reload. We have tested with a custom adventure and it shows up for us. Could you tell us which steps you are doing so we could potentially reproduce?

Felienne:
> > I have made a few errors as a student, but I am not yet able to see them in the dashboard:
> > image
> > Can you help me understand what steps to take to see something?
> 
> We could not replicate exactly what you have I am afraid. In theory (upon reload) one should see the last program run of the student. So logging in as a student in an incognito tab while the dashboard is open, and then running a few programs (both correct and purposefully wrong) should show up in the dashboard upon reload. We have tested with a custom adventure and it shows up for us. Could you tell us which steps you are doing so we could potentially reproduce?

A relogin fixed it!

image

One more question: what does the graph on the left mean? I suspect it is the trajectory of submissions? I find it a bit hard to grasp as a line graph since it is not really a relationship

HcKide:
> > > I have made a few errors as a student, but I am not yet able to see them in the dashboard:
> > > image
> > > Can you help me understand what steps to take to see something?
> > 
> > 
> > We could not replicate exactly what you have I am afraid. In theory (upon reload) one should see the last program run of the student. So logging in as a student in an incognito tab while the dashboard is open, and then running a few programs (both correct and purposefully wrong) should show up in the dashboard upon reload. We have tested with a custom adventure and it shows up for us. Could you tell us which steps you are doing so we could potentially reproduce?
> 
> A relogin fixed it!
> image
> 
> One more question: what does the graph on the left mean? I suspect it is the trajectory of submissions? I find it a bit hard to grasp as a line graph since it is not really a relationship

Almost, it is supposed to be a history of program runs, not necessarily submissions. From feedback gathered from our usability study it was fairly clear for most participants.

HcKide:
> I have made a few errors as a student, but I am not yet able to see them in the dashboard:
> image
> 
> Can you help me understand what steps to take to see something?

actually, there seems to be a bug for specifically the `echo before ask` error not showing up. The other error messages do show up from what I've tested.

Felienne:
> > One more question: what does the graph on the left mean? I suspect it is the trajectory of submissions? I find it a bit hard to grasp as a line graph since it is not really a relationship
> 
> Almost, it is supposed to be a history of program runs, not necessarily submissions. From feedback gathered from our usability study it was fairly clear for most participants.

I think the numbers on the lower axis throw me off a bit, what do they indicate? Run 1, run 2? In what timescale, in total? In the recent time (in this lesson)?

app.py
```diff
@@ -702,8 +702,9 @@ def transpile_add_stats(code, level, lang_):
         return result
     except Exception as ex:
         class_name = get_class_name(ex)
+        error_message = hedy_error_to_response(ex)['Error']
         statistics.add(username, lambda id_: DATABASE.add_program_stats(
-            id_, level, number_of_lines, class_name))
+            id_, level, number_of_lines, class_name, error_message))
```
Felienne, line 707: why do we need to add the error message here? We have it already in the program, so it can be fetched immediately form the id, in principle?

Felienne: The current dashboard already tracks error messages, so maybe this might even already be stored, but I am not 100% sure how that works.

website/database.py
```diff
@@ -773,14 +778,30 @@ def get_quiz_stats(self, ids, start=None, end=None):
         data = [QUIZ_STATS.get_many({"id": i, "week": dynamo.Between(start_week, end_week)}) for i in ids]
         return functools.reduce(operator.iconcat, data, [])
 
-    def add_program_stats(self, id, level, number_of_lines, exception):
+    def add_program_stats(self, id, level, number_of_lines, exception, error_message=None):
         key = {"id#level": f"{id}#{level}", "week": self.to_year_week(date.today())}
-
         add_attributes = {"id": id, "level": level, "number_of_lines": number_of_lines}
+        program_stats = PROGRAM_STATS.get_many({"id": id, "week": self.to_year_week(date.today())})
```
Felienne, line 784: A little bit of a comment that this code is used in the new dashboard would be nice

Felienne:
> > I have made a few errors as a student, but I am not yet able to see them in the dashboard:
> > image
> > Can you help me understand what steps to take to see something?
> 
> actually, there seems to be a bug for specifically the `echo before ask` error not showing up. The other error messages do show up from what I've tested.

Ah what a specific issue! I can also dive into this if needed!

static/js/statistics.ts
```diff
@@ -491,3 +493,145 @@ function updateChart(elementId: string, datasets: any[] ) {
...
+      plugins: {
+        legend: {
+            display: false
+        }
+      }
+    }
+  });
+}
+
+export function enable_level_class_overview(level: string) {
```
jpelay, line 567: This file is dedicated for functions related to statistics and chart updates. This functions relates to the control of the page in a teacher's page, so its natural place is in the file `teachers.ts`

jpelay: Another thing I noticed is that these buttons require a refresh in order to select the levels that I want. I think is better to do this without the need to press a button to refressh, you can trigger a partial refresh using HTMX (the preferred method to handle interaction in Hedy currently) or use JavaScript and show and hide elements.

static/js/statistics.ts
```diff
@@ -491,3 +493,145 @@ function updateChart(elementId: string, datasets: any[] ) {
...
+
+export function select_levels_class_overview(class_id: string) {
+  let levels: (string | undefined)[] = [];
+  $('.level-select-button').each(function() {
+      if ($(this).hasClass("green-btn")) {
+          levels.push(<string>$(this).val());
+      }
+  });
+
+  $.ajax({
```
jpelay, line 585: Maybe is a bit late for this, but I'd love this to be done using HTMX.

website/statistics.py
```diff
@@ -122,6 +126,610 @@ def get_program_stats(self, user):
...
+            customizations = {
+                "id": class_id,
+                "levels": [i for i in range(1, HEDY_MAX_LEVEL + 1)],
+                "opening_dates": {},
+                "other_settings": [],
+                "level_thresholds": {},
+                "sorted_adventures": sorted_adventures
+            }
+            self.db.update_class_customizations(customizations)
+            class_customization = self.db.get_class_customizations(class_id)
```
jpelay, line 173: If you are saving the customizations object you don't need to to get it from the database. This way you can save it right away with the field `dashboard_customization`

website/statistics.py
```diff
@@ -122,6 +126,586 @@ def get_program_stats(self, user):
...
+                {
+                    "student": student_username,
+                    "in_progress": quiz_in_progress,
+                    "finished": quiz_finished
+                }
+            )
+        quiz_info = _get_quiz_info(quiz_stats)
+
+        attempted_adventures = {}
+        for level in range(1, HEDY_MAX_LEVEL+1):
```
jpelay, line 382: I see this code is duplicated, can you put it in a function?

website/statistics.py
```diff
@@ -340,6 +924,181 @@ def _calc_error_rate(fail, success):
...
+        teacher_adventures_formatted[adventure['id']] = adventure["name"]
+
+    selected_adventures = {}
+    for level, adventure_list in customizations['sorted_adventures'].items():
+        adventures_for_level = []
+        for adventure in list(adventure_list):
+            adventure_key = adventure['name']
+
+            students_in_progress = []
+            for d in last_adventures[int(level) - 1]:
```
jpelay, line 971: I don't understand what the `d` variable level means, can you use a more descriptive variable name?

website/statistics.py
```diff
@@ -340,6 +924,181 @@ def _calc_error_rate(fail, success):
     return (failed * 100) / max(1, failed + successful)
 
 
+def _determine_bool(bool_str):
+    if bool_str == "True":
+        return True
+    return False
```
jpelay, line 930: This is rather silly, but a cleaner way to do this is the following:

```suggestion
return bool_str == "True":
```

templates/class-live-student.html
```diff
@@ -0,0 +1,97 @@
...
+
+
+
+            <!-- Last ran code -->
+            <div class="w-full text-center turn-pre-into-ace mb-4">
+                <h5 class="font-normal pb-1">{{_('Last_program')}}</h5>
+                {% if student_programs %}
+                    <pre level="{{ student_programs[0].level }}">{{ student_programs[0].code }}</pre>
+                {% else %}
+                    <p>No program run yet</p>
```
jpelay, line 59: Don't forget to turn this into a babel string

Felienne: Done! (reused a string, does not fit perfectly but I think it is ok)

translations/en/LC_MESSAGES/messages.po
Felienne: Fixed this myself!

Felienne:
Hi @jpelay!

I picked up a few of your suggestions, if you can move what you think is needed to HTMX, I think we could merge and launch!

Felienne:
> > Hi @jpelay!
> > I picked up a few of your suggestions, if you can move what you think is needed to HTMX, I think we could merge and launch!
> 
> I'm on it!

ooooooow no the conflicts!

jpelay:
> > > Hi @jpelay!
> > > I picked up a few of your suggestions, if you can move what you think is needed to HTMX, I think we could merge and launch!
> > 
> > 
> > I'm on it!
> 
> ooooooow no the conflicts!

working on them!

jpelay:
So with these changes I'm introducing HTMX  to load the levels and also the students. This way we don't need to update the whole page when selecting one of these options.

Another important part of the changes I'm introducing is that we don't need server variables to know weather to show UI elements, and rather I'm using Hyperscript for that. Therefore not needing a full refresh when hiding UI elements.

jpelay:
@Felienne With this last changes this is ready for deployment!

Felienne:
Hi Jesus! 

It seems there are two failing Cypress tests related to running the code. I have not checked out the issues locally but maybe you can take a peek?

jpelay (dismissed):

app.py
jpelay: We are indeed tracking the error messages already as exceptions, so the tracking of the program used to look like this:
```
    {
      "id#level": "student1#1",
      "week": "2023-28",
      "id": "student1",
      "level": 1,
      "number_of_lines": 12,
      "successful_runs": 5,
      "IncompleteCommandException": 2,
      "LonelyEchoException": 2
    }
```

And after this PR it looks like this:

```
    {
      "id#level": "student1#1",
      "week": "2023-28",
      "id": "student1",
      "level": 1,
      "number_of_lines": 9,
      "successful_runs": 4,
      "chart_history": [
        1,
        1,
        1,
        1,
        0,
        0,
        0,
        0,
        0,
        0
      ],
      "error_history": [
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print.",
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print.",
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print.",
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print.",
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print.",
        "Oops! You forgot a bit of code! On line 10, you need to enter text behind print."
      ],
      "IncompleteCommandException": 6
    }
```

So I guess we are indeed tracking the same thing twice and adding much more weight to the DB? What do you think Miss Felienne?

jpelay:
image
Changed the UI of the Pop-up a bit!

app.py
Felienne: Thanks for clarifying! Which table is that information from? Is it the logs?

I see how we will need to store the information to be able to chart.
It will make the DB a bit bigger but I don't tend to worry about it much. What could be worrisome:

is whether we have to do a lot of fetching this data, that will make the app slower
the fact that we are saving the error message: "Oops! You forgot a bit of code! On line 10" How will that work if people switch languages, or when the teacher has a different language set than a student (which I do)? Can we make that more generic? I am not sure why we even need to store that but I am not very up to date on this PR.

jpelay: That information comes from the table `program-stats`. Also, just did a test setting the language to Spanish to one of the students, and indeed this PR only works for English. That is due to this function:

https://vscode.dev/github/hedyorg/hedy/blob/radboud-new-media-lab/website/statistics.py#L848-L855
```Python
    def misconception_hit(self, error):
        # Currently not in use
        for misconception, keywords in self.misconception_groups.items():
            # Check if the current error is different from the last error;
            # errors that fall in same misconception group are considered same errors
            if error and any(keyword in error.lower() for keyword in keywords):
                return misconception
        return None
```

That checks if the words in the error match a group of missconceptions they created:
```Python
self.misconception_groups = {
            'Not a current level command': ['level'],
            'Incorrect use of command': ['cannot'],
            'Incorrect use of variable': ['variable'],
            'Unwanted spaces': ['Spaces', 'confuse', 'computers'],
            'Forgot commandos': ['forgot'],
            'Empty program': ['empty program'],
            'Typed something that is not allowed': ['entered', 'allowed'],
            'Echo and ask mismatch': ['echo before an ask', 'echo without an ask'],
        }
```

I think maybe we can replace this to use the Exceptions we are already storing, since it will work for every language and also are cheaper to fetch

jpelay: I think this is the only place where they use the exceptions they store in the DB. I do not completely understand the algorithm they used to gather the misconceptions, but I think I can change the way they get a misconception hit

Felienne: > That information comes from the table `program-stats`. 

Got it! That is a nice addition to the data we have here (the last few errors are findable now but it requires a lot of querying) so I am fins with adding this info to the DB.

> Also, just did a test setting the language to Spanish to one of the students, and indeed this PR only works for English. That is due to this function:
> 
> https://vscode.dev/github/hedyorg/hedy/blob/radboud-new-media-lab/website/statistics.py#L848-L855
> 
> ```python
>     def misconception_hit(self, error):
>         # Currently not in use
>         for misconception, keywords in self.misconception_groups.items():
>             # Check if the current error is different from the last error;
>             # errors that fall in same misconception group are considered same errors
>             if error and any(keyword in error.lower() for keyword in keywords):
>                 return misconception
>         return None
> ```
> 
> That checks if the words in the error match a group of missconceptions they created:
> 
> ```python
> self.misconception_groups = {
>             'Not a current level command': ['level'],
>             'Incorrect use of command': ['cannot'],
>             'Incorrect use of variable': ['variable'],
>             'Unwanted spaces': ['Spaces', 'confuse', 'computers'],
>             'Forgot commandos': ['forgot'],
>             'Empty program': ['empty program'],
>             'Typed something that is not allowed': ['entered', 'allowed'],
>             'Echo and ask mismatch': ['echo before an ask', 'echo without an ask'],
>         }
> ```

> I think maybe we can replace this to use the Exceptions we are already storing, since it will work for every language and also are cheaper to fetch

Ah yeah I see, this is almost a 1-1 mapping on our exceptions, so let's match these and make it more language independent and future proof (since we might also change phrasing of messages!)

jpelay: Ok! So I'm going to replace this with the exceptions we are already storing, or something similar, since that is the easiest way to make it language and phrasing independent.

jpelay:
I changed the way the misconceptions are gathered. I changed the name to exception types (suggestion by @Felienne) and also made them language independent by using the already tracked exceptions per program. This means that the `error_history` field is not used anymore and it can be removed what do you think @Felienne. 

Also, I haven't realized before that the commor exceptions types per class were stored in the customizations table. This is a bad idea, because they are two conceptually different entities, and also if the teacher deletes the customizations they will also delete this important bit of their class information! So I think this should go to the `classes` table. I wil ldo that tomorrow.

Felienne:
> I changed the way the misconceptions are gathered. I changed the name to exception types (suggestion by @Felienne) and also made them language independent by using the already tracked exceptions per program. This means that the `error_history` field is not used anymore and it can be removed what do you think @Felienne.

Yes, better to remove that field it was a bit of a weird way of storing!

> Also, I haven't realized before that the commor exceptions types per class were stored in the customizations table. This is a bad idea, because they are two conceptually different entities, and also if the teacher deletes the customizations they will also delete this important bit of their class information! So I think this should go to the `classes` table. 

Ah what a good catch @jpelay!! This indeed should be stored elsewhere. Maybe on the student rather than on the class (or on the program) after which we can gather them per student/per class (potentially with some caching in the DB otherwise it'll be too expensive?)

> I wil ldo that tomorrow.

<3

jpelay:
> Ah what a good catch @jpelay!! This indeed should be stored elsewhere. Maybe on the student rather than on the class (or on the program) after which we can gather them per student/per class (potentially with some caching in the DB otherwise it'll be too expensive?)

Sorry, I don't I don't fully understand this 😅. For clarification you mean storing the groups of exceptions a student have incurred per student right?

Right now they're stored line this:

```json
"class": {
  "common_errors": [
      {
          "label": "invalid_command_exception",
          "active": 1,
          "students": ["student1", "student2"]
      }
   ]
}
```

In which table do you think we could store this information? These groups are already gathered every time the teacher access the Class live page, they are actually being recomputed every time, they're stored in the database only in case the teacher decides to ignore that specific group of exceptions.

jpelay:
I created a new table where we store the common errors for a class. This table has the following structure:

```
"class_errors": [
    {
      "id": "5c39c2a936f24db1a4935c52fab77cd7", // the class id
      "errors": [
        {
          "id": 0,
          "label": "invalid_command_exception", // the label of the exception group. Must be messages.pot
          "active": 0, // weather the exception is shown to the teacher
          "students": [
            "student1" // list of students that have incurred in this group of exceptions
          ]
        }
      ]
    }
]
```

Also, I created a new function for better handling of the class customizations. I tested the page with classes with no customizations and it works now.

website/statistics.py
```diff
@@ -290,20 +279,651 @@ def get_grid_info(self, user, class_id, level):
...
+                # forcefully overwrite oldest error despite not being resolved and set oldest half of the db to
+                # inactive to free up space
+                # Todo: could use a better way to handle this
+                new_id = 0
+
+                for i in range(self.MAX_COMMON_ERRORS // 2):
+                    common_errors['errors'][i]['active'] = 0
+                self.db.update_class_errors(common_errors)
+
+        return new_id
```
jpelay, line 844: I don't know if this is really necessary. This enforces a limit of 10 exception groups per class, so if the students of a class have made exceptions belonging to 10 different groups, it'll delete the 5 previous ones to make room for new. I think this is for not to clutter the UI. What do you think, @Felienne, do we leave this?

Felienne:
Hi @karlijnbok!!

It took a while but we are ready to merge this, and it will appear on the website soon! Thanks a lot on behalf of the Hedy team!!

Felienne (approved):
Let's go!

PR #4173, opened 2023-03-23
[FIX] Fix loading teacher adventures

rix0rrr:
This PR fixes two issues around teacher adventure customizations:

- Class customizations had the `teacher_adventures` field, which indicate what adventures are available in a particular class. This list of adventures had been replaced with the `adventure_order` set, and the UI that would fill the `teacher_adventures` field has been removed, so the field wasn't being set anymore. However, the `load_customized_adventures` function was still assuming that it only needed to load teacher adventures from the `teacher_adventures` field (and hence, since that field would always be empty, no adventures would be loaded).

- Second, on the class customization page, adventures wouldn't show up in the "available adventures" dropdown right away. That's because the code that would add options to the dropdown wasn't being called on page load, but only when the user twiddled the other parts of the user interface.

This PR addresses both of these issues.

**How to test**

Create an adventure, create a class, add a student to the class, add an adventure to the class, log in as the student, observe the adventure.


Felienne (approved):

website/database.py
```diff
@@ -30,10 +30,45 @@
...
+# - levels (int[]): the levels available in this class
+# - opening_dates ({ str -> str }): key is level nr as string, value is an ISO date
+# - other_settings (str[]): string list with values like "hide_quiz", "hide_parsons"
+# - sorted_adventures ({ str -> { from_teacher: bool, name: str }[] }):
+#     for every level (key as string) the adventures to show, in order. If from_teacher
+#     is False, the name of a built-in adventure. If from_teacher is true, name is the
+#     id of a adventure in the ADVENTURES table. The id may refer to an adventure that
+#     has been deleted. In that case, it should be ignored.
+#
+# These not so much:
```
jpelay, line 63: Hi Rico! Thank you for taking care of this bug. This fields represent the following:

1. `level_thresholds`: the grade threshold to unlock the next level. I think it could be changed to just a field. Since it now only have the quiz field, maybe it could be changed to only be an int, instead of an object.

2. `adventures`: yes this field is deprecated, and when a teacher tries to customize a class that have the old method of saving customizations, there's a script that automatically converts them to the new way of saving them https://github.com/hedyorg/hedy/blob/1a2b9a4c7662e003d1d5d1be1e0b4771ddd2ac85/website/for_teachers.py#L221

PR #4175, opened 2023-03-24
[FIX] An error occurs when running another user's program

rix0rrr:
We now try to do saving as part of running -- but the upshot is that when you run another user's program, the save fails (because you don't have permissions to overwrite it), which also fails the run.

Catch and ignore this particular error.

**How to test**

Run a student's program


Felienne (approved):
Thanks for the quick fix!

Issue #3537, opened 2022-11-01
[CONTENT] Use {} everywhere around keywords


Felienne:
As pointed out in #3514 some languages do not yet use {} around keywords for legacy reasons:

https://github.com/hedyorg/hedy/blob/1dfa0d9e11dee5646fb79c7d7960ad27be504ca9/content/adventures/de.yaml#L1402

Would be great if someone could fix that with a bit of prepping (here or in Weblate, both is fine by me!)

ana3999:
Hello!
My partner and I are looking to contribute to Hedy and we thought that this issue might be a good first issue to tackle to familiarize us with the codebase! Would it be possible for us to claim this issue? 
Thank you so much!

Felienne:
Absolutely, we'd be very happy if someone picks this up. It would be nice
if you could do it in small steps (for example in a few languages at a
time, or for a few levels at a time) to minimize conflicts with other
translators in Weblate (our translation system)

Let us know if you need any help (either here or in Discord!)

ana3999:
Thank you so much Felienne! 
We are extremely excited to start working on this task! I have reached out to you on Discord about a couple questions we had about this issue

jpelay:
> Hi @jpelay!!
> 
> I know you were working on this today while translating, can we close it, are you convinced you caught them all or do we need to keep it open so we don't forget to check later?

The issue I was working on was different I think! Instead of languages not having the {} around keywords, it was caused by {} being around keywords in normal text! For example (not real, just an example):

'You can {ask} the user for clarification'
'{if} the user says yes, print hey, {else}...'

@TiBiBa said that this was caused by an automatic script and wasn't intetional. What I can do is review it in English and remove the {} where they don't make sense. But I think that's a separate issue from this.

PR #4182, opened 2023-03-25
[CONTENT] verification of the use of double quotation marks

### LINKED ISSUE #2269: [CONTENT] verification of the use of double quotation marks
In case the PR #2249  would be accepted, the quizzes would have to be checked to ensure that the answers are correct. And this in all languages.

There is this question that will have to be corrected, but there are surely others.
image

antoineveldhoven:
**Description**

I did a full check of content/quizzes/*.yaml to see if there where any options left with double quotations.
- Changed an option in "nb_NO"; used Google Translate to help me with the translation
- Rewrote all options for "nl" in order to be in line with other languages.

**Fixes #2269**

**How to test**

Go to Level 4; Quiz; Continue to question 2 check both Dutch and Norsk

**Checklist**
  

antoineveldhoven:
@prvInSpace helped me with a translation, will update this asap, do not merge yet.

antoineveldhoven:
@prvInSpace Thnx for helping me out with the translation!

Can you confirm "Det er ingen fnutter her." is something like "There are no quotation marks here!"

prvInSpace:
> @prvInSpace Thnx for helping me out with the translation!
> 
> Can you confirm "Det er ingen fnutter her." is something like "There are no quotation marks here!"

Yes, that translation is correct

antoineveldhoven:
Lgtm

Felienne:
Thanks!! I am marking in Weblate-lock and will merge #4151 first since sometimes Weblate gets conflicts with manually edited yamls

antoineveldhoven:
> Thanks!! I am marking in Weblate-lock and will merge #4151 first since sometimes Weblate gets conflicts with manually edited yamls

Ok, can somebody explain me in more detail the process of these changes / Weblate; perhaps it is something for the next Contributors Meetup.

Same goes for issue #3537; I want to pick this up as well, wanted to suggest to write a tool like /tools/check-yaml-structure.py in order to check if all languages confirm to a base file and do some sorting in order to increase readability, but I can use some more in depth info regarding the regeneration of these files and the used conventions etc.

Felienne:
> > Thanks!! I am marking in Weblate-lock and will merge #4151 first since sometimes Weblate gets conflicts with manually edited yamls
> 
> Ok, can somebody explain me in more detail the process of these changes / Weblate; perhaps it is something for the next Contributors Meetup.

The basic idea is that if someone translates a string in weblate, a PR is generated, and changes are pushed on the open PR until we merge it. If at the same time, manual translations are made in another PR, that can create merge conflicts (like any two PRs working on the same files) The issue is that conflicts with Weblate are harder to handle, since it will cause Weblate to go into locked state meaning no one can translate anymore (contrary to regular conflicts that are only annoying for their authors).

We therefore have the convention of marking content changes with label weblate-lock so it is easier to see. Every now and then, I merge the Weblate PR and then also quickly push in the other content PRs before someone translates and a new PR appears. I hope that helps! If not we can totally explain more in the meetup next week.

> Same goes for issue #3537; I want to pick this up as well, wanted to suggest to write a tool like /tools/check-yaml-structure.py in order to check if all languages confirm to a base file and do some sorting in order to increase readability, but I can use some more in depth info regarding the regeneration of these files and the used conventions etc.

Thanks for the initiative to look into #3537, it'd be great if we have a tool to check consistency there too! The idea is that all keywords occur in {} so they can be swapped out for local versions in the UI. F.e. in Dutch I can do this:

image

But if I switch the switcher, I see:

image

This is enabled by the {}. The issue with the content is that we did not always have this switching (way back when we only supported English keywords!) so not in all cases were the {} added around keywords. It was hard to automate, since you can also have a sentence like "we are going to print hello".

What is also relevant to know is that content can be outdated. Sometimes someone translates, but we update lateron and then we leave the translations in place until they are updated. Half/old translated is better than reverting to English, is usually our thinking. So expect to find a lot of inconsistencies with a checker, of which only some are actually missing {}.

Hope that helps (if not, do reach out on Discord, or let's discuss next week!)

PR #4189, opened 2023-03-28
[FIX] Fix customization showing wrong first level 

### LINKED ISSUE #4178: [BUG] Teacher customization shows default on first load
**Describe the bug**
When I first load my customization page, my chosen customizations do not show.

I have an adventure (this is user Hedy on alpha) called Henkie

image

If I go to my class Testklas 123, it is not there:

image

But then I select level 2 and then again level 1, and there it is:

image

It is not a problem for the students in the class, it seems to only show for the teacher.

jpelay:
The customizations page recently was modified to show by default the minimum available level. It caused an error showing the wrong minimum level when in the database the levels were not stored in order. Now, instead of comparing to the loop index, we compare to the level number.

**Fixes #4178**

**How to test**

Go to dev_database.json and change the order of the field sorted_adventures levels and put some other level first. It should still show the minimum level correctly.


Felienne:
Thanks for the fix @jpelay! Is this a fix for #4178 or is this a different issue?

jpelay:
> Thanks for the fix @jpelay! Is this a fix for #4178 or is this a different issue?

Yes! It is a fix for that issue!

Felienne (approved):
I cant reproduce this issue locally (but it also does not seem to break new things) so I will merge and test on alpha where I know the issue exists and how to replicate it!

Felienne:
I just checked on alpha and indeed the issue as I notices it before has been fixed!! 

image

PR #4191, opened 2023-03-29
[FIX] Parsons puzzles were not randomized

### LINKED ISSUE #3475: [BUG] Parsons are given in right order
**Describe the bug**
Parsons codelines are in correct order. The lines of code given to drag/drop in the right order, are given in the right order. Not every now and then, but allways. Restarting in new browser results in the same order. I seems the fisherYatesShuffle in Parsons.ts doesn't shuffle very well.

**Expected behavior**
I expect the lines to be presented at random order, this is to easy.

antoineveldhoven:
**Description**

The fisherYatesShuffle did it's work already; the shuffle_code_lines function returned an array with numeric keys, which was nicely shuffled, but retained the numeric sortorder, which was the correct order.

**Fixes #3475**

**How to test**

Go to any level, open the parsons tab, observe sort order; refresh and observe again.

**Checklist**
  
### DIFF
--- a/static/js/parsons.ts
+++ b/static/js/parsons.ts
@@ -103,9 +103,9 @@ function shuffle_code_lines(code: string) {
     let shuffled: Record<string, string> = {};
     let keys = Object.keys(code_lines);
     fisherYatesShuffle(keys);
-    for (const k of keys) {
-        shuffled[k] = code_lines[k];
-    }
+    keys.forEach((k, i) => {
+        shuffled[i] = code_lines[k];
+    });
     return shuffled;
 }
 


Felienne (approved):
Thanks a bunch @antoineveldhoven! I had not noticed before that this was broken again (meaning that we should add this to the long list of things that should be front-end tested!)

antoineveldhoven:
> Thanks a bunch @antoineveldhoven! I had not noticed before that this was broken again (meaning that we should add this to the long list of things that should be front-end tested!)

I'm afraid I introduced a new bug; the order is indeed random, but the key association was needed apparently in order to figure out what the correct order should be.

(Same goes for me; I've tested the order, but not the result, by bad!)

PR #4194, opened 2023-03-30
[FIX] Shuffle parsons code lines, maintain original index.

antoineveldhoven:
**Description**
#4191 Introduced a new issue; due to the index that was used to validate the results.

This simplifies the radomisation and maintains the original index.

**How to test**
In any level, go to parsons, with each click on the tab the options should be randomized; submitting the the right order will now be accepted.


Felienne (approved):
Thanks, this indeed fixes the issue of the lines being marked wrong!

image

PR #4184, opened 2023-03-26
[CONTENT] Split keywords into adventures

MarleenGilsing:
Splits new keywords into separate adventures allowing for a better flow in the learning trajectory, f.e. in level 2 is, ask and sleep get their own tabs:

image

PR #4199, opened 2023-03-30
[CONTENT] Outline for slides level 1 to 18

MarleenGilsing:
This is an outline for the slides for level 1 to 18 in English and Dutch.

### DIFF
--- a/content/slides/nl.yaml
+++ b/content/slides/nl.yaml
@@ -640,13 +640,13 @@ levels:
       3:
         header: "Symbolen"
         text: |
-            Je kunt ook deze twee symbolen gebruiken:
+          Je kunt ook deze twee symbolen gebruiken:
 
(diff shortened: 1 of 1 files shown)


Felienne:
> Will try to do that next week, but I need some help with that first.

And let me know if help here is still needed. The default now is that each code block will use the startcode of the adventure you link. Esp since we are now splitting out the commands in new adventures in #4184.

I will take a look at #4184, let's finish that so we can more easily use the startcode from there!

MarleenGilsing:
I've made slides upto level 11 now (and 13). It will be easier to do if the new split adventures are on the website, so i'll finish then (shouldnt take too long). And I'll also check what I've made now.

Felienne:
Now that https://github.com/hedyorg/hedy/pull/4184 is finished, this can be picked up!

PR #4200, opened 2023-03-31
[CHORE] Remove calls to jsonbin

Felienne:
In the beginning of Hedy, we logged to jsonbin for simplicity, but we have since professionalized and now also log to s3. We can stop to also log to jsonbin (we did both because I was scared of losing logs and then we never removed jsonbin)


website/jsonbin.py
```diff
@@ -1,126 +0,0 @@
...
-
-    def log(self, obj):
-        S3_LOG_QUEUE.add(obj)
-
-
-S3_LOG_QUEUE = log_queue.LogQueue("parse", batch_window_s=300)
-S3_LOG_QUEUE.try_load_emergency_saves()
-
-
-def emergency_shutdown():
```
rix0rrr, line 124: I'm guessing this function is being called from somewhere else. I think the gunicorn wrapper. You're going to need to update the import in that other place as well.

Felienne: ah yeah sorry, the validator caught that too, fixed now!

rix0rrr (approved):

PR #4208, opened 2023-04-04
functions integrated in existing levels

jlole:
functions integrated in existing levels for testing purposes

check the readme in `function-files/`


Felienne:
Lovely work @jlole!!

A few things to think about (not only for you, for me also!)

* Syntax highlighting
Do we want to have the same colors for functions as we already have? I can imagine that we'd use something different for both or for one of the functions. Maybe a blue? Let's play with that a bit. 

* Examples/adventures
What are the examples that we will use to introduce this? I have often used songs as context, Vader Jacob, my Bonnie lies over the ocean to show functions, see for example this lesson: https://www.dropbox.com/s/hf7ys4r48qaem75/Les_7a_009.pdf?dl=0 (Dutch)

* Trajectory of levels
Do we add a new levels for all different options (with/out parameters and with/out return) or do with add this to existing levels? And in what order? I don't immediately have an answer, we will have to play with examples a bit too

functions-files/script_keywords.py
```diff
@@ -0,0 +1,18 @@
+import os
```
Felienne, line 1: @jlole Do we need to keep this in the repo? I think now that we have added the keywords everywhere, it can be dropped?

Felienne:
ow and you also mentioned that you want to add some exceptions? But I guess we can do that in a separate PR so this does not linger too long? I left one small question, if you answer that I think we can go ahead and merge.

functions-files/script_keywords.py
jlole: yes it can be dropped, I will remove it

Felienne:
ok. let's merge this for now!I have made an issue for the exceptions (for you) and for the content and migration (for me)

Felienne (approved):

PR #4210, opened 2023-04-04
[TRANSLATION] Add Hungarian translations

zslim:
**Description**

* Fixes a mistranslation that caught my eye: General (on the achievement page) was translated to "neme", which means gender. Changing that to "Általános".
* Adds several other translations.

**Note**

Why I'm not using Weblate: it was easier here to find the exact mistranslation that I wanted to fix. Then I just translated things that seemed straightforward enough. I'll use Weblate next time, sorry for the inconvenience.

### DIFF
--- a/translations/hu/LC_MESSAGES/messages.po
+++ b/translations/hu/LC_MESSAGES/messages.po
@@ -728,7 +728,7 @@ msgstr "Please select a valid gender, choose (Female, Male, Other)."
 
 #, fuzzy
 msgid "general"
-msgstr "Neme"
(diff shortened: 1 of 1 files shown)


Felienne:
Hi @zslim!

Welcome to Hedy! Let us know if there is anything we can help you with.

No problem to translate here instead of in Weblate. The only consequence is that merging might take us a bit longer since we want to merge Weblate first.

Felienne (approved):
Weblate is closed so we can now merge this, thanks again @zslim!

PR #4224, opened 2023-04-09
[REFACTOR] Reimplement quiz without JavaScript

rix0rrr:
This PR is the start of a project that aims to remove as much JavaScript from the site as possible. 

## Motivation

Many features in Hedy are implemented using JavaScript that JavaScript isn't needed for. Some examples:

* Navigation: Many places do `window.location = "..."` where an `<a>` tag would do the same.
* Form submissions: many places read all values of `<input>` fields and send them to the server in an AJAX request, where a `<form>` can submit that data to the server equally well.
* Showing and hiding elements: we output a lot of HTML elements hidden, given them an `id="something"`, and then use JavaScript to dynamically show and hide those elements (`$('#something').show();`). Flask's Jinja templates can do the same more easily using `{% if %}` tags.

There are a couple of places where JavaScript is unavoidable, but for most features it should be rare.

There are two purposes that JavaScript is commonly used for that this PR introduces replacements for:

* **Refreshing part of the page instead of the entire page**: this is especially necessary because of our tab-based layout. For this purpose, this PR introduces [HTMX](https://htmx.org). HTMX is a technology that allows you to put things like `<div hx-get="/hello">` in the HTML, and when someone clicks that `<div>` its contents are replaced with a server fetch from the `/hello` URL. The server will return plain HTML, but just the part of the page that needs to go into that `<div>`. There are many customization options, but this is the basic idea.
* **Doing light interactivity in response to clicks**: instead of writing custom JavaScript for every page, this PR introduces [hyperscript](https://hyperscript.org): you write a HyperTalk-like script in a special `_` tag in the HTML to achieve this interactivity. For a trivial example: `<div _="on click add .outlined to me">`. No additional JavaScript necessary!

## This PR

This PR introduces the new technologies mentioned above and reimplements the quiz using them.

Here are the key integration points:

* See `quiz/incl-begin.html`:
  * To make the quiz behave like it's loaded into an `<iframe>`, we add a `<div hx-target="this">` as the first child in the tab. This will make sure that all HTMX requests replace the contents of the tab.
  * Links look like `<a hx-get="/quiz/begin/5">`, `<button hx-post="/quiz/submit_answer">`. This will do a regular HTTP request to that server endpoint, which renders regular HTML that gets sent back. The HTML is swapped into the `hx-target` div.
* See `static/js/htmx-integration.ts`:
  * We sometimes execute some JavaScript on load to attach event handlers to special HTML tags. This code now also needs to run on the HTML that is swapped into the page by HTMX (because the regular page load event is long in the past). With Hyperscript the need for custom JavaScript on load handlers should go down, but we still need to turn our `<pre>` tags into read-only Ace editors, so that the code in then can be syntax highlighted. In this file we define an event handler for `htmx.onLoad`, which runs on every DOM content swap to create Ace editors from them.
  * We also sometimes want to run some JavaScript in response to server loads. Specifically, if a user earned achievements by performing an action, we want to show those. How that works is we put the achievements into `session['pending_achievements']`. At the very first opportunity, we send the achievements in the session there to the client in a header: `HX-Trigger: { "displayAchievements": [...] }`, and HTMX is going to run the JavaScript we hook up to those events: `htmx.on('displayAchievements', (ev) =>  .... )`.
* See `quiz.py`:
  * The route handlers for the quiz routes are regular Flask routes in Python that mutate state in `session` to track the user's progress through the quiz. They have been divided into `POST` handlers that mutate state, and `GET` handlers to visualize the current state (e.g., show the current question or show the result of answering the previous question). `POST` handlers redirect to `GET` handlers to avoid the classic `F5 -> do you want to resubmit this form` problem that occurs if you render a page as a response to a `POST` handler. *Technically* this won't happen when the page is loaded using HTMX since `F5` means refreshes the parent page, not the most recent AJAX event, but it feels like good HTTP practice anyway.
* See `hx-question.html`:
  * This is the template for a quiz page that is intended to be rendered using HTMX. It `{% extends hx_layout %}`: note that `hx_layout` is a variable that at runtime will be either `hx-layout-yes.html` or `hx-layout-no.html`. `hx-layout-yes` will be chosen if the page is ACTUALLY requested using HTMX, and it won't render anything other than the contents of the block, ready to be injected into the page. `hx-layout-no.html` in contrast renders the tags to load scripts, CSS and a minimal page frame. This version can be used to render a usable version of the page directly into the browser, which is easy to refresh using F5 and so makes it easy to work on the layout of the page without having to click through the entire quiz.
  * See the hyperscript that replace custom JavaScript to make the quiz buttons focusable by swapping HTML classes around:

```html
                <div ...
                    _="on click if I do not match .incorrect-option then
                        add .hidden to .pick-answer-button
                        remove .hidden from .pick-answer-button in me
                        remove .focused-option from .focused-option
                        add .focused-option to me
                    end"
                >
```

## Also in this PR

- Invert the logic of `no-copy-button`: remove that class everywhere and replace it with a `show-copy-button` (because we need it in fewer places than we don't need it).

## Slight redesign

The quiz pages have been slightly redesigned to make them a little cleaner. New looks:

image

image

image

**How to test**

The quiz still works


reedspool:
Woah worlds colliding for me - I love HTMX and hyperscript. Very cool!

reedspool:
Lots of file changes, I'm afraid I don't have the time right now to review all of this. But at a general glance, I love the direction, and as I said I'm a big fan of the technologies!

TiBiBa:
> Thanks for the look @reedspool! I wasn't sure how controversial this would be, so I'm happy to have you on board 😁.
> 
> 
> 
> I'm still waiting for @jpelay and @TiBiBa to also be on board with this becoming the new standard, as they are also doing a lot of coding work and reviews in this project. We have to all agree on the direction.
> 
> 
> 
> If and when we all agree, I'll put in some work to write up a developer guide as well.

I will take a look tonight! I'm not familiar with both libraries and a bit hesistant as it will increase the learning curve for new contributors. Especially students working on thesis projects.

But, can't say I don't like it without checking it out! Will let you know my thoughts later today.

rix0rrr:
> as it will increase the learning curve for new contributors

I think in fact it's going to make it better. We can cut out learning JavaScript and TypeScript and jQuery, and how to effectively organize code with these technologies (especially the last part seems to be a problem in practice).

Instead, contributors will basically only have to know HTML and CSS and Jinja2 templates (and a couple of magic attributes that we can explain in documentation).

TiBiBa:
> > as it will increase the learning curve for new contributors
> 
> I think in fact it's going to make it better. We can cut out learning JavaScript and TypeScript and jQuery, and how to effectively organize code with these technologies (especially the last part seems to be a problem in practice).
> 
> Instead, contributors will basically only have to know HTML and CSS and Jinja2 templates (and a couple of magic attributes that we can explain in documentation).

I've looked into the libraries and the proposed changes and am still a bit skeptical on the added value of re-writing parts of our code into these two libraries. I do agree that they clean-up our codebase but I, as a contributor without prior knowledge to these libraries, find some of the changes hard to understand and comprehend the underlying file dependencies. Something that is always clear due be in the current structure due to separation of concerns by file.

Given this I still stick to my point that the use of more external libraries might scare away new contributors and introduce complexity. I can see this being really helpful for small straightforward class toggles and interactive tweaks, but how do we work with more complex front-end interactions? The current Javascript, jQuery solution might not be best: but it is straight-forward and a low-level entry for new contributors to add functionality to our front-end.

For me personally this is more of an overall approach question: what do we want Hedy to be? A bit of a hacky but working research project that is continuously growing by students projects or a more professional programming teaching website that is more efficiently programmed? I think this decision is not up to me and I don't want to take sides in this discussion.

static/js/app.ts
```diff
@@ -119,10 +120,11 @@ export interface InitializeAppOptions {
  */
 export function initializeApp(options: InitializeAppOptions) {
   theLevel = options.level;
+  theKeywordLanguage = options.keywordLanguage;
   initializeSyntaxHighlighter({
     keywordLanguage: options.keywordLanguage,
   });
-  initializeHighlightedCodeBlocks(options.keywordLanguage);
+  initializeHighlightedCodeBlocks(document.body);
```
Felienne, line 127: This might require a bit of an explanation (either here or in the docs) the old version clearly explains what we are fetching but what is in the body and why do we use it here?

static/js/app.ts
```diff
@@ -39,6 +39,7 @@ let askPromptOpen = false;
 let theAdventures: Record<string, Adventure> = {};
 let theLevel: number = 0;
 let theLanguage: string = '';
+let theKeywordLanguage: string = 'en';
```
Felienne, line 42: Why do we initialise language to '' but theKeywordLanguage to 'en'? Seems a bit weird, should we explain this?

static/js/message-translations.ts
```diff
@@ -501,7 +501,7 @@ export const TRANSLATIONS: Record<string, Record<MessageKey, string>> = {
     "songs": "🎵,🎶",
     "teacher_welcome": "Welcome to Hedy! Your are now the proud owner of a teachers account which allows you to create classes and invite students.",
     "turtle": "🐢",
-    "unsaved_class_changes": "Biztosan elhagyod ezt az oldalt? Mentetlen változtatásaid vannak, amik így elveszhetnek."
+    "unsaved_class_changes": "There are unsaved changes, are you sure you want to leave this page?"
```
Felienne, line 504: this is probably an accidental reversal?

Felienne:
> I've looked into the libraries and the proposed changes and am still a bit skeptical on the added value of re-writing parts of our code into these two libraries. I do agree that they clean-up our codebase but I, as a contributor without prior knowledge to these libraries, find some of the changes hard to understand and comprehend the underlying file dependencies. Something that is always clear due be in the current structure due to separation of concerns by file.

Yeah I understand @TiBiBa! You know a lot about what is going on already so for you this is a big change but I think in the end it will be better. The issue with the code as we have it now is that is probably quite extensible, in that students and other newcomers can easily expand the javascript, but it does not really communicate how we want stuff to be done, resulting in code styles and decisions that vary a lot, assumptions being made that do not turn out to be true in all cases, and relatively often in broken things. This rewrite not only adds new dependencies but also cleans up in the sense that similar things are more similar and different things are more different.

The new approach, I agree, will have a bit of a steeper learning curve, because people will need to know HTMX which they are less likely to already know than javascript. But, once they have that, it is also more clear how to do things, and the codebase hopefully will be more consistent, which, in a different sense, also helps learnability! 

> For me personally this is more of an overall approach question: what do we want Hedy to be? A bit of a hacky but working research project that is continuously growing by students projects or a more professional programming teaching website that is more efficiently programmed? I think this decision is not up to me and I don't want to take sides in this discussion.

That is a great question and trade off to discuss, and we have discussed that in a contributors meetup earlier in the year, and we decided that cleaning up the code base is one of the code concerns of the coming months (in addition to: making teacher features more usable). Yes, we started out as a "hacky but working research project" (or even: not working haha) but we now have over 300.000 visitors each month, and teachers increasingly rely on Hedy to be working. So we must make sure that future expansions, by students and other newcomers, do not destroy working features and do not make testing harder, even if that comes at the price of a bit more learning html before getting started.

jpelay:
Hello @rix0rrr! 

Even though this is indeed a big change, and would mean that we need to learn new things, I believe this is a step in the right direction. I see this as a sort of middle ground between using a framework such a react or vue, and pure JavaScript. I think one of the big problems we currently face in our code base is that we make different decisions and follow different styles, so this step towards a homogenization of the front-end code base is really useful.

As for students coming to Hedy, I think we need to also work towards a code base that can be maintained and is reliable and professional. I think Hedy is heading towards a project that is as much a research project as it is a learning environment for kids. And I feel that asking students to learn HTMX and a couple other technologies is a manageable task. We also ask them to learn Lark for example, in the back-end, and they manage just fine.

With all this said, I would really appreciate the code guide to incorporate this into Hedy!

templates/quiz/hx-question.html
```diff
@@ -0,0 +1,103 @@
...
+
+        <div class="grid grid-cols-1 md:grid-cols-2 gap-x-12 gap-y-16 mb-8 w-full max-w-screen-lg mx-auto">
+            {% for choice in question.choices %}
+            <div>
+                <div class="turn-pre-into-ace relative py-4 px-1 border-4 rounded-lg shadow-lg
+                    {{container_colors.next()}}
+                    {% if choice.number == progress.last_wrong_answer %}incorrect-option{% endif %}
+                    transition
+                    "
+                    _="on click if I do not match .incorrect-option then
```
jpelay, line 65: Is this a language specific to HTMX?

reedspool: It's a separate language made by the same people as HTMX called hyperscript.org

HTMX is focused on handling user interaction by calling a server. Hyperscript focuses on in-browser interaction to fill in the gaps where a server request wouldn't work or wouldn't make sense

rix0rrr:
I'm going to add a test and then this can go in. After talking to Felienne a bit, I have more things to do, but we don't need to block this change on those.

rix0rrr:
Okay, ready to merge! I recommend having a look at `quiz.cy.js` for fun, and maybe as a strategy in the future that can replace the use of `cy.wait()`.

tests/cypress/e2e/hedy_page/quiz.cy.js
```diff
@@ -0,0 +1,59 @@
...
+        randomlyClickThroughUntilFinished(expectedCounter + 1);
+        return;
+      }
+
+      // Otherwise click a random answer
+      cy.get('*[data-cy="quiz-choice"]:not(.incorrect-option)').then(($buttons) => {
+        return Cypress._.sample($buttons.toArray());
+      }).click();
+      cy.get('button.pick-answer-button:visible').click();
+      randomlyClickThroughUntilFinished(expectedCounter + 1);
```
jpelay, line 56: Love this, so the recursivity prevents the use of cy.wait()? Creative :D

jpelay (approved):

jpelay:
@rix0rrr I have a question about using this in the customizations, or elsewhere. Is it recommended to update the database which every user interaction? For example, when deleting an adventure, or disabling a level, now instead of saving that state in JavaScript, or in the HTML, would we save it right away to the DB?

rix0rrr:
@jpelay we can do one of two things:

- Save in the database immediately. I think this is what most web apps do these days... the "bunch of modifications, hit **save** to apply" model feels a bit 90s to me. On the other hand, we *do* want people to have a way to recover from mistakes. Disabling is not horrible because it can be immediately undone, but "delete" either requires a confirmation, or a state change that can be undone later. Perhaps updating a text block does require clicking a "save" button.
- If the "apply all changes at once" model is important, then we can put the queued changes into `session`, and apply them to the database in one go when the user finally hits "save" (don't forget to remove from `session` again then).

tests/cypress/e2e/hedy_page/quiz.cy.js
rix0rrr: The `cy.wait()` is normally there because we need to wait until the page refresh is definitely "done", and we guess at a time in which that swap has happened. If we know that the current page **definitely does not** contain a particular element, and the page we're waiting for **definitely does have** a particular element, we can use `cy.get(<that element>)`, and it will finish as soon as the element is found (or wait up until a maximum of 4 seconds before giving up).

In our case, the page counter prevents the use of `cy.wait()`.  On every page load, there will be hidden element with `data-cy="ctr:0"`, `data-cy="ctr:1"`, `data-cy="ctr:2"`, etc. If we are currently on page `ctr:1`, then we know that we can wait for `ctr:2` to become available to know that the new content has been swapped in.

--------

The recursion is because we can't use a `while()` loop because of the async nature of Cypress. All cypress commands aren't executed immediately, they are queued and execute "later", when the page has finished refreshing. That means while the `while()` loop is executing, nothing has loaded yet. We must do our actual work inside the `.then(() => { ... })` callback, and the only way to loop in those kinds of situations is to use recursion to start a new async operation.

rix0rrr:
Also responded here: https://github.com/hedyorg/hedy/pull/4224#discussion_r1167452473

jpelay:
> I think for most of our cases, saving in the database immediately should be fine. What do you think?

I think so too! I'm mostly worried about the strain it would have on the server, but since this is NoSQL it shouldn't be a big problem, right?

PR #4227, opened 2023-04-11
[RU] Implemented clickable popup windows for the common errors section of the dashboard

HcKide:
The card with common errors/misconceptions now has clickable elements (apart from the reolve which will be a separate issue). 

Relevant technical choices:
- The popup has its own html file which extends the live-page
- it creates a transparent overlay with a slight top margin so the navbar is still clickable but the other elements are not
- macro was changed to include the parameter used in the href link

Test instructions:
- Run hedy locally, login as teacher or admin, navigate to for_teachers/class1/live-dashboard
- Click on any of the three items in the 'Students who are struggling' card
- A page should open looking like this:

Capture

### DIFF
--- a/templates/class-live-popup.html
+++ b/templates/class-live-popup.html
@@ -2,10 +2,6 @@
 
 {% block popup_content %}
 
-<div>
-
-</div>
-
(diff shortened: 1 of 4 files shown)


karlijnbok:
Looks good! Test passed 👍

PR #4236, opened 2023-04-14
[FIX] Scroll output to bottom when a prompt appears

rix0rrr:
It used to be that when the `ask` prompt pops up and there has been enough output already, the ask modal will cover the output the program has generated so far.

Do an additional "scroll to the bottom" when we open up the modal, so that the last lines (usually containing the question you need to answer) are visible.

It looks like there was an attempt to do this already, but it was added to the wrong event, so I removed it.

## Before

image

## After

image

**How to test**

Emit a large amount of text, followed by a prompt. Observe that the last line is visible.


Felienne:
I think this fixes #2464?

Felienne (approved):

Felienne (approved):

PR #4238, opened 2023-04-16
[UI] added grid_overview button in the class page "for-teachers/class/class_id" that forwards into /class/grid_overview/cl…

akadahsalim:
**Description**
A new button under teacher's view "for-teachers/class/class_id" that forwards into a new page "grid_overview/class/class_id" that will hold the grid. Another button in the new page that allows you to go back to the class page "for-teachers/class/5c39c2a936f24db1a4935c52fab77cd7"

First step in regards to discussion number 4008

**How to test**
Log into teacher's view -> for_teachers -> click on one of the classes -> grid_overview button. You should end up at http://localhost:8080/grid_overview/class/class_id. You can also test the other button by and you should end up back in the class page


templates/class-grid.html
```diff
@@ -0,0 +1,47 @@
+{% extends "auth.html" %}
+{% block regular_content %}
+<div class="flex flex-col gap-2">
+  <h2>{{_('Grid overview')}}: {{class_info.name}}</h2>
+  <div class="flex flex-row items-center justify-center mr-2 w-1/12 ml-4">
+    <select class="py-2 px-1 mt-2 h-10 text-center w-full" name="levels" id="menu">
+        {% for i in range(1, max_level + 1) %}
+            <option name="selected_level" id="select-{{i}}"
+                    value="{{ i }}" hx-trigger="change" hx-get="/grid-overview/change-level?level={{ i }}" hx-target="#level" {% if i == min_level %} selected {% endif %} >
```
rix0rrr, line 9:
```
hx-get="/grid-overview/change-level?level={{ i }}" hx-target="#level"
```

Will replace the HTML element `<div id="level">` with the contents of fetching `http://hedywebsite/grid-overview/change-level?level=5`.

I don't see a div with `id="level"`, so there's nothing for HTMX to place the content into. This is probably why you're not seeing anything happen, but if you want to be sure open the Developer Console (`Ctrl-Shift-i` or right-click anywhere on the page and click **Inspect Element**), then check that the network request is being done and that HTMX hasn't printed errors on the JavaScript console.

It also looks like the `/grid-overview/change-level` route is rendering the `class-grid.html` template, renders a whole page (with `<html>` and `<body>` tags and everything), not just a page fragment to be dynamically loaded into a larger page.

You can do one of two things:

## 1. Refresh part of the page when the level changes

If you want a change in the dropdown to refresh only a small part of the page, use HTMX to request a Flask route that only renders a fraction of the page, namely the part that should be rendered into the `<div>` you point to. You will have:

```
/grid-overview/class/1234               -- loads the containing page and sets up the HTMX dropdown
/grid-overview/class/1234/grid/5   -- loads only the grid and nothing else for class 1234 and level 5
```

And then using HTMX:

```
/grid-overview/class/1234  --[ dynamically loads ]--> /grid-overview/class/1234/grid/5
```

## 2. Refresh the entire page when the level changes

You can also encode the level into the page load of the grid-overview page itself, and do a full page refresh to a different level when the dropdown changes.

```
/grid-overview/class/1234                 -- loads the grid for class 1234 for level=1
/grid-overview/class/1234?level=5   -- loads the grid for class 1234 for level=5
```

And then the dropdown just needs to trigger a refresh of the page.

You can do that like this:

```html
<form method="GET" action="/grid-overview/class/{{ class.id }}">
  <select name="level">
    ...
    <option value="5">Level 5</option>
    ...
  </select>

  <input type="submit" value="Change">
</form>
```

If you want the page to switch immediately as the user picks a dropdown value, so they don't have to click the submit button, you can do that with a bit of hyperscript:

```html
  <select name="level"
    _="on change "
```

Felienne:
Thanks for the tips @rix0rrr! I am not sure what the preferred option would be (not just for this PR but also going forward). We can discuss in the meeting today with @jpelay!

templates/class-grid.html
akadahsalim: Hi @rix0rrr. Thank you so much for the tips. The first method seemed like a more convenient way and it's what I ended up using.

Felienne:
Hy @akadahsalim!! I just took a look and it works great!!

There just seems to be an issue when you already have customizations?

image

I can't reproduce this error easily (it does not happen in the clean local test database) but a bit of defensive programming is a good idea here!

Felienne:
Ow I just realized I can share my local db with you! Here it is: https://www.dropbox.com/s/3wg0h3q2kypwtie/dev_database.json?dl=0

That should allow you to reproduce the issue!

akadahsalim:
> Ow I just realized I can share my local db with you! Here it is: https://www.dropbox.com/s/3wg0h3q2kypwtie/dev_database.json?dl=0
> 
> That should allow you to reproduce the issue!

Capture

I managed to fix this in the following segment. I used the indices because it came into my mind that that data of that student might be in a different order than the row order.
I also ran into an issue where a specific checkboxes where not updated after a click even though the database was being updated. After diving into it, it seems that some entries have the same data with different IDs. 
Capture
An example is above. This breaks stuff and I am not sure how could this be possible. It was from the json file you sent. https://www.dropbox.com/s/3wg0h3q2kypwtie/dev_database.json?dl=0
Is this a normal thing? The only thought I have is that after every run of that program the student make, it generates new data?

website/statistics.py
```diff
@@ -101,6 +104,106 @@ def get_class_stats(self, user, class_id):
...
+        level = request.args.get('level')
+        student_index = request.args.get('student_index', type=int)
+        adventure_index = request.args.get('adventure_index', type=int)
+        students, class_, teacher_adventures, class_adventures_formatted, ticked_adventures = self.get_grid_info(
+            user, class_id)
+        matrix_values = self.get_matrix_values(students, class_adventures_formatted, ticked_adventures, level)
+
+        for program in ticked_adventures[students[student_index]]:
+            if level and class_adventures_formatted[level][adventure_index] in program:
+                program_id = program[0]
```
Felienne, line 164: Does this need to be a numerical index? It makes the code a bit hard to read and a bit shaky. Can we do a dictionary instead so it is more clear what you are fetching?

Felienne:
> I managed to fix this in the following segment. I used the indices because it came into my mind that that data of that student might be in a different order than the row order. 

I don't think I understand this (and I also don't seem to see a new commit?)

> I also ran into an issue where a specific checkboxes where not updated after a click even though the database was being updated. After diving into it, it seems that some entries have the same data with different IDs. Capture An example is above. This breaks stuff and I am not sure how could this be possible. It was from the json file you sent. https://www.dropbox.com/s/3wg0h3q2kypwtie/dev_database.json?dl=0 Is this a normal thing? The only thought I have is that after every run of that program the student make, it generates new data?

Interesting! It might have something to do with us having moved from manual saving to automatic saving, but if this happens locally, it might happen in the server too. Can we do something like taking the newest program when multiple exist?

akadahsalim:
> > I managed to fix this in the following segment. I used the indices because it came into my mind that that data of that student might be in a different order than the row order.
> 
> I don't think I understand this (and I also don't seem to see a new commit?)
> 
> > I also ran into an issue where a specific checkboxes where not updated after a click even though the database was being updated. After diving into it, it seems that some entries have the same data with different IDs. Capture An example is above. This breaks stuff and I am not sure how could this be possible. It was from the json file you sent. https://www.dropbox.com/s/3wg0h3q2kypwtie/dev_database.json?dl=0 Is this a normal thing? The only thought I have is that after every run of that program the student make, it generates new data?
> 
> Interesting! It might have something to do with us having moved from manual saving to automatic saving, but if this happens locally, it might happen in the server too. Can we do something like taking the newest program when multiple exist?


With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.

I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?

Felienne:
> With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.

Great! I will try it locally again to see if it now works!

> I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?

I assume with "submit" you mean "save" here, when a new program appears in the database? That will happen a lot since we have auto-save now, so even if a kids leaves the tab open it might occur. 

So my suggestion is that once a program is ticked off (true) it will remain true for ever and never change back. Could you make that change?

akadahsalim:
> > With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.
> 
> Great! I will try it locally again to see if it now works!
> 
> > I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?
> 
> I assume with "submit" you mean "save" here, when a new program appears in the database? That will happen a lot since we have auto-save now, so even if a kids leaves the tab open it might occur.
> 
> So my suggestion is that once a program is ticked off (true) it will remain true for ever and never change back. Could you make that change?

Does this mean that the true checkbox is also disabled? I think this is doable but a very expensive change because now I need to fetch all the programs for that student at that level and adventure name to find the program that has a "ticked" column when rendering the page, unless we store the IDs of the true cells somewhere. I also think that teachers are prone to missclicking and in that case they can never change that.

akadahsalim:
> > > With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.
> > 
> > 
> > Great! I will try it locally again to see if it now works!
> > > I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?
> > 
> > 
> > I assume with "submit" you mean "save" here, when a new program appears in the database? That will happen a lot since we have auto-save now, so even if a kids leaves the tab open it might occur.
> > So my suggestion is that once a program is ticked off (true) it will remain true for ever and never change back. Could you make that change?
> 
> Does this mean that the true checkbox is also disabled? I think this is doable but a very expensive change because now I need to fetch all the programs for that student at that level and adventure name to find the program that has a "ticked" column when rendering the page, unless we store the IDs of the true cells somewhere. I also think that teachers are prone to missclicking and in that case they can never revert that.

What I suggest is that I keep fetching the latest program and when an auto save is done, the "ticked" value of that program is passed onto the new latest. Not sure if this is possible but it feels like it and it's fairly optimal in my opinion

Felienne:
> > > With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.
> > 
> > 
> > Great! I will try it locally again to see if it now works!
> > > I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?
> > 
> > 
> > I assume with "submit" you mean "save" here, when a new program appears in the database? That will happen a lot since we have auto-save now, so even if a kids leaves the tab open it might occur.
> > So my suggestion is that once a program is ticked off (true) it will remain true for ever and never change back. Could you make that change?

> Does this mean that the true checkbox is also disabled?
> I think this is doable but a very expensive change because now I need to fetch all the programs for that student at that level and adventure name to find the program that has a "ticked" column when rendering the page, unless we store the IDs of the true cells somewhere. I also think that teachers are prone to missclicking and in that case they can never revert that.

No I see how we would want to keep the option open to teachers to correct misclicks.

> What I suggest is that I keep fetching the latest program and when an auto save is done, the "ticked" value of that program is passed onto the new latest. Not sure if this is possible but it feels like it and it's fairly optimal in my opinion.

Ah I understand now..., you are fetching the last program, and placing the true/false flag on that! I had not fully realized and I understand in that architecture, this change is expensive, since you will have to fetch all programs! I think your solution is creative, but also creates code that is a bit hard to understand going forward. I think a more elegant and readable solution os to create a new table StudentAdventures with Id-AdventureName-Level as key and a Ticked (True or False) value as second column. We could also add a program ID, so a teacher could (in the future) navigate to the program. 

I think that represents better in the database what we are trying to achieve. After all, we want to tick off the fact that a kid has finished an adventure, not that they did it with a certain programs per se.

With that table, we use it as follows: we populate the grid that the teacher sees, we first see if there is a program, if not, the tick mark is disabled (as we have it now). We then see if there is a ticked value true from the other table, if so, we show a True/tricked tickmark. 

I hope that is understandable! If not we can discuss tomorrow. I also understand it it is too much work, then we can always go for your option which is simpler to build but harder to grok.

akadahsalim:
> > > > With this new commit. I updated the list to be a dict to be more clear what I am trying to fetch. The issue initally was happening because I was fetching all the programs of the user without considering the customizations. This is now fixed by just checking whether the fetched program is in the list of customzied adventures. I also updated the fetched program to be the latest by calling self.db.last_level_programs_for_user(student, level) instead of self.db.programs_for_user.
> > > 
> > > 
> > > Great! I will try it locally again to see if it now works!
> > > > I believe now when a student submits a new program. The grid is automatically changed to be false at that cell even if it was true before. I've been thinking about a solution to this as it might be confusing for our teachers to see that a cell has magically changed. Any suggestions?
> > > 
> > > 
> > > I assume with "submit" you mean "save" here, when a new program appears in the database? That will happen a lot since we have auto-save now, so even if a kids leaves the tab open it might occur.
> > > So my suggestion is that once a program is ticked off (true) it will remain true for ever and never change back. Could you make that change?
> 
> > Does this mean that the true checkbox is also disabled?
> > I think this is doable but a very expensive change because now I need to fetch all the programs for that student at that level and adventure name to find the program that has a "ticked" column when rendering the page, unless we store the IDs of the true cells somewhere. I also think that teachers are prone to missclicking and in that case they can never revert that.
> 
> No I see how we would want to keep the option open to teachers to correct misclicks.
> 
> > What I suggest is that I keep fetching the latest program and when an auto save is done, the "ticked" value of that program is passed onto the new latest. Not sure if this is possible but it feels like it and it's fairly optimal in my opinion.
> 
> Ah I understand now..., you are fetching the last program, and placing the true/false flag on that! I had not fully realized and I understand in that architecture, this change is expensive, since you will have to fetch all programs! I think your solution is creative, but also creates code that is a bit hard to understand going forward. I think a more elegant and readable solution os to create a new table StudentAdventures with Id-AdventureName-Level as key and a Ticked (True or False) value as second column. We could also add a program ID, so a teacher could (in the future) navigate to the program.
> 
> I think that represents better in the database what we are trying to achieve. After all, we want to tick off the fact that a kid has finished an adventure, not that they did it with a certain programs per se.
> 
> With that table, we use it as follows: we populate the grid that the teacher sees, we first see if there is a program, if not, the tick mark is disabled (as we have it now). We then see if there is a ticked value true from the other table, if so, we show a True/tricked tickmark.
> 
> I hope that is understandable! If not we can discuss tomorrow. I also understand it it is too much work, then we can always go for your option which is simpler to build but harder to grok.

With this new commit. I created a new table "student_adventures" that has an "id" and "ticked" columns. IDs are of the following format "studentID-adventureName-level". I first check whether the student has a program in the programs table at that cell. If so, I fetch the new table to check the value of the 'ticked' column, If the tables don't match for example there's a program but no data in the new table then it's created which I believe helps a lot when populating the data for new classroms. Then the value is swapped upon click. Also updated a small fragment of the code which was confusing to understand. This is the json I tested on:
http://jsonblob.com/1115345383592312832

Let me know if you spot any bugs and wether there's anything that needs to be modified.

Felienne:
Thanks for the update!! This seems to work perfectly!

The only thing that would be nice is if we'd also store the program id in the table, and maybe even allow teachers to navigate there directly from the grid page? But I think we can also do that in a separate PR (or I can do it myself if it is too much work)

Felienne:
> I would absolutely be happy to work on this. I just need to know a few things regarding the autosave feature and the stored data in the programs table. It is not inituitve knowing which program ID I want to store. I believe that we should store the ID of the latest submitted program. This makes it way easier because it was whatever the student submitted and not what the autosave triggered. Remember that the grid data is updated only if the page is requested and I cant't keep up with how many times the autosave was triggered. We still have the problem of a student not submitting the program and in that case I can store the ID of the latest program. Let me know whether this is something I should proceed with.

Thanks for wanting to pick this up!

Just to clarify:

* Submitting in Hedy means using the paper airplane icon, freezing the program of future edits in that level/tab combo. This is done for grading purposes usually.
* Running (and waiting a few minutes) triggers and autosave, overwriting the current program in the DB. So there there is always only 1 program per student-level-adventure combo. 

What I would like to save is the latter, not the former, because we surely want to keep the option open for kids to continue programming in a tab after the program has been ticked off by a teacher. I understand this might mean that the program the teacher navigates to a *different* program then they ticked off before (because another autosave has been triggered) but that is ok for me.

So ticking off is a lighter activity than freezing/submitting. It is just "yes this kid has done a sensible thing for this adventure"

Hope that helps!

akadahsalim:
With these new commits, the new table now contains a program ID column. The grid now also contains a new icon in the cells that have an enabled checkboxes. This allows the teachers to navigate to the program they went to tickoff.

jpelay:
Hi, @akadahsalim! I'm beginning to test your code but this error popped out:

image

I assume this has to do with the database, because this was a class already created. When I created a new database it worked fine! Is this something wrong in my side of does this happen to you as well?

templates/class-grid.html
```diff
@@ -0,0 +1,33 @@
+{% extends "auth.html" %}
+{% block regular_content %}
+<div class="flex flex-col gap-2">
+  <h2>{{_('Grid overview')}}: {{class_info.name}}</h2>
```
jpelay, line 4: Can you please remove the space here and something like this

templates/customize-grid/grid-levels.html
```diff
@@ -0,0 +1,47 @@
...
+      {% for row in matrix_values %}
+        <tr>
+          {% set student_index = loop.index0 %}
+          {% set student = class_info['students'][student_index] %}
+          <th>{{ student }}</th>
+          {% for checkbox in row %}
+            {% set adventure_index = loop.index0 %}
+            {% set adventure_id = adventure_names[class_adventures[level][adventure_index]] %}
+            {% set student_adventure_id = student ~ '-' ~ adventure_id ~ '-' ~ level %}
+            {% if checkbox == 0 %}
```
jpelay, line 21: I think you can save a lot of code here by joining the conditions `checkbox == 0` and `checkbox == 1` and checking and and decide to check or not inside the html element with something like


```
<input type="checkbox" name="myCheckbox" class="checkbox-wrapper" {% if checkbox == 1%}value="checked"{% endif %} [...]>
```

templates/customize-grid/grid-levels.html
```diff
@@ -0,0 +1,47 @@
...
+              <th class="table-header">
+                <div>
+                  <input type="checkbox" name="myCheckbox" class="checkbox-wrapper" value="checked" hx-post="/grid_overview/class/{{ class_info.id }}/level?level={{ level }}&student_index={{ student_index }}&adventure_index={{ adventure_index }}" hx-target="#level" hx-trigger="change">
```
jpelay, line 24: Can you also give change the name `myCheckbox` for something more descriptive, here and anywhere else it appears.

website/database.py
```diff
@@ -268,6 +268,19 @@ def delete_program_by_id(self, id):
...
 
+    def student_adventure_by_id(self, id):
+        # Fetch a student adventure with id formatted as studentID-adventureName-level
+        return STUDENT_ADVENTURES.get({"id": id})
+
+    def update_student_adventure(self, id, ticked):
+        # Swap the ticked value when a request is sent
+        return STUDENT_ADVENTURES.update({"id": id}, {"ticked": not ticked})
+
+    def store_student_adventure(self, adventure):
```
jpelay, line 279: Adventure is a name that is so used across the entire codebase and is often used to mean an adventure that will be displayed in the index, can you use a more descriptive name, maybe `student_adventure`?

website/statistics.py
```diff
@@ -101,6 +104,103 @@ def get_class_stats(self, user, class_id):
...
+                                             page_title=gettext("title_class grid_overview"),
+                                             )
+
+    def get_matrix_values(self, students, class_adventures_formatted, ticked_adventures, level):
+        rendered_adventures = class_adventures_formatted.get(level)
+        grid_values = []
+        for student in students:
+            clone = [None for _ in range(len(class_adventures_formatted[level]))]
+            student_list = ticked_adventures.get(student)
+            if student_list and rendered_adventures:
```
jpelay, line 195: Im not sure if we can reduce the nesting here a bit

website/statistics.py
```diff
@@ -121,6 +221,59 @@ def get_program_stats(self, user):
...
+        class_adventures_formatted = {}
+        for key, value in class_adventures.items():
+            adventure_list = []
+            for adventure in value:
+                adventure_list.append(adventure_names[adventure['name']])
+            class_adventures_formatted[key] = adventure_list
+
+        ticked_adventures = {}
+        student_adventures = {}
+        for student in students:
```
jpelay, line 253: We can also try to reduce the nesting here a little bit as well?

jpelay (changes requested):
Hi, @akadahsalim! everything seems really cool and I liked it. There were a handful of suggestion that I'd like you to address. Thank you!

akadahsalim:
> Hi, @akadahsalim! I'm beginning to test your code but this error popped out:
> 
> image
> 
> I assume this has to do with the database, because this was a class already created. When I created a new class in the database it worked fine! Is this something wrong in my side of does this happen to you as well?

This is indeed happening because of the database. In the database you're using which I assume is the one generated by running feed_dev_database.sh, there is an adventure in this specific class that has the key 'next'. This adventure is not returned from hedy_content.Adventures so when I look for a match to find the value of this key 'next', the error is returned. First I fixed this by adding a static to check to not add the adventure if the key is 'next' but then I thought that is something not ideal as it happens only in this classroom.
I am unsure why would that be the case for this adventure but an interesting thing also is that when I go to customize_class page, the adventure is not there and is wiped from the database. Should I add the check back or just assume that this is something that will never happen as the adventure has a key but no value. For example we have: 'default': Introduction but 'next' is not appearing. I hope this was sufficcient information. let me know how to proceed with this.

website/statistics.py
akadahsalim: This does two things indeed and can be split. However, each split will be almost identical in terms of what is being iterated through but the returned variables will be the same. I think it could've been done in a less complex way but refactoring this means refactoring a lot of the code.

website/statistics.py
akadahsalim: Changed it a tiny bit. Let me know your thoughts on the new commit

website/statistics.py
```diff
@@ -119,6 +119,11 @@ def render_class_grid_overview(self, user, class_id):
         if not class_ or (class_["teacher"] != user["username"] and not is_admin(user)):
             return utils.error_page(error=404, ui_message=gettext("no_such_class"))
 
+        max_width = max(len(str(element)) for row in matrix_values for element in row)
```
akadahsalim: This was removed in a further commit

jpelay:
> This is indeed happening because of the database. In the database you're using which I assume is the one generated by running feed_dev_database.sh, there is an adventure in this specific class that has the key 'next'. This adventure is not returned from hedy_content.Adventures so when I look for a match to find the value of this key 'next', the error is returned. First I fixed this by adding a static check to not add the adventure if the key is 'next' but then I thought that is something not ideal as it happens only in this classroom. I am unsure why would that be the case for this adventure but an interesting thing also is that when I go to customize_class page, the adventure is not there and is wiped from the database. Should I add the check back or just assume that this is something that will never happen as the adventure has a key but no value. For example we have: 'default': Introduction but 'next' is not appearing. I hope this was sufficcient information. let me know how to proceed with this.

That adventure is a bit of legacy code that is no longer needed, that's why is later removed from the customizations when you enter the page. However, I'm afraid that some classes created before might have that adventure, so can you please  add a condition for skipping it?

akadahsalim:
> 

Fixed in the new commit

jpelay:
> Fixed in the new commit
It works perfect now! Thanks

Felienne:
Hi @akadahsalim!

I think this is good to go, but because it is still a draft I wanted to double check if we can go ahead and merge it.

jpelay (approved):

Felienne:
Hi @jpelay! Just dropping a comment here, I would love to make this grid a bit more like the one we have in #4166. Shall we make an issue for that (to be discussed today also!)

jpelay:
> Hi @jpelay! Just dropping a comment here, I would love to make this grid a bit more like the one we have in #4166. Shall we make an issue for that (to be discussed today also!)

Hi! You mean like this Miss Felienne?

image

Felienne:
Yes more or less like that if that is possible!

PR #4245, opened 2023-04-18
[UI] Drop second header

### LINKED ISSUE #4217: [UI idea] Mini issue: remove header 'example hedy code'
Since I'm on the topic of removing clutter from the interface: Can we remove the header that says 'Example Hedy code". 
It's unnessecary info. We could keep the first header, but if there's multiple examples, we don't need the second or third header.  

image

Felienne:
Fixes #4217 (although this also inspired #4221 with more experimentation)


Felienne:
#4221 is experimenting a bit with removing even more headers but I am pretty sure we can remove all but the first one, in any case!

PR #4246, opened 2023-04-18
[RU] Resolve buttons are now functional

HcKide:
[To be approved by Radboud NML members, for #4166 ]

The resolve function on RADboard are now functional. 
When clicked they show a prompt asking to confirm, if yes is pressed, that item is removed from the feed. 

Technical Choices:
Behind the scenes the list items are loaded dynamically by reading from a small json database. 
This is the file `radboard_error_data.json`. For now this file should serve the purpose of dummy data and testing purposes, eventually the error data should probably be stored in the dynamo database as is used on deploy. 
When yes is pressed on the resolve prompt, a js function is triggered that sends a delete request to a route which is found in the statistics.py file [remove_common_error_item()]. In here the active status of the specific error item is set to 0, this means it won't be rendered anymore in the frontend. Then the database is reloaded into memory, otherwise the frontend loads in the outdated version. 
In general, I did a bunch of reformatting and structuring code to make it nicer to work with. Most notably I split up the live stats and normal stats into two classes/modules. 

How to test:
Run hedy locally on this branch, navigate to live dashboard, click resolve button in the common error section. A prompt should show.
If no is pressed, nothing should happen. If yes is pressed that item should dissappear. 
Since the database file is changed, make sure to discard any changes made to your `radboard_error_data.json` file.


Felienne:
Thanks for the update @HcKide!! I will check it out later today!

I am not sure though how this relates to #4166? Is it an addition to it, or does it come in place of?

Felienne:
> @Felienne , hi this is just an issue we had made our own and this is how we integrate them in our new-media-lab branch. Since these PRs are only for the new-media-lab branch and do not touch the rest of the repo (yet), we thought it smart to only have our team members look at them and approve merges. Also because we know the direction that we want our project to head in. I probably should have indicated that this PR is only for members of NML in the description, my apologies. Essentially its just a feature/addition for #4166. The penultimate merge for #4166, will have to be approved by you and the team of course. But since it's barely functional yet this will have to wait for later, hence it's still a draft.
> 
> If you rather have us tackle these PRs in a different way, we would be happy to comply.

Thanks for the clarification! It is probably good if you add to the first message that this branch depends on #4166 (we also sometimes have PRs that can only be merged after other PRs) I also see that you are now saying that your other team members will merge this (which they technically *can*) but we would like to have the final say in merging new/experimental features, so do leave that to use for now.

website/statistics.py
```diff
@@ -51,24 +52,97 @@ def render_class_stats(self, user, class_id):
...
+
+    @route("/live_stats/class/<class_id>", methods=["GET"])
+    @requires_login
+    def render_live_stats(self, user, class_id):
 
-        # card 3 boolean
-        show_c3 = request.args.get("show_c3", default="True", type=str)
-        show_c3 = _determine_bool(show_c3)
+        collapse, show_c1, show_c2, show_c3 = _args_checks()
+        # retrieve common errors from the database, but only for the correct class_id
```
karlijnbok, line 144:
```suggestion
        # Retrieve common errors from the database for class
```

website/statistics.py
```diff
@@ -94,10 +168,10 @@ def render_live_stats(self, user, class_id):
         return render_template(
             "class-live-stats.html",
             class_info={"id": class_id, "students": students, "collapse": collapse,
-                        "show_c1": show_c1, "show_c2": show_c2, "show_c3": show_c3},
+                        "show_c1": show_c1, "show_c2": show_c2, "show_c3": show_c3,
```
karlijnbok, line 171: Maybe not for this PR, but I'd say `show_c{x}` should not be a part of `class_info`, but could be a separate parameter maybe? What do you think?

website/statistics.py
```diff
@@ -107,17 +181,9 @@ def render_student_details(self, user, class_id):
...
-        show_c1 = request.args.get("show_c1", default="True", type=str)
-        show_c1 = _determine_bool(show_c1)
-
-        show_c2 = request.args.get("show_c2", default="True", type=str)
-        show_c2 = _determine_bool(show_c2)
-
-        show_c3 = request.args.get("show_c3", default="True", type=str)
-        show_c3 = _determine_bool(show_c3)
+        collapse, show_c1, show_c2, show_c3 = _args_checks()
+        # retrieve common errors from the database, but only for the correct class_id
```
karlijnbok, line 185:
```suggestion
        # Retrieve common errors from the database for class
```

website/statistics.py
```diff
@@ -151,30 +217,21 @@ def render_student_details(self, user, class_id):
         return render_template(
             "student-space.html",
             class_info={"id": class_id, "students": students, "student": selected_student, "collapse": collapse,
-                        "show_c1": show_c1, "show_c2": show_c2, "show_c3": show_c3},
+                        "show_c1": show_c1, "show_c2": show_c2, "show_c3": show_c3, "common_errors": common_errors},
```
karlijnbok, line 220: See comment above

website/statistics.py
```diff
@@ -151,30 +217,21 @@ def render_student_details(self, user, class_id):
...
             current_page='my-profile',
-            page_title=gettext("title_class live_statistics"),
-            javascript_page_options=dict(page='class-live-stats')
+            page_title=gettext("title_class live_statistics")
         )
 
     @route("/live_stats/class/<class_id>/pop_up", methods=["GET"])
     @requires_login
-    def render_pop_up(self, user, class_id):
+    def common_error_details(self, user, class_id):
```
karlijnbok, line 227:
```suggestion
    def render_common_error_items(self, user, class_id):
```

website/statistics.py
```diff
@@ -151,30 +217,21 @@ def render_student_details(self, user, class_id):
...
-        show_c1 = request.args.get("show_c1", default="True", type=str)
-        show_c1 = _determine_bool(show_c1)
-
-        show_c2 = request.args.get("show_c2", default="True", type=str)
-        show_c2 = _determine_bool(show_c2)
-
-        show_c3 = request.args.get("show_c3", default="True", type=str)
-        show_c3 = _determine_bool(show_c3)
+        collapse, show_c1, show_c2, show_c3 = _args_checks()
+        # retrieve common errors from the database, but only for the correct class_id
```
karlijnbok, line 233:
```suggestion
        # Retrieve common errors from the database for class
```

website/statistics.py
```diff
@@ -204,80 +261,26 @@ def render_pop_up(self, user, class_id):
...
-        if not class_ or (class_["teacher"] != user["username"] and not is_admin(user)):
-            return utils.error_page(error=404, ui_message=gettext("no_such_class"))
-
-        students = sorted(class_.get("students", []))
-        return render_template(
-            "class-logs.html",
-            class_info={"id": class_id, "students": students},
-            current_page="my-profile",
-            page_title=gettext("title_class logs"),
+                        "show_c1": show_c1, "show_c2": show_c2, "show_c3": show_c3, "program_results": result,
```
karlijnbok, line 264: See comment above about `show_c{x}`

website/statistics.py
```diff
@@ -504,6 +507,25 @@ def _determine_bool(bool_str):
     return False
 
 
+def _args_checks():
```
karlijnbok, line 510: I think we should maybe make this name a bit more specific, to state which args we're checking exactly. Something like `_get_args_live_stats()` or `_get_show_collapse_args()` ?

karlijnbok (commented):
CR: I really like the code improvements, it's looking a lot cleaner 🙌. And you got typescript working on the live stats page 🥳. 
I have some small comments mainly about naming, but otherwise I think it looks very good.

karlijnbok:
One thing to consider (not for this PR): a number is currently changed in the database to resolve or hide the error items. We need to consider whether we instead would like to remove these. It might be good to keep a log of the error items, so keeping them in the database would be the better option. However, currently we are not using the data for anything so perhaps it is just wasted space, which means removing them would be the better option. Or we should consider adding an option to easily clear the error items in the database. Something to consider for the future ;)

karlijnbok:
CR & test OK 👍

karlijnbok:
CR & test OK 👍

Issue #3955, opened 2023-01-18
[UI idea] Improvements to adventures reordering


jpelay:
**Idea incl level**
After #3854 is merged we need to address these improvements:

Originally posted by @Felienne 
>    Now that we have the new layout, I think we can also move quiz and puzzle into the "dragger" instead of a tickbox below?
    We need to change the text for the enable/disable buttons (I can do that)
    If there are no adventures available, can we remove the drowdown? And empty one is a bit weird. We can even consider a button there with "add you own"
    We now have 2 places for all levels (in addition to the dragger) the top enable/disable and the dates list. Can we make that into one? Maybe just the date lists with a tick mark?

Also we need to add a Cypress test for the new drag & drop functionality!

**Additional context**
Add any other context about the idea here.

Felienne:
Hi @Annelein!

Maybe many of these (or all?) have been fixed already? If so feel free to close right away.

Annelein:
@jpelay I don't think there is an Cypress test for dragging the adventures right? 

And then all thats left is this: _`If there are no adventures available, can we remove the drowdown? And empty one is a bit weird.`_

jpelay:
> @jpelay I don't think there is an Cypress test for dragging the adventures right?
No, we used to use our own solution for the dragging and dropping, but we later switched to a library and it doesn't work welk with Cypress.

> And then all thats left is this: _`If there are no adventures available, can we remove the drowdown? And empty one is a bit weird.`_

I think I can tackle this in #5540 or maybe in another PR but using that same element.

jpelay:
> @jpelay I don't think there is an Cypress test for dragging the adventures right?

No, we used to use our own solution for the dragging and dropping, but we later switched to a library and it doesn't work welk with Cypress.
 
> And then all thats left is this: _`If there are no adventures available, can we remove the drowdown? And empty one is a bit weird.`_

I think I can tackle this in #5540 or maybe in another PR but using that same element.

Annelein:
@jpelay allright! Should I just create a new issue for removing the dropdown?
And also a new issue for testing the drag and drop for in the future?

Annelein:
> [@jpelay](https://github.com/jpelay) allright! Should I just create a new issue for removing the dropdown? And also a new issue for testing the drag and drop for in the future?
> 
> Then we can close this one.

@jpelay 👆🏼

PR #4252, opened 2023-04-18
[UI] Improvements to class customization

### LINKED ISSUE #3976: [INTERFACE] Customize class page still needs some love
**Describe the bug**

Hi Jesus, 
I love the new way of selecting adventures for teachers! 

I was just trying it out and I think this page can still use some improvements. 

image

I started this issue thinking I had selected level 1 and 3, but I couldn't customize them. Now I'm thinking about it, I probably deselected them instead of selecting them, changing the issue into a - much easier to solve- graphics problem. It's not clear to the user whether an adventure is selected or deselected (blue and green don't intuitively mean anything), so maybe we can change the deselected levels to a lock picture? 

I also still think this page is very busy and overwhelming for teachers. Here are some clean-up suggestions.

image
This explanation can be deleted, since we have a much better way of selecting adventures now. 

image
Maybe we can change the texts in the dropdown menus to make them fit more beautifully. I would suggest 'Level' in the first menu and 'add adventure' in the second. 

image
Can we maybe make these features foldable? Especially the openingdates are distracting to me (and not that important for teachers to see on the first glance). 

Lastly 'Unlock level tesholds' might need a little line of text as explanation.

jpelay:
**Description**

This PR started as a way to address the issues raised in #3976 and #3955.

The improvements that has been made so far:

1. Removes the buttons in favor of a switch to disable the levels.
2. Adds a status badge to show the status of the current level we're editing.
3. _Adds button to add an adventure right in the customizations page._ (It was decided that this feature will be tackled in a future PR)
4. Removes lengthy explanation at the top of the page. 

Now, once we merged #4224 and added HTMX to Hedy, this is an opportunity to remove a large part of the Javascript code that makes this page work.

**Fixes (partially) #3976, #3955**

**How to test**

Todo: add Cypress tests

**Checklist**
  

jpelay:
Even though there were improvements in the page, the main focus of this PR was the HTMX rewrite. This part of the page was rewritten in HTMX and the messy JavaScript was removed:

image

This means that a call to the server is made in the following actions:
1. Switching a level
2. Removing an adventure
3. Adding an adventure
4. Dragging an adventure

And also when pressing the buttons: Restore default adventures and remove customization.

To sort the adventures in the dragger, the sortable.js library was used as explained in this htmx tutorial: https://htmx.org/examples/sortable/.  I also added the library: Jinja Partials because this library allows us to use partial templates inside other templates that can also be called from the Python code.

In the future I want to improve the animation when switching levels and also add an additional button for adding an adventure

jpelay:
So the front-end now pass! There's one that I'm skipping, that relates to the drag and drop functionality, and will leave as an open issue until I (or someone else) can tackle it.

jpelay:
Im working with the Cypress tests, but they work in my computer (ha, classic!)

image

static/js/htmx-integration.ts
```diff
@@ -38,6 +39,13 @@ htmx.defineExtension('disable-element', {
  */
 htmx.onLoad((content) => {
     initializeHighlightedCodeBlocks(content);
+    var sortable : HTMLElement =  content.querySelector('.sortable') as HTMLElement;
```
rix0rrr, line 42: In the [example here](https://htmx.org/examples/sortable/) they're also taking care of the possibility of there being more than one `.sortable` element. Something we'd want to do as well, just to be sure?

static/js/htmx.d.ts
```diff
@@ -20,7 +20,7 @@ export function addClass(elt: Element, clazz: string, delay?: number): void;
  * @param path the URL path to make the AJAX
  * @param element the element to target (defaults to the **body**)
  */
-export function ajax(verb: string, path: string, element: Element): void;
+export function ajax(verb: string, path: string, element: Element): Promise;
```
rix0rrr, line 23:
```suggestion
export function ajax(verb: string, path: string, element: Element): Promise<void>;
```

static/js/htmx.d.ts
```diff
@@ -46,7 +46,7 @@ export function ajax(
     verb: string,
     path: string,
     context: Partial<{ source: any; event: any; handler: any; target: any; values: any; headers: any }>
-): void;
+): Promise;
```
rix0rrr, line 49:
```suggestion
): Promise<void>;
```

static/js/teachers.ts
```diff
@@ -399,116 +382,93 @@ export function save_customizations(class_id: string) {
...
+        // Restore all the options other than the adventures.
+        // The adventures will be restored to the default using an HTMX call to the server
+        $('.other_settings_checkbox').prop('checked', false);
+        // Remove the value from all input fields -> reset to text to show placeholder
+        $('.opening_date_input').prop("type", "text")
+                                .blur()
+                                .val('')
+                                .prop('disabled', false)
+                                .attr('placeholder', ClientMessages.directly_available)
+                                .attr('placeholder', ClientMessages.directly_available)
```
rix0rrr, line 405: This line has been duplicated

templates/customize-class.html
```diff
@@ -1,92 +1,46 @@
...
-                        </div>
-                    {% endfor %}
-                {% endif %}
-            </div>
-            <div class="flex flex-row justify-center items-center ml-2 w-2/12 mr-4">
-                <select class="py-2 px-1 w-full mt-2 h-10 text-center" name="levels" id="available" data-cy="available_adventures_current_level">
-                    <option value="none" selected>{{_('available_adventures_level')}} -</option>
                 </select>
             </div>
+            {{ render_partial('customize-class/sortable-adventures.html',
```
rix0rrr, line 35: Ah, there it is! Can you explain to me what this does differently from the built-in `render_template`?

rix0rrr (commented):
Well this thing is huge! In all honesty I find it a bit daunting to review 🥲. So, I'm doing what people do in that case: pick out a few things to  comment on or have questions about, and ship the rest unseen! 😆

But seriously: a couple of questions and nitpicks, but in general I see this is moving in the right direction! I trust you, and I trust that we can all pick up and fix any issues that might come up 👍.

So LGTM!

website/for_teachers.py
```diff
@@ -184,13 +185,137 @@ def get_class(self, user, class_id):
...
+        if not is_teacher(user) and not is_admin(user):
+            return utils.error_page(error=403, ui_message=gettext("retrieve_class_error"))
+        Class = self.db.get_class(session['class_id'])
+        if not Class or (Class["teacher"] != user["username"] and not is_admin(user)):
+            return utils.error_page(error=404, ui_message=gettext("no_such_class"))
+
+        level = request.args.get('level')
+        customizations, adventure_names, available_adventures, _ = self.get_class_info(user, session['class_id'])
+
+        return render_partial('customize-class/sortable-adventures.html',
```
rix0rrr, line 227: Does `render_partial` here do anything different than `render_template` ?

templates/customize-class/sortable-adventures.html
```diff
@@ -0,0 +1,37 @@
+<div class="flex flex-row w-full" id="adventure-dragger">
```
rix0rrr, line 1: Can we come up with a naming convention for files that are intentionally partials, and designed to be retrieved via HTMX?

In the past I did `hx-<name>.html`, but we can also go for `partial-<name>.html` or something.

templates/customize-class.html
```diff
@@ -1,92 +1,46 @@
-{% extends "auth.html" %}
-
+{% extends hx_layout %}
```
rix0rrr, line 1: Is `customize-class.html` intended to be retrieved via HTMX? If not, `hx_layout` doesn't do anything.

If yes, can we do something with the naming convention to indicate that? `hx-<name>.html` or somesuch?

rix0rrr (commented):
Okay a couple more 😉

static/js/htmx-integration.ts
jpelay: Ok! Will do!

static/js/teachers.ts
jpelay: Thanks!!

website/for_teachers.py
```diff
@@ -184,13 +185,137 @@ def get_class(self, user, class_id):
...
+            adventures_default_order=hedy_content.ADVENTURE_ORDER_PER_LEVEL,
+            current_page="for-teachers",
+            min_level=min_level,
+            class_id=class_id,
+            javascript_page_options=dict(
+                page='customize-class',
+                class_id=class_id
+            ))
+
+    @route("/change-level", methods=["GET"])
```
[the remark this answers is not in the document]
jpelay: I changed it to `/get-customization-level`, what do you think about it?

static/js/teachers.ts
```diff
@@ -399,116 +382,93 @@ export function save_customizations(class_id: string) {
...
-        const name = <string>$(this).attr('adventure');
-        if (name == adventure_name && $(this).is(":visible")) {
-            if (first_input) {
-                checked = $(this).prop("checked");
-                first_input = false;
-            }
-            $(this).prop("checked", !checked);
-        }
-    });
+export function restore_adventures_to_default(prompt: string) {
```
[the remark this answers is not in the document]
jpelay: I'm using this function to fire the htmx call from the button, which will in return call the function from python. I do it this way since we need to get the current level that the user is editing, so when they restore the button won't be kicked out to level 1. Also, because we need to display the modal.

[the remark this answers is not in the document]
rix0rrr: Aha, I see!

> I'm using this function to fire the htmx call from the button

I would say that ideally this should not be necessary. The "htmx way" to do this would be to have a button like this:

```
<button hx-post="/teacher-adventures/{{ level }}/restore">
```

Now I'm guessing you're probably doing this with JavaScript because I think your page layout might look something like this:

image

So, you need some external mechanism to transfer the level number from the selector to the reset button in order to trigger the appropriate reset.

------------

However, if you put the button *inside* the HTML fragment that HTMX renders, it's all a lot easier! When loading the HTML fragment that will go into the orange box, we *know* that we are rending `/adventures/5`, so we *know* that `{{ level }} == 5`, so we can immediately substitute that into the URL for the reset button:

image

Here is the trick: you render the buttons that take actions on data along with the data you are rendering! That way, you know all the identifiers you need to know in order to generate the correct URLs.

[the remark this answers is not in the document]
rix0rrr: The above pattern holds if you need to replace the entire content area, for example for a button like "reset". 

If you have a list of items you want to be able to edit, you could do the same for a "save all" button, or you could have individual "saves" (automatic or via a button click) for individual elements. Both are fine patterns.

In both cases you'd have a `<form>` to hold the data you want to send to the server, and in both cases you'd probably re-render the exact same page after saving, with potentially a "Saved" announcement rendered in as well.

image

website/for_teachers.py
jpelay: Sorry for not commenting this yesterday! 

We discussed it in the last Hedy meeting, basically this function works like a macro or an include, in the sense that you can include this template inside another template; but neither macro nor include can be called from python, you can only call `render_template`, `render_partial` can be called so its a bit of a combination. Another justification, that is crutial, is that when called inside a template, you can pass values to the partial template. The author justifies the existence of this library here: https://github.com/mikeckennedy/jinja_partials/issues/1 and I found it in this talk that he gave: https://www.youtube.com/watch?v=kNXVFB5eQfo

I think that this is a good library moving forward, because makes working with Flask and HTMX much more easier and flexible. Also, don't seem to be a library that can brake randomly due to an update, as it seems that development finished. I'm also very willing to listen to your comments in the use of this library.

templates/customize-class/sortable-adventures.html
jpelay: Sure! I'm going with your convention since it's the one already in place!

templates/customize-class.html
jpelay: Ohhh, perfect! I included it because I thought that its use was that a page that uses HTMX calls needed to include it. Will remove it then!

templates/customize-class/sortable-adventures.html
Felienne: I strongly prefer partial since it is a lot more descriptive of what it does, but let's merge this now, and I will make an issue separately since this PR is already open so long.

website/for_teachers.py
Felienne: Thanks for the explanation!! We should also move this to the docs, will you remember this or should I make an issue?

jpelay:
@rix0rrr, I couldn't manage to remove the `htmx.ajax` when is used with the "Remove customizations" button, because I do need to edit parts of the HTML that aren't drawn with HTMX. I think in the future we can draw everything with HTMX, but I think this is a good middle ground.

In the Restore Adventures button I was able to remove the Javascript call. Another thing that I had to replace besides knowing the current level, was to trigger the confirm modal. So I added a new htmx modal following their tutorial using hyperscript (@Felienne if this gets aproved I'll add this too to our wiki!). 

It looks like this now:

image

website/for_teachers.py
rix0rrr: I’m still a little confused. Don’t let that bother you or anything, it’s not a reason to block anything on. Let’s get this majestic piece of work merged!

But here’s where I’m at:

- I see how `render_partial` is nicer in a template than `{% include %}`
- I don’t see how `render_partial` is different from `render_template` in Python code
- I don’t see how `render_partial` helps with macros called from Python.

rix0rrr (approved):

PR #4253, opened 2023-04-19
[RU] moved collapse button, changed colour

HcKide:
[To be approved by members of the NML course, for #4166 ]

Move the collapse button of the first section to the top. This was because when testing it with people they found that to be more intuitive rather than below. Moreover, I changed the colour to blue rather than green to have a bit of separation to the 'Back to class' button

How to test:
- run hedy locally on this branch, navigate to live stats page
- verify that Expand button is blue
- when pressing Expand button, verify that collapse button shows up at the top of the section rather than bottom, that it is positioned correctly and is blue
image

### DIFF
--- a/templates/class-live-stats.html
+++ b/templates/class-live-stats.html
@@ -43,7 +43,7 @@
     </div>
     {% if class_info.show_c1 %}
     <div class="relative">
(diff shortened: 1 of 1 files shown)


karlijnbok:
Looks good 👍

PR #4263, opened 2023-04-20
[CONTENT] Adds quiz Level 14 in En

MarleenGilsing:
Made a quiz for level 14. It seems to work, but it also seems to give the wrong colors for right and wrong. It might be because I'm running it on the local host


Felienne:
There was a small issue in the indentation, this is why the wrong colors were shown, I fixed that now.

Felienne (approved):

Felienne (approved):

Issue #4258, opened 2023-04-19
[BUG] newlines are gone from quiz answers


Felienne:
This is due to the quiz rewrite in #4224, I think:

<img width="1131" alt="image" src="https://user-images.githubusercontent.com/1003685/233117378-94355137-b70b-45b1-932e-70044a0dc264.png">

axvr:
Hi, I think the change made in #4259 to fix this issue has caused another one.  Some of the quiz answers contain `</br>`s.  (E.g. level 1 quiz question 3 and some questions on the level 3 quiz.)

<img width="1126" alt="image" src="https://user-images.githubusercontent.com/26504626/234373833-25ac23e6-a874-450f-915e-6da1aba9ad2d.png">

Felienne:
> Hi, I think the change made in #4259 to fix this issue has caused another one. Some of the quiz answers contain `<br />`s. (E.g. level 1 quiz question 3 and some questions on the level 3 quiz.) I saw it on Firefox, I haven't tested other browsers.
> 
> <img alt="image" width="1126" src="https://user-images.githubusercontent.com/26504626/234373833-25ac23e6-a874-450f-915e-6da1aba9ad2d.png">

Ah thanks for noticing! I reverting the fix and will look into a beter solution!

Felienne:
Fixed by #4270

PR #4270, opened 2023-04-25
[FIX] Make the newlines in the quiz text render properly

rix0rrr:
MarkDown by default doesn't consider newlines in the input text significant: multiple lines of text following another add up to a single paragraph.

It's possible to force newlines by ending a line with two spaces.

While no one is going to do this in the YAML, we can use a filter to replace all newlines with two spaces followed by a newline, to put the input into a format for the `|commonmark` filter that will render the newlines correctly.

Also in this PR: get rid of the now unused `quiz.ts`, which should have been removed as part of #4224, but was forgotten.

**How to test**

Find a quiz question where the answers have newlines in them. See that the newlines are there.


Felienne:
yes this fixes the original issue (#4258):

image

But also does not do the ugly <br> thing:

image

Felienne (approved):

Felienne:
yes this fixes the original issue (#4258):

image

But also does not do the ugly <br> thing:

image

Issue #4240, opened 2023-04-18
[LANGUAGE idea] Add exceptions for functions


Felienne:
In https://github.com/hedyorg/hedy/pull/4208 we have introduced functions (very cool!), but there was mention of some exceptions that you wanted to add @jlole?

Felienne:
Closed by #4280

PR #4280, opened 2023-05-04
 [LANGUAGE] Nested functions

jlole:
**Description**

**Adds** exception for nested functions
**Fixes** #4240


grammars/level11-Additions.lark
```diff
@@ -1,11 +1,13 @@
-command:+= for_loop | call | define
+command:+= for_loop | call | error_nested_define
 
 //new : for loop
 for_loop: _FOR var _IN _RANGE (INT | var_access) _TO (INT | var_access) _EOL (_SPACE command) (_EOL _SPACE command)*  _EOL? _END_BLOCK
 
 // FUNCTIONS ======================================== =============================
 // check line 1: command:+= call | define
+_non_empty_program: _EOL* (define | command | error_invalid) _SPACE* (_EOL+ (define | command) _SPACE*)* _EOL* //lines may end on spaces and might be separated by many newlines
```
Felienne, line 8: Thanks for the addition @jlole!

I don't love the fact that define is a bit hidden here, it is a command, could we not merge it with command in line 1? I guess there is a reason for this (a conflict in parsing with another command, because you put it before) but we can redefine `command` here, like in level 6, so it is more clear where the conflict is.

Felienne (changes requested):
Can we make it more clear where define needs to be in the command order?

grammars/level11-Additions.lark
jlole: The reason for this is that if `define` is merged with `command`, I cannot check if a function definition is nested. This way, `define` is only allowed outside of any blocks (can't define a function in another function or in an if statement)

Felienne (approved):

PR #4283, opened 2023-05-08
[CONTENT, WIP!!] Adds 2 basic function adventures

Felienne:
I am experimenting with adventures for functions based on https://www.dropbox.com/s/hf7ys4r48qaem75/Les_7a_009.pdf?dl=0 but this should not be merged yet since we want to split this into a separate level (probably at the end of the EU school year).


Felienne:
This cannot be merged because of too many yaml issues, and @MarleenGilsing is working on function adventures elsewhere (in #4528).

PR #4284, opened 2023-05-09
[UI] Adds function to render maximum level

### LINKED ISSUE #4219: [UI] Start at highest level achieved
From a teacher survey:

"Alternatively, when students are in a class that has Levels locked behind completing quizzes, it would be great if they would start back on their highest unlocked level when they log in. Students sit there clicking the "Go to Level 2", "Go to Level 3", etc. buttons for a long time at the start of class."

jpelay:
**Description**

Adds a new function to render the maximum level reached by the user in a quiz when they access the /hedy path.

**Fixes #4219*

**How to test**
Log in as a one of the students in the database (student1 to student5), and complete a quiz. Then go to the Hedy root and see if the next level is shown!


TiBiBa (approved):
Looks good to me! Not sure why the `get_highest_quiz_level()` function doesn't return 0 or None, as this would make more sense. But that is not the aim of this PR.

jpelay:
Oops there's a cypress test error, will fix it.

PR #4285, opened 2023-05-10
db script and adventures markdown file

jlole:
Add the script for future level migration to the repo
Add the markdown file containing some adventures using functions

### DIFF
--- a/functions-files/ADVENTURES.md
+++ b/functions-files/ADVENTURES.md
@@ -1,0 +1,95 @@
+# Adventures for Hedy Functions
+## Level 11
+### Introduction to functions
+```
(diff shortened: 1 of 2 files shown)


Felienne:
Thanks for the adventures and the migration script @jlole!! I see it is a draft so I am not sure what extras you are planning but there are other tables we need to consider, most notably the custom adventures that teachers have created for classes in table "adventures" and what is shown in a class in table "class_customizations".

And data about quizzen in quiz-stats and quizAnswers.

But if you don't get that, that is totally understandable!

Felienne:
We are going to merge the content into main and leave the levels as is (see: https://github.com/hedyorg/hedy/issues/4469)

PR #4290, opened 2023-05-13
[FIX] Class customization dropdown closes immediately after opening

rix0rrr:
The level selector dropdown behaves differently in different browsers:

- In Chrome, it opens up and immediately closes.
- In Safari, it doesn't seem to do anything.

This is due to the event we were listening for: `hx-trigger="click"`.

On Chrome, the click event that opens the dropdown immediately triggers the HTMX request. On Safari, `click` doesn't seem to be triggered at all.

Change the event to `hx-trigger="input"`; this gets fired as soon as a value is selected.

**How to test**

Go to the class customization page. Click the level dropdown. It stays open, and clicking an option loads that level.

### DIFF
--- a/templates/customize-class.html
+++ b/templates/customize-class.html
@@ -1,7 +1,7 @@
 {% extends "auth.html" %}
 {% block regular_content %}
 <div class="flex flex-col gap-2">
-    <h2>{{_('customize_class')}}: {{class_info.name}}</h2>    
+    <h2>{{_('customize_class')}}: {{class_info.name}}</h2>
     <div class="flex flex-col mt-2 border border-gray-400 rounded-lg">
         <h3 class="px-4">{{_('select_adventures')}}</h3>
         <div class="flex flex-row">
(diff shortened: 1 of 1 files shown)


jpelay:
> The level selector dropdown behaves differently in different browsers:
> 
> * In Chrome, it opens up and immediately closes.
> * In Safari, it doesn't seem to do anything.
> 
> This is due to the event we were listening for: `hx-trigger="click"`.
> 
> On Chrome, the click event that opens the dropdown immediately triggers the HTMX request. On Safari, `click` doesn't seem to be triggered at all.
> 
> Change the event to `hx-trigger="input"`; this gets fired as soon as a value is selected.
> 
> **How to test**
> 
> Go to the class customization page. Click the level dropdown. It stays open, and clicking an option loads that level.

Thank you so much for this! I don't have Safari on my machine, I tried it in Chrome and Firefox and it seemed to worked ok there. But I use Linux, so maybe it's different on Mac OS. Next time I'll make sure to also test on Windows, since is the machine I have access to!

Felienne:
Superseded by #4293

PR #4291, opened 2023-05-13
[FIX] Attempt to fix a 500 server error

rix0rrr:
In the server log, I've seen this exception:

```
<ERROR>: Exception on /for-teachers/get-customization-level [GET]

  .....

  File "/app/website/for_teachers.py", line 225, in change_dropdown_level
    customizations, adventure_names, availableadventures,  = self.get_class_info(user, session['class_id'])
  File "/app/website/for_teachers.py", line 347, in get_class_info
    available_adventures = self.get_unused_adventures(customizations, teacher_adventures)
  File "/app/website/for_teachers.py", line 398, in get_unused_adventures
    customizations['sorted_adventures'][adventure['level']]:
KeyError: 1
```

The issue seems to be that `adventure['level']` is the integer `1` and it is not in the `sorted_adventures` dictionary.

I'm not sure how we guarantee that `sorted_adventures` will always have an entry for all levels. It would seem safer to do `customizations['sorted_adventures'].get(adventure['level'], [])` so that we at least fall back to an empty list if the key is not there. I didn't do that, as I don't know the invariants that we expect here.

However, it seems that in the code a couple of lines above, we expect to be indexing `sorted_adventures` with a string, as we are explicitly converting `level` to a string there:
`customizations['sorted_adventures'][str(level)]`

My guess is that we were expecting `adventure['level']` to be a string, but it is in fact an `int`.

The simplest probable fix would be to stringify this value as well, which this PR is doing.

Unfortunately I also don't know how to reproduce this error, so I can't tell you how to test this.

### DIFF
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -395,7 +395,7 @@ class ForTeachersModule(WebsiteModule):
 
         for adventure in teacher_adventures:
             if {"name": adventure['id'], "from_teacher": True} not in \
-                    customizations['sorted_adventures'][adventure['level']]:
+                    customizations['sorted_adventures'][str(adventure['level'])]:
                 available_adventures[int(adventure['level'])].append(
                     {"name": adventure['id'], "from_teacher": True})
 


jpelay:
The key is in the function `get_class_info` that is called when you enter in the customize-class page. If it has customizations, then it'll retrieve them; otherwise it'll create a fresh customization object and save it in the database, that's how we guarantee that it always have one. Here you can see that: https://github.com/hedyorg/hedy/blob/5862c1eb0aae200f57f90022e134605122ec0275/website/for_teachers.py#L359-L377

And it's called here: https://github.com/hedyorg/hedy/blob/5862c1eb0aae200f57f90022e134605122ec0275/website/for_teachers.py#L196

That's why later on I don't check if it has the fields I need or not, I assume that it has been created because otherwise you wouldn't be able to access that endpoint.

jpelay:
Btw, is there a way we can take a peak to the alpha enviroment database? I would love to take a peak to the fields, and see how the error might be originating. Miss @Felienne handed me the api key, but it still didn't allow me to read the database from my machine.

jpelay:
> Btw, is there a way we can take a peak to the alpha enviroment database? I would love look at it and see how the error might be originating. Miss @Felienne handed me the api key, but it still didn't allow me to read the database from my machine.

jpelay:
I just made a test with my classes in alpha. I deleted all my custom adventures, and then tried to access the customization page and it worked. Then added a new adventure and then it worked too, so I think it's got to do that the old teacher adventures have an int field while the new ones have an str field (probably after your fix in that page?) so I think this indeed solves the problem!

rix0rrr:
Yes historically much of the code has played a bit fast and loose with treating level as `str` or `int`. Both occur, and for safety I just cast it to whatever I need it to be everywhere, because you can't trust any existing code :).

Felienne (approved):

PR #4292, opened 2023-05-13
[FIX] [TRANSLATIONS] Fix translations

jpelay:
The PR #4252 introduced errors in the translations that I'm fixing here in English. The fixes I made in English didn't seem to trickle down to the other languages.


jpelay:
Ups, didn't think all the other commits will be shown here!

PR #4293, opened 2023-05-13
[FIX] Cypress test

rix0rrr:
This seems to fix the failing cypress test on my machine.

I'm not sure why, exactly (if it keeps on working on GitHub).


jpelay:
Thanks for fixing the tests!

PR #4294, opened 2023-05-14
[CHORE] Detect .po file problems during the PR build

rix0rrr:
For reasons we're not quite sure of, sometimes translations get lost while generating `.po` files. A situation that happens sometimes is that we end up with the following in a file:

```
msgid "mandatory_mode"
msgstr ""

(...a lot further down...)

\#~ msgid "mandatory_mode"
\#~ msgstr "Mandatory Developer's Mode
```

For some reason the correct translation is commented out and replaced with an empty string one.

This PR adds a script that will detect this situation and fail the build on the PR that tries to accidentally introduce this situation.

**How to test**

The build on this PR will currently fail, as the `main` branch is currently in this broken state.

Once the main branch is fixed and merged into this PR, the build will start passing.

### DIFF
--- a/build-tools/github/validate
+++ b/build-tools/github/validate
@@ -16,5 +16,6 @@ else
   $scriptdir/validate-tests
 fi
 $scriptdir/validate-e2e
+python3 $scriptdir/validate-pofiles
 
 echo "Everything is great! 🍰"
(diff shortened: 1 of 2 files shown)


Felienne (approved):
I tested and indeed of the files are ok, the tests pass!

PR #4295, opened 2023-05-15
[FIX] Reset po/mo files

Felienne:
`main` currently has inconsistent pot/po/mo files with some duplicate keys, we are not sure yet why. 

This PR resets them all to the state of #4275. We can use this: 

1) to get back into a working state
2) to test whether #4294 indeed passes if all is well

I will update my progress here so @jpelay can follow along.


Felienne:
I have rerun extract and then update, and many keys disappear again to the bottom of the file but:

1) There are no more duplicate in them and
2) These keys seem indeed to no longer be in use, grepping from them in the code base only finds the po files

I am a bit afraid to merge this, so maybe @jpelay can take a look at this? Are those keys indeed removed?

jpelay:
> I translated the new keys but
> 
> ```
> msgid "updating"
> msgstr ""
> ```
> 
> I did not really understand, and it is hard to grep for, so it still needs a translation.

This one is used when you are editing the adventures in the dragger, or changing levels, i.e: when you are firing an HTMX request. I will change it's `msgid` to reflect this!

jpelay (approved):

PR #4296, opened 2023-05-15
[FEAT] Display a warning when the user is not logged in

rix0rrr:
Warn users if they have been spending a notable amount of effort on a program and they are not logged in, that their work may be lost if they don't log in.

The rules are currently a multitude, because we don't know what will work best. If they either:

- Have click "Run" X times; or
- Their program is longer than Y lines; or
- They've spent more than Z minutes on one tab

Then we show the warning.

image

**How to test**

Be not logged in, hit the Run button a bunch of times.


translations/nl/LC_MESSAGES/messages.po
```diff
@@ -777,8 +777,11 @@ msgstr "Inloggen"
 msgid "login_long"
 msgstr "Inloggen op jouw account"
 
+msgid "login_to_save_your_work"
+msgstr "Log in om je werk op te slaan"
+
 msgid "logout"
-msgstr "Log uit"
+msgstr "Log out"
```
Felienne, line 784:
```suggestion
msgstr "Log uit"
```

Felienne (approved):

PR #4298, opened 2023-05-16
[RU] Student details panel 

HcKide:
[To be merged by members of the NML team only!]

The student details panel (the panel that shows if you click on a student name in the list) now shows representative data thats from the database. 
It has a graph made with chart.js that shows the history of runs (errors and successes), it shows the last code that was run and a potential error that was made in the last run. See screenshot.

Screenshot from 2023-05-16 19-52-38

Relevant technical choices:
The chart uses chart.js and hence is completely in typescript. 
Since the database only shows the last run in the session I had to add more data whenever a run is done by a student. 
It is called chart_history and is saved under program-stats in the database. In order to not flood the database there is a maximum length to the chart_history object. 
The last program and error are both retrieved from the database by simply retrieving the last run program. 
Lastly, the whole page refreshes each minute now. This is so the data in the details section actually updates (although there is also a refresh button). 
While this is not exactly the vision of a live dashboard this is a quick fix for that now, and will likely also be useful for the rest of the features we're adding down the line. 

How to test:
Run hedy locally, navigate to this branch
Log in as teacher and go to live dashboard
In a different tab (use incognito mode so session isn't shared) log in as a student of choice and solve some levels / write erroneous code on purpose
You will see the data update in the details section for that student if all goes well.

### DIFF
--- a/website/statistics.py
+++ b/website/statistics.py
@@ -176,7 +176,6 @@ class LiveStatisticsModule(WebsiteModule):
             )
             student_names.append(student_username)
 
-
         # Data for student overview card
         if hedy_content.Adventures(g.lang).has_adventures():
             adventures = hedy_content.Adventures(g.lang).get_adventure_keyname_name_levels()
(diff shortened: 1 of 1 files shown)


sonimegha123:
I did the code review and it works as expected. Nice work!

PR #4299, opened 2023-05-16
[UI] Fix appeareance of adventure dragger when overflowing

jpelay:
There was an issue when editting level five in the customizations:

image

It looks terrible because the text in the button is taking too much space, so now this PR adds a tooltip and a new label to the button, which is now 'Reset' looking like this now:

image

Another thing that this PR fixes is the overflowing behavior, when you resized the screen, the `selects` also resized and become tiny. Now they have a minimum width and the whole div becomes scrollable:

image

In Firefox, this causes a bit of a problem because Firefox has a different (and in my opinion worst) approach to scroll bars, being able to just use the outer scroll bar:

https://github.com/hedyorg/hedy/assets/45865185/b0696e14-a5db-49bf-8fcf-fffd9cc6df6f

Lastly, also changes the value of the `level_future` key in Babel so it reads more nicely:

image

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-05-15 11:37-0400\n"
+"POT-Creation-Date: 2023-05-16 11:32-0400\n"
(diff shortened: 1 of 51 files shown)


Felienne:
This works, so let's just merge and ship it!

But I was thinking, the reset button is a bit hidden, in Chrome no scroll bar is shown, so you have to know to scroll:

image

Could we place it above the dropdown:

image

Felienne (approved):

PR #4215, opened 2023-04-06
[SLIDES] New theme, flatter code block, no local storage

Felienne:
Some assorted changed to the slides:
* New theme aligning with Hedy colors
* Makes the code block flatter (less high) in the slides view
* Grabs code in the /raw mode from the yaml and not from local storage

PR #4302, opened 2023-05-18
[FIX] The output window grows with all output

rix0rrr:
In https://github.com/hedyorg/hedy/pull/4215, the `height: 22rem` style attribute was removed from the program output, which stops it from restricting its height and makes it grow with the output.

Revert that.


Felienne (approved):
Whoops, sorry!

PR #4303, opened 2023-05-20
[CHORE] Rename templates that are used with HTMX

jpelay:
**Description**

Renames the templates that are used with HTMX, following 2 criteria (that will be added to the wiki):
1. The files that are partial, or meant to be included inside another template using an HTMX call use the prefix `partial-`
2. The files that aren't partials, have the prefix `htmx-`

I believe that is worth adding the extra couple of characters in the last case to make it more obvious that we are dealing with an HTMX file.


Felienne:
There sadly is a small conflict now, I will see if I can fix it!

Felienne (approved):

PR #4304, opened 2023-05-20
[CHORE] Remove headers from .po files

rix0rrr:
The headers in the `.po` files don't really add a lot of value, but the time stamps in them do cause merge conflicts constantly.

Remove the headers, update the commands in `CONTRIBUTING.md` to include the `--omit-header` flag, and add a validation script that will fail if someone runs a babel command from shell history and the headers are re-added.


TiBiBa (approved):
Nice change, all looks good to me!

PR #4311, opened 2023-05-24
Ru 'Common Errors' explanation textbox

n0abaij:
**RU / New Media Lab issue**

**Description**
Next to the title of the 'Common Errors' card, there is now a question mark icon. If you hover over it, a text box pops up with extra information about the AI detection system, and a disclaimer that it is not a foolproof system. 

**Relevant technical choices**

- Used a tooltip for the hover-textbox functionality
- The text for the text box is now directly pasted in `class-live-stats.html`. In the future, the text can be added as a message in the relevant translation files.

**How to test**

- Check if the textbox only becomes visible when hovering over the question mark itself (not the area around it)
- Check if the textbox appears natural and is aligned correctly with the question mark icon


sonimegha123:
Reviewed the code and it works as expected.

PR #4314, opened 2023-05-25
Answers for level 1 in the teacher manual

MarleenGilsing:
Hi @Felienne, 

I made this using the structure of the common mistakes part of the teacher manual. If you like it this way, I can make it for all the other levels as well.


Felienne:
Thanks @MarleenGilsing!

Contrary to the other parts of content, here you will have to manually copy the new content from English to the other languages or Weblate will get angry and delete stuff (this is why the tests are failing).

Felienne:
Otherwise, this is a great change!! 

I was also thinking, can we make a new user account (Hedy_answers or something), save all the answers there and link to them? That will make it easier for people to find them.

MarleenGilsing:
We could merge this to the website. It's the answers upto level 3 now. But as we discussed, I think linking to the answer accounts might be a better idea, because this was quite a hassle.

MarleenGilsing:
@Felienne I have made #4372 now and I think it might be a better way of showing the answers than this idea. It might have been a bit of waste of my time to make this PR then, but I think it will look weird to have the answers of level 1-3 looking like this and then referring to the user account for all the other answrs. 

I will manually copy the new content of that answer-account PR to the other languages, because the tests are failing there too.

MarleenGilsing:
We decided to use the answer account instead of this way, so I'm closing this PR. It should not be merged.

PR #4316, opened 2023-05-26
[FIX] removes the gap

Felienne:
The programmer's mode has a "gap":

image

This shows I should stay out of the css :D 

This PR fixes it :)


Felienne:
Not sure why this has weird old commits in it, trying again

PR #4318, opened 2023-05-26
[LANGUAGE] Functions into level 12

jlole:
**Description**

All functionality of functions are now added in level 12
**Adds** highlighting tests for functions 
**Adds** nested functions exception


Felienne:
Due to the merging of your other PR, there are now a few issues, could you take a look @jlole?

Felienne (approved):
Well done!! I will add some more adventures so we can make sure this launches with the next deploy 🥳

PR #4322, opened 2023-05-31
Ru legendandlist

AriyanT98:
Created a hoverable legend for the Student Details section which lets the user know about the different bubbles and how they can be used


static/css/generated.full.css
```diff
@@ -1126,7 +1126,7 @@ code {
   bottom: 0.5rem;
   left: 0.5rem;
 }
-
+Fz
```
HcKide, line 1129: I think this is a small mistake

templates/class-live-stats.html
```diff
@@ -136,9 +136,25 @@ <h2 class="text-center">Class Overview</h2>
...
+                        Student List
+                    </h2>
+                    <div class ="relative">
+                        <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-6 h-6">
+                            <path stroke-linecap="round" stroke-linejoin="round" d="M9.879 7.519c1.171-1.025 3.071-1.025 4.242 0 1.172 1.025 1.172 2.687 0 3.712-.203.179-.43.326-.67.442-.745.361-1.45.999-1.45 1.827v.75M21 12a9 9 0 11-18 0 9 9 0 0118 0zm-9 5.25h.008v.008H12v-.008z" />
+                        </svg>
+                        <div class='text-left tooltip opacity-0 hover:opacity-100 duration-300 absolute inset-0 z-10 flex justify-left items-top -mr-6 mt-0.5'>
+                            <span class='tooltiptext text-xs font-medium flex justify-left absolute rounded shadow-lg p-1 w-max' style="color:black"> 
+                                <ul style="list-style: none;"> 
+                                    <li><button type="button" class="h-3 w-3 border-2 border-blue-800 rounded-full"></button> - Current Adventure</li>
```
HcKide, line 150: use these class settings for button to reflect actual student list: `h-6 w-6 m-1 border-4 border-blue-500 rounded-full bg-gray-400`

templates/class-live-stats.html
```diff
@@ -136,9 +136,25 @@ <h2 class="text-center">Class Overview</h2>
...
+                                    <li><button type="button" class="h-3 w-3 rounded-full bg-green-600"></button> - Quiz</li>
```
HcKide, line 151: Instead of `Quiz` maybe say like, `Finished Quiz`

templates/class-live-stats.html
```diff
@@ -136,9 +136,25 @@ <h2 class="text-center">Class Overview</h2>
...
+                                    <li><button type="button" class="h-3 w-3 rounded-full bg-blue-500"></button> - Attempted</li>
```
HcKide, line 152: use these class settings `h-6 w-6 m-1 rounded-full bg-blue-500`

templates/class-live-stats.html
```diff
@@ -136,9 +136,25 @@ <h2 class="text-center">Class Overview</h2>
...
+                                </ul>
```
HcKide, line 153: you could optionally include a fourth bubble that is gray: `h-6 w-6 m-1 rounded-full bg-gray-500` which says something like: `Not Attempted`

HcKide (changes requested):
See comments for which changes I suggest

templates/class-live-stats.html
AriyanT98: The issue with bigger bubble sizes will be that the legend will be too big as a pop up, we just need to show the user the color changes and thus the size doesnt need to be same as original ones

AriyanT98: But i can chnage them to the  one you asked for .

PR #4325, opened 2023-05-31
[RU] misconception detection

HcKide:
To be merged ONLY by members of NML!

The Misconception Detection / Common Errors is now functional. 
That includes that when students make a certain number of continuous errors they are flagged and a pop up appears in the feed on the dashboard of the teacher (page needs to be reloaded)

Moreover, all functionality that was written to a secondary database has been rewritten to use only the primary database. 

Relevant technical choices:
- In database, the key `error_history` was added under `program-stats` which allows to see the chronological order in which errors were made
- In database, the keys `dashboard_customization` >`selected_levels` and `dashboard_customization` > `common_errors` were added to load in relevant customization to the dashboard
- app.py was changed to include error message when writing program stats to database

How to test:
- reset dev_database
- open Hedy on this branch, navigate to dashboard
- open incognito tab, login as student
- make three or more of the same error in a row as student
- reload teacher dashboard, misconception item should show up in common errors
- click on it and see the correct student as flagged
- student name should also be clickable
- see if an item is still resolvable and dissappears
- click on level selector and see if everything still works as intended (to test whether primary db logic is correct)


karlijnbok:
Works great :)

PR #4329, opened 2023-06-05
[UI] New tag color for tags about command adventures

### LINKED ISSUE #4312: [UI idea] Color new keyword tabs
In https://github.com/hedyorg/hedy/pull/4184 we add new tabs for keywords:

image

It would be lovely if we could give these tabs a different color for clarify, maybe a shade of "Hedy keyword" pink?

And not just in this view, I think we'd also need it in the customization page so teachers know what they are removing.

jpelay:
**Description**

Adds a new type of tag both in the normal level page and also in the customization page, for this I did the following:

1. Adding  a new type of object  `SortedAdventure` which will serve as the object being passed to the templates. Therefore, not needing changes in the databse in we want to change aspects of the tabs. For this I changed the filename `types.py` to  `frontend_types.py` and added a new file `server_types.py` for this type of data that doesn't need to be shared with the front-end, releveant since using HTMX prevents for handing data in the front-end anyway.

2. Added a set that has the Keywords Adventures so we know which types of adventures should have a different color.

3. Edited the templates for the tabs and the customize class page.

4.  Fixed the flaky failing test in  `customize_clas.cy.js`

This is how the tab looks on the customize class page:
image

**Fixes #4312**

**How to test**

You can visit the index page and also the customize class page!


Felienne:
> 4. Fixed the flaky failing test in  `customize_clas.cy.js`

🥳!!

templates/customize-class/hx-sortable-adventures.html
```diff
@@ -30,12 +28,12 @@
             hx-indicator="#indicator">
         <option value="none" selected disabled>{{_('adventures')}}</option>
         {% for adventure in available_adventures[level|int] %}
-            <option value="{{ adventure['name'] }}"> 
-                {{ adventure_names[adventure['name']] }} 
+            <option value="{{ adventure.short_name }}">
```
Felienne, line 31: Love how this makes the HTML more easy to read!

website/frontend_types.py
```diff
@@ -101,4 +102,5 @@ def from_teacher_adventure_database_row(row):
             save_name=row['name'],
             start_code='',  # Teacher adventures don't seem to have this
```
Felienne, line 103: True! These start codes in general need a bit more thinking! I will open an issue about that later!

PR #4331, opened 2023-06-06
[FIX] Remove reference to session storage

### LINKED ISSUE #4330: [BUG] Programs on the view page are sometimes loaded from session storage
**Describe the bug**

Felienne got into a situation where the view page was showing the same program every time, and it was being loaded from session storage (with an `undefined` in the key).

We only use local storage, not session storage, and the `view` page doesn't need to load from LS anyway either.

Figure out what's going on.

rix0rrr:
This code dates from 3 years ago, was a first attempt at code persistence, and isn't very useful anymore. Let's get rid of it.

Fixes #4330.

this is fine

**How to test**

Everything works exactly the same (but with less confusion reviewing).

### DIFF
--- a/static/js/app.ts
+++ b/static/js/app.ts
@@ -274,41 +274,24 @@ function initializeMainEditor($editor: JQuery) {
 
   window.Range = ace.require('ace/range').Range // get reference to ace/range
 
-  // Load existing code from session, if it exists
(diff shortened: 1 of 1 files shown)


jpelay (approved):
Perfect!

PR #4333, opened 2023-06-08
[REFACTORING] Use dictionary comprehension for request header by replacing 'for loop' to make it more efficient

shubham-kshetre:
Improving the efficiency and readability of the code by optimizing the creation of the request_headers dictionary in the before_request_proxy method.

The current implementation of the before_request_proxy() method iterates over the request.headers to exclude the header with the key "host" when populating the request_headers dictionary. This process can be simplified and made more efficient by using a dictionary comprehension.

**How to test**
* The provided script tests/ab_test_setup.sh sets up an A/B testing environment for the server. 
* Run the script
* The script will start two instances of the server in A/B testing setup.


Felienne:
Thanks for the PR @shubham-kshetre but as per our guidelines, we can't accept it, see https://github.com/hedyorg/hedy/blob/main/CONTRIBUTING.md

Also, this code has not been in use for a long while, a better refactoring is probably to clean it up in a different way, this is not necessarily useful.

PR #4335, opened 2023-06-11
[UI] Fix unreadable quiz output

rix0rrr:
The quiz has an `output` field that gets rendered to show the output of a hypothetical program. The output was being rendered as black text on a dark blue background, and barely readable.

Make it readable and treat the text as `<pre>`formatted.

Also change the width of code and output containers to look slightly nicer.

**How to test**

http://localhost:8080/quiz/preview-question/9/5 will show the difference.

Before:

image

After:

image


Felienne (commented):
Check!

image

Felienne:
(note to self) We might want to make output blocks *in* answers also white on black (in a separate PR)

image

Felienne (approved):

PR #4338, opened 2023-06-13
[FIX] Remove deprecated extensions from `babel.cfg`

### LINKED ISSUE #4337: [BUG] Babel extract commands fail
**Describe the bug**
When running `pybabel extract -F babel.cfg -o messages.pot . --no-location --sort-output` it returns this error:

```
AttributeError: module 'jinja2.ext' has no attribute 'autoescape'
```

According to this StackOverflow question: https://stackoverflow.com/questions/72651555/attributeerror-module-jinja2-ext-has-no-attribute-autoescape-while-trying-t

This line needs to be removed from `babel.cfg`:

```
extensions=jinja2.ext.autoescape,jinja2.ext.with_
```

jpelay:
**Description**

Removes the line: 
```
extensions=jinja2.ext.autoescape,jinja2.ext.with_
```
from `babel.cfg` which caused an error when executing the extracting phase. This happens because those extensions were deprecated and we recently updated flask to a newer version not realizing this.

**Fixes #4337 **

**How to test**

I tested that everything worked again by adding a test key and seeing if it was extracted and compiled.


rix0rrr (approved):
Thanks, good find!

Issue #3922, opened 2023-01-10
[BUG] Long code is partly blocked while code is running


Felienne:
**Describe the bug**
A long program sometimes is not shown entirely when running. I am not sure yet what causes this, I suspect it has to do with waiting (with `ask` or `sleep`), because long programs with only `print` do not have this issue. 

Here is an example which surely has the issue:
https://hedy.org/hedy/81afd509383c4300894d016f618a5a68/view

Run the program in Chrome in macos and after a while you will see this:

**Add a screenshot (optional)**
<img width="1518" alt="image" src="https://user-images.githubusercontent.com/1003685/211575094-04cd46f1-6463-45fa-9517-3721a6b92dc9.png">

PR #4363, opened 2023-06-27
[FIX] Reset height fix

Felienne:
This bug has a fun backstory :) 

#4215 introduced an issue where code now expands the output box, then #4302 fixed that, but caused a new issue (a gap in the programmer's mode) which #4317 fixes, but that again caused the output expansion to resurface. This reverts the original change that was not even needed since we have the if in the raw already.

Lesson learned: I should not touch the front-end without adult supervision :)


Felienne:
And this PR also seems to fix #3922 also I am not sure why (it predates all the height-shenanigans!)

jpelay (approved):

PR #4320, opened 2023-05-29
[FEATURE] Skipping Faulty Code

ToniSkulj:
**Feature**
This PR is made for making skipping faulty code possible in Hedy.

**Example**
A simple program written in level 1

afbeelding

Notice that lines 3 and 6 are skipped, the skipped statements are underlined with a red dashed line, these contain a faulty ```print``` and ```echo``` statement, the rest of the program is still executed! When a user clicks on the skipped section the 'original' error message is shown.

When clicking on ```prind Mistake``` :
afbeelding

When clicking on ```eko Hello``` :
afbeelding

**Bug - FIXED** 
There seems to be a weird bug. After an error message is shown that is not skipped the program keeps returning an Exception instead of skipping faulty code. But as soon as the local server is restarted it works again.

You can replicate this by first execting ```print hello``` in level 5 and then ```prind 'hello'```. it will skip the faulty code but return no source map due to a TypeError exception, after restarting the local server it will work again as intended.

TO-DO:
- [x] Fix bug mentioned above
- [x] Change the way we test handling exceptions
- [x] Change the way we test the source mapper

PR #4368, opened 2023-06-28
[BUG] Make source map not use grammars-Total

### LINKED ISSUE #4367: [BUG] Sourcemap Dependent on grammars-Total
The source map is dependent on the merged grammars in **grammars-Total**. 
I thought that these were also generated after deployment, but I was wrong 😅. 

This causes the skipping faulty #4320 to not work the very first time the level is executed which is undesired (I think?).
(btw, I found this bug by testing skipping faulty on alpha)

ToniSkulj:
Fixes #4367


ToniSkulj:
Hey @Felienne! Do you think we can merge this when you have the time? Alpha is still bugged out after merging #4320  (sorry for that), this PR fixes that

afbeelding

Felienne (approved):
Ah sorry, here you go! Let me know if this fixes the issues on Alpha!

ToniSkulj:
> Ah sorry, here you go! Let me know if this fixes the issues on Alpha!

Thank you, will do!

PR #4372, opened 2023-06-29
Adds a link to the answers account to the teachers manual

MarleenGilsing:
It works, but it seems like there are only 5 programs visible on the public page. Can we change this so all the programs are accessible (maybe only for this account)?


Felienne (changes requested):
Let's update the url

MarleenGilsing:
@Felienne Done

Felienne (approved):
Good to go!

PR #4383, opened 2023-07-10
[FIX] Fix crash in grid page and in class customization

jpelay:
There's an error both in the customizations and in the new grid page. I'm not sure if this is a case just of spurious data in the alpha environment or if this also happens in production. Either way this bit of defensive programming should do the trick.

### DIFF
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -399,6 +399,8 @@ class ForTeachersModule(WebsiteModule):
 
         for level, sorted_adventures in customizations['sorted_adventures'].items():
             for adventure in sorted_adventures:
+                if adventure['name'] not in adventure_names:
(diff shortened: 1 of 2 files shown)


Felienne (approved):
Yeah the customizations have a way of crashing a lot! More defenses are a smart idea regardless of whether it happens in the current start of the site. Maybe we should make an issue to add a few more defenses in other places too?

jpelay:
> Yeah the customizations have a way of crashing a lot! More defenses are a smart idea regardless of whether it happens in the current start of the site. Maybe we should make an issue to add a few more defenses in other places too?

Yes! We definitely should. We can't trust the DB to be in a proper state, then. An assumption I made when I made the customizations

PR #4386, opened 2023-07-18
[FIX] Fixes incorrect template name for htmx pop-up

### LINKED ISSUE #4385: [BUG] Reset adventures button not working
After #4303, when we changed the names of several templates, the Reset adventures button in the customizations isn't working.

jpelay:
Fixes #4385 adding the correct name of the template and also adding a test.


Felienne (approved):
Yes this fixes the 500 error!

PR #4392, opened 2023-07-21
[UI] Clicky buttons

### LINKED ISSUE #4391: [UI idea] Make buttons clicky
**Idea incl level**
I think it'd be fun and more useful for our users if the buttons are clicky. They already appear to pop out from the screen with the border, the next step is to be able to press them like real buttons.

These would be the three states of the button:

1. Base state:
image

2. Hover state:
image

3.  Click state:
image

**Additional context**

jpelay:
**Description**

Changes the classes of all the buttons to make them appear as if you can physically click them. Besides being fun, it's also an indication for the users that the button has been clicked.

**Fixes #4391**

**How to test**

You can test the buttons


Felienne (approved):
What a lovely small addition to the UX @jpelay

PR #4397, opened 2023-07-25
[UI] Add pagination to profile page

### LINKED ISSUE #4378: [UI idea] Show all programs by user
In #4372 we add a link to the user account with answers to the exercises, but by default profile pages only show 5 programs, can we make them show all programs (paginated to save long loading times)

jpelay:
**Description**

Adds pagination to the profile page, limiting to 10 programs per page. It also improves the UI by using the same template for programs as the one used in the /explore page, for this I created a new file that contains that template in this way is easier to keep track of future changes to that UI.

**Fixes #4378**

This is how the page looks now. Timon did such a good work with that design of the programs, that I think is a waste not to use it in this page too!:

image


jpelay:
Still a little bit of work needed before I'm comfortable setting this PR as open!

jpelay:
I took the freedom to redesign the User page and this is how it looks now!
image

Felienne:
I love the new layout @jpelay, but it seems we misse a few programs, my profile says 5 but I only see one.

image

You screenshot actually also shows this problem because it says 6 but there is only 3 shown.

jpelay:
> 

Hello Miss Felienne! This actually happens because in this page we're only showing Shared programs, and not all programs created, so the user actually has to set those programs as shared!

Felienne:
Ah of course, silly me! I will merge this, but I was thinking it would also be nice to have the link to the public profile in the dropdown:

image

I will make an issue

Felienne (approved):
Let's go and pray for a limited number of Weblate problems!

jpelay:
> Ah of course, silly me! I will merge this, but I was thinking it would also be nice to have the link to the public profile in the dropdown:
> image
> 
> I will make an issue

That's a great idea!

PR #4400, opened 2023-07-28
[FIX] Fixes Hedy's choice programs not being shown

### LINKED ISSUE #4399: [BUG] Hedy's choice programs are not being shown
When you go the Explore page, the Hedy's choice program are not being shown. Hedy's choice programs should be shown when the query arguments are not set, like in this case. However we're not showing them:

image

jpelay:
**Description**

Changes the assignment of the `filtered_level` variable that goes to the template. This variable was being converted to `string`, but when there was no `level` argument in the query string, and therefore hold  the value of `None` when it was passed to the template it actually was `'None'` a string. This PR fixes that behaviour.

**Fixes #4399 **


Felienne (approved):

Felienne:
Indeed puts Hedy's choice back in:

image

Should we add a front-end test so we don't break this again @jpelay?

PR #4057, opened 2023-02-23
[FEAT] Save all programs automatically

rix0rrr:
Instead of children having to click a button to save their programs, all programs will be saved automatically.
    
Introduce new UI elements to support current sharing and handing in workflows.

## Notes on this PR

We are nowhere near done, this PR is in a preview state to show the progress.

The PR is built on top of the previous refactoring PR, so will show a lot of changes that aren't actually new here (these will disappear after the previous PR is merged and this branch is updated from main).

## Implementation

The feature is going to work as follows:

- In the current model, we are expecting to have either 0 or 1 programs saved for each `(user, level, adventure)` combination (In the limit, all users will have a program for every `(level, adventure)` combination).
- On page load of the code page (for a logged in user):
  - If no program is requested, the backend will send at most one saved program with each adventure to the frontend. If it so happens there there is more than one program for a `(user, level, adventure)` combination, we will pick the most recently saved one.
  - If a program is specifically requested by id, we will give that program preference over any other program in the same `(user, level, adventure)` combination.
  - If a program is specifically requested requested by id for a tab that no longer exists, we invent a tab (by its shortname, without a story) with the program loaded into it.
  - If no program is specifically requested by id, but a program is available in localstorage for the current `(level, adventure)` combination, we will load that instead (covering the case of writing code before logging in).
- During interaction with the page:
  - If the user has modified the code since the last save point AND (has stopped typing for 5 seconds OR switches tabs OR clicks the run button OR changes the program title [on blur] OR clicks share/hand in), the code is saved.
  - If the user is not logged in, saving entails writing to localStorage.
  - If the user is logged in, saving will write to the server database. If the save was triggered by clicking the Run button, saving will be done as part of the parse (since the program is sent to the server anyway). If the save was triggered by a client action, the code will be sent to the server via a separate POST. A save to the server clears the corresponding localStorage entry.
- The new sharing/hand in dialogs replace the equivalent functionality that exists today. If you are not currently logged in, they will prompt you to log in instead. That login will send you back to the code page with the right tab focused, giving the child an opportunity to click the button again. (As a consequence of the behavior described above, the program will be retrieved from local storage, but stored to the server upon clicking this button).

## TODO

- [x] Translations
- [x] JavaScript to make the popups work
- [x] Show/hide UI elements based on user being logged in/being in a class, etc
- [x] Make sure that the login flow works and sends the user back to the right tab with the program still there
- [x] Backend and database work to support this
[...]

PR #4402, opened 2023-08-01
[FIX] Fix share button in Programs page

### LINKED ISSUE #4398: [BUG] Share button in my programs doesn't work
Screenshot of console: 

image

jpelay:
**Description**

Restores the `share_program` function that was deleted in #4057, and also adapted it to the way Rico was calling this type of function. Besides that, so this doesn't happen again I also added a test for this feature.

**Fixes #4398**

**How to test**

You can save a program, go to the Programs page and then share and unshare it! You can also run the cypress tests!


Felienne (approved):
Thanks for the fix and for adding tests for all the small broken things in the front-end! Step by step we are getting more robust!

jpelay:
> Thanks for the fix and for adding tests for all the small broken things in the front-end! Step by step we are getting more robust!

It's my pleasure! Yes, that's the objective. There are lots of small things that need more testing. I found these mistakes by testing around developing other features.

PR #4413, opened 2023-08-16
[FIX?] Remove Skip Line feature

jpelay:
Since the skip lines are failing in ways we don't understand yet, and it hinders our ability to ship new features, this PR temporarily comments some features of the skip lines code. But what do you think of removing them in their entirety? It would look better, but in the future, it'd be harder to reincorporate the feature. What do you think?


jpelay:
Please don't accept this for a day or two more, so I can keep looking into the underlying issue. Because I can't  reproduce it locally yet.

Felienne:
Hi @ToniSkulj and @jpelay!

Thanks for the hard work on the fixes! I will merge this soon, deploy and then test the fix in #4414

Felienne (approved):

PR #4414, opened 2023-08-19
[FIX] Skipping Faulty lines bug

### LINKED ISSUE #4412: [BUG] A weird issue in skip line!
It seems skip line is causing issues on main (I am hearing from teachers) so I will be rolling it back shortly.

The issue is hard to reproduce, but it seems sometimes you get an error message from another program, like this:

image
 
or:

image

ToniSkulj:
Should fix #4412 & #4413, I have not been able to reproduce the issue locally anymore.


jpelay:
Hi, Tonny! Thank you so much for this fix. I've testing it locally and I also can't reproduce the error and everything seems to be working fine. I have a quick question regarding the tests that you changed, now that they don't end up in the exception, but rather they return code processed by the Skip Faulty feature, I was wondering if there's still a way to not process the code using that feature, and let the test be successful if the desired exception is thrown. 

Other than that this fix works great!

ToniSkulj:
Hi @jpelay! My pleasure, you did the majority of the work by finding a way to reproduce the issue!
>  I was wondering if there's still a way to not process the code using that feature, and let the test be successful if the desired exception is thrown.

Yes, this is possible. The _single_level_tester_ and _multi_level_tester_ have a skip_faulty parameter now. If set to True (the default) then the skipping feature will be used, otherwise mistakes will not be skipped and therefore will raise an exception.
Hope this answers your question!

Felienne:
> Hi @jpelay! My pleasure, you did the majority of the work by finding a way to reproduce the issue!
> 
> > I was wondering if there's still a way to not process the code using that feature, and let the test be successful if the desired exception is thrown.
> 
> Yes, this is possible. The _single_level_tester_ and _multi_level_tester_ have a skip_faulty parameter now. If set to True (the default) then the skipping feature will be used, otherwise mistakes will not be skipped and therefore will raise an exception. Hope this answers your question!

Hi @jpelay!

I think it would be good to merge this, can you approve if you are ok with everything? Two open questions for me:

1. Is the issue you asked about above fixed?
2. If we merge this, the feature will still be turned off right? Because  we do still want to think about how to communciate it to teachers and we don't want to change this in the middle of the term.
Of so, let's merge this!

jpelay:
> 1. Is the issue you asked about above fixed?
> 2. If we merge this, the feature will still be turned off right? Because  we do still want to think about how to communciate it to teachers and we don't want to change this in the middle of the term.
>    Of so, let's merge this!

1. Yes!
2. No, if we merge this the feature will be active again, that's why I haven't approved it. Maybe @ToniSkulj can take care of that :pray:

ToniSkulj:
>     2. No, if we merge this the feature will be active again, that's why I haven't approved it. 
>     Maybe @ToniSkulj can take care of that 🙏 Otherwise I can do it too! 😄

Sure @jpelay, I will take care of that this weekend!
I was thinking about an environment variable to enable/disable the feature.

Felienne:
> > ```
> > 2. No, if we merge this the feature will be active again, that's why I haven't approved it. 
> > Maybe @ToniSkulj can take care of that 🙏 Otherwise I can do it too! 😄
> > ```
> 
> Sure @jpelay, I will take care of that this weekend! I was thinking about an environment variable to enable/disable the feature.

Thanks!! Any kind of feature flag would be perfect, env var is surely fine by me!

ToniSkulj:
Hi @Felienne! [Add feature env flag](https://github.com/hedyorg/hedy/pull/4414/commits/eaf7265748b73e37c01bc3a4c31e91032ba5b586) adds a feature flag that defaults to False for skipping mistakes.

Also @jpelay, I changed the way that the "clicking the mistake to see the error message" works. It used to rely on the editor but now it does not, making the transition to CodeMirror in the future easier.

jpelay:
Hi, @ToniSkulj!

> Also @jpelay, I changed the way that the "clicking the mistake to see the error message" works. It used to rely on the editor but now it does not, making the transition to CodeMirror in the future easier.

Thank you so much for this!

I've been trying the changes in the branch and I just have a couple of comments!

When the environment variable is unset, and therefore skip faulty is disabled, when you try tun a program with errors repeatedly, the error box disappears, and after a couple of runs it stops showing:


https://github.com/hedyorg/hedy/assets/45865185/6726aa9a-bb0f-404b-be17-9c5c9031d2bc


Secondly, and this is a lot minor and I can change this myself in the future, but worth mentioning:

I think since every line has their own error associated with it, when you click in a line with no errors the error box should disappear.

image


I think I recall this used to be the case? If not, what do you think about this idea?

ToniSkulj:
Hi @jpelay !! No problem😃

> When the environment variable is unset, and therefore skip faulty is disabled, when you try tun a program with errors repeatedly, the error box disappears, and after a couple of runs it stops showing:

Good catch! I removed de fading of the error when hiding it. This resolves the issue in my browser (Firefox). Let me know if the issue is also fixed for your browser.

> I think since every line has their own error associated with it, when you click in a line with no errors the error box should disappear.

I 100% agree with you. I added that too, if a user clicks anywhere on the editor the error will be hidden.

Felienne:
Hi @jpelay!

I think this can be merged right? Feel free to approve!

app.py
```diff
@@ -2129,28 +2125,23 @@ def public_user_page(username):
...
         return utils.error_page(error=404, ui_message=gettext('user_not_private'))
     user_public_info = DATABASE.get_public_profile_settings(username)
-    page = request.args.get('page', default=None, type=str)
     if user_public_info:
-        user_programs = DATABASE.public_programs_for_user(username,
-                                                          limit=10,
-                                                          pagination_token=page)
-        next_page_token = user_programs.next_page_token
-        user_programs = normalize_public_programs(user_programs)
+        user_programs = DATABASE.public_programs_for_user(username)
```
jpelay, line 2129: I think these are unwanted changes right? I'm gonna revert them ;D

jpelay (approved):
Hi @ToniSkulj!

Thank you so much for fixing this insidious bug! And making it work with the changes to the editor interface as well. There were a couple of unintended changes, but I fixed those, and now we're good to go! ;D

jpelay:
> Great work you two!!! Hope this will all be compatible with CodeMirror (but it no, we will deal with that too!)

I'll deal with Code Mirror!

PR #4417, opened 2023-08-21
[FIX] Grid page failing

jpelay:
Classes with old programs were failing because some of them did not have the field `adventure_name`. This PR adds an if when retrieving programs that have this characteristic. It also deletes the debug prints that were added in #4416.


Felienne (approved):

Felienne:
Thanks for the quick fix!!

PR #4419, opened 2023-08-22
[FIX] Grid not working when there are no customizations

jpelay:
**Description**

The Grid page fails when it tries to load customizations for a class that doesn't have any, of course this causes an error because it's then trying to access a field that isn't there. This PR fixes that mistake by creating a new default customizations object in case there ins't one. 

**How to test**
You can run `./feed_dev_database` and then delete the only customizations object in the database. Access the Grid for that class and it should work


Felienne (approved):
Yes, this does the job, thanks!

PR #4422, opened 2023-08-24
Adds a warning to a intentionally not working example code

MarleenGilsing:
Adds a warning to a intentionally not working example code


Felienne (approved):
Nice, thanks! Good balance between not confusing people and making the point!

Issue #4424, opened 2023-08-24
[CONTENT] Answer account problems


MarleenGilsing:
I've created this account that contains al the answers, so teachers can use it to check the answers to all the exercises. 
https://hedycode.com/user/hedy_answers 

Only problem is that if you click on this link, only the last 5 program of this account are shown. Is there any way we could make all the programs visible for this account? 

And maybe - if this is not too much trouble to make- make the page only available for teachers, so students can't use it?

Felienne:
Duplicate of #4378, will be closed soon by #4397

PR #4425, opened 2023-08-24
[CONTENT] Teacher manual improvements NL EN

MarleenGilsing:
The teacher manual is fully up to date again! :)

I would only like to add the answer account to it, but #4424 

Some tests might get angry because I've added some headers that will not exist yet in other languages. 
Weblock will definitely be angry, haha.


MarleenGilsing:
@Felienne This one is done. I changed it for all languages, but the vi and bg languages seem to have a different structure. I couldn't figure out how to change everything in those 2 languages.

Felienne:
ok, I hope Turkish is ok! If so, we can merge this now!

Felienne (approved):

yilmazdurmaz:
It is hard for me to follow all these changes, especially when you move parts up and down in yaml files. are you making these changes manually or through a program?
thus, I am counting on them appearing on Weblate as "needs editing".

by the way, I noticed there were a lot of content changes on the source but these did not change on other languages that got their strings from auto-translation (actually auto-copy from source). Some quiz questions were completely different than the source. do you have an easy way to detect them and sync to latest?

Felienne:
> It is hard for me to follow all these changes, especially when you move parts up and down in yaml files. are you making these changes manually or through a program? thus, I am counting on them appearing on Weblate as "needs editing".

Yes, Weblate should see then as "needs editing". Let me know if anything is amiss.

> by the way, I noticed there were a lot of content changes on the source but these did not change on other languages that got their strings from auto-translation (actually auto-copy from source). Some quiz questions were completely different than the source. do you have an easy way to detect them and sync to latest?

Yes this is a known issue in Weblate sadly, but there is no easy way to fix this, other than a manual find and replace.

PR #4431, opened 2023-08-24
[FIX]Quiz level 14, question 8 answer corrected.

Muhammadummerr:
The Answer to question 8 in quiz of level 14 has been corrected.


Felienne (approved):

Felienne:
Ah I now understand the issue! This PR fixes the issue, but #4439 does that too, and that does it in a bit more logical way (rollercoasters tend to have minimum lengths) So I will revert this because #4439 has also been accepted.

Sorry for the confusion!

PR #4437, opened 2023-08-30
[FIX] QUIZ 14 ANSWER 8 UPDATED IN MULTIPLE LANGUAGES.

Muhammadummerr:
QUIZ 14 ANSWER 8 UPDATED IN MULTIPLE LANGUAGES.


Felienne:
Thanks for your contribution @Muhammadummerr!

I suspect this issue in fact might have already been fixed by https://github.com/hedyorg/hedy/pull/4439, could that be the case? There we change the code rather than the answer. I am setting this to draft (so we don't keep it on the list op open PRs).

Let me know if you need more info!

Felienne:
See discussion in https://github.com/hedyorg/hedy/pull/4431

PR #4438, opened 2023-08-30
[FIX] [TESTING] Adds wait to restore customizations call

jpelay:
**Description**

Sometimes we have tests that fail somewhat randomly; the most recent occurrence of this was [in this test run](https://github.com/hedyorg/hedy/actions/runs/6013879982/job/16312322160), where the test  `Disabling current level displays a message` from the customize page class failed. That's a bit weird, because that test just run a bit of Javascript and doesn't do any API calls. So my process was to add a wait to the  `/restore-customizations` POST in Python, since that piece of code is executed before each test of this file, and I could consistently see some other tests of this same group fail, not the one that failed in the run mentioned earlier though, but I think it really has to do with that. My solution was to add an intercept and a wait to the API call so if it takes a little bit longer to respond it won't cause issues.

I suggest maybe we can try this and if this test fails again we can keep debugging.


Felienne (approved):
I just had another flaky tests, let's see if this helps!

jpelay:
> I just had another flaky tests, let's see if this helps!

Can you tell me in which PR?

PR #4449, opened 2023-09-02
[TESTING] Skip all default tabs rather than use a hard-coded list

Felienne:
We used to hard-code all names, but that list is getting too long and this is nicer!


Felienne:
Merging this since it will likely fix the open Weblate PR (and will make subsequent ones less likely to error)

PR #3119, opened 2022-08-02
No hashing needed!

jpelay:
**Description**

A few months ago the Skulpt team merged the change that allowed them to parse non ASCII strings as variables, allowing us to finally ditch hashing our variables. Therefore, this PR introduces 3 changes:

* Updates Skulpt.
* Changes the functions `hash_var` and `hash_needed` in `hedy.py`. Now we still need preprocessing for variables, because some of them might be named like python keywords. We keep a function called `escape_var` for that, if the variable is named like a keyword it prepends a _ to it.
* Updates the tests that used hashing.

**Fixes #959**

**How to test**

* You can recompile the front-end and see the generated code.

**Checklist**
Done? Check if you have it all in place using this list:*
  
- [ ] Describes changes in the format above (present tense)
- [ ] Links to an existing issue or discussion 
- [ ] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #4465, opened 2023-09-08
[LANGUAGE, UI] Integrate Skulpts debugger to Hedy

### LINKED ISSUE #4427: [LANGUAGE idea] Integrate Skulpt debugging capabilities into Hedy
As one of the first steps towards #3397 we need to use Skulpt to be able to stop the execution of the program, and move forwards (possibly bacwards) through it. 

As of right now, the current debugger does not actually stop the execution of the program, but rather compiles it line by line, essentially treating it as a new program for each line. This approach, albeit very clever and creative, doesn't work the moment we introduce nesting and blocks to Hedy. So we need to integrate the debugger from the ground up and deprecate this old debugger.

The support that Skulpt has for this is rather barebones, with a [single PR](https://github.com/skulpt/skulpt/pull/559/files#diff-65f3f9492cf15bdc2d7f9fa7b98e7f06c434a8980624724caa721b6195c6779b) 7 years ago introducing the feature. But we know this is possible and feassible because in https://webtigerjython.ethz.ch they implemented a debugger with Skulpt, so our task is to use this as the base to set up and get the debugger working as a first step. 

It's important to note, too, that WebTigerJython implementaton of the debugger they manipulated Skulpt code and changed it, I'm not sure if we need to that for Hedy, most likely yes, because as I said, the support and documentation for the debugger is very limited. As a point in favor of manipulating Skulpts code is that the debugger won't be a part of Skulpts core functionality, and any error's here won't affect the way in which the programs are executed (they said so themselves when approving that PR). Also, since Skulpt is also Open Source, perhaps we can actually merge this as part of their code and have it officially supported there and not maintain ourselves a fork.

jpelay:
**Description**

Skulpt has an integrated debugger that is shipped as a separate entity and is envisioned to work like pdb. The basis that is there, however, can help us a great deal in getting our own debugger to work. For this, first we need to integrate the debugger that comes with Skulpt and edit it so it can work with our code. This will not affect Skulpt, since we are just extending it, so any problems that might come with the debugger will not affect the normal execution of code.

However, by doing this, we are taking the responsibility to maintain this piece of code, but I think is very much worth the cost. 

The code in this PR will most likely be a little ugly since this is a proof of concept about getting this to work, and then I'll tackle the task of improving the code and making it adhere to our standards.

**fixes #4427**


Felienne:
OMG I LOVE WE ARE EXPLORING THIS!!

I am ccing @juliabolt, she is also exploring more work on the debugger (since yesterday, so she is very new :)) It might make sense in a few weeks to meet (she might also drop into the contributors meeting in a while)

jpelay:
> OMG I LOVE WE ARE EXPLORING THIS!!
> 
> I am ccing @juliabolt, she is also exploring more work on the debugger (since yesterday, so she is very new :)) It might make sense in a few weeks to meet (she might also drop into the contributors meeting in a while)

Great! Ok! I'm gonna keep working on this and will keep you updated. I also have a functioning version of the debugger from a past version of the Skulpt site, so I can provide that as well, if you'd like

jpelay:
It works now for print!
https://github.com/hedyorg/hedy/assets/45865185/5782c3c0-3696-464c-b99e-4de0bc585b75

jpelay:
Some bad news, the debugger doesn't work with the input function we supply to Skulpt, it only works when the input function is a simple alert from Javascript


https://github.com/hedyorg/hedy/assets/45865185/228802ca-9d04-4cc1-a0c8-fec5eb02c96c

Felienne:
> Some bad news, the debugger doesn't work with the input function we supply to Skulpt, it only works when the input function is a simple alert from Javascript
> 
>  debug.3.mp4

Ah what a pity! Could we somehow catch the input and replace it with our box? Something like:

* Show our box first, store inout
* Run the simple alert, enter the stored input (without the user seeing)
* Continue?

But I am not at all sure these things are feasible!

jpelay:
> Ah what a pity! Could we somehow catch the input and replace it with our box? Something like:
> 
> * Show our box first, store inout
> * Run the simple alert, enter the stored input (without the user seeing)
> * Continue?
> 
> But I am not at all sure these things are feasible!

That's a very good idea, not sure if it's feasible due to the program having to stop in order to do that, but I can try!

jpelay:
So I managed to fixed the problems and it works with ask now!
https://github.com/hedyorg/hedy/assets/45865185/59c5b017-96fe-42a6-854a-592461e462c1

jpelay:
I need to work with the integration between the source mapper and the debugger, but if you hardcode the breakpoints, debugging nested code is possible!

https://github.com/hedyorg/hedy/assets/45865185/84aab654-3a40-4b51-a422-28bf2260c22d

Felienne:
> I need to work with the integration between the source mapper and the debugger, but if you hardcode the breakpoints, debugging nested code is possible!
> 
>  debugger.repeat.block.mp4

OOOWWW so great!!!! This is going to help students so much!!

jpelay:
I'm testing the debugger with each statement to see where I can place the breakpoints in a way that makes sense and works. So far everything works, except for sleeps after repeats. It even works for nested code like this, in the console you can see the current python line being executed:

https://github.com/hedyorg/hedy/assets/45865185/be492190-f9b9-44be-9868-678db97fe18f

jpelay:
After testing the debugger with every statement in Hedy it works for almost everything, except the PyGame. Although I think I need to look into it more to decide definitely whether it works with the pygame or not. But it works for everything else!!! In every level!!!

jpelay:
Note to self:  Need to change the way the editor handle runtime errors and mimic the way Skulpt handles them in a normal execution path.

jpelay:
Hi @Felienne! I have a question about how can we proceed with respect to marking in which Python lines to place the breakpoints.  I have come out with two proposals: 

1. Mark the lines with a comment at compile time. This would consists on a flag being set in the `ConvertToPython` class, if the flag is set to true we would generate something like this:

```
__trtl = 100
try:
  __trtl = int(__trtl)
except ValueError:
  raise Exception(f'....')
t.forward(min(600, __trtl) if __trtl > 0 else max(-600, __trtl)) # ___BREAKPOINT___
time.sleep(0.1)
__trtl = 45
try:
  __trtl = int(__trtl)
except ValueError:
  raise Exception(f'...')
t.right(min(600, __trtl) if __trtl > 0 else max(-600, __trtl)) # ___BREAKPOINT___
```

This has the advantage of not needing any post processing in the Javascript, because we'd only need to iterate over the code, find the lines with `# __BREAKPOINT__` in them, and set the breakpoint there.

However, I do understand that by doing this we can be adding a failure point in one of our most critical systems, possibly causing issues with the transpilation process if a bug is introduced. 


2. Process the lines in the front-end, deciding where to put the breakpoints by looking at the source map and then processing the Python code. In this case the source map would look like this:

```
    "0": {
        "hedy_range": {...},
        "python_range": {...},
        "error": null | exception
        "command": "var" | "assign" | "if" | "while" | "for" | ...
    }
}
```

Then, in Typescript, we'd iterate in the sourcemap, and depending on the level and which command we're processing, decide where to put the breakpoint. For example, in the case of a  `repeat` on level 7:

```typescript
if (statement.command === 'repeat) {
    python_code = getCodeFromSourceMap(statement.python_range) // get the Python lines as an array
    for (line in python_code) { 
        if (line.match('/for __i__ in range.+/')) {
            // place breakpoint in this line
        }
    }
}
``` 
Some commands, however, like the turtle commands or the if in level 5-7 in the source map appear as  the rule `command`, having we to distinguish between them by processing their respectives strings. 

As you can see this method involves more code and we'd be dealing with the Python code without knowing the structure of the underlying code, as opposed as doing it directly at compile time. We'd have to do a ton of string processing for each command of each level, at least when they change. Another downside of this method is that when changing the generated Python code in the transpiler, we'd have to edit this processing as well. However, this would mean that the transpiler code would be unchanged.

What do you think?

Felienne:
Hi @jpelay!

Thanks for the extensive explanation!! This really helped me to quickly understand the two options. I would propose option 1 for two reasons:

1) The Python transpilation part (while being our most critical system as you say) is the best tested part of the code, I don't fear so much that it will break, and if it break, we will now soon from users. Breaking javascript is trickier to test and to detect for users too. 
2) In general, our line recently has been "more Python, less javascript" (with the HTMX change, for example) More things in one system for writing and testing helps to get a better grip on the code base.

From your text I also read (maybe in between the lines) that you'd prefer option 1? I see a lot of downsides mentioned in option 2. If your main worry for 1 is breaking the transpiler, I think we should take the risk because it is well tested and well understoond.

jpelay:
> From your text I also read (maybe in between the lines) that you'd prefer option 1? I see a lot of downsides mentioned in option 2. If your main worry for 1 is breaking the transpiler, I think we should take the risk because it is well tested and well understood.

Yes, I greatly prefer option 1! But wanted to consult with you before committing to one of these two options! Thank you! I will start working on this right now; if I keep this pace I think we might have a functioning prototype by our next meeting.

jpelay:
There's still some work left to do, but it kinda works!!!!!!!

https://github.com/hedyorg/hedy/assets/45865185/8c682094-219c-49f6-8b61-521851b35c03

jpelay:
This PR is now ready for review! There are some things that I want to improve and change, but I think we can leave that for a new PR (I need to have this finished so I can present it in my dissertation)

Felienne:
This seems to work GREAT from the front-end @jpelay!! Finally a debugger for if's and loops!!!!

One tiny UX issue I saw is that if you start the debugger, you get the confetti immediately:

image

Maybe we can suppress it when running from the debugger? Or at least until the final line?

jpelay:
> This seems to work GREAT from the front-end @jpelay!! Finally a debugger for if's and loops!!!!
> 
> One tiny UX issue I saw is that if you start the debugger, you get the confetti immediately:
> 
> image
> Maybe we can suppress it when running from the debugger? Or at least until the final line?

Yes, of course!

Felienne:
Wooohoooooo!!! All seems to be in order! I will deploy now, and then merge, so we can test for a bit in alpha just to be sure.

Felienne:
Hahaha did you rewrite this Skulpt file entirely :D 

image

static/js/app.ts
```diff
@@ -18,6 +18,7 @@ import { HedyEditor, EditorType } from './editor';
 import { HedyAceEditorCreator } from './ace-editor';
 import { stopDebug } from "./debugging";
 
+export let theGlobalDebugger: any;
```
Felienne, line 21: Would it make sense (not in this PR but in the future!) to also make an interface for the debugger like we do for the editor?

static/vendor/skulpt.min.js
Felienne, line 1: Is this a new version of Skulpt (if so, which one) or is this the one we were already supporting (1.3.0 I think? https://github.com/hedyorg/hedy/pull/753/files)? Would be good to document which version we are on, just in case

static/js/app.ts
```diff
@@ -853,20 +871,11 @@ export function runPythonProgram(this: any, code: string, sourceMap: any, hasTur
...
-  const variableButton = outputDiv.find('#variable_button');
-  const variables = outputDiv.find('#variables');
-  outputDiv.empty();
-  outputDiv.append(variableButton);
-  outputDiv.append(variables);
-
-  const storage = window.localStorage;
   let skulptExternalLibraries:{[index: string]:any} = {
       './extensions.js': {
         path: theStaticRoot + "/vendor/skulpt-stdlib-extensions.js",
```
Felienne, line 876: In light of our new changes, might it be safer to load all Skuplt recourses ourselves? I worry a bit that they might update stuff that breaks our new code (but I can't oversee whether that is a reasonable worry, what do you think @jpelay?)

We added this CDN link in https://github.com/hedyorg/hedy/pull/4490 for performance reasons (so we will make @rix0rrr sad haha)

static/vendor/skulpt.min.js
jpelay: This is a version of Skulpt I built from their source. I also recall we updated that Skulpt version to a more recent one to be able to use Unicode variable names! We did that in this PR: https://github.com/hedyorg/hedy/pull/3119/files

jpelay: I still need to add the fork to our organization, and include that in the documentation.

Felienne: > This is a version of Skulpt I built from their source. I also recall we updated that Skulpt version to a more recent one to be able to use Unicode variable names! We did that in this PR: https://github.com/hedyorg/hedy/pull/3119/files

Ah yeah, you are right, that is even more recent. We do want to keep that version haha, or we will have to hash all variable names again!

Felienne: > I still need to add the fork to our organization, and include that in the documentation.

I guess we want to do that in a next PR? Shall we make an issue so we don't forget?

static/js/app.ts
jpelay: Sorry, Felienne, but I don't fully understand what you mean by load all Skulpt resources ourselves? I didn't remove that part of the code, the `storage = window.localStorage` was being used to know if we were in debug mode, no need to do that since now we have two clearly differentiated execution paths!

static/js/app.ts
jpelay: Yes! I'm planning on two things: leave the debugger in JavaScript and write an vendor interface like we do for some other Skulpt code we execute . Or, convert it entilery to TypeScript.

The interface we use for Skulpt is here ((https://github.com/hedyorg/hedy/blob/4c4d24ee0322f37205f0bed2c1a5e44e7f4a1d39/static/js/vendor.d.ts#L4-L40))

static/vendor/skulpt.min.js
jpelay: Opened #4685

static/js/app.ts
Felienne: Ah never mind, my bad! I thought thestaticroot was referring to a generic CDN hosting libraries (something like https://developers.google.com/speed/libraries) but I see now it just refers to our own repo, so we can leave it as is!

Felienne (approved):
Let's gooooooooo!!!!

Issue #4455, opened 2023-09-04
[BUG] Flaky Cypress Tests


jpelay:
**Describe the bug**
In [this job ](https://github.com/hedyorg/hedy/actions/runs/6076542251/job/16484723178?pr=4435)there were 2 failing Cypress unrelated to what's being done in that PR. When I rerun the tests again on GH and local they worked again, so I'm making this issue as a note to self to investigate this.

The two failing tests are the following:
[BUG] Flaky Cypress Tests
 
```
1 failing

  1) Is able to download login credentials
       "before each" hook for "Passes":
     CypressError: Timed out retrying after 4000ms: `cy.type()` failed because this element is not visible:

`<strong id="username" class="inline-block w-1/2 px-2">teacher1</strong>`

This element `<strong#username.inline-block.w-1/2.px-2>` is not visible because its parent `<div#profile-change-body.profile-section-body>` has CSS property: `display: none`

 Running:  for-teacher_page/class_page/logs_page/search_button.cy.js                     (82 of 84)


  Is able to press the search button
    1) Passes
Cannot find coverage file /home/runner/work/hedy/hedy/tests/.nyc_output/out.json
Skipping coverage report


  0 passing (10s)
  1 failing

  1) Is able to press the search button
       Passes:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `.view_class`, but never found it.
      at Context.eval (webpack:///./cypress/e2e/for-teacher_page/class_page/logs_page/search_button.cy.js:10:7)


```

**Paste the Hedy code & level**
It is super helpful if we can copy-paste the Hedy code to test, so please paste the code here, and don't forget to tell us what level you were in.

**Add a screenshot (optional)**
Make a picture or screenshot to show the issue. Tip! You can make a screenshot and simply paste the image into GitHub with ctrl-v.

**What machine and browser you were using (optional)**
If the issue concerns things in the website, let us know:
- What computer you are using (Windows, Mac, Linux?)
- What browser you were using (Chrome, Edge, Safari)

jpelay:
The test in the file `editor_box.cy` inside the `describe` Test editor box functionality` are constantly failing. We need to remove them for now, and then keep working on them.

Felienne:
#4529 might have fixed some issues, but there still is a weird flaky test in the test for creating a class (if I see it occur, I will add more details here) So we need to keep this open for now.

Felienne:
Here it is again in #4581 

[...]

PR #4468, opened 2023-09-08
[FIX] Highlighted commands are shown in errors again

jpelay:
**Description**

There's a bug right now in Hedy were the keywords in the errors are not being highlighted:

image

This is due to a syntax error in `additional.css`. Now is back to normal:

image

### DIFF
--- a/tests/cypress/e2e/hedy_page/editor_box.cy.js
+++ b/tests/cypress/e2e/hedy_page/editor_box.cy.js
@@ -28,7 +28,7 @@ describe('Test editor box functionality', () => {
     cy.get('#editor > .ace_scroller > .ace_content').click();
     // We wait until the editor is focused
     // TODO: replace this wait. The editor takes a while to be focused
-    cy.wait(1000);
+    cy.wait(2500);
     cy.focused().clear();
   });
   
--- a/tests/cypress/e2e/hedy_page/editor_box.cy.js
+++ b/tests/cypress/e2e/hedy_page/editor_box.cy.js
@@ -61,4 +61,26 @@ describe('Test editor box functionality', () => {
     cy.get('#output').should('contain.text', 'Hello world');
     cy.get('#ask-modal').should('be.visible');
   });
+
+  it ('When making an error the error modal should be shown', () => {
+    cy.get('#editor').type('echo');
+    cy.get('#editor > .ace_scroller > .ace_content').should('contain.text', 'echo');
+    cy.get('#runit').click();
+
+    cy.get('#errorbox').should('be.visible');
+    // The error should be about the lonely echo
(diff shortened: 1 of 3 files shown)


Felienne:
Haha I just noticed this while working on something and I thought *I* had broken it :) Thanks for the fix!

jpelay:
> Haha I just noticed this while working on something and I thought _I_ had broken it :) Thanks for the fix!

hahaha No worries, I'm trying to get the tests to work and then it'll be ready!

jpelay:
The test were failing in the command line because of timing. I looked up for a while, but couldn't get it to work without the wait. I left a TODO comment to tackle this in the future.

tests/cypress/e2e/hedy_page/editor_box.cy.js
```diff
@@ -26,13 +26,15 @@ describe('Test editor box functionality', () => {
     cy.visit(`${Cypress.env('hedy_page')}#default`);
     // click on textaread to get focus
     cy.get('#editor > .ace_scroller > .ace_content').click();
-    // empty textarea
+    // We wait until the editor is focused
+    // TODO: replace this wait. The editor takes a while to be focused
+    cy.wait(1000);
```
Felienne, line 31:
```suggestion
    cy.wait(2500);
```

Felienne:
Let's see if a higher time works!

jpelay:
> Let's see if a higher time works!

Yes let's see! Hopefully it will! For me it also worked without waits in the command line using the option `--browser chrome` but I'm not sure if that's feasible on Github Actions

Felienne:
> > Let's see if a higher time works!
> 
> Yes let's see! Hopefully it will! For me it also worked without waits in the command line using the option `--browser chrome` but I'm not sure if that's feasible on Github Actions

I don't know either, so I am not sure how we can move this forward at the moment

jpelay:
> I don't know either, so I am not sure how we can move this forward at the moment

According to Cypress, yes we can, but since I'm working in the debugger at the moment, I'll tackle this on Monday! Do I set this as a draft in the meantime?

jpelay:
The test that was failing was a flaky cypress test that I already opened an issue for in #4455.

Felienne (approved):
Yes, works again and now with passing tests!

image

Will also deploy this now cause it is a bit annoying that is is broken.

jpelay:
Sigh, these test failed in https://github.com/hedyorg/hedy/actions/runs/6158212111 :c in #4471

Felienne:
> Sigh, these test failed in https://github.com/hedyorg/hedy/actions/runs/6158212111 :c in #4471

Let's not worry about it all too much, we can discuss tonight!

PR #4479, opened 2023-09-13
[UI] Adds CodeMirror as editor

### LINKED ISSUE #4461: [UI idea] Integrate CodeMirror in Hedy
In order to solve #3823 we need to replace our current editor, which is Ace, for CodeMirror. At least for languages that don't use the Latin Alphabet.

This first issue pertains only the inclusion of the most basic capabilities of CodeMirror into Hedy. Furthermore we'd want to include syntax highlighting, markers, etc.

jpelay:
**Description**

Adds CodeMirror as editor in Hedy. For now it replaces Ace everywhere, but we're discussing only doing it for languages where Ace is not usable. The functions with `//pass` in them are temporary, as I work through them, and the same goes for debug `console.log`s. The style that I'm using is the default Dark Style provided from Code Mirror, but this can be leter edited to mimic Monokai, which is the one we use for Ace.

The editor, as of right now, can be used with the big caveat that it doesn't have syntax highlighting.

Things that I still need to do:
- [x] Properly configure the editor to hande right-to-left languages.
- [x] Add markers to the editor, crucial for highlighting the line that has an error after compiling.
- [x] Integrate the debugger with CodeMirror, for this the previous task needs to be done.
- [x] Investigate about adding syntax highlighting, and how long can it take.

**Fixes #4461**

Tip, if you use the word `fixes` before the issue number in this description, the related issue will automatically close then the PR is merged!

**How to test**

Cypress tests will invariably fail, because the tests for the editor that we've written use Ace native classes and, as such, are really coupled with Ace. I need to work on this once the work with the editor is completed.

As of right now, you can test it by going to the editor and testing its different options!


jpelay:
This is how the editor looks right now!
image

jpelay:
I think is also worth it that I mention here some of the key differences between Ace and CodeMirror. The later uses a functional mindset in its configuration and the way it handles updates in the state of the editor. In Ace you have an instance of `AceAjax.Editor` and this object has different functions that manipulates it's state. For example, if I want to change the contents of the editor in Ace I can just do this:

```typescript
this._editor.setValue(content, MOVE_CURSOR_TO_END);
```

Whereas in CodeMirror you have to make a transaction that will hold the value of the new contents for the editor and then dispatch that transtaction to actually mute the state. That looks like this:

```typescript
let transaction = this.view.state.update({ changes: { from: 0, to: this.view.state.doc.length, insert: content } });
this.view.dispatch(transaction);
```

This philosophy extends to things like the style of the editor, or even if it can be edited or not. For example, CodeMirror doesn't take the styles or contents of its HTML parent, so the styles applied to it don't trickle down to the editor, therefore we need to change the pixel value for the style of the theme applied to the editor, and then dispatch that change as a reconfiguration, that look like this:

```typescript
this.themeStyles['&'].height = `${newHeight}px`;
this.view.dispatch({
    effects: this.theme.reconfigure(EditorView.theme(this.themeStyles))
});
```

This makes the editor a little bit more difficult to work with, but has the benefits of being more configurable.

jpelay:
Relevant thing that I just found out about CodeMirror: the Tab key doesn't add indentation. They explain their reasons not to do so here: https://codemirror.net/examples/tab/, but basically it's because of inclusion, due to some users navigating the site without a pointing device, and recurring to tab in order to move between items. 

We can change this behavior, but is an important consideration whether we want to or not, what do you guys think?

jpelay:
Hi @Felienne!

This is my plan to move this PR forward. For us to be able to start working on the markers (these are the decorations in the editor such as highlighting errors, or a debugger line) in this PR, #4508 needs to be merged. Therefore in this PR I will work focus now on including a button to switch between CodeMirror and Ace when the user has set the Arab language. 

Since I think that will be relatively easy and I want to keep working on this, I will work on the markers in a branch on my fork: https://github.com/jpelay/hedy/tree/codemiror-markers, based on the improvements to the interface made in  #4508. This way I will be able to learn how markers work in CodeMirror without cluttering this PR with unmerged commits. 

What do you think about this plan? Is it ok?

Felienne:
> Hi @Felienne!
> 
> This is my plan to move this PR forward. For us to be able to start working on the markers (these are the decorations in the editor such as highlighting errors, or a debugger line) in this PR, #4508 needs to be merged. Therefore in this PR I will work focus now on including a button to switch between CodeMirror and Ace when the user has set the Arab language.
> 
> Since I think that will be relatively easy and I want to keep working on this, I will work on the markers in a branch on my fork: https://github.com/jpelay/hedy/tree/codemiror-markers, based on the improvements to the interface made in #4508. This way I will be able to learn how markers work in CodeMirror without cluttering this PR with unmerged commits.
> 
> What do you think about this plan? Is it ok?

Sounds good! But I also just approved #4508, so if you are ready, you can continue there.

jpelay:
Advancements with error hihglighting! It does not look pretty with CodeMirror actual theme, so I'll change that :3

https://github.com/hedyorg/hedy/assets/45865185/49ea368b-a0bc-4d76-86f2-f9a5fc728b9f

jpelay:
There's a bug right now where sometimes we know the line of the error, but we don't color it, because I think our logic to highlight the word must be flawed. Ace probably fails silently and just does not highlight the error, but in CodeMirror it errors out:
For example in this code:

image

The error is in line 3, position 9 (the print), but the 9 refers to the last character of the print. I replicated the logic that we use in Ace in CodeMirror and it tells me this:

image

As you can see in Ace this error is not being highlighted:

image


I'm going to change make it work here in CodeMirror, and I have the feeling that fix will also work for Ace, and will change that in a separate PR. At least if we can't find color the appropriate word, we should highlight the entire line.

jpelay:
I found something today that is quite the conundrum!

There's a certain feature in CodeMirror that requires us to have ES6 or more. This feature is the following: in order to show the :sleeping: emoji in the gutter we need to instantiate an abstract class like this:

```
const deactivateGutterMarker = new class extends GutterMarker {
    toDOM() { return document.createTextNode("😴") }
}
```
This is exactly taken from CodeMirror's documentation.  

When we compile this and then try to access the page it gives us this error:

```
Uncaught TypeError: Class constructor Gi cannot be invoked without 'new'
    at new e (cm-decorations.ts:133:36)
    at cm-decorations.ts:133:32
    at appbundle.js:1:778
    at cm-editor.ts:8:1
    at appbundle.js:1:778
    at app.ts:19:1
    at appbundle.js:1:778
    at htmx-integration.ts:4:1
    at appbundle.js:1:778
    at index.ts:14:1
```

After digging a bit, I found that we needed to increase our compilation target to ES6, but if we do this, Ace is the one that fails (apart from something in the statistics module, but I was able to easily solve that one).

I'm looking into solving this issue before moving forward with this project. Luckily this PR is almost finished!

jpelay:
Hi, @Felienne!  The decorations are done!! It means that the only thing that isn't working on the main editor is the syntax highlighting, but everything else is done!

I think we know might want to discuss how we want to move this project forwards: whether to wait until the syntax highlighting is done to ship this change, or maybe add a button for languages that have problems with the current editor.

Felienne:
I can't believe we are almost ready to ship this....!!

jpelay:
Hi, @Felienne! I think that the vast majority of the necessary work to make CodeMirror work in Hedy is done!

What have I done:

- The editor works for every level, syntax highlighting included.
- The syntax highlighter works for every language
- The debugger, skip line feature, and other markers works as well.
- I have included a script to generate the parsers from the grammar files.
- The cypress tests have been adapted to work with CodeMirror.

What is yet to be done?:
- Generate highlighting tests for the grammar. This in essence will consists of adding tests for the Lezer parsers. We can't use the highlighter tests that we already have, because the way CodeMirror highlights is fundamentally different than Ace, so it's not possible.
- Replace the read only editors from Ace to CodeMirror, and maybe remove Ace in its entirety?
- A grammar merger like the one we use for Lark.

build-tools/heroku/generate-lezer-parsers
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eu
+
+# Generate the lezer parsers from the grammars specified in  highlighting/lezer-grammars/
+for level in $(seq 1 18);
```
rix0rrr, line 5: Can't we do this via pattern matching? That way, when we add levels, we don't need to remember to update this number here.

Maybe something like:

```bash
for grammarFile in ../../highlighting/lezer-grammers/level*.grammar; do
  baseName=$(basename "$grammarFile")
  tsFile=${baseName%.grammar}-parser.ts
  npx lezer-generator --typeScript "$grammarFile" -o ../../static/js/lezer-parser/$tsFile
done
```

static/js/app.ts
```diff
@@ -272,12 +276,13 @@ export function initializeCodePage(options: InitializeCodePageOptions) {
...
 
   editor.on('change', () => {
-    theLocalSaveWarning.setProgramLength(theGlobalEditor.contents.split('\n').length);
-    theGlobalEditor.clearIncorrectLines();
+    theLocalSaveWarning.setProgramLength(theGlobalEditor.contents.split('\n').length);    
   });
 
   // If prompt is shown and user enters text in the editor, hide the prompt.
   editor.on('change', function() {
+    // TODO: When the user clicks back in the editor when an ask is open
```
rix0rrr, line 284: Iiinteresting. I thought we used to have the behavior that the ask would be closed if they did that?

static/js/app.ts
```diff
@@ -286,10 +291,29 @@ function attachMainEditorEvents(editor: HedyEditor) {
...
+        // Hide error, warning or okbox
+        error.hide();
+      } else {
+        // Show error for this line
+        let mapError = theGlobalSourcemap[index];
+        error.hide();
+        error.show(ClientMessages['Transpile_error'], mapError.error);
+      }
+    }
+  })
```
rix0rrr, line 315: Couldn't this whole block be a feature of `HedyCodeMirrorEditor` ?

```ts
new HedyCodeMirrorEditor({
  // Given an error manager, the editor can handle it?
  errorManager: error,
});
```

Or is the problem you need access to the source map?

static/js/app.ts
```diff
@@ -389,7 +413,7 @@ export function initializeHighlightedCodeBlocks(where: Element) {
       // Otherwise, the teacher manual Frequent Mistakes page is SUPER SLOW to load.
       onElementBecomesVisible(preview, () => {
         // Create this example editor
-        const exampleEditor = editorCreator.initializeReadOnlyEditor(preview, dir);
+        const exampleEditor = aceEditorCreator.initializeReadOnlyEditor(preview, dir);
```
rix0rrr, line 416: Almost got rid of this :)

static/js/app.ts
```diff
@@ -1590,11 +1617,12 @@ export function toggle_developers_mode(enforced: boolean) {
     $('#editor-area').removeClass('mt-5');
     $('#code_editor').css('height', 36 + "em");
     $('#code_output').css('height', 36 + "em");
-    theGlobalEditor.resize();
+    theGlobalEditor.resize(576);
```
rix0rrr, line 1620: Fixed pixel sizes are a bit ewww. Is this necessary?

static/js/cm-decorations.ts
rix0rrr, line 1: This whole file looks quite mind-bending 😲 nice job on figuring this out!

static/js/debugging.ts
```diff
@@ -344,10 +349,12 @@ export function incrementDebugLine() {
         const repeatMatches = repeatRe.exec(line);
         const elseMatches = elseRe.exec(line);
         if (ifMatches || repeatMatches || elseMatches) {
-          theGlobalEditor.setDebuggerCurrentLine(map.hedy_range.from_line - 1, 
-            map.hedy_range.from_column - 1, map.hedy_range.to_column - 1);
+          console.log('inside if repeat else')
+          console.log(map)
```
rix0rrr, line 353: Oops? 😝

static/js/debugging.ts
```diff
@@ -344,10 +349,12 @@ export function incrementDebugLine() {
...
+          theGlobalEditor.setDebuggerCurrentLine(map.hedy_range.from_line, 
+            map.hedy_range.from_column, map.hedy_range.to_column - 1);
         } else {
-          theGlobalEditor.setDebuggerCurrentLine(map.hedy_range.from_line - 1);
+          theGlobalEditor.setDebuggerCurrentLine(map.hedy_range.from_line);
```
rix0rrr, line 357: Because line numbers are now 1-indexed across both editors?

static/js/lezer-parsers/language-packages.ts
```diff
@@ -0,0 +1,241 @@
...
+import { parser as level15Parser } from './level15-parser'
+import { parser as level16Parser } from './level16-parser'
+import { parser as level17Parser } from './level17-parser'
+import { parser as level18Parser} from './level18-parser'
+import { tags as t } from "@lezer/highlight";
+import { LRParser } from '@lezer/lr';
+
+interface languageSupport {
+    parser: LRParser,
+    styleTags: Record<string, any>
```
rix0rrr, line 23: Should the any be some sort of `Token` type?

static/js/lezer-parsers/language-packages.ts
```diff
@@ -0,0 +1,241 @@
...
+
+export let languagePerLevel: Record<number, languageSupport> ={
+    1: {
+        parser: level1Parser,
+        styleTags: {
+            "print ask echo forward turn color": t.keyword,
+            "Text": t.name,
+            Comment: t.lineComment,
+            "Command/ErrorInvalid/Text": t.invalid,
+        }
```
rix0rrr, line 34: There's a bunch of repetition in these `styleTags`. I assume that the values have to be the same in every level?

Is it required that we only define the keys that actually exist in every level, or is it possible to just define a full map of everything?

If it is really the case that we must only supply the keys that are actually used, then we could use some additional information to filter down from the big map?

Something like this (but not exactly this):

```ts
const ALL_STYLE_TAGS = {
          "print forward turn color ask is echo sleep Op at random remove from add to Comma": t.keyword,   
          Comment: t.lineComment,
          "Text": t.name,
          "Command/ErrorInvalid/Text": t.invalid,
      };


function specificStyleTags(...tagNames: string[]): Record<string, any> {
  // flatMap allows us to map and filter at the same time
  return Object.fromEntries(Object.entries(ALL_STYLE_TAGS).flatMap((namesString, token) => {
    const names = namesString.split(/\s+/).filter(n => tagNames.includes(n));
    if (names.length === 0) {
      // Nothing left
      return [];
    }
    // the 'as const' here is because I'm not sure TypeScript will typecheck this correctly without it
    return [[names, token] as const];
  });
}
```

static/js/modal.ts
```diff
@@ -200,56 +200,71 @@ export const success = {
...
 
   show(caption: string) {
     $('#okbox .caption').text(caption);
     $('#okbox').show();
-    editor?.resize();
-    setTimeout(function(){     $('#okbox').hide();
-    editor?.resize(); }, 3000);
+    const height = computeEditorHeight('okbox');
+    console.log(height); 
+    console.log(editor)
```
rix0rrr, line 221: Oops!

rix0rrr:
By the way, all of my comments are suggestions! Apart from the `console.log`s you probably want to remove, no need to block on any of them. They could all be follow-ups.

build-tools/heroku/generate-lezer-parsers
jpelay: Great idea!!! I'm very new to bash scripting, so this didn't occur to me.

static/js/app.ts
jpelay: It does closes, but the editor loses focus again after typing a something. I opened #4742 to remind me of this issue!

static/js/app.ts
jpelay: Nice, idea!

static/js/app.ts
jpelay: Yes, don't love them. But resizing this editor is quite difficult, it doesn't quite interact with the parent in the same way that Ace does. To be able to get rid of this, I think in the future we can implement the errors as CodeMirror widgets.

static/js/cm-decorations.ts
jpelay: Thanks! Luckily they have some great documentation at CodeMirror.

static/js/lezer-parsers/language-packages.ts
jpelay: True!

static/js/debugging.ts
jpelay: Indeed!

static/js/lezer-parsers/language-packages.ts
jpelay: You were right that you can supply keys that aren't actually used. Took advantage of that and simplified this a lot!

tests/cypress/e2e/lezer-tests/level_1.cy.js
jpelay, line 1: What do you think about testing the highlighter like this @Felienne?  I think in the future I might make a multilevel tester, like the one we have for the Lark gramars. 

Also, do you think it's a priority to test the highlighter for other languages, like we do with Lark as well?

jpelay: This questions can of course be debated with everyone on our meeting :smile:

Felienne: This seems similar to what we had but more readable, so great!

And yeah I can imagine we want multi-level testers just to save you work?

Felienne:
> I think is also worth it that I mention here some of the key differences between Ace and CodeMirror. The later uses a functional mindset in its configuration and the way it handles updates in the state of the editor. In Ace you have an instance of `AceAjax.Editor` and this object has different functions that manipulates it's state. For example, if I want to change the contents of the editor in Ace I can just do this:
> 
> ```ts
> this._editor.setValue(content, MOVE_CURSOR_TO_END);
> ```
> 
> Whereas in CodeMirror you have to make a transaction that will hold the value of the new contents for the editor and then dispatch that transtaction to actually mute the state. That looks like this:
> 
> ```ts
> let transaction = this.view.state.update({ changes: { from: 0, to: this.view.state.doc.length, insert: content } });
> this.view.dispatch(transaction);
> ```
> 
> This philosophy extends to things like the style of the editor, or even if it can be edited or not. For example, CodeMirror doesn't take the styles or contents of its HTML parent, so the styles applied to it don't trickle down to the editor, therefore we need to change the pixel value for the style of the theme applied to the editor, and then dispatch that change as a reconfiguration, that look like this:
> 
> ```ts
> this.themeStyles['&'].height = `${newHeight}px`;
> this.view.dispatch({
>     effects: this.theme.reconfigure(EditorView.theme(this.themeStyles))
> });
> ```
> 
> This makes the editor a little bit more difficult to work with, but has the benefits of being more configurable.

Just bumping this here so we remember to add something like this to the docs?

Felienne (approved):
LET's GOOOOOOOOOO! We can add the rest of the tests later, as discussed today in the meeting

jpelay:
> LET's GOOOOOOOOOO! We can add the rest of the tests later, as discussed today in the meeting

AAAAAAAAA I'm free hahaha

PR #4456, opened 2023-09-05
[UI] Add public profile link to account dropdown

Ilia-Gershen:
**Description**

Removes open profile link button in profile settings and adds a new button in profile dropdown menu which links to open profile

**Fixes #4445**

**How to test**

* Login as teacher1 or any other teacher account
* Click on the account in the top right side of the site and see the dropdown
* A new menu item should be located under the account setting and before the exit button
* In account setting the green button should disappear 

**Checklist**
Done? Check if you have it all in place using this list: (mark with x if done)

- [ x] Contains one of the PR categories in the name
- [ x] Describes changes in the format above
- [ x] Links to an existing issue or discussion
- [ x] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #4492, opened 2023-09-14
[UI] Adds public profile link to account dropdown

### LINKED ISSUE #4445: [UI idea] Add public profile to account dropdown
If users choose to do so, they have a public profile with info and programs, for example: http://localhost:8080/user/teacher1

It would be nice to have the link to the public profile in the dropdown:

image

We could then also remove the corresponding button from the account page, because I find a button for a link a bit weird:

image

jpelay:
**Description**

This PR is based on the work of @Ilia-Gershen on #4456. I couldn't commit changes to that branch, so I had to make a new PR.

Apart from the changes made by Ilia, I also added `has_public_profile` as a variable available in all templates, therefore we can access this variable when deciding whether to show or not the My public profile, button.

**Fixes #4445**

**How to test**

If you're a user with a public profile, the option should be shown in the menu!


translations/bg/LC_MESSAGES/messages.po
```diff
@@ -910,13 +909,13 @@ msgstr "Highest quiz score"
...
 msgstr ""
-"On this page you can look through programs created by other Hedy users. You "
-"can filter on both a Hedy level and adventure. Click on \"View program\" to "
-"open a program and run it. Programs with a red header contain a mistake. You "
-"can still open the program, but running it will result in an error. You can "
-"of course try to fix it! If the creator has a public profile you can click "
-"their username to visit their profile. There you will find all their shared "
-"programs and much more!"
+"On this page you can look through programs created by other Hedy users. "
+"You can filter on both a Hedy level and adventure. Click on \"View "
```
Felienne, line 913: Hi @jpelay!

Do we understand why this PR rewrites the layout of all po files (it might mess up Weblate?)

jpelay: Oh I didn't realize that when submitting the PR. I just copy-pasted the commands listed in the wiki, so I'm not sure why this happened. I'll try this again and fix it!

jpelay: I've been looking into this and it has to do with the width of the output lines. For some reason maybe this has changed? According to their docs the maximum output is 76, but if I pass that width in this branch, everything still changes. Maybe we can add a `--no-wrap` option or set the maximum amount of `width` ourselves.

jpelay: Ok! I'll do that tomorrow!

jpelay: So I have the answer as to why this keeps happening: the way Weblate and gettext calculate the length of the string is fundamentally different, one uses byte lenght and the other character length, there is an [issue for that](https://github.com/WeblateOrg/weblate/issues/6350) on the Weblate repository that is marked as wontfix, so no matter what width we choose, there will always be a difference in the way the lines are wrapped. 

The way I see we can fix this is by installing a plug-in to Weblate and set the option to not wrap lines ([apparently someone had problems](https://github.com/WeblateOrg/weblate/issues/6350) with that too, but I'm not sure if they messed up something or if its actually an issue, still they marked it as wontfix), but as it concerns to this particular PR I don't think there's a way to prevent the changes done to the line spacing, at least in the state that main is right now, perhaps after applying the plug-in and if the lines aren't wrapped I can apply the same option here and there won't be all of those differences.

What would like to do? Can I install the plug-in to Weblate or are there other considerations?

Felienne: Hi @jpelay! Great find about the plugin! I don't think this is a pressing issue, let's rather focus on the other open plans for now, but I will make an issue so we can pick it up in the future.

Felienne (approved):

PR #4504, opened 2023-09-16
[FIX] Improve memory usage by removing parser caching

rix0rrr:
Add facilities for memory leak debugging. Run the server with:

```
$ env PYTHONTRACEMALLOC=10 python3 app.py
```

To show statistics about leaked memory when the server is stopped with Ctrl-C.

After some analysis, it seems like the Lark parser cache is holding on to more memory than is necessary. Remove the cache, taking a hit of some CPU work to build the parser every time, in order to reduce memory usage.

**How to test**

This change is a bit speculative: after deploying, we should keep an eye on the server metrics to make sure response times don't blow up.


Felienne (approved):
Let's deploy and keep an eye on the stats

PR #4506, opened 2023-09-18
Adds translation files for public profile menu element

jpelay:
**Description**

In #4492 I forgot to mention that I reversed the changes to the translation files, and that I needed to include them before merging! This PR adds those files.

**How to test**

In English the menu should look like this:

image


Felienne:
Ah sorry I could have seen that (I am usually too lazy to run babel and check keys).

messages.pot
```diff
@@ -287,9 +290,18 @@ msgstr ""
 msgid "comma"
 msgstr ""
 
+msgid "command_not_available_yet_exception"
```
Felienne, line 293: I don't really understand why the exceptions also hop around in this PR, did we move them earlier?

jpelay: These exceptions are for the live statistics module! They represent the translations for the exception groups that we discussed a while back for that PR. Maybe it was an error from my part, but `messages.pot` isn't checked in in that PR (#4166) and maybe that is what caused the mess. Here's where I edited them: https://github.com/hedyorg/hedy/pull/4166/files#diff-7d430e5ce7e9bf543efa95783df17746ec4ddcbeb38599d013266447bf075b2c

Felienne: Ah yes, got it!

Felienne:
Oops, I made it worse with the merge! I will take a look later @jpelay, don't worry about it!

Felienne (approved):

PR #4520, opened 2023-09-21
Lost content level 12

MarleenGilsing:
Fixes issue #4491


Felienne:
Ha @MarleenGilsing!

Ah sorry, I intentionally removed this because I found the intro of level 12 to be very long and thus quite overwhelming. But that was before the split keyword so maybe it is now better? 

Or, maybe we can put this in a separate tab with an adventure? Or in a keyword tab for `+`?

Felienne (approved):
Ah I see this is added to the maths tab! Sounds perfect like this!

PR #4522, opened 2023-09-21
[CONTENT] add /start route

Felienne:


Felienne:
Merging this (partial) work for now, so we can merge in the yaml fix script. Will make a new draft PR to work on lateron!

PR #4525, opened 2023-09-21
[FIX] Fix error in Live Statistics

jpelay:
*Description**

Fixes the bug that caused the Live Statistics page to crash. This happened due to not filtering out teacher adventures that were deleted. This PR fixes that by filtering them, and also removes those adventures from the customizations table when the teacher accesses the Customization Class page.

**How to test**

1. Create a teacher adventure
2. Add it to a class
3. Delete the adventure
4. Access Live Statistics

It shouldn't crash! If you do the same in Alpha it will


Felienne (approved):
Yes, indeed works! Let's see if it also fixes the issue I ran into in my own class (but I think it is a performance problem)

jpelay:
> Yes, indeed works! Let's see if it also fixes the issue I ran into in my own class (but I think it is a performance problem)

Probably not, but we'll work on that issue as well!!

PR #4526, opened 2023-09-21
[FIX] Fixes Skulpt extensions not working on dev and Alpha

### LINKED ISSUE #4524: [BUG] Clear not working on Alpha
**Describe the bug**

image

I must have messed up something when working in the Ts, I'll handle this!

jpelay:
**Description**

Changes the static root from `'/'` to `'' ` when we are in a dev enviroment. The problem was that when we didn't point to a CDN and the contents were delivered from the server the paths looked like this: `//vendor/`  this caused Skulpt not to find those extensions and fail. 

**Fixes #4524**

**How to test**

Try the `clear` or pygame comands.


rix0rrr (approved):

rix0rrr:
Good catch, thanks!

PR #4527, opened 2023-09-21
[CONTENT, FIX] stop fetching intro from manual

### LINKED ISSUE #4523: [CONTENT] What's the use of 'key' in the teacher manual?
In the teacher manual is the header 'key'used quite often I don;t know what it does. But if it's just a bookmark for us, we should tell the translators not to translate it, so we could easily navigate in documents that are in a foreign language. 

image

Felienne:
In a previous translation pr, keys were translated into Turkish, this breaks things:

image


Felienne:
Leaving this as draft PR until someone (maybe me :) has time to make bullets 2 and 3 into issues so we don't forget.

Felienne:
Update: it is only a problem if it's the intro key, since that is the one we grab:

https://github.com/hedyorg/hedy/blob/92c82e9cadca2ed3885aeada82b55d4706e2a787/website/for_teachers.py#L80

Now that I grasp this, it is probably better to change the code and not rely on the name of the first section! But for now let's do this so that the manuals aren't broken anymore.

rix0rrr (approved):

rix0rrr (approved):

PR #4529, opened 2023-09-21
[FIX] Fix cypress test race conditions

rix0rrr:
This test, which tests the Ace editor, works reliably on our workstations, and fails reliably on GitHub Actions. I still don't know how that's possible, but I have found a way to rewrite the test that seems to succeed.

The following all doesn't work:

```ts
cy.<whateverElement>.clear();
cy.wait(<howeverlong>);
cy.<whateverElement>.type('{moveToStart}{del}{del}{del}{del}'...);
```

The following seems to work:

```ts
cy.<whateverElement>.type('{moveToEnd}{backspace}{backspace}{backspace}{backspace}'...);
```

I can only guess that `type()` special symbols like `{moveToStart}` and `{selectAll}` don't work, for whatever reason (`.clear()` is an alias for `type('{selectAll}{del}')`). BUT, it seems to be the case that our cursor gets dropped at the end of the text, and we can erase the text by sending in a whole lot of `{backspace}`s.

**How to test**

The GitHub actions should not fail :)


tests/cypress/e2e/hedy_page/editor_box.cy.js
```diff
@@ -90,4 +94,14 @@ describe('Test editor box functionality', () => {
...
+});
+
+/**
+ * Wait for the Ace editor to finish initializing all the HTML elements and attach the listeners
+ * 
+ * If we start typing in the editor immediately, it may happen before the editor has
+ * loaded. Waiting for HTML elements that are created as the editor loads will fix that.
+ */
+function waitForAceToLoad() {
+  cy.get('#editor > .ace_scroller > .ace_content').should('be.visible');
```
jpelay, line 106: Yes, I tried this and it didn't work. You can try locally by running the spec in the command line! It works in the UI but not there! Can you try with `--browser chrome` it worked for me locally too.

Felienne:
Still failing 😭

tests/cypress/e2e/hedy_page/editor_box.cy.js
rix0rrr: I've been able to repro the failure once on my machine, and then it never failed again 😢

rix0rrr: It's interesting though that the `Language XX should run` tests never seem to fail, though they do something quite similar.

rix0rrr:
Looking at this, it actually looks like on my machine I was hitting a different race condition than the GitHub runners are hitting 🤣

tests/cypress/e2e/hedy_page/editor_box.cy.js
jpelay: > It's interesting though that the `Language XX should run` tests never seem to fail, though they do something quite similar.

I know right? That's why I was baffled that these tests where failing so much.

rix0rrr:
💡 It looks like every *found* text starts with `print hello world`, which I think might be the starting code that we start out with. It looks like the `clear()` command doesn't come through! Or maybe the `.clear()` command goes through too early, before we load the `starting_code` into the text area!

rix0rrr:
💡 Another insight. We are all assuming that the `Language XX should run` tests succeed and the other tests fail, HOWEVER they are not testing the same thing!

```ts
    it(`Language ${language} should run`, () => {
      cy.get('#editor').type('print Hello world');
      cy.get('#editor > .ace_scroller > .ace_content').should('contain.text', 'print Hello world');
    });

  it('Ask modal should hold input and the answer should be shown in output', () => {
    cy.get('#editor').type('print Hello world\nask Hello!\necho');
    cy.get('#editor > .ace_scroller > .ace_content').should('have.text', 'print Hello worldask Hello!echo');
  });
```

The difference is `should('contain.text')` vs `should('have.text')`! 

Probably, the actual text in the first test is also `print hello worldprint Hello world`, but since we are only checking for a substring, the test still passes!

(So also, we cannot rely on the code there as an example!)

rix0rrr:
I tried:

```ts
cy.focused().type('{moveToStart}{del}{del}{del}{del}{del}{del}...')
```

And that didn't work, but instead:

```ts
cy.focused().type('{moveToEnd}{backspace}{backspace}{backspace}{backspace}...')
```

DOES work.

Probably `{moveToStart}` and `{moveToEnd}` don't actually do anything, but it just so happens the cursor starts at the end of the text and backspacing it all away is a viable way to clear the input.

Felienne:
@jpelay will you take a look and approve this? It looks good to me but maybe you want to read and understand the whole solution?

jpelay:
> @jpelay will you take a look and approve this? It looks good to me but maybe you want to read and understand the whole solution?

Hi! I did check it, was just waiting for the tests to pass again before approving 😄  

@rix0rrr this was genius detective work, thank you so much!!!!

jpelay (approved):

PR #4532, opened 2023-09-22
[FIX] Fix levels in customizations not being set properly

jpelay:
**Description**

When storing the customizations settings we where using the same variable for both the levels in the live statistics module, and for the class customizations. Changing that to a use a new variable name fixes the problem.

**How to test**

1. Go to class customizations.
2. Select a few levels and unlock them.
3. Recharge the page

The levels shown should be the ones you selected.

### DIFF
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -610,7 +610,8 @@ class ForTeachersModule(WebsiteModule):
 
         customizations = self.db.get_class_customizations(class_id)
         dashboard = customizations.get('dashboard_customization', {})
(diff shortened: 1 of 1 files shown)


Felienne:
Thanks for the speedy fix @jpelay! I will make an issue to add a test for this too, because it was a pretty annoying bug for teachers.

Felienne (approved):

jpelay:
> Thanks for the speedy fix @jpelay! I will make an issue to add a test for this too, because it was a pretty annoying bug for teachers.

Great! No problem!

PR #4533, opened 2023-09-22
[FIX] Fix error highlighting in Ace

jpelay:
**Description**

I noted this when working in the CodeMirror migration: when a column is provided to the highlightError function, sometimes it fails to highlight a word and just doesn't show anything.

This happens because when we try to get the length of the word we want to highlight, we  actually get an empty string. Since I'm not 100% sure that getting the word before the column given by Skulpt when processing the error would be correct, I decided was best to highlight the whole line. 

For example, let's look at this code in level 14:

```
name = ask 'What is your name?'
print 'dfdf'
if name == 'hedy'
    prit 'You are cool!'
```

The compiler returns this error:

> The code you entered is not valid Hedy code. There is a mistake on line 4, at position 9. You typed a space, but that is not allowed.

Position 9 refers to the last character after `prit`. When we do our calculations to highlight the word, we actually look for the positions **after** the one that is given by the transpiler, splitting them by spaces so we get something like this:

```
['', ' ', "'You", ' ', 'are', ' ', "cool!'"]
```

Since we get first element of that array, which has length zero, we don't actually highlight anything.

So, my question is, is it safe that if we get an empty string like this, to highlight the last word before the column given by the compiler? In this case, it would correctly be `prit`, is this true for every error?

**How to test**

You can try the error I made here, but in general many errors after level 4 should be highlighted now.


Felienne:
Wow, nice find, and what a great PR, very clear what the issue is and why it occurs!!

Indeed this fixes the issue!

image

Felienne (approved):

jpelay:
> Wow, nice find, and what a great PR, very clear what the issue is and why it occurs!!
> 
> Indeed this fixes the issue!
> image

Thank you!!! I'm glad to fix these errors 😄

Issue #4475, opened 2023-09-13
[CHORE] Remove old stats page and logs


Felienne:
In #4474 we remove calls to a lot of backend code that we used to create stats and query logs. This should be removed because it has been replaced by the new dashboard (and the logs were actually only used ever by me, and I can also query the real DB).

How to get started with this issue:

* Python: The calls we have removed call the routes `/stats/class/` and `/logs/class/`, so you could search the Python code and find references to these. They are in statistics.py, but note that that file also contains routes related to the grid, which *is* still in use
* Typescript: I know there is also code in the front-end to render graphs, but I am not 100% sure where. Figuring that out is step 2 :)

Felienne:
Hi @jtwaleson!

This too could be a nice issue to pick up, which is also going to remove a lot of front-end code, so will help you get a better grasp on things in the js/ts world

Felienne:
> > * Typescript: I know there is also code in the front-end to render graphs, but I am not 100% sure where. Figuring that out is step 2 :)
> 
> This code lives in `statistics.ts`, which also contains code for the new grid that is in use.

Ah thanks! So we need to take a bit of care in removing the relevant stuff.

jtwaleson:
Great, I'll take a look in the next couple of days :) Thanks for the context @jpelay .

jtwaleson:
@Felienne @jpelay That was a bit of a struggle, but I think it's nearly done. Could one of you take a look and see if this is the right direction? See my branch here: https://github.com/jtwaleson/hedy/tree/4475-remove-admin-stats-preview 

I didn't include the compiled typescript and css yet, but it does compile fine and seems to run fine locally. BUT: I don't have the production statistics and locally there is not a lot of data, so I can't see with the `teacher1` account whether all the stats work.

Felienne:
> @Felienne @jpelay That was a bit of a struggle, but I think it's nearly done. Could one of you take a look and see if this is the right direction? See my branch here: https://github.com/jtwaleson/hedy/tree/4475-remove-admin-stats-preview
> 
> I didn't include the compiled typescript and css yet, but it does compile fine and seems to run fine locally. BUT: I don't have the production statistics and locally there is not a lot of data, so I can't see with the `teacher1` account whether all the stats work.

Ah cool @jtwaleson! Have you tried the local test db, it is small but it has a few runs. Can you make a draft PR out of your branch, that makes it easier to look at your code and run it locally.

jtwaleson:
[...]

PR #4534, opened 2023-09-22
[REFACTORING] use a json file for the babel caching

jtwaleson:
**Description**

Improve the developer experience of contributors by removing code generation which sometimes conflicts with linters / autoformatters.

We were generating python code for the i18n config in the frontend. Now we serialize to JSON and parse it instead, the python "API" stays the same.

The problems with generating python code were:
1) differences in the output when running `npm run build` (different formatting depending on the local system)
2) different ordering of the dicts depending on the local system
3) rewrites in the autopep8 pre-commit hook

With this commit we rewrite it to indented & ordered JSON so differences are simpler to review. Also the serialization is using 100% standard tooling instead of code generation.

**Fixes ....**

I ran into this when compiling frontend code and running `npm run build` while fixing https://github.com/hedyorg/hedy/issues/4475 .

I don't believe creating a separate issue for issues like this (which would create the exact same content as the PR) is needed, but if I'm wrong about this please let me know.

**How to test**

Check out the branch locally and check if multiple languages still display correctly

**Checklist**


jtwaleson:
@rix0rrr I saw that you made the current setup and thus are best to judge if this situation would be fine or whether it has any unintended consequences. I humbly believe it's an improvement and am happy to stand corrected :)

rix0rrr (approved):
👍 Good shout, I like it

PR #4551, opened 2023-09-25
[FIX] Make subscribe page language dropdown use right name for languages

rix0rrr:
Previously, the subscribe page used the English word for every language, and defaulted to "Afrikaans".

Make the dropdown better: don't default to the first language in the list, and use the language's own word to describe itself ("Deutsch" instead of "German", etc).

**How to test**

Visit http://localhost:8080/subscribe, open the dropdown


jpelay:
Thanks for this @rix0rrr! Works perfectly for me!

jpelay (approved):

PR #4556, opened 2023-09-26
[UI, FIX] adjusts hide and disables save warning popup in slides

### LINKED ISSUE #4513: [BUG] Slides lack buttons
Oops, we broke the slides, see: https://hedy.org/slides#/1/0/0

And observe we no longer show the run button or the ladybug-debugger button:

image

As far as I know (when I was working on them), slides use the `/raw` route:

https://hedy.org/adventure/story/1/raw#story

And that does show the buttons, albeit very low.

image

Not sure what went wrong here, I will take a look cause I am the one who built this! But we do also need a few tests on the slides in any case (maybe a separate issue for someone else, since I never made Cypress tests before)

Felienne:
Fixes #4513. Also disables the save pop up which is quite annoying in slides mode.

As per https://github.com/hedyorg/hedy/pull/4363 I should not touch the front-end without supervision, so it is probably a good idea if @jpelay checks if this indeed works, and maybe to add a test for the slides and/or for raw? :D


Felienne:
Thanks for taking a look @jpelay and making sure I don't break stuff. In the future (no rush!) it would be nice to also have cypress tests for slides.

jpelay:
> Thanks for taking a look @jpelay and making sure I don't break stuff. In the future (no rush!) it would be nice to also have cypress tests for slides.

Hi, yes I was taking a look before the meeting, bit haven't fully tried it yet! Gonna review it between today/tomorrow and also maybe add some tests right away!

rix0rrr (approved):

PR #4573, opened 2023-10-01
[CHORE] run snippet generation on changes

### LINKED ISSUE #4546: [CHORE] Run hash generation on each PR?
We now periodically (daily) run a GitHub action to generate hashes for all snippets (See #3757 and #3873). I am not sure the why is documented somewhere so let me do it here briefly:

The repo contains a lot of code snippets, in /adventures and in /parsons for example, and we want to make sure those snippets don't have errors in them. Therefore we test them against the grammar files. But running all snippets tests takes a very long time (1.5 hours!) and that is annoying, esp for PRs that have nothing to do with the grammars.

So, we run the tests, and if we have seen a code snippet with this version of the Hedy language, we store the hash, so we know to skip it later on. That shrinks the time for tests from 90 minutes to about 10. Every night we regenerate all hashes so we remain up to date.
 
However, there are now 2 issues:
* The commits clog the history a bit
* (more important) if we merge a grammar PR, for the whole day the tests are slow until the evening (this happened yesterday for example)

So, we could also run the generation on each PR, but only do it when a grammar change was made (we used to have code for this, see https://github.com/hedyorg/hedy/pull/3895/files where it was removed). Doing it on each PR irrespective of changes to files would make each PR really slow. 

I once started work on this: https://github.com/hedyorg/hedy/pull/4078 but then I gave up because I was not sure enough about the Actions semantics, but it would add value.

jtwaleson:
**Description**

Before this change we ran the snippet generation daily. This resulted in commits even when there was no activity. As the snippets file is a bit big and has binary contents, the history of the repository grew a lot.

Now we switch to only updating the hash generation when one of the files that invalidates the entire cache is updated.

We do not do this on PRs as we can not/ do not want to push commits to forked repositories.

After we merge a PR to main, and it had one of the files affecting snippet generating updated, we make a separate commit that updates the snippet hashes file.

Fixes #4546 .

**How to test**

This is a bit sub-optimal, but you can only test this change with quite some effort (creating a fork etc). However, I did some of these tests already and I think the pragmatic way to test is: read the workflow change critically, see if we agree with the proposed flow, merge it and fix any problem as we go along.

After merging we should check this:
- merge a PR that does NOT affect hedy.py or the grammars -> observe that the action does not run
- also merge a PR that does affect hedy.py or the grammars -> observe that a commit is added after all the tests are executed

**Checklist**

### DIFF
--- a/tests/Tester.py
+++ b/tests/Tester.py
@@ -66,6 +66,7 @@ class HedyTester(unittest.TestCase):
         ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
         directory = os.path.join(ROOT_DIR, 'grammars')
 
+        # should be kept in sync with .github/workflows/generate-snippet-hashes.yml
         files_affecting_parsing = (
             [os.path.join(directory, filename) for filename in os.listdir(directory)] +
             [os.path.join(ROOT_DIR, 'hedy.py')]
--- a/.github/workflows/generate-snippet-hashes.yml
+++ b/.github/workflows/generate-snippet-hashes.yml
@@ -1,37 +1,42 @@
+# Every PR that contains a change to the hash we use in tests/Tester.py
+# should update the hashes file to speed up the unit test CI run.
+
+# We do not do this on PRs as we can not/ do not want to push commits to
+# forked repositories.
+
+# After we merge a PR to main, and it had one of the files affecting
+# snippet generating updated, we make a separate commit that updates the
+# snippet hashes file.
+
 env:
   save_snippet_hashes: True
 
-name: Periodically generate hashes
+name: Generate hashes after every push to main that affects hedy core
 on:
-  schedule:
-    - cron: "0 3 * * *" # Run this job every day at 3am
-  workflow_dispatch: {} # Allow running the workflow on-demand
+  push:
+    branches: ["main"]
+    # these paths should be kept in sync with tests/Tester.py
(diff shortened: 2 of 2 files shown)


tests/Tester.py
```diff
@@ -66,6 +66,7 @@ def setUpClass(cls):
         ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
         directory = os.path.join(ROOT_DIR, 'grammars')
 
+        # should be kept in sync with .github/workflows/generate-snippet-hashes.yml
```
Felienne, line 69: I am not sure I love this comment, I think what you mean to say is: these files *are* kept in sync with .github/workflows/generate-snippet-hashes.yml, is that correct? Then maybe we can document the mechanism here through which that happens (in the on-build script, right?)

jtwaleson: Oh sorry for the confusion. The tests/Tester.py file compiles a list of files that affect how hedy works, to determine if a test can be skipped. Any change to any of these files will cause the entire snippet file to be invalidated.

This list is now also included in the `.github/worksflows/generate-snippet-hashes.yml` file, to detect if we should regenerate the cached file. So what I meant with the comment is that we manually have to keep the list in sync. Any time we change the "set of files that determine how hedy works" we have to update in two places.

Let me know if that makes sense and/or whether the comment needs more explanation.

Felienne (approved):

PR #4574, opened 2023-10-01
[REFACTOR] Faster CI

### LINKED ISSUE #4546: [CHORE] Run hash generation on each PR?

jtwaleson:
## Description

Fixes #4546 in a fundamentally improved way in 3 parts, plus one part with some suggestions for the future.

This speeds up the unit tests:
- non-cached run: was 1.5 hours -> now roughly 35 minutes https://github.com/jtwaleson/hedy/actions/runs/6371719946/attempts/1
- cached run: was 11 minutes -> now roughly 7 minutes https://github.com/jtwaleson/hedy/actions/runs/6371719946/attempts/2

Follow-up work to split the work out into multiple jobs is underway, reducing the total time to 7 minutes uncached and 3 minutes cached. I'll create a separate PR for that once this is merged, as it is dependent on this work, but I don't want to make the PR too big.

**Part 1) using a github cache**
   
We previously used a snippet hash pickled file to see if snippets had
been executed successfully already on this version of hedy.

This was fine but had two downsides:
- the snippet file changed frequently and was checked into git, leading
  to frequent non-functional changes
- using multiple processes to speed up the CI run was not possible
  as each process would try to overwrite the snippet hash file

Now we introduce pytest-xdist to speed up the test runs by using 2
processes concurrently. We can attempt to go to 4 but the Github runners
have 2 cores and can go out of memory if we go to more concurrency.

We also use a directory based cache instead of pickling a set. We use
concurrency-friendly OS primitives to see if a file exists (open and 
os.path.isfile). This way multiple processes can read/write the cache
at the same time.

This cache is .gitignore'd so it's not stored in the repository. It is
however stored as a github cache entry. The key uses the hedy language
definition as an input and then hashes it with sha256sum. This means that
any change to the hedy files will invalidate the cache, which is what we
want.

This removes the need for the periodically generating of the snippet
hashes, so that workflow is removed.

As the entire hedy language definition was calculated with every test
run, and this was hashed with md5. We change this a bit to use the more
modern and standard but still fast sha1. The sha1 for the entire
language is only used once to calculate a short digest for the language.
That digest is then concatenated to the snippets to form the individual
test snippets. This shaves off only a couple of milliseconds and a
couple of KB of memory usage, but it's nice.

**Part 2) using translation key caching**

With every single unit test that translated error messages this function
was hit, which would parse the yaml files from disk. A simple cache
makes this 10x as fast, shaving about 15 seconds from the CI run.

When we work on the language locally and introduce new keywords, we
might now need to restart the process, but I believe this should
already be covered by flask.

**Part 3) use functools.cache for parser cache**

Will pick this up in a future PR.

~~We used custom cache for the parsers. This was fine but we can use
out-of-the-box tooling just as well.~~

~~functools.cache was introduced in python 3.9 and is a nice one-liner.~~

~~However, after this, unit tests were failing as the cache key was
dependent on a "hidden" variable, sourcemap.skip_faulty.~~
    
~~This also makes that relationship transparent.~~

**Part 4) Added some TODOs for future improvements**

We run some tests sequentially which could be done with a `matrix` approach in github actions. This way they run parallel.

## How to test

- Thorough code review
- Observe CI test runs and see the improvement

## Checklist


jtwaleson:
Ready for review!

Felienne:
thanks for this @jtwaleson! I made a few small comments, but the most important thing I think is, how do I use this locally? Will that just work out of the box or are there steps to be taken? I would like to test the local running since I often use that!

jtwaleson:
> thanks for this @jtwaleson! I made a few small comments, but the most important thing I think is, how do I use this locally? Will that just work out of the box or are there steps to be taken? I would like to test the local running since I often use that!

It will work locally out of the box! A directory `.test-cache` will be created automatically.

tests/Tester.py
```diff
@@ -237,7 +232,9 @@ def single_level_tester(
     ):
         if level is None:  # no level set (from the multi-tester)? grap current level from class
             level = self.level
-        if not self.snippet_already_tested_with_current_hedy_version(code, level):
+        if self.snippet_already_tested_with_current_hedy_version(code, level):
```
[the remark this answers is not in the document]
jtwaleson: Ok, will change it :)

I have a slight preference for this way, so you see in one glance that we return if we hit the cache, else you have to scroll down to see what happens in the "else" case.

rix0rrr:
Going off your initial PR body:

> This cache is .gitignore'd so it's not stored in the repository. It is
however stored as a github cache entry. The key uses the hedy language
definition as an input and then hashes it with sha256sum. This means that
any change to the hedy files will invalidate the cache, which is what we
want.

Is this still true, or has this been changed in the mean time? I'm asking because making local testing fast is also a goal.

> When we work on the language locally and introduce new keywords, we
might now need to restart the process, but I believe this should
already be covered by flask.

Flask only restarts if any Python code has changed, not if the YAML has changed. If you use the `YamlFile` helper it should all work out automatically?

> Part 3) use functools.cache for parser cache

In light of #4581, I guess this is no longer necessary?

.github/workflows/unittests.yml
```diff
@@ -6,9 +6,7 @@ name: Unit tests
 
 on:
   push:
-    branches: [ main ]
   pull_request:
-    branches: [ main ]
```
rix0rrr, line 11: Is this change not going to run the unittests twice for every PR push? Once for the `push` event, once for the `pull_request` event?

tests/Tester.py
```diff
@@ -48,6 +51,28 @@ def __init__(self, source_range: SourceRange, exception_type: type(Exception)):
         self.exception_type = exception_type
 
 
+@cache
+def get_hedy_source_hash():
+    directory = os.path.join(ROOT_DIR, 'grammars')
+
+    # should be kept in sync with .github/workflows/generate-snippet-hashes.yml
```
rix0rrr, line 58: This seems a bit dangerous. Why do we have the hash in both again? 

Can we not have it in one place?

.gitignore
```diff
@@ -173,3 +173,6 @@ grammars-Total/*.lark
 
 # Prod css doesn't have to be checked in, only leads to merge conflicts
 ./static/css/generated.css
+
+# hedy test cache
+.test-cache
```
rix0rrr, line 178: I see this is not checked in after all. Does that not mean my initial checkout + running the tests is going to take a long time?

jtwaleson:
Thanks for taking the time to review @rix0rrr !

> Going off your initial PR body:
> 
> > This cache is .gitignore'd so it's not stored in the repository. It is
> > however stored as a github cache entry. The key uses the hedy language
> > definition as an input and then hashes it with sha256sum. This means that
> > any change to the hedy files will invalidate the cache, which is what we
> > want.
> 
> Is this still true, or has this been changed in the mean time? I'm asking because making local testing fast is also a goal.

Yes this is still true. We don't check the test cache into the repository, but locally it does live in the same directory, hence the .gitignore addition. If you run the test suite once, then the cache will be populated. Note that if you change any of the grammars or hedy.py it would be invalidated in any case. With this change we can run in parallel (`python -m pytest -n 8` if you have 8 CPUs), this speeds testing up a lot. I believe most local testing is much faster with this change than without it :)

> > When we work on the language locally and introduce new keywords, we
> > might now need to restart the process, but I believe this should
> > already be covered by flask.
> 
> Flask only restarts if any Python code has changed, not if the YAML has changed. If you use the `YamlFile` helper it should all work out automatically?

Good point, and a bit sloppy that I didn't look into that. I'll see if I can add this!

> > Part 3) use functools.cache for parser cache
> 
> In light of #4581, I guess this is no longer necessary?
> 

Indeed but I only figured that out after this PR ;) We can merge this first and then #4581 as it builds upon a small refactor with `skip_faulty` that was done here.

> > Part 4) Added some TODOs for future improvements
> 
> Is adding TODOs to the code in this way valuable? Could the same note not live in your head, a sticky on your monitor, or a GitHub issue? Also, not really sure that a matrix job is intended to parallelize heterogeneous tasks in this way. If you wanted to parallelize them, you could use `make` or `bash` for that. But how much time is there to be gained?

:shrug: it depends a bit on the style of the repo I guess. I saw more TODOs used in this way and I think it would make the detective work of other contributors in the future a bit easier with these mental notes. Happy to remove them though :) Maybe we can make this an entry in the contributor page on the wiki so it's clear what's expected in PRs.

About the parallelization: a lot of time is gained! Non-cached unit tests can run in 5m33s which I ran a couple of days ago: https://github.com/jtwaleson/hedy/actions/runs/6373381325/attempts/1 . The cached run makes it 3m16s. This would take between 30 minutes and 1 hour without parallellism. Github uses runners with 2 CPUs which this PR already maxes out with `python -m pytest -n 2`. If we divide the load over multiple runners like we do with matrix it's simply a lot faster.

Maybe matrix was not designed with this in mind, but if we can use to speed things up I'm not complaining ;) We're already maxing out the single runner.

Btw for the cost aspect: As runners are charged by the minute it's about the same costs too if multiple runners execute for shorter periods of time.

rix0rrr:
> About the parallelization

I mean the proposed `matrix` parallelization of the various validation tasks, not the parallelization of the unit test runner. I'm down with that change.

jtwaleson:
> > About the parallelization
> 
> I mean the proposed `matrix` parallelization of the various validation tasks, not the parallelization of the unit test runner. I'm down with that change.

Oh I read heterogeneous but my brain made it homogeneous... Good to hear :)

I also prefer to use a different job for different kind of validation steps. In the PoC I used matrix and stuffed all `other` tests in one of the matrix slots, but that was hacky and wouldn't make it into the final code.

.gitignore
jtwaleson: See my comment below https://github.com/hedyorg/hedy/pull/4574#issuecomment-1745072702 . The initial checkout + running the test will be slow-ish, but as you can run with `-n CPUs` and we are optimizing the parser caching too I think local testing will be fine. As always I'm happy to change my mind :)

tests/Tester.py
jtwaleson: This comment needs to go as we remove that workflow in this PR. Sorry about that.

rix0rrr:
> > > Part 3) use functools.cache for parser cache
> > 
> > 
> > In light of #4581, I guess this is no longer necessary?
> 
> Indeed but I only figured that out after this PR ;) We can merge this first and then #4581 as it builds upon a small refactor with `skip_faulty` that was done here.

If you wouldn't mind, I'd rather just split up the changes into different PRs. That means removing that change from this PR and fully keeping it in the other one. It's just easier to reason about that way (in my opinion)

rix0rrr:
> I believe most local testing is much faster with this change than without it :)

I believe it. I'm just saying that keeping the cache server-side, and requiring running an additional command to prime it, is a bit of a departure from our current design goals, which is (mostly): have Python, check out the repo, ready to go.

Felienne should come in here and decide what is acceptable for new user experience.

jtwaleson:
> Flask only restarts if any Python code has changed, not if the YAML has changed. If you use the `YamlFile` helper it should all work out automatically?

Wow that's a cool helper! Found a little problem with it though, as the place I was using it mutates the data structure and the YamlFile re-uses a cached version. So anywhere where we mutate the contents of the yamlfile is scary!

It would probably makes sense to `freeze` this with something like https://github.com/diegojromerolopez/gelidum to prevent this from happening.

Any pointers on how to proceed with this? For now I just cloned the object with `copy.deepcopy` so it works, but maybe an issue to `freeze` the output of YamlFile is in order.

jtwaleson:
> > > > Part 3) use functools.cache for parser cache
> > > 
> > > 
> > > In light of #4581, I guess this is no longer necessary?
> > 
> > 
> > Indeed but I only figured that out after this PR ;) We can merge this first and then #4581 as it builds upon a small refactor with `skip_faulty` that was done here.
> 
> If you wouldn't mind, I'd rather just split up the changes into different PRs. That means removing that change from this PR and fully keeping it in the other one. It's just easier to reason about that way (in my opinion)

Makes sense! I didn't know I would be able to fix it in a much better way so soon. I removed the commit and left the old cache in place.

tests/Tester.py
jtwaleson: Removed.

jtwaleson:
> > I believe most local testing is much faster with this change than without it :)
> 
> I believe it. I'm just saying that keeping the cache server-side, and requiring running an additional command to prime it, is a bit of a departure from our current design goals, which is (mostly): have Python, check out the repo, ready to go.
> 
> Felienne should come in here and decide what is acceptable for new contributor experience.

Mostly agreed. I'll make some before / after timing runs, put them in a nice table and then @Felienne can do the work of a BDFL ;)

Felienne:
> Is adding TODOs to the code in this way valuable? Could the same note not live in your head, a sticky on your monitor, or a GitHub issue? 

I don't oppose TODOs in the code like this, we have a few of them, and they can serve as nice reminders work needs to be done, but traditionally we have added both initials, date and a PR number to them. Even though that info is present in the repo history, it is nice to have present without a search in case of question.

Felienne:
> > I believe it. I'm just saying that keeping the cache server-side, and requiring running an additional command to prime it, is a bit of a departure from our current design goals, which is (mostly): have Python, check out the repo, ready to go.
> > Felienne should come in here and decide what is acceptable for new contributor experience.
> 
> Mostly agreed. I'll make some before / after timing runs, put them in a nice table and then @Felienne can do the work of a BDFL ;)

I am a bit confused, I thought nothing extra was needed (this is what @jtwaleson answered [https://github.com/hedyorg/hedy/pull/4574#issuecomment-1744519483](here) when I asked about it) but now I read from @rix0rrr that an extra command is needed?

Do let me know so I can chime in.

jtwaleson:
> > > I believe it. I'm just saying that keeping the cache server-side, and requiring running an additional command to prime it, is a bit of a departure from our current design goals, which is (mostly): have Python, check out the repo, ready to go.
> > > Felienne should come in here and decide what is acceptable for new contributor experience.
> > 
> > 
> > Mostly agreed. I'll make some before / after timing runs, put them in a nice table and then @Felienne can do the work of a BDFL ;)
> 
> I am a bit confused, I thought nothing extra was needed (this is what @jtwaleson answered [https://github.com/hedyorg/hedy/pull/4574#issuecomment-1744519483](here) when I asked about it) but now I read from @rix0rrr that an extra command is needed?
> 
> Do let me know so I can chime in.

Sorry for the confusion: no special command is needed, it's just that the first time you run pytest locally it will be slower.  Just as slow as when you make any change to the grammars or hedy.py as that invalidates the cache.

I believe the only downside to this change is this scenario:
- a developer clones the repo
- runs pytest locally without changing the grammars or hedy.py
- tests are now done in +/- 13 minutes instead of 2 minutes (right now on main)
- if you run pytest again (no changes to hedy.py or the grammars) and it runs in under 1 minute

In CI any branch (also from forks) can re-use the cache from the main branch, so there it will be fast every time.

I'm currently running benchmarks and will post the results when it's done.

jtwaleson:
> > Is adding TODOs to the code in this way valuable? Could the same note not live in your head, a sticky on your monitor, or a GitHub issue?
> 
> I don't oppose TODOs in the code like this, we have a few of them, and they can serve as nice reminders work needs to be done, but traditionally we have added both initials, date and a PR number to them. Even though that info is present in the repo history, it is nice to have present without a search in case of question.

Nice. Added to this PR.

Hope you don't mind the force pushing, I like to keep the commit history in logical blocks.

rix0rrr:
I'm interested in the difference between E and G. Looks like running on 8 cores gives us only a 2x speedup?

jtwaleson:
> I'm interested in the difference between E and G. Looks like running on 8 cores gives us only a 2x speedup?

I suspect it's due to the memory caches having to be populated separately. The proposed disk cache in #4581 would make this a bit better.

rix0rrr:
> I suspect it's due to the memory caches having to be populated separately.

Mmm. I buy that.

.github/workflows/unittests.yml
rix0rrr: Prepare to be disappointed

jtwaleson:
@Felienne @rix0rrr let me know when you come to a conclusion on this and/or have questions :)

rix0rrr (approved):

jtwaleson:
Cool! Final results after both #4581 and this one got merged: Fresh unit tests in CI are now done in 18:36
https://github.com/hedyorg/hedy/actions/runs/6439991689/attempts/1

and a cached run takes 6:54 https://github.com/hedyorg/hedy/actions/runs/6439991689/attempts/2

Big improvement from 1 hour (fresh) and 11 minutes (cached)!

Thanks for the reviews!

PR #4575, opened 2023-10-01
[FIX] Remove one parser pass on every `transpile` invocation

rix0rrr:
The `transpile_inner` function was calling `all_commands()`, which would do a parse + transform. Since the `transpile_inner` function already had done a parse, we can run the counting immediately on the AST, which saves doing another parse.

This saves just under half the time of `/parse` on my machine.

Did not remove the `all_commands()` function altogether, since it still gets used in `test_level_01.py`, and in the `/commands/<program>` route.

**How to test**

Everything should work as before, just a bit quicker.


Felienne (approved):
Thanks for cleaning up after me <3

PR #4576, opened 2023-10-01
[FIX] Re-disable parser caching

rix0rrr:
Re-disables parser caching, to reduce memory usage.

Originally proposed in #4504, then reverted later in #4543 because we thought it made parse times a lot worse. 

Reverting it didn't make a difference, so the reason for slower parses must have been something else. For that reason, disable caching again, because saving memory is going to allow for more concurrency on a single machine.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -11,7 +11,6 @@ import warnings
 import hedy
 import hedy_translation
 from hedy_content import ALL_KEYWORD_LANGUAGES
-import utils
 from collections import namedtuple
 import re
 import regex
@@ -2609,20 +2608,11 @@ def get_keywords_for_language(language):
     return keywords
 
 
-PARSER_CACHE = {}
-
-
 def get_parser(level, lang="en", keep_all_tokens=False):
     """Return the Lark parser for a given level.
     """
-    key = str(level) + "." + lang + '.' + str(keep_all_tokens) + '.' + str(source_map.skip_faulty)
-    existing = PARSER_CACHE.get(key)
-    if existing and not utils.is_debug_mode():
-        return existing
     grammar = create_grammar(level, lang)
-    ret = Lark(grammar, regex=True, propagate_positions=True, keep_all_tokens=keep_all_tokens)  # ambiguity='explicit'
-    PARSER_CACHE[key] = ret
-    return ret
+    return Lark(grammar, regex=True, propagate_positions=True, keep_all_tokens=keep_all_tokens)  # ambiguity='explicit'
 
 
 ParseResult = namedtuple('ParseResult', ['code', 'source_map', 'has_turtle', 'has_pygame', 'commands'])


jtwaleson:
I was working on this as well in #4574 and simplified the caching mechanism using functools.cache . I think we definitely want this in CI right? Creating the parsers every time is going to be very slow unless I'm missing some other optimization that I haven't seen yet.

If we want to reduce the memory, maybe we can try to use `functools.lru_cache` with a small maxsize of about 10-50?

As an alternative I looked into doing `Lark.save(...)` and `Lark.load(...)` but got a pickle error when saving. This would be great as it would allow cross-process caching and reducing the memory overhead.

rix0rrr:
>  Creating the parsers every time is going to be very slow 

Well... does it have to be, though? I had a hunch it wouldn't be.

In all fairness, I didn't test it before you challenged me on this. So I tested it just now using `ab`. I benchmarked this on top of https://github.com/hedyorg/hedy/pull/4575, not on `main`, to get a good sense for impact on when the additional parse is gone. `example_program.json` below contains an request to parse an example program from that level, captured by the browser's Network tab.

```shell
ab -p example_program.json -T application/json -c 1 -n 100 http://127.0.0.1:8080/parse
```

## Level 14

Level 14 has quite an expensive parse.

With the cache:

```
Time per request:       258.719 [ms] (mean)

Percentage of the requests served within a certain time (ms)
  50%    259
  66%    263
  75%    265
  80%    267
  90%    270
  95%    271
  98%    279
  99%    310
 100%    310 (longest request)
```

Without the cache:

```
Time per request:       252.508 [ms] (mean)

  50%    254
  66%    256
  75%    257
  80%    259
  90%    260
  95%    263
  98%    270
  99%    304
 100%    304 (longest request)
```

## Level 3

I also tested level 3, which is a lot easier to parse:

```
# With cache
Time per request:       56.480 [ms] (mean)
99%     79

# Without cache
Time per request:       55.553 [ms] (mean)
99%     90
```

## Conclusion

So I think the conclusion is: there might be a couple of milliseconds difference with and without the cache, but compared to the cost of the entire operation it's not worth worrying about. 

That cache probably should not have been there in the first place, it doesn't really do much.

jtwaleson:
Ok I'll take a look when I have a chance. However, I did notice that the unit test job is extremely slow with this change: https://github.com/hedyorg/hedy/actions/runs/6371998627/job/17294420773 . Not sure what's going on there.

jtwaleson:
Cool @rix0rrr , I rebased after #4575 got merged and it speeds up things a bit more.

My benchmark with pytest shows a serious degradation if we remove the cache, so please don't merge this yet. I ran this on the branch with my open PR #4574, every test ran with a cleared snippet hash cache.

Here are the numbers:

    python -m pytest tests/test_level/test_level_01.py
    
    On #4574 without #4575 merged in yet
    - with cache: 22s
    - with cache commented out: 1m46s
    
    On #4754 rebased on #4757
    - with cache: 20s
    - with cache commented out: 1m20s



    python -m pytest tests/test_level/test_level_14.py
    
    Only ran on the rebased version as it was only an improvement.
    - with cache: 31s
    - with cache commented out: 5m22s

rix0rrr:
Good find. Do you have a sense for how many times the parser was instantiated across these tests?

Felienne:
> Good find. Do you have a sense for how many times the parser was instantiated across these tests?

Every parse gets the parser, so probably about 18.000 times.... once for each snippet we test.

Felienne:
However, we introduced the caching only a few weeks ago, and before it was never this excessively slow, so there must also be something weird going on now!

rix0rrr:
>  the unit test job is extremely slow with this change

Wow, 5h45. That is ludricously slow!

Felienne:
> > the unit test job is extremely slow with this change
> 
> Wow, 5h45. That is ludricously slow!
> 
> That would be ~20,000 seconds / ~18,000 parses = 1s per parser construction. Is that realistic? If so, I must have messed up my testing above quite badly.

Yeah that is weird right? We used to have this not-caching for ever and then it was never this bad!

jtwaleson:
> However, we introduced the caching only a few weeks ago, and before it was never this excessively slow, so there must also be something weird going on now!

I looked at the git history, the cache was introduced in May 2021 with 4c8da41eb50e6fa054c171f6344b76a9f1c8c5c0 , so it has been running for quite some time in production.

However the memory costs also went up with all the permutations. First it was just 18 levels (so 18 parsers), then levels * languages, then levels(18) * languages (47) * skip_faulty (2) * keep_all_tokens (2) = 3384 parsers to be kept in memory...

Lark does contain a `cache` which caches the parsers to disk. That would be ideal, especially combined with a small in-memory lru_cache! However, the cache only supports the `lalr` parser, hedy uses `earley`. I have no clue how this works, but maybe it's an idea to provide a patch to Lark to add support for pickling the `earley` parsers. Alternatively it could be an option to switch Hedy to the `lalr` parser (could be a stupid suggestion, as again, I don't know how this works).

The cheapest solution would probably be to buy a bigger dyno, or switch to an lru_cache and take the performance hit.

Felienne:
> > However, we introduced the caching only a few weeks ago, and before it was never this excessively slow, so there must also be something weird going on now!
> 
> I looked at the git history, the cache was introduced in May 2021 with [4c8da41](https://github.com/hedyorg/hedy/commit/4c8da41eb50e6fa054c171f6344b76a9f1c8c5c0) , so it has been running for quite some time in production.
> 
> However the memory costs also went up with all the permutations. First it was just 18 levels (so 18 parsers), then levels * languages, then levels(18) * languages (47) * skip_faulty (2) * keep_all_tokens (2) = 3384 parsers to be kept in memory...
> 
> Lark does contain a `cache` which caches the parsers to disk. That would be ideal, especially combined with a small in-memory lru_cache! However, the cache only supports the `lalr` parser, hedy uses `earley`. I have no clue how this works, but maybe it's an idea to provide a patch to Lark to add support for pickling the `earley` parsers. Alternatively it could be an option to switch Hedy to the `lalr` parser (could be a stupid suggestion, as again, I don't know how this works).
> 
> The cheapest solution would probably be to buy a bigger dyno, or switch to an lru_cache and take the performance hit.

I thought we (re)enabled it here: https://github.com/hedyorg/hedy/pull/4543?

jtwaleson:
> > the unit test job is extremely slow with this change
> 
> Wow, 5h45. That is ludricously slow!
> 
> That would be ~20,000 seconds / ~18,000 parses = 1s per parser construction. Is that realistic? If so, I must have messed up my testing above quite badly.

From running my test suite it seems like that is realistic, but I have limited experience with the project.

> > > However, we introduced the caching only a few weeks ago, and before it was never this excessively slow, so there must also be something weird going on now!
> > 
> > 
> > I looked at the git history, the cache was introduced in May 2021 with [4c8da41](https://github.com/hedyorg/hedy/commit/4c8da41eb50e6fa054c171f6344b76a9f1c8c5c0) , so it has been running for quite some time in production.
> > However the memory costs also went up with all the permutations. First it was just 18 levels (so 18 parsers), then levels * languages, then levels(18) * languages (47) * skip_faulty (2) * keep_all_tokens (2) = 3384 parsers to be kept in memory...
> > Lark does contain a `cache` which caches the parsers to disk. That would be ideal, especially combined with a small in-memory lru_cache! However, the cache only supports the `lalr` parser, hedy uses `earley`. I have no clue how this works, but maybe it's an idea to provide a patch to Lark to add support for pickling the `earley` parsers. Alternatively it could be an option to switch Hedy to the `lalr` parser (could be a stupid suggestion, as again, I don't know how this works).
> > The cheapest solution would probably be to buy a bigger dyno, or switch to an lru_cache and take the performance hit.
> 
> I thought we (re)enabled it here: #4543?

Yes, so it was only inactive for about a week on the main branch. Maybe the snippet hashes masked the issue a bit, as that avoids the parser completely? I'll check out the performance in a commit from that week to see if it's actually slow.

jtwaleson:
Yes the commit that disabled the cache was very slow at 4+ hrs:
https://github.com/hedyorg/hedy/actions/runs/6212537459/job/16862767198

When it was re-enabled it was "fast" again at 1hr:
https://github.com/hedyorg/hedy/actions/runs/6291792681/job/17080518457

The PRs in between were reasonably fast as the snippet_hashes were re-used and no changes to the grammars or hedy.py were made.

Also note that the workflow to regenerate hashes did contain a line to remove the snippet hashes file, but that was executed before the repo was checked out, so it did not do anything, meaning the pickle file only grew in size. This is resolved in #4573.

Felienne:
> I found a way to pickle the parsers so we can cache them to disk :)
> 
> This increases performance considerably and removes the out of memory issues. Let me know if/when a PR would be welcome.

Yes, let's try it, and see if it helps in production! There also is one more change I have been wanting to make to the grammars itself to optimize further, I will look into that too.

rix0rrr:
Let's close this one

PR #4582, opened 2023-10-03
[LANGUAGE] Optimizes grammars

Felienne:
This PR does two things:

* Removes deleted productions from subsequent grammar merges
* Adds concrete keywords in after merging instead of before

**Description**

This makes grammar composition quicker since the merge function does not have to search the long keywords lists (which will never have special merge options `=-` and `=+` and `>`) and it also makes parsing itself quicker since redundant lines are now longer needing to be processed.

**How to test**

The tests should still all pass and hopefully, quicker!


Felienne:
There is a weird issue where some invalid lines get priority -100 (like error_invalid itself) I am not sure why that happens, it is not a problem, but some rules that can be removed, aren't yet. So I will play with it a bit more to also remove these.

For the longer away future, the grammar merging does a lot of string operations, which can be expensive, more speedups could be gained, I suspect, by using a better data structure like a dictionary or set, instead of looping over lists of strings. But that is for the future!

rix0rrr (approved):

PR #4584, opened 2023-10-03
[UI] Remove tick mark referring to third-party agreement 

### LINKED ISSUE #4583: [UI] remove tickmark from profile
In #4538 I removed the tickmark from the signup page but oops... I forgot we have it in profile info also:

image

Should have done a ctrl-f!

This is not a hard issue to fix, go to templates/profile.html and remove the code associated with the tickmark, something like this:

```
<input type="checkbox" id="agree_third_party" name="agree_third_party">
<label class="text-center" for="agree_third_party">agree_third_party</label>
```

And maybe also the `<div>` around it. And you will probably also want to do a code search for third party :) (Do as I say... not as I do 🤣)

ghubrakesh:
Fixes #4583: [UI] remove tickmark from profile 
* [LANGUAGE] -> HTML
* [FIX] -> removed the 'agree_third_party' checkbox as required in issue #4583

```diff
- <div class="flex flex-row gap-4 items-center border-gray-400 border p-4 rounded-lg">
-   <input type="checkbox" name="agree_terms" id="agree_terms" required class="" {% if public_settings %}checked{% endif %}>
-    <label for="agree_terms" class="cursor-pointer w-full">{{_('public_profile_info')}}</label>
- <div>
```


Felienne:
Thanks @ghubrakesh!! Did you by any chance also do a search of the repo if it is used elsewhere?

Felienne:
> Thanks @ghubrakesh!! Did you by any chance also do a search of the repo if it is used elsewhere?

never mind, I did a full search and removed a few more, I think we are now good to go! Thanks again for your first Hedy PR!!

ghubrakesh:
> > Thanks @ghubrakesh!! Did you by any chance also do a search of the repo if it is used elsewhere?
> 
> never mind, I did a full search and removed a few more, I think we are now good to go! Thanks again for your first Hedy PR!!

Oh, I am sorry @Felienne, I did not get the notification for your comment. I hope the issue has been resolved completely by now.

templates/profile.html
```diff
@@ -59,10 +59,6 @@ <h3 class="px-2 pt-0 pb-2 my-0 font-semibold">{{_('favourite_program')}}</h3>
                         </select>
                      </div>
                  {% endif %}
-                 <div class="flex flex-row gap-4 items-center border-gray-400 border p-4 rounded-lg">
-                    <input type="checkbox" name="agree_terms" id="agree_terms" required class="" {% if public_settings %}checked{% endif %}>
-                    <label for="agree_terms" class="cursor-pointer w-full">{{_('public_profile_info')}}</label>
-                 </div>
```
jpelay, line 65: Can you revert this change, please? This tick mark is used if a user wants their profile to be public, and we don't want to remove this one!

jpelay: The lines we want to remove are these ones: 

https://github.com/hedyorg/hedy/blob/f7823ac8ca3780dd680c3edef97a274a4f940afb/templates/profile.html#L145-L148

Can you please restore the lines you deleted and delete these ones instead? I can also do this if you want, but I don't have permissions to commit to your branch :c Thank you very much for your help @ghubrakesh!!

jpelay: See how we're looking to remove the ones that have the string `agree_third_party`. Wich are present in the lines I attached in 3 ways: the id of the input, the for in the label, and also in this weird string `{{_('agree_third_party')}}`, which is a label that will later be replaced with the appropriate translation in whichever language the user has set.

The one you removed only has the string: `agree_terms` which is for agreeing to have a public profile! It's perfectly reasonable to get confused, more so being your first time working with our codebase!

jpelay (changes requested):
Thank you very much for this PR @ghubrakesh! I left a couple of comments because you accidentaly removed code that is still in use and not the tick mark we wanted! I also attached which lines you have to delete; if you need any further help don't hesitate in contacting us!

templates/profile.html
ghubrakesh: Okay @jpelay, I will restore the changes made by me (I will add the deleted lines again) but what about the changes made by @Felienne? Should I ignore them and add the lines deleted just by me?

jpelay: Yes! The changes made by Felienne are correct, and you shouldn't revert them; just the ones I told you about.

jpelay (approved):
Perfect! Thank you very much, again, @ghubrakesh! This looks great!

Felienne:
The error we are getting now is unrelated, and hopefully will be fixed soon!

PR #4589, opened 2023-10-04
[FIX] fix the github workflow for generating snippet hashes

jtwaleson:
**Description**

After my attempt to make the snippet generation change-based in #4573 there's good news and bad news. The good is that the trigger works, and a PR that had a new `hedy.py` file triggered the snippet generation. The bad news is that the automatic commit action failed due to a permission issue.

https://github.com/hedyorg/hedy/actions/runs/6409135403/job/17399693158

Reading the workflow logic this does not make sense to me and feels like a bug in the git-auto-commit-action but it is what it is :shrug: . As we had no simple way of testing this we decided to merge anyway and solve problems as we go along. This fixes the issue.

**How to test**

- merge this and see if the next change to grammars and/or hedy.py manages to commit the snippet hashes correctly.

**Checklist**

### DIFF
--- a/.github/workflows/generate-snippet-hashes.yml
+++ b/.github/workflows/generate-snippet-hashes.yml
@@ -28,6 +28,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        # We _really_ need to pass the token here -- the commit action below
+        # will not overwrite the token to push. We removed it in
+        # 82fd39a9e215246a37f652aef416bb1600d849db and it failed badly.
+        token: ${{ secrets.HEDY_BOT_TOKEN }}
     - uses: actions/setup-python@v3
       with:
         python-version: 3.9


rix0rrr (approved):
I want to say that comment was there for a reason 😜

rix0rrr:
By the way you've also removed the `fetch-depth`. Why is that?

jtwaleson:
> By the way you've also removed the `fetch-depth`. Why is that?

Fetching with fetch-depth: 0 takes 40+ seconds. Without it only a fraction of that. The only reason I could think of keeping it would be the commit action (maybe it needs an entire repo before pushing?) but looking at the documentation of that action there's no need for it. It's not in any of their examples, and looking at the failure right now it fails on authorization. I'm pretty confident that means that it would be able to push.

jtwaleson:
> I want to say that comment was there for a reason 😜

:see_no_evil: :hear_no_evil: :speak_no_evil: It was with good intentions. ;)

PR #4592, opened 2023-10-04
[FIX] fix flaky cypress tests

### LINKED ISSUE #4455: [BUG] Flaky Cypress Tests

jtwaleson:
**Description**

When looking at the errors at https://github.com/hedyorg/hedy/actions/runs/6392606011/job/17350202761?pr=4581 I noticed that going to the logged in page sometimes redirects to the profile page, meaning the session is already logged in.

Cypress should run with test isolation but I'm not 100% sure how it's configured. The simplest thing I can think of is to clear the cookies + session + localstorage whenever we call login. That way we are sure that we have a fresh session.

I built a mini-test locally that tries to login 100x, which failed every time at some point. With this change it succeeds.

This fixes #4455 .

**How to test**

- Very difficult as it is flaky, but if you like you can wrap the login action locally 100x and seeing it fail consistently, and succeeding after this change.

**Checklist**


Felienne (approved):
Let's see if this helps! Thanks for the detailed work!

jpelay:
> Let's see if this helps! Thanks for the detailed work!

Nooo I was just going to make a comment hahaha :sweat_smile:

jpelay:
Hi, Jouke!

I tried this in a CodeSpace and it didn't work for me, Cypress returned this error:

image


This was how I edited the login function:

```
export function login(username, password) {
    for (let i = 0; i < 100; i++) {
        cy.clearCookies();
        cy.clearAllLocalStorage()
        cy.clearAllSessionStorage();
        goToLogin();
        cy.get('#username').type(username);
        cy.get('#password').type(password);
        cy.get('#login_button').click();
        cy.wait(500);
    }
}
```

I'm i testing it right? Or I'm I doing something incorrectly?

jtwaleson:
@jpelay I placed the loop in a separate test that called the loginForTeacher function and then logout. I didn’t see the error then. When I have some time and am back at my computer I’ll check out your case, thanks for verifying!

In any case the (hopeful) fix does not hurt, worst case it doesn’t fix the flakyness.

jpelay:
> @jpelay I placed the loop in a separate test that called the loginForTeacher function and then logout. I didn’t see the error then. When I have some time and am back at my computer I’ll check out your case, thanks for verifying!
> 
> In any case the (hopeful) fix does not hurt, worst case it doesn’t fix the flakyness.

Yes you're right! Let's hope that this indeed fixes the error :D

PR #4594, opened 2023-10-05
Fixes restaurant code problem

MarleenGilsing:


yilmazdurmaz:
Though fixes 3 languages, I am not sure if this is the correct way to fix the issue.  Check below for a little more detail about this.

I guess changing translation files here in source would also require Weblate lock. Instead, fixing this on Weblate for English would be faster. It will auto-mark other languages as "needs editing". https://hosted.weblate.org/translate/hedy/adventures/en/?&offset=543

---
regex on vscode: `\{if\}[^{]*\{is\}[^{]* _ [^{]*\{is\}`

- cy (2728), en (2817), ko (2785) : got updated 6 months ago
- de (2769), es (2770), sv (2802), tr (2843) : translation started later

These has no problems for following reasons:
- nl (2784) : translated before lines lost
- kmr (2732), sr (2709) : translations older than 6 months, but story did not even catch up: 

All others don't have the issue because they are not ever touched and hold another code from ancient(!) times. we actually need a huge clean-up for them to update to the latest.

Felienne:
> All others don't have the issue because they are not ever touched and hold another code from ancient(!) times. we actually need a huge clean-up for them to update to the latest.

Yes, we are aware of this issue. Sadly we depend on volunteers (like you) for all languages other than en and nl, so some languages lag behind for a long time. We can't keep up with all languages, so sometimes we fix only in our own lanugages.

Felienne:
> > All others don't have the issue because they are not ever touched and hold another code from ancient(!) times. we actually need a huge clean-up for them to update to the latest.
> 
> Yes, we are aware of this issue. Sadly we depend on volunteers (like you) for all languages other than en and nl, so some languages lag behind for a long time. We can't keep up with all languages, so sometimes we fix only in our own lanugages.

In response to this, I have updated our Wiki to document that we maintain En and NL and that other languages can be outdated: https://github.com/hedyorg/hedy/wiki/Hedy-Translation-Tutorial#hedy-team-languages

I have also updated our community guidelines to help you understand how we communicate in this project: https://github.com/hedyorg/hedy/wiki#how-to-communicate

Felienne (approved):

PR #4607, opened 2023-10-07
Skip faulty only when more than 1 mistake is made [UI]

### LINKED ISSUE #4404: [UI idea] Should we skip a line if there is only one?
f.e. this program:

image

Does it make sense to skip the line here? I would propose to just give the error in case we have one line with an error, that saves a click and is more clear. What do you think @jpelay and @ToniSkulj?

ToniSkulj:
Fixes #4404

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -2827,6 +2827,10 @@ def transpile(input_string, level, lang="en", skip_faulty=True):
             try:
                 source_map.set_skip_faulty(True)
                 transpile_result = transpile_inner_with_skipping_faulty(input_string, level, lang)
+
+                if source_map.get_amount_of_mistakes() == 1:
+                    raise original_error  # raise original exception if only 1 mistake
+
             except Exception:
                 raise original_error  # we could not skip faulty code, raise original exception
         else:
--- a/hedy_sourcemap.py
+++ b/hedy_sourcemap.py
@@ -102,6 +102,7 @@ class SourceMap:
     level = 0
 
     skip_faulty = False
+    amount_of_mistakes = 0
 
     exceptions_not_to_skip = (
         exceptions.UnsupportedStringValue,
@@ -159,6 +160,9 @@ class SourceMap:
 
             python_code_mapped.append(python_source_code.code)
 
+    def get_amount_of_mistakes(self):
+        return self.amount_of_mistakes
+
     def get_grammar_rules(self):
         script_dir = path.abspath(path.dirname(__file__))
 
@@ -179,6 +183,7 @@ class SourceMap:
     def clear(self):
         self.map.clear()
         self.level = 0
+        self.amount_of_mistakes = 0
         self.language = 'en'
         self.hedy_code = ''
         self.python_code = ''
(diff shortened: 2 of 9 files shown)


Felienne:
Hi @ToniSkulj!

I think I was not clear enough in the issue description! I meant not skipping when there is just one line in the whole program:

```
prind Hallo
```

Because if we skip there, nothing happens. 

In case we have one line with an error but more lines, skipping does make since I think:

ToniSkulj:
Hi @Felienne!
> I meant not skipping when there is just one line in the whole

Ahh that makes much more sense! I did not think this through 😅

> Hope that makes sense and sorry for the confusion.

No worries, will change that, stay tuned!

ToniSkulj:
Hi @Felienne!,

> I meant not skipping when there is just one line in the whole

[Do not skip if 1 line in total](https://github.com/hedyorg/hedy/pull/4607/commits/609ab6f5d1787e6e8ef5025340dc72bda8361718) fixes this now.

Felienne:
Hi @jpelay! Could you test and see if this works? Thanks!

hedy.py
```diff
@@ -2823,7 +2823,9 @@ def transpile(input_string, level, lang="en", skip_faulty=True):
         transpile_result = transpile_inner(input_string, level, lang, populate_source_map=True)
 
     except Exception as original_error:
-        if getenv('ENABLE_SKIP_FAULTY', False) and skip_faulty:
+        hedy_amount_lines = len(input_string.split('\n'))
```
hasan-sh, line 2826: If you try to run only one line that contains an error, we wanna show it. But now you try to skip it if there's two lines; one with code and another empty. Maybe `.strip()` would solve the issue.
image

hedy.py
```diff
@@ -2823,7 +2823,9 @@ def transpile(input_string, level, lang="en", skip_faulty=True):
...
+
+        if getenv('ENABLE_SKIP_FAULTY', False) and skip_faulty and hedy_amount_lines > 1:
```
hasan-sh, line 2828: @Felienne is the `ENABLE_SKIP_FAULTY` env variable set to true on production? Because I had to set it myself to see the effect of the changes!

jpelay: The variable is set to false in production! This is because this feature changes a lot of the core of Hedy, so we want to ship it to teachers once we test extensively and also we want to prepare a proper announcement.

hedy.py
ToniSkulj: Hi @hasan-sh, Amazing catch and you were right! [Strip Hedy code](https://github.com/hedyorg/hedy/pull/4607/commits/b6c2ddd533278b2e2681ec2244e95ee289ddb7d6) solves this issue.

hedy.py
hasan-sh: Hey Jesus, thanks for replying and that makes sense.

jpelay (approved):
Hi @ToniSkulj! This looks greats. Thanks @hasan-sh for picking up the error with the empty lines!

Issue #3520, opened 2022-10-28
[CONTENT] Level 8 turtle code is wrong


LeslieMurphy:
**Describe the issue**
Level 8 turtle code is wrong - English.  It loops way too many times!

Current code

```
angles = ask 'How many angles should I draw?'
angle = 360 / angles
repeat angle times
    turn _
    forward _

```
Correct code

```
angles = ask 'How many angles should I draw?'
angle = 360 / angles
repeat angles times
    turn angle
    forward 40
```

Suggested enhancement to make the turtle colorful

```
angles = ask 'How many angles should I draw?'
colors =  black, blue, brown, gray, green, orange, pink, purple, red, yellow
angle = 360 / angles
repeat angles times
    turn angle
    color colors at random
    forward 40
```

Felienne:
Ah great catch! The angle to angles change should indeed happen, do you know how to do that in Weblate? That is the easiest way to prevent conflicts. If not you can do it in a PR here too.

The second suggestion could be added as a third code example but there already are some color examples before and I worry it would become a bit long. Of course you could add it as your own adventure in your class!

LeslieMurphy:
I will try to figure out the weblate platform, when I get a chance.  Thanks!

Felienne:
I am not sure this has been with through Weblate in the meantime. Could you check @MarleenGilsing?

MarleenGilsing:
Changed all the old English to the new (correct) content in #4622

Felienne:
A fix is still needed for a few other languages, see https://github.com/hedyorg/hedy/pull/4622#issuecomment-1759680183 @yilmazdurmaz has kindly volunteered to pick this up

yilmazdurmaz:
Weblate commits in PR #4625 now has the all-language correction. So, I am closing the issue. feel free to reopen, or add comments, if something is not right.

PR #4622, opened 2023-10-12
Fix turtle level 8 - issue 3520

MarleenGilsing:
Addresses #3520 in English


Felienne:
> simple but lovely touch.
> 
> side note: this string for these following language codes have previously been translated and use wrong variable names.
> 
> es, hu, id, ar, pt_BR, nb_NO, eo, sq, da,
> 
> you may try to fix them by copy/pasting correct variable name if your tool allows fast finding in these files, or I may later edit them on Weblate. add a comment about the decision before merging so I won't miss it.

Would be great if you can fix this later on Weblate, I will leave the issue open so we don't forget.

Felienne (approved):

PR #4623, opened 2023-10-12
[CONTENT] Debugging adventures level 1 - 10

MarleenGilsing:
Debugging adventure for level 1 - 10 in English. 

The tests are obviously gonna be angry that the examples contain bad codes. I haven't added them to the other languages yet, I tought I'd do that once I'm do all the way done through level 18. @Felienne I don't know if you want to put these changes through before that already, or if you'd rather wait until the other levels are ready (in that case you could convert this to a draft).


Felienne:
> Debugging adventure for level 1 - 10 in English.

Love this @MarleenGilsing!! Really great work.

> The tests are obviously gonna be angry that the examples contain bad codes. 

My fixes above should take care of that!

> I haven't added them to the other languages yet, I tought I'd do that once I'm do all the way done through level 18. @Felienne I don't know if you want to put these changes through before that already, or if you'd rather wait until the other levels are ready (in that case you could convert this to a draft).

If the tests are happy, I am ok with releasing this so other languages can start translating this!

MarleenGilsing:
I have added new content upto and including level 18. So it's ready to go!

@Felienne Am I right to think that I don;t have to copy-paste it to the other languages because all of the content is brand new? Or should I still do that?

Felienne:
> I have added new content upto and including level 18. So it's ready to go!

Awesome!! There is still a test failing (but this is not your fault). I will ask Jesus to dive in because I also don't understand the issue.

> @Felienne Am I right to think that I don;t have to copy-paste it to the other languages because all of the content is brand new? Or should I still do that?

Yes, it will go automatically (I will keep an eye on it)

Felienne:
Ok, I am diving in here again because it is annoying in all sorts of ways that this is stuck!

It seems to be the second teacher test:

image

@jpelay also found out that there are differences locally and on github, so my best guess is that the test database does not have the right information to test a second teacher and that someone messes up things?

Felienne:
Maybe @hasan-sh can take a peak since it might be related to the new second teacher feature?

hasan-sh:
> @jpelay also found out that there are differences locally and on github

What differences?



> the test database does not have the right information to test a second teacher and that someone messes up things?

I've added 3 more teachers (check [here](https://github.com/hedyorg/hedy/blob/main/data-for-testing.json#L123-L205)) just for testing.


I cannot reproduce the error locally. But after some digging, here's the error in the logs:
> This element `<div.tab.z-10.whitespace-nowrap.flex.items-center.justify-left.relative>` is not visible because its content is being clipped by one of its parent elements, which has a CSS property of overflow: `hidden`, `scroll` or `auto`

This element is appended to the current adventures by clicking on the available adventure dropdown. So, the problem is that it's invisible to the screen and but exists in the list. Shall i fix it in this pr?

jpelay:
> > @jpelay also found out that there are differences locally and on github
> 
> What differences?

Some tests were failing locally, but maybe that was because I didn't do feed the dev database first!

Felienne:
> I cannot reproduce the error locally. But after some digging, it occurs in both customize_class.cy.js and customize_class_second_teacher.cy.js here's the error in the logs:

Thanks for diving in!!

> This element is appended to the current adventures by clicking on the available adventure dropdown. So, the problem is that it's invisible to the screen and but exists in the list. 

Is that because the list is getting too long?

> Shall i fix it in this pr?

If you can, yes! Merging a PR with failing tests is asking for issues down the line when something really breaks.

 We can maybe change the test itself? It is not so important that the thing is visible, as long as it is in the list?

Felienne:
Thanks for the fix @hasan-sh!!

Felienne (approved):

PR #4626, opened 2023-10-14
[FIX] Generalize the fix for quotes in french exception messages

### LINKED ISSUE #4610: [BUG] "bad input" error in French workspace for lists or "at random"
**Describe the bug**

The code below is the fortune teller program from level 5 (stripped off braces from English keywords). This code works without any problem if I set preferred language to German, Dutch, Turkish. But it gives "bad input" error for French as @etienneCharignon  commented on Weblate.

I dig a bit more and played with this code on Hedy. It seems to be either on parsing lists, or the use of "at random", but surely for French at least, as this simple code does not work either:
````
amis is Jonathan, Lucile, David
print amis at random
````
> Quelque chose s’est mal passé en exécutant ce programme.

Level 5 - fortune teller

```
amis is Jonathan, Lucile, David
print "Je peux prédire si tu vas gagner à la loterie demain !"
personne is ask "Qui es-tu ?"
bonnes_réponses is Hourra ! Tu as gagné !, Tu vas certainement gagner !, Nous avons un gagnant !
mauvaises_réponses is Pas de chance ! Essaye encore !, C`est une autre personne qui gagnera, Tu as perdu !
if personne in amis print bonnes_réponses at random
else print mauvaises_réponses at random
```

screenshot-hedy-alpha herokuapp com-2023 10 07-18_18_00

- Windows
- Edge

### PR DESCRIPTION
**Pull request by etienneCharignon**
**Description**

In French, it is common to have quotes in messages, this commit generalizes the correction that has been done at level 1 but that was also present at level 16.

If you tried the following code at level 16, you had the same error as the one reported by the issue #4610 instead of having a nice error message

```
fruit = []
fruit[1] = "pomme"
```

Capture d’écran 2023-10-14 à 15 26 29

**Adds** I created the function `translate_index_error` responsible for translating a possible IndexError exception and escaping all possible quotes in the error message.

Being new to Hedy, I did not really know where to put that utilities function. Please give me any instruction in the comments.

**Fixes #4610**

**How to test**

Run the following code in French, at level 16 : 

```
fruits = []
fruits[1] = "pomme"
```

The reported error message should be: 
`Tu as essayé d'accéder à la liste fruits, celle-ci est vide ou sans index.`
instead of
`bad input`

**Checklist**

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -174,6 +174,16 @@ def needs_colon(rule):
     return f'{rule[0:pos]} _COLON {rule[pos:]}'
 
 
+def _translate_index_error(code, list_name):
+    exception_text = gettext('catch_index_exception').replace('{list_name}', style_command(list_name))
+    return textwrap.dedent(f"""\
+        try:
+          {code}
+        except IndexError:
+          raise Exception({repr(exception_text)})
+        """)
+
+
 PREPROCESS_RULES = {
     'needs_colon': needs_colon
 }
@@ -1493,17 +1503,8 @@ class ConvertToPython_1(ConvertToPython):
                     list_args.append(group[3])
                     lists_names.append(group[4])
         code = ""
-        exception_text_template = gettext('catch_index_exception')
         for i, list_name in enumerate(lists_names):
-            exception_text = exception_text_template.replace('{list_name}', style_command(list_name))
-            if "'" in exception_text:
-                exception_text = exception_text.replace("'", "\\'")
-            code += textwrap.dedent(f"""\
-            try:
-              {list_args[i]}
-            except IndexError:
-              raise Exception('{exception_text}')
-            """)
+            code += _translate_index_error(list_args[i], list_name)
         return code
 
 
@@ -2312,13 +2313,7 @@ class ConvertToPython_16(ConvertToPython_15):
     def change_list_item(self, meta, args):
         left_side = args[0] + '[' + args[1] + '-1]'
         right_side = args[2]
-        exception_text = gettext('catch_index_exception').replace('{list_name}', style_command(args[0]))
-        exception = textwrap.dedent(f"""\
-        try:
-          {left_side}
-        except IndexError:
-          raise Exception('{exception_text}')
-        """)
+        exception = _translate_index_error(left_side, args[0])
         return exception + left_side + ' = ' + right_side
 
     def ifs(self, meta, args):

### CODE REVIEW

**Comment by Felienne** [comment]
Thanks for the fix @etienneCharignon and welcome to the project. As for the place of the function, we don't really have a good guideline for that at the moment. For me this is fine, maybe @jtwaleson would also like to chime in?

File: hedy.py [hunk 1]
```diff
@@ -174,6 +174,17 @@ def needs_colon(rule):
     return f'{rule[0:pos]} _COLON {rule[pos:]}'
 
 
+def translate_index_error(code, liste_name):
```
jtwaleson, line 177: `liste_name` -> `list_name`?

jtwaleson, line 177: If it's a util method only used internally in `hedy.py` it's standard python practice to add an underscore to the method. So `_translate_index_error` . That signals to the developers that it's a one-of util method, but python doesn't treat it any different.

hedy.py
```diff
@@ -174,6 +174,17 @@ def needs_colon(rule):
...
+    exception_text = gettext('catch_index_exception').replace('{list_name}', style_command(liste_name))
+    exception_text = exception_text.replace("'", "\\'")
```
jtwaleson, line 179: we can leave this out if we use `repr` below.

hedy.py
```diff
@@ -174,6 +174,17 @@ def needs_colon(rule):
...
+    return textwrap.dedent(f"""\
+        try:
+          {code}
+        except IndexError:
+          raise Exception('{exception_text}')
```
jtwaleson, line 184: @Felienne Not strictly part of this change, so not sure if we want to include it, but we can make this `raise Exception({repr(exception_text)})` to properly escape all string oddities. Then we no longer need the `replace` method 
for the single quote escaping above. Let me know if you want to add this here.

jtwaleson:
@etienneCharignon I think we're almost good to go, one typo and the other is the naming convention for the methods. We don't have a clear way to put these functions in and we know that `hedy.py` is a bit bloated at 3k+ lines, but that's something we can work on some other time. Right now adding the method "somewhere" in hedy.py is perfectly fine. Thanks so much for the contribution!

@Felienne I added one suggestion that I'd need your opinion on but it's not really part of this PR so we can also decide to not pick it up or postpone it.

hedy.py
Felienne: Lovely idea, that would allow us to catch other issues too. I am ok with you pushing this on this branch but it is also ok of course if @etienneCharignon does that!

hedy.py
etienneCharignon: oops, the french word "liste" got me !

hedy.py
etienneCharignon: very well ! That's what I wanted to do initaily but did not know how to do it.

etienneCharignon:
@jtwaleson I took all your comments into account.

jtwaleson (approved):
Great, thank you @etienneCharignon !

Felienne:
> @jtwaleson I took all your comments into account.

Thanks!! Congrats on getting your first Hedy PR merged! If you want to pick up more work, we'd love to see you on Discord and/or in our biweekly contributors meeting on Tuesday (tomorrow 7:30 CET, see Discord).

etienneCharignon:
@Felienne Unfortunately, I will not be available today at 7:30pm. But I'm interested in participating. Maybe next time!

PR #4635, opened 2023-10-17
[CONTENT] fixes some typos in pages

Felienne:
I noticed some issues in the pages, but of course this is a weblate lock!

### DIFF
--- a/content/adventures/ca.yaml
+++ b/content/adventures/ca.yaml
@@ -789,7 +789,7 @@ adventures:
                     Som-hi!, anem a la següent pestanya a descobrir la nova comanda!
                 example_code: |
                     ```
-                    contrassenya {is}{ask}'Quina és la contrassenya correcta?'
+                    contrasenya {is} {ask} 'Quina és la contrassenya correcta?'
                     ```
                 start_code: |-
                     nom {is} {ask} 'Quin és el teu nom?'
--- a/content/adventures/it.yaml
+++ b/content/adventures/it.yaml
@@ -66,16 +66,19 @@ adventures:
                     The command `{add}` can also come in handy in your story.
                 example_code_2: |
                     ```
-                    {print} Lui sente un suono
-                    animali {is} 🐿, 🦔, 🦇, 🦉
-                    animale {is} {ask} Chi pensi che sia?
-                    {add} animale {to_list} animali
-                    {print} è stato l'animale {at} {random}
+                    {print} He hears a sound
+                    animals {is} 🐿, 🦔, 🦇, 🦉
+                    animal {is} {ask} What do you think it is?
+                    {add} animal {to_list} animals
+                    {print} it was an animals {at} {random}
                     ```
                 story_text_3: |
                     Questo è un esempio del comando `{remove}` nella tua storia
 
-                    Riesci a usare i nuovi tre comandi `{at} {random}`{random}, `{add} {to}` e `{remove} {from}` nella tua storia?
+                    ### Esercizio
+                    Copia la tua storia dai livelli precedenti a questo livello.
+                    In questo livello hai imparato 3 nuovi comandi `{at} {random}`, `{add} {to}` e `{remove} {from}`.
+                    Aggiungi nuove righe di codice alla tua storia, in modo che tutti i nuovi comandi siano presenti almeno una volta nella tua storia.
                 example_code_3: |
                     ```
                     {print} His backpack got way too heavy.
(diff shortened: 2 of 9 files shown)


yilmazdurmaz:
@Felienne with this PR, I saw a worrying problem with Weblate: changes made here are not marked to be "needs editing".

I was thinking I got 100% but the Turkish file you edited showed me I have at least 1 incorrect translation because of this missing notification.

Can you please check if Weblate has a setting to mark changes from repository for further editing?

Felienne:
> @Felienne with this PR, I saw a worrying problem with Weblate: changes made here are not marked to be "needs editing".

Ah good to know!

> I was thinking I got 100% but the Turkish file you edited showed me I have at least 1 incorrect translation because of this missing notification.

Ah, sorry! In the project speed and size as we have it now, sadly, we sometimes do need fixes made outside of Weblate.

> Can you please check if Weblate has a setting to mark changes from repository for further editing?

This is a question for @Mark-Giesen:

Can we mark strings that are changed from a non-Webalte PR as needs editing?

yilmazdurmaz:
> Ah, sorry! In the project speed and size as we have it now, sadly, we sometimes do need fixes made outside of Weblate.

That's absolutely fine as the content continues to evolve. If we can solve this "needs editing" problem, then it will be of greater help to correcting and synching towards other languages.

Felienne:
> > Ah, sorry! In the project speed and size as we have it now, sadly, we sometimes do need fixes made outside of Weblate.
> 
> That's absolutely fine as the content continues to evolve. If we can solve this "needs editing" problem, then it will be of greater help to correcting and synching towards other languages.

yes let's see if Mark knows a solution!

Mark-Giesen:
> Can we mark strings that are changed from a non-Weblate PR as needs editing?

Yes we can and we do. I've been editing "Needs editing" strings in Dutch this week, usually because of small changes in the source. In these cases Weblates shows both the old en new source strings and highlights the diffs, so the translator sees what changes have been made. This "change detection" is done by Weblate itself if possible, the keys of YAML files are known and if the text changes, it is marked for edit in all other languages. 
In PO files this is done by the pybabel tool during extract or update phase, changed strings are marked as fuzzy and this mark is use by Weblate as trigger to mark the string for edit.
If our translation process is followed these changes should be updated nicely. The only faulty route I can think of is if someone translates right into PO files without setting the fuzzy flag. This way Weblate can't see if this is a new string or an altered one.

Mark-Giesen:
Weblate is a little smarter than you gave it credit for. In the Turkish translation, the change in English is correctly overwritten. You only had one line translated out of the complete text. Weblate noticed the original EN part and patched the change. The problem here seems to be the translation was not finished yet, while someone clicked "save and continue".  In Dutch, Weblate couldn't do this, because that specific line was allready translated. 

How smart these partial updates are in Weblate I don't know.

yilmazdurmaz:
> Weblate noticed the original EN part and patched the change

actually, (sorry for bad news) this was not Weblate. Felienne manually made those corrections in all language files within this PR (unless I missed something)

yilmazdurmaz:
new observation: the history of this string for Dutch on the link above, there is no record of editing through Weblate as you supply the translation here within the repo. yet it has "need editing" mark unlike the Turkish version which saw no editing since my last one except pushes from repo.

of course, I might be wrong so heads up, but it is wise to first check settings for languages. start comparing Dutch and Turkish. 

PS: I am preparing to run a local Weblate instance. If I get success, I may pinpoint such problems (if any). Until then, please bear with me.

Mark-Giesen:
Maybe we're talking about different things? I think Felienne changed "costumizing" into "customizing" in the Englisch version. Your Turkish version now has this change as well, it has "customizing" correctly in the last sentence. So there was no need for "Need editing" to be set, the change was applied.
In the Dutch version, this was not possible because the old "costumizing" was allready translated.

yilmazdurmaz:
> Maybe we're talking about different things?

I think we are just looking at the problem in different angles and our views differ a bit :)

let me add another thing that might have caused Dutch version to have the "needs editing" mark.

screenshot-hosted weblate org-2023 10 18-22_59_43

"I think" this is what happened back then: 3 weeks ago, this string has changed to this new long version, and Dutch component was already translated in the repo. When uploaded to Weblate, it did not like the problems you see in the image and automatically marked it for editing. On the other hand, Turkish version did not have these problems, passed the checks, and never marked.

If this is true, then there was (and is) no settings to make the mark in the first place. I don't know, however, how to test this possibility.

yilmazdurmaz:
oh, by the way, the whole string actually "needs" to be edited as it is "not" really translated after it was added 3 weeks ago, and I was not aware of it since then, until I see this PR, because of this marking issue.

Mark-Giesen:
To test if it works, you can change a string in English in Weblate, all languages with translations should have the "need editing" set after that (can take en few minutes). If that works, you could try the same with an string in the yaml file in our repo. I did something with a test_string last week, you could try something like that. Tests take some time though.

yilmazdurmaz:
> you can change a string in English in Weblate

There is no problem with changing within Weblate. I made a lot of small changes like that and it marks them fine. The problem, I think, is with the changes directly pushed from here.

for changing a yaml file here, "why not". for the moment, my brain is in a bit of lazy night mod for such a trial, so it may take a few hours to decide a workflow. but basically, these two are the things on mind:
- select an English string that needs a small typo such as backticks, and make edit here on a yaml file
- select a Turkish string that already needs a bit of touch, and make edit here on a yaml file
- push the two files and see what Weblate's response is.

I will ping you in it when I make that PR

PR #4643, opened 2023-10-18
[TESTING] Test for the "needs editing" problem of weblate

yilmazdurmaz:
**Description**

@Mark-Giesen  @Felienne 
As we discussed in #4635, the changes are simple and includes only 2 files to test the "needs editing" problem of Weblate.

_Expectation_

- Changes made to English should be reflected to all other languages and they should be marked as "needs editing"
- Changes for other languages, small or big, should be seen by Weblate as "needs editing"
- so that translators can see and react to changes and start proof reading.

**How to test**

For the merge, automatic tests should be enough. But it possibly needs Weblate lock & sync.

for the actual result, these three links must have "needs editing" checkbox active after the sync:
- https://hosted.weblate.org/translate/hedy/adventures/tr/?checksum=3b28260d4fdbc59b
- https://hosted.weblate.org/translate/hedy/adventures/tr/?checksum=0ac7882083bb6913
- https://hosted.weblate.org/translate/hedy/adventures/tr/?checksum=648220eade026e9c

**Checklist**

- [+] Contains one of the PR categories in the name
- [+] Describes changes in the format above
- [+] Links to an existing issue or discussion
- [+ ] Has a "How to test" section

### DIFF
--- a/content/adventures/en.yaml
+++ b/content/adventures/en.yaml
@@ -1838,8 +1838,8 @@ adventures:
                     When you've finished the code, try to create a code of your own and use a question that you've thought of yourself.
                 example_code_2: |
                     ```
-                    animals is dog, cow, sheep
-                    answer is ask 'What is your favorite animal?'
+                    animals {is} dog, cow, sheep
+                    answer {is} {ask} 'What is your favorite animal?'
                     _ answer _ animals _ 'Mine too!'
                     _ _ 'My favorite animals are dogs, cows and sheep'
                     ```
--- a/content/adventures/en.yaml
+++ b/content/adventures/en.yaml
@@ -859,7 +859,7 @@ adventures:
                     ```
                     answer = {ask} 'Are you ready to learn something new?'
                     {if} answer {is} yes
-                        {print} 'Great! You can learn to use the repeat commando in the if command!'
+                        {print} 'Great! You can learn to use the repeat command in the if command!'
                         {print} 'Hooray!'
                         {print} 'Hooray!'
                         {print} 'Hooray!'
--- a/content/adventures/tr.yaml
+++ b/content/adventures/tr.yaml
@@ -30,7 +30,7 @@ adventures:
                 start_code: '{print} Hikayeniz burada başlıyor'
             2:
                 story_text: |
-                    2. seviyede, hikayenizi daha eğlenceli hale getirebilirsiniz. Ana karakterinizin adı artık cümle içinde herhangi bir yerde olabilir.
+                    Bu 2. seviyede, hikayenizi daha eğlenceli hale getirebilirsiniz. Ana karakterinizin adı artık cümle içinde herhangi bir yerde olabilir.
 
                     Bunun için birazcık daha fazla programlamanız gerekiyor. İlk önce ana karakterinize isim vermelisiniz.
 


yilmazdurmaz:
Although I named the PR as testing but the changes are already needed, so they are here to keep even if my test fails.
- "content/adventure/en.yaml" was missing curly braces around
- "content/adventure/tr.yaml" was seen by Hedy as a start of the list.

yilmazdurmaz:
I added a third fix to test typo correction.

The first change on the English is about enclosing keywords, but I think this can be marked because Weblate checks the braces.

The other one is a typo correction, but it could be a big change that would easily pass Weblate checks.

Let's see what happens.

Mark-Giesen (approved):

Mark-Giesen:
On first and third link you'll see the new and old EN versions and the TR version. The items are marked "needs editing"
Second link shows approved TR text, no marking of "needs editing". This is correct behaviour, since source did not change.

All three behave like they're ment to.

yilmazdurmaz:
> Second link shows approved TR text, no marking of "needs editing". This is correct behaviour, since the source did not change.

@Mark-Giesen  So, we narrowed the problem down to this one.

Unfortunately, this behavior is "incorrect".

Just like #4635, this causes drifts and lets inconsistent translations go unnoticed. Last month, I was confident I had translated Turkish to 100%, but now I know at least 1 string slipped without any notice.

The "needs editing" check on Weblate signals us, translators, that something has changed and should be proof-read whether it is correct or not.

Can you please check Weblate's settings about this?

yilmazdurmaz:
> You might change something in Weblate while another translator changes it in an other way in yaml which you'll never now and your change will be lost

this is exactly what the current problem is without this marking. Discourage ways other than Weblate. Weblate itself provides tools such as direct and partial file uploads on it. Instead of committing to the repo, one can add to Weblate, and it is much easier to track than repo operations.

> It doesn't matter to us if someone translates via the yaml in our repo, or via Weblate.

And this is the heart of the problem: Do not let translations be made directly to the repo. That is why we have the Weblate. And if you do, this needs to be "marked for edit". I know you guys maintain English and Dutch and do translations here. But this should be discouraged for others.

> We won't change this behaviour though, because this is by design.

I don't know who composed the "we" and what the circumstances were at the time. Yet it is clear this problem we are discussing here is one important problem for translators to maintain translations synched with the source.

@Felienne As we have a dispute on the topic, where can I open this to further discussion?

Felienne:
> And this is the heart of the problem: Do not let translations be made directly to the repo. That is why we have the Weblate. And if you do, this needs to be "marked for edit". I know you guys maintain English and Dutch and do translations here. But this should be discouraged for others.

In essence the situation is as you describe, it is discouraged, but not prohibited. Sometimes people (maybe people not even knowing that Weblate exists!) make wanted changes here. We don't want to discourage them by saying: thanks for this fix, now do it again on an entire new platform that you have never seen (and it not so user friendly).
 
> I don't know who composed the "we" and what the circumstances were at the time. Yet it is clear this problem we are discussing here is one important problem for translators to maintain translations synched with the source.
> 
> @Felienne As we have a dispute on the topic, where can I open this to further discussion?

Hi @yilmazdurmaz, no further discussion is needed, the "we" here represents the core team. The Weblate setup is designed with several constraints and while this is indeed a problem, the alternative is also unwanted since it limits power that we have for fixes outside of Weblate.

I know it is annoying for you! But if you want to keep up to date on the source, there is the Github history that you can read, where everything is documented always, so it tells you what changes could be needed.

Issue #3610, opened 2022-11-09
[FEAT] Adds support for second teacher of a class


TiBiBa:
**Description**
As requested it would be nice to add support for teachers to add a second teacher to their classes. It would however still be nice to make sure that only the "main" teacher can make crucial decisions like renaming, deleting and/or duplicating the class. We do have to "puzzle" a bit as the classes of a teacher are stored in the database as well as on the user profile. In this case we have to alter the profile itself as well.

Maybe we can use the user invite for class structure to implement this "invite as teacher" idea?

Felienne:
This could be a fin issue for @hasan-sh to get started with, since it will help get to know our teacher backend. There are a few things to consider though, I think we want to go for a "first teacher" and "second teacher" where maybe only the first teacher can, for example, delete the class, or remove students? And we have, of course, to think of interesting edge cases, what is a secondary teacher removes their account? What is a first teacher does this? So a bit of "scoping out" is needed before the start of this.

hasan-sh:
Since the "second teacher" may or may not have all capabilities, I'm thinking of the following user story:

The first teacher creates a _class_. They can create and customize classes, adventures, and add/invite/create students. Among the aforementioned things the teacher is able to control, they may add a second teacher (or a teacher assistent; since several may be added in the future!) with _certain capabilities_. The teacher assigns one of the following labels to their assistent: viewer (default), moderator, admin. 

A viewer can:
- [ ] view statistics; the progress of that class and students. 

A moderator (allowed to perform specific actions) can:
- [ ] view adventures, live stats and progress so far (explore)
- [ ] add/invite/remove students
- [ ] customize the class; add/remove/assign adventures
- [ ] @Felienne maybe there's more actions to be considered? But this is what I found after loging in as a teacher and playing around a bit!

> These actions, as you mentioned, should be specified by the teacher (so, ui needed).  
> I think the first two would be logical to be the default for the moderator, while the third is not and needs to be determined by the teacher. 

Finally, an admin is granted with all capabilities of the teacher (really second teacher!) 


[...]

PR #4644, opened 2023-10-19
[FEAT] second teacher

hasan-sh:
**Description**
#3610 

Depends on #4681

**How to test**
go to classes and invtie second teachers.

**Checklist**

- [x] invite/remove second teacher
- [x] update `profile` page
- [x] update `class-overview` page
- [x] update `customize-class` page
- [x] update `customize-adventure` page
- [x] show programs of each teacher in class
- [x] show adventures of each teacher in class
- [ ] add tests
    - [x] could invite teachers
    - [x] could customize and edit class
    - [x] could only view adventures
    - [ ] could only view programs

**Summary: Updates and Rules**

- second teachers can be added from the `/fore-teachers/class/[class_id]` page.
- only teachers can be second teachers
- second teachers can do everything the main teacher can do except deleting the class.
- second teachers can create, read and update a class
- upon creating adventures the ownership is simply to the creator (current user)
- adventures of all second teachers in all classes of the current teacher are visible to the teacher
- adventures of teachers are visible to second teachers
- second teacher cannot see other second teachers' adventures unless they are in the same class.
- adventures can only be updated/deleted by the creator.
- second teachers can customize any class they are in (granting them the same rights as the respective primary teacher.)


content/adventures/en.yaml
```diff
@@ -859,7 +859,7 @@ adventures:
                     ```
```
Felienne, line 859: Hmmm these changes are going to upset weblate probably, but also... why are they in this PR? These seem uprelated? I don't know how they got here, do you understand @hasan-sh?

Felienne: (maybe a merge from main when the conflicts are addressed will solve this?)

README.md
```diff
@@ -5,7 +5,7 @@
 ![](https://img.shields.io/github/issues/hedyorg/hedy)
 
 ![](static/images/Hedy-logo-96x96-round.png)
-
+<div id="top"></div>
```
Felienne, line 8: This too is a recent change on main, and as fas as I understand should not be here.

Felienne:
I have correctly merged main in @hasan-sh! I set it to draft because it is now missing your new strings and translations. 

Can you add those by running extract again? (if not no problem, than I will finish it later)

hasan-sh:
> Can you re-add them?

Sure, will do so tomorrow morning ;)

Felienne (changes requested):
Hi @hasan-sh!

I tried the feature and it works great!! Teachers will be very happy!

I love also how you have thought of many details, like adding the role in the invite list:

image

There are still a few tests failing, hope you can look into those! 

In addition to that, a few comments from my first trial run.

* I am not sure if I am supposed to type a username or email address here? Or maybe both work? Can we add a label?
image

* inviting a non-existing teacher gives a weird error:
image

* this is a bit of a "stretch goal" (let's do that in a separate PR!) but maybe we want to show the teachers here too: 
image
 
* this too is a nice to have, but maybe after the adventure name here we can also add the author (maybe between brackets?) 
image

* I was not sure what this label is:
image
 (but maybe I'd understand if the strings are filled)

Felienne:
Hi @hasan-sh! I think in principle, this is good to go (it seems to work well) but we need to look at the tests, do they catch a real problem, are they breaking for no reason (that sometimes happens) or are they in need of some changes due to the PR?

Let us know if you need help fixing the tests!

hasan-sh:
> Hi @hasan-sh! I think in principle, this is good to go (it seems to work well) but we need to look at the tests, do they catch a real problem, are they breaking for no reason (that sometimes happens) or are they in need of some changes due to the PR?
> 
> Let us know if you need help fixing the tests!

I changed my approach on testing the feature; some tests relied on others. Now they run independenlty! All tests passed locally! However, the error that the github action yields is: `Os is not defined`. I think it's some missing env variable or some other small thing!

Felienne:
> > Hi @hasan-sh! I think in principle, this is good to go (it seems to work well) but we need to look at the tests, do they catch a real problem, are they breaking for no reason (that sometimes happens) or are they in need of some changes due to the PR?
> > Let us know if you need help fixing the tests!
> 
> I changed my approach on testing the feature; some tests relied on others. Now they run independenlty! All tests passed locally! However, the error that the github action yields is: `Os is not defined`. I think it's some missing env variable or some other small thing!

For me also the e2e tests pass locally (I dud not try Cypress) so indeed I guess something small/weird is amiss, but we still need to see how to get this fixed since we don't want to merge broken tests. I don't know what the issue is myself.

Do have time to dive in and do you see a path forward? Because of all the changes and commits it is very hard for someone else to understand what happens (or maybe @jpelay has a brilliant plan?)

hasan-sh:
> Do have time to dive in and do you see a path forward? Because of all the changes and commits it is very hard for someone else to understand what happens (or maybe @jpelay has a brilliant plan?)

Yes, I do. Just pushed the tests for adding tags to adventures! Will look into this!

Felienne:
Woohoo, all good to go! I will do Weblate first and then this one.

Felienne (approved):

PR #4652, opened 2023-10-19
[UI] Implement buttons to switch tabs

### LINKED ISSUE #4606: [UI] Confusing go to level 2 button
The discussion below makes a great point! The funny thing is, that the "go to level" button predates the tabs, so it made sense in the beginning and we just never changed it!

But the correct logic should totally be to go to the next tab, and only show "next level" at the last one!

### Discussed in https://github.com/hedyorg/hedy/discussions/4358

<sup>Originally posted by **markusgeert** June 21, 2023</sup>
The button `Go to level 2` can be confusing for beginners, since they expect to go to the next tab. Maybe a better idea would be to replace it with a `Go to next tab` button, and add the button for going to the next level at the end of the tab list? </div>

TiBiBa:
In this PR we re-style the regress and advance buttons for levels. Currently the buttons below the output window directly navigate the user to the previous or next level. This has been confusing for students and they aim to go to the next adventure. To fix this we replace the level buttons with regress / advance adventure buttons and implement a levels dropdown to allow directly switching a level.

The quiz already has a "go to next level" button within their own UI so this is automatically solved. We do however not provide a solution yet for a "next level" button on the last adventure dynamically. Do we need this?

This PR fixes #4606


jpelay:
Hello, @TiBiBa! It was a bit unfortunate that we had to choose from two PRs implementing the same feature, but we ultimately decided to move forward with #4709, because that PR implements it with HTMX. However, that one is missing the dropdown, hope it is ok to take it from this PR and move it there?

I'm really sorry Timon, but in any case we are really thankful for this PR and your work on it.

TiBiBa:
> Hello, @TiBiBa! It was a bit unfortunate that we had to choose from two PRs implementing the same feature, but we ultimately decided to move forward with #4709, because that PR implements it with HTMX, and we will close this one. However, that one is missing the dropdown, hope it is ok to take it from this PR and move it there?
> 
> I'm really sorry Timon, but in any case we are really thankful for this PR and your work on it.

No problem, feel free to take the dropdown code and implement it within #4709.

PR #4657, opened 2023-10-20
Polish translations for some part of adventures

note:
I've extended the Polish translations so adventures up to level 3 (including 3) is fully covered.

Sorry for not following the suggested structure of the PR, but couldn't find one for translation work.

I didn't use Weblate because with editing file and running the website with Docker I feel more in control, with Weblate I somehow couldn't map translation to the actual content


Felienne:
Welcome to Hedy @note! Doing translations in a PR is also fine, if you have experience coding that is easier. 

Maybe indeed we should make a template for translation PRs. I am labeling this with weblate-lock, since Weblate gets in conflict if we merge content here, so I will first merge the latest weblate and then this.

Thanks for picking up Polish!

Felienne:
Hi @note! Are you planning to continue all adventures on this branch/PR? In that case, it might be better to convert it to a draft until you are ready (or until you are ready for a while, I can imagine you don't want to do all adventures in one go, it is a lot!)

note:
Hi @Felienne, I don't mind either way, but to keep it simple let's assume I am done with branch and will open another once this one is merged and accumulated some local changes :)

Felienne:
> Hi @Felienne, I don't mind either way, but to keep it simple let's assume I am done with this branch and will open another once this one is merged and accumulated some local changes :)

ok! The "problem" is that I have to merge it in when weblate is complete (which happens never since people translate basically around the clock haha) so I have to lock weblate, merge this and open weblate, which is why I care so much about knowing when its ready

note:
> ok! The "problem" is that I have to merge it in when weblate is complete (which happens never since people translate basically around the clock haha) so I have to lock weblate, merge this and open weblate, which is why I care so much about knowing when its ready

Makes sense, thank you for clarification!

Felienne:
ok, weblate it merged, this is going in!

Felienne (approved):

PR #4663, opened 2023-10-24
[LANGUAGE] improved error productions for at

akseron:
Changes in error production in level 4 lark file and check_var_usage in main hedy file to support better syntax error messages for the 'at' command. 

Fixes https://github.com/hedyorg/hedy/issues/4780


translations/en/LC_MESSAGES/messages.po
```diff
@@ -145,6 +145,10 @@ msgstr ""
 "finish with a quotation mark. You forgot that for the text "
 "{unquotedtext}."
 
+msgid "Misspelled At Command"
```
Felienne, line 148: Hi @akseron!

It is not good practice to add to the po files directly, that will create problems. Instead, you have to add the new exception here: 

https://github.com/hedyorg/hedy/blob/5ba17c5c738ecb015e68b2e05dd77c5429dd6d0c/content/error-messages.txt#L10

And then run the extract script (see wiki) and the strings will be entered into the pot and po files automatically.

Felienne: See here: https://github.com/hedyorg/hedy/wiki/Hedy-Development-Process#adding-new-translation-keys

exceptions.py
```diff
@@ -233,6 +233,15 @@ def __init__(self, level, line_number, unquotedtext=None):
                          line_number=line_number)
 
 
+class MisspelledAtCommand(HedyException):
+    # TODO adjust the parameters, will only put this in en/messages.po
```
Felienne, line 237: Hi @akseron!

Is this something you want to do in this PR or can we merge this now and do this later on?

In any case, can you add a date and your initials to the TODO so we know who to ask in the future? :)

akseron: Oh I had forgotten to remove that, that was part of an earlier part of the code adjustment.
Will keep in mind to do so!

Felienne: Does that mean it can be removed? If so, please do :) 

I have fixed the po issues, so I think.... we might be ready to merge?!

akseron: Hahaha, I did want to ask about one more thing before that. In the testing in tests/test_level/test_level_04.py I had changed the parameter for skipping faulty code, however I'm not 100% sure that I did it in the right place. If that is correct it should be ready to merge :D

tests/test_level/test_level_04.py
```diff
@@ -572,6 +572,20 @@ def test_print_list_access_index_var(self):
...
         )
 
+    def test_list_access_misspelled_at_gives_error(self):
+        code = textwrap.dedent("""\
+        dieren is Hond, Kat, Kangoeroe
+        print dieren at random""")
+
+        self.multi_level_tester(
+            max_level=5,
+            # set to false for now, probably is incorrect so a look would be appreciated!
```
Felienne, line 582: This is correct actually, it disables a special setting we have for skipping error lines. The only thing is that I prefer False for false over 0.

tests/test_level/test_level_04.py
```diff
@@ -572,6 +572,20 @@ def test_print_list_access_index_var(self):
...
+            skip_faulty=0,
```
Felienne, line 583:
```suggestion
            skip_faulty=False,
```

Felienne:
Hi @akseron!

This can be merged if you update the PR description (the first message here) to describe briefly what you did and why (and whether there is an open issue related to this PR)

hedy.py
```diff
@@ -724,19 +724,22 @@ def equality_check(self, tree):
...
 
     def repeat(self, tree):
+        print("repeat")
         command = Command.repeat
         allowed_types = get_allowed_types(command, self.level)
         self.check_type_allowed(command, allowed_types, tree.children[0], tree.meta)
         return self.to_typed_tree(tree, HedyType.none)
 
     def for_list(self, tree):
+        print("for_list")
```
Felienne, line 734: I think we want to remove all these debugging prints before merging

Felienne (changes requested):

tests/test_level/test_level_04.py
```diff
@@ -572,6 +572,19 @@ def test_print_list_access_index_var(self):
             extra_check_function=check_in_list
         )
 
+    def test_list_access_misspelled_at_gives_error(self):
+        code = textwrap.dedent("""\
+        dieren is Hond, Kat, Kangoeroe
+        print dieren at random""")
```
Felienne, line 578: This should be "ad" I think? Not sure why this test was not failing, it should (since the error is not raised). I do get an error locally so maybe something is up with the tests?

tests/test_level/test_level_04.py
```diff
@@ -572,6 +572,19 @@ def test_print_list_access_index_var(self):
             extra_check_function=check_in_list
         )
 
+    def test_list_access_misspelled_at_gives_error(self):
+        code = textwrap.dedent("""\
+        dieren is Hond, Kat, Kangoeroe
```
Felienne, line 577:
```suggestion
        dieren is Hond, Kat, Kangoeroe, Kaketoe
```

Felienne (approved):
I removed the original comment documenting this issue from over 2 years ago:

image

With that, we are good to go!!

Felienne:
And congrats on getting your first contribution accepted @akseron! It was quite a ride, but you did it!!

PR #4666, opened 2023-10-25
Added "Back To Top" button

PrajwalGraj:
Added a "Back To Top" button at the end of Readme.md file which will act as a scroll button .When clicked it will redirect us to the top of the file

* [UI] -> improvements and changes of the Hedy user interface👾


Felienne (approved):
Ah nice addition! I wonder if we can do this in the Wiki too? Do you know @PrajwalGraj?

Felienne:
And welcome to Hedy! If you want to pick up other things, let us know!

PR #4677, opened 2023-10-27
[LANGUAGE] Convert calc to int in sleep level 12

### LINKED ISSUE #4672: [BUG] Improve error message
I found this error message, that I think can be better. 

image

Felienne:
Fixes #4672 

This PR does three things to improve weird error messages.

Firstly, it escapes the library name time, fixing error number three named in #4672:

image

It also converts sleep input to integers, so that you can use fractions (error number 1):

image

And it allows floats as input for sleep in level 12 and up, fixing error 2:

image

**How to test**
I added tests for all cases


Felienne:
Hi @ToniSkulj!

This change to the handling of sleep messes up one of your tests :( I have disabled it for now (since I don't exactly understand why it fails, it does not seem to see an inner error production?) Would be great if you can take a look, either in this PR or in a fresh one.

Thanks!

ToniSkulj:
> Would be great if you can take a look, either in this PR or in a fresh one.

Hi @Felienne! Of course! [bc2ec53](https://github.com/hedyorg/hedy/pull/4677/commits/bc2ec5361e7e9687a08d07c06154a5086e387b5e) fixes the test in level 16.

Essentially what happens is that when skipping error_list_access_at the transpiling is accidentally still correct.
Fixed it by checking if we skipped an error and if so we can manually raise an error now.

jpelay:
Hi @Felienne! It all works great, but since we are allowing float variables to be used with sleep, we probably want float literals to be allowed to

image

Felienne:
> Hi @Felienne! It all works great, but since we are allowing float variables to be used with sleep, we probably want float literals to be allowed to
> 
> image

Ah great catch, should be easy to allow. I will add that too in this PR!

Felienne:
Should be ready to go now @jpelay!

jpelay (approved):

PR #4686, opened 2023-10-30
[FIX] Fix error messages in debugger, and temporary fix for cypress tests

jpelay:
After merging #4465 some Cypress tests were failing. This was because an ad-hoc calcutation I did to figure out the current debugging line. It worked fine locally and even in the tests in the PR, but for some reason failed later on, this PR deabilates that check and I will work on it later this week.

Also, after testing in Alpha I found out I wasn't showing the error messages when an error ocurred during execution. 

How to test?
Go to level 3 and write this code:
```
l is 1, 2
print l at 3
```
Then debug it and when you reach the second line this error should be shown:

image

### DIFF
--- a/static/js/app.ts
+++ b/static/js/app.ts
@@ -1130,12 +1130,11 @@ export function runPythonProgram(this: any, code: string, sourceMap: any, hasTur
         if (cb) cb ();
       }
     ).catch(function(err: any) {
-      console.error(err)
       const errorMessage = errorMessageFromSkulptError(err) || null;
       if (!errorMessage) {
         throw null;
       }
-      throw new Error(errorMessage);
+      throw new Error(errorMessage);      
     });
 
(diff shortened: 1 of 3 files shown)


Felienne (approved):
Debugger still works locally, hope this fixes the alpha issue!

jpelay:
> Debugger still works locally, hope this fixes the alpha issue!

I think that perhaps the problem you mentioned to me was actually a matter of local cache! I sent the link to several friends and they said that it worked on their browsers.

Felienne:
> > Debugger still works locally, hope this fixes the alpha issue!
> 
> I think that perhaps the problem you mentioned to me was actually a matter of local cache! I sent the link to several friends and they said that it worked on their browsers.

I think you are right! Current alpha has no issues!
image

PR #4699, opened 2023-11-02
Check for use of = and is

MarleenGilsing:
Replaced a lot of {is} with = in example codes in all languages


Felienne (approved):

Felienne:
Let's hope this does not anger weblate!

PR #4702, opened 2023-11-02
[CONTENT] Teacher manual: multilangual mode, teacher statistics and after hedy

### LINKED ISSUE #4550: [CONTENT] Explain multilingual mode in teacher's manual
We don't explain anywhere that you can mix en keywords and your own language! ould be good to add to the teacher's manual. We can explicitly also add it as a way to transfer to Python after Hedy ("Don't forget to switch to English, you can do that in earlier levels too). Maybe that transfer should even be its own section!

    @Felienne  I have searched the source material on Weblate for the words "mix" and "keyword". There are "program in English, and also in your own language" (I shortened in meaning), but none gives any clue about "you can mix a keyword's language (English and your own) provided you keep their position" or something like that.

_Originally posted by @yilmazdurmaz in https://github.com/hedyorg/hedy/issues/4516#issuecomment-1733649189_
            

And (too lazy to open another issue, sorry!)

Another small issue that I noticed:

* "Per december 2021 is het verplicht op een voorkeurstaal te kiezen." -> I think we can safely remove that, all our new users know this by now.

### LINKED ISSUE #4569: [CONTENT] Add a new chapter in the teacher's manual "after Hedy"
This is more of a long-term issue for @MarleenGilsing, it would be nice to have a section in the teacher's manual called "after Hedy" where we explain in a not more depth how to go to Python, or even to something else.

MarleenGilsing:
fixes #4550 and fixes #4569 

@Felienne Could you check the 'After Hedy' section? I don't know if you have some more tips for the teachers who want to continue in Python?

Also, I saw that the teacher statistic is live now! Great! I will add some more explanation to the teacher manual. I have already replaced the old information with a COMING SOON. This way there is no old info on the website, and I could easily replace it in all languages at once (as long as the translators don't translate this part, so @yilmazdurmaz please don't translate this yet, i know how quick you are haha ;) ).


Felienne:
> @Felienne Could you check the 'After Hedy' section? I don't know if you have some more tips for the teachers who want to continue in Python?

For now, this is ok! 

> Also, I saw that the teacher statistic is live now! Great! I will add some more explanation to the teacher manual. I have already replaced the old information with a COMING SOON. This way there is no old info on the website, and I could easily replace it in all languages at once (as long as the translators don't translate this part, so @yilmazdurmaz please don't translate this yet, i know how quick you are haha ;) ).

It is a good idea to make an issue (so you don't forget) but it is a bit too early since it has some issues we still need to work on, so the interface might change!

Felienne (approved):

PR #4709, opened 2023-11-03
[UI] Go to tabs not to levels

hasan-sh:
See #4606

### DIFF
--- a/templates/incl/editor-and-output.html
+++ b/templates/incl/editor-and-output.html
@@ -200,7 +200,7 @@
             <button class="green-btn previous-tab ltr:ml-1 rtl:mr-1" 
               href="#"
               _="on load or click from .tab
-                set :els to getPreviousAndNext()
+                set :els to hedyApp.getPreviousAndNext()
                 if :els[0]
                   set *display to 'block'
                   set the *justify-content of #pagination_container to 'space-between' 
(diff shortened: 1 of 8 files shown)


Felienne:
Ow no @hasan-sh, @TiBiBa continued on his PR today also: #4652

Felienne:
Sorry for the double work, I will let @jpelay decide who we proceed with the two PRs (I know he is looking at #4652 now)

hasan-sh:
> Sorry for the double work, I will let @jpelay decide who we proceed with the two PRs (I know he is looking at #4652 now)


Good that I pushed this PR before talking to Annelein since she wanted to do the dropdown which Timon implemented already!! Don't mind going with his i think we should just stress minimizing the usage of jquery moving forward!

templates/incl/editor-and-output.html
```diff
@@ -196,46 +196,35 @@
...
+              href="#"
+              _="on load or click from .tab
+                set :els to getPreviousAndNext()
+                if :els[1]
+                  set *display to 'block'
+                  set nextContent to :els[1].innerHTML.trim()
+                else
+                  set *display to 'none'
+                end
+                put nextContent + ' &LongRightArrow;' into my innerHTML
```
jpelay, line 225: Can you change this `&LongRightArrow` for `&LongLeftArrow` in case we are in a RTL language, like Arab?

templates/incl/editor-and-output.html
```diff
@@ -330,3 +319,17 @@
 });
 </script>
 {% endif %}
+
+<script>
+  function getPreviousAndNext(el) {
```
jpelay, line 324: Can you take this code to `app.ts` or a similar file where this would conceptually belong? We strive to include as much code as possible inside our TypeScript files! I also like that you didn't use JQuery, but pure JavaScript instead, do you think this is the way moving forward?

tests/cypress/e2e/hedy_page/back_to_level_button.cy.js
```diff
@@ -2,15 +2,7 @@ import {goToHedyLevel2Page, goToHedyPage} from "../tools/navigation/nav";
...
-      goToHedyLevel2Page();
-      cy.get('#prev_level_button').click();
-      cy.url().should('include', Cypress.env('hedy_page'));
-
-      // Test when code is changed
-      goToHedyLevel2Page();
-      cy.get('#editor').type('hello');
-      cy.get('#prev_level_button').click();
-      cy.url().should('include', Cypress.env('hedy_page'));
+      // make sure to text navigating to next levels
```
jpelay, line 5: Can you add this a TODO comment with an explanation as to why we are removing this test? Some IDEs gather TODO comments and also add some context for someone who is just reading the code and doesn't have knowledge of this PR.

tests/cypress/e2e/hedy_page/to_level_button.cy.js
```diff
@@ -2,15 +2,7 @@ import {goToHedyPage} from "../tools/navigation/nav";
...
-      goToHedyPage();
-      cy.get('#next_level_button').click();
-      cy.url().should('include', Cypress.env('hedy_level2_page'));
-
-      // Test when code is changed
-      goToHedyPage();
-      cy.get('#editor').type('hello');
-      cy.get('#next_level_button').click();
-      cy.url().should('include', Cypress.env('hedy_level2_page'));
+      // make sure to text navigating to next levels
```
jpelay, line 5: Same here!

tests/cypress/e2e/hedy_page/to_text_tabs_button.cy.js
```diff
@@ -0,0 +1,23 @@
+import {goToHedyPage} from "../tools/navigation/nav";
+
+describe('Navigating through the tabs with the buttons', () => {
```
jpelay, line 3: Thanks for adding this one! :D

jpelay (changes requested):
Hello,  @hasan-sh! Thank you for this PR, it looks and works great. I really like that you used HTMX to implement the feature, which is why @Felienne and I decided to move forward with this PR, instead of #4652. We think implementing features with HTMX as much as possible is the way forward.

That being said, there's one thing that we need to take from #4652, and that is the level drowpdown, can you bring that feature to this PR? And then we will be ready to go :smile: There are also a couple of things that I think can be improved a little, nothing too serious.

Again, great work, and thanks!

templates/incl/editor-and-output.html
hasan-sh: Good one! Will add it

templates/incl/editor-and-output.html
hasan-sh: I think pure JS with the power of hyperscript and htmx is enough currently; until we reach a point we need something more powerful like React. Jquery has far more cons than pros and I really dislike it..

tests/cypress/e2e/hedy_page/back_to_level_button.cy.js
hasan-sh: Great will do!

Felienne:
This is now integrated into #4714

PR #4714, opened 2023-11-06
Hour-of-Code route

### LINKED ISSUE #4705: [CHORE] Add tracking pixel
We need to add a tracking pixel from Hour of Code (https://code.org/api/hour/begin_hedy_text.png) somewhere on the link we will create for them, info:

**Tracking Pixel**
In order to help us more accurately track participation in the Hour of Code, please include a 1-pixel tracking image on the first page of your Hour of Code activity. Place the pixel-image on the start page (if you have a landing page before the real start of your activity, place the pixel-image on the start page instead of the landing page). Please do not place the pixel on any other page. 

If an activity is only offered as an app download, or a lesson plan download (or link to a PDF or Google Doc), there is no need to include the tracking pixel and you can ignore this 

I am still double checking the link, will update later, but we can start to think about how/where to add it.

### LINKED ISSUE #4706: [UI idea] Create link based customization for hour of code
We want to create something like this link: https://hedy.org/hour-of-code/1?keyword_language=da&language=da

Which then shows a subset of adventures. 

Where to define the set? We could do it in the code in hedy_content.py for now? An alternative could be to add tags to adventures in the yaml, and load based on those. That is more flexible and future proof, but might require more work.

I think we now also really need to pick up #4606 that @TiBiBa was working on since people coming to Hedy from the outside will surely be confused by that legacy button!

hasan-sh:
Fixes #4706 and fixes #4705

Adventures subsets can now be passed through the route. For instance, the following route will give back the corresponding translated adventures of level 8: 
> /hour-of-code/8?**subset=fortune,turtle**&language=en

Also, the `keyword_language` and `language` parameters work as expected. 

@SabinaChita, this part is relevant for the newsletter:

**This PR also changes the button from "level X" to "next adventure name"**

### DIFF
--- a/app.py
+++ b/app.py
@@ -36,7 +36,7 @@ from safe_format import safe_format
 from config import config
 from website.flask_helpers import render_template, proper_tojson, JinjaCompatibleJsonProvider
 from hedy_content import (ADVENTURE_ORDER_PER_LEVEL, KEYWORDS_ADVENTURES, ALL_KEYWORD_LANGUAGES,
-                          ALL_LANGUAGES, COUNTRIES, HOC_ADVENTURE_PER_LEVEL)
+                          ALL_LANGUAGES, COUNTRIES, HOUR_OF_CODE_ADVENTURES)
 
 from logging_config import LOGGING_CONFIG
 from utils import dump_yaml_rt, is_debug_mode, load_yaml_rt, timems, version, strip_accents
@@ -1128,7 +1128,7 @@ def hour_of_code(level, program_id=None):
         loaded_program = Program.from_database_row(result)
 
     subset = [adv.strip() for adv in request.args.get("subset", "").strip().split(",")]
-    subset = subset if subset[0] else HOC_ADVENTURE_PER_LEVEL[level]
+    subset = subset if subset[0] else HOUR_OF_CODE_ADVENTURES[level]
     adventures = load_adventures_for_level(level, subset=subset)
 
     if not adventures:
--- a/hedy_content.py
+++ b/hedy_content.py
@@ -264,8 +264,7 @@ ADVENTURE_ORDER_PER_LEVEL = {
     ]
 }
 
-# Felienne, you can change what show here!
-HOC_ADVENTURE_PER_LEVEL = {
+HOUR_OF_CODE_ADVENTURES = {
     1: [
         'print_command',
         'parrot',


Felienne:
> Adventures subsets can now be passed through the route. For instance, the following route will give back the corresponding translated adventures of level 8:
> 
> > /hour-of-code/8?**subset=fortune,turtle**&language=en

Ah cool, that will allow for a lot of other customization as well (f.e. for a workshop!)
 
> Also, the `keyword_language` and `language` paramenters work as expected.

Great!

Felienne:
I think we can merge this for now, including the quiz, we can always remove that later-on.

Felienne (changes requested):
Hi @hasan-sh!

I just noticed we show the puzzle, not the quiz in the route:

image

The puzzles aren't the most awesome part of the UI at this point, they have a few annoying issues, so I don't want to showcase those at the moment. Can we swap the puzzle for the quiz? With that this should be good to go!

app.py
```diff
@@ -1100,6 +1103,187 @@ def teacher_tutorial(user):
...
+
+    commands = hedy.commands_per_level.get(level)
+    return render_template(
+        "code-page.html",
+        level_nr=str(level_number),
+        level=level_number,
+        current_page='Hour of Code',
+        prev_level=None,
+        next_level=None,
+        # prev_level=level_number - 1 if level_number > 1 else None,
```
Felienne, line 1257: I do think we want to have the button in this view too (once is has changed into "next adventure" where appropriate)

app.py
```diff
@@ -1100,6 +1103,187 @@ def teacher_tutorial(user):
...
+        # next_level=level_number + 1 if level_number < max_level else None,
+        # customizations=customizations,
+        hide_cheatsheet=hide_cheatsheet,
+        # enforce_developers_mode=enforce_developers_mode,
```
Felienne, line 1261: This way of disabling parameters by commenting them out is not what we commonly do (I think setting the variable to False occurs more often) I actually think what you do here is more clear, you have the information that it is not in use right in view, just want to point out that this is not what is used elsewhere (maybe it should be?)

app.py
hasan-sh: I will merge #4709 once it's done.

hasan-sh: But we don't really want them to navigate through next levels, do we?

app.py
hasan-sh: Overall in the codebase indeed that's not the case. For all PRs i worked on, I had to find out where the variables are passed to the view and where not, that made it harder to find why certain part of the view renders or doesn't. 

I think we should comment them out to inform other devs that these may be used, see https://jinja.palletsprojects.com/en/3.0.x/templates/#variables
> If a variable or attribute does not exist, you will get back an undefined value. What you can do with that kind of value depends on the application configuration: the default behavior is to evaluate to an empty string if printed or iterated over, and to fail for every other operation.


So, if we assign an unused variable to false and we iterate over it, the rendering will fail. Whereas it wouldn't if we didn't assign it at all.

Felienne: Great find, thanks! Shall we open an issue to make other renderings similar (not even for you, this is a great first issue for a newcomer too)

Felienne:
Great work team!! 

I love how we put the next name on the button:

image

(I envisioned just "next" but this is nicer)

Felienne:
I fixed a small issue in the tutorial, here was a similar problem in the hour-of-code:

image

So after we merge this, we really do need a cypress tests for the hour-of-code route too.

Felienne:
The only thing still needed I think is the tracking pixel (that @hasan-sh wanted to add to this PR)

hasan-sh:
> I fixed a small issue in the tutorial, here was a similar problem in the hour-of-code:
> 
> image
> So after we merge this, we really do need a cypress tests for the hour-of-code route too.

Great! I'll add tests today

hedy_content.py
```diff
@@ -264,6 +264,240 @@
     ]
 }
 
+# Felienne, you can change what show here!
```
Felienne, line 267: Perfect!

Felienne (approved):

PR #4724, opened 2023-11-08
[FIX] Fix the `filtered_programs_for_user` call

rix0rrr:
In #4644, the call signature of `filtered_programs_for_user` was changed to replace the function `public_programs_for_user`. The goal was to be able to supply the query `get_many({ "username": "henk", "public": 1 })`, with the goal to retrieve all public programs by user `henk`.

Unfortunately, there is no index on the pair `(username, public)`. Instead, we have indexes on `(username)` and on `(public, date)`.

The first is used to retrieve all programs by a single user, the second is used to retrieve all public programs (for the "Explore" page). To get the public programs for a single user, instead what we do is retrieve all programs for a single user, and filter them down in Python.

Because of a bug in the database layer, trying to retrieve programs by `{ "username": "henk", "public": 1 }` uses the `(public, date)` index. However, the database layer does not detect that in that particular query the field `date` is missing, and that the field `username` is unnecessarily passed. The query incorrectly succeeds in the local database emulation mode so this goes undetected during testing, and when passed to an actual DynamoDB database the server returns an error.

This currently breaks the profile page, as well as some other pages probably.

Add `public` as an optional keyword argument to `filtered_programs_for_user` and do the filtering locally instead.

**How to test**

The following pages should work:

* `/my-profile`
* `/programs`
* `/class/<class_id>/programs/<username>`

### DIFF
--- a/app.py
+++ b/app.py
@@ -954,7 +954,7 @@ def programs_page(user):
     filter = request.args.get('filter', default=None, type=str)
     submitted = True if filter == 'submitted' else None
 
-    result = DATABASE.filtered_programs_for_user({"username": from_user or username},
+    result = DATABASE.filtered_programs_for_user(from_user or username,
                                                  level=level,
                                                  adventure=adventure,
                                                  submitted=submitted,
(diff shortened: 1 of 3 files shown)


jpelay (approved):
Thanks for the quick fix Rico :smile:

jpelay:
Working correctly on Alpha :partying_face: 

image

PR #4732, opened 2023-11-09
[CONTENT] Adds teacher statistics to manual

MarleenGilsing:
This is brand new content, only in English, so it shouldn't upset weblate, but might upset the tests?


MarleenGilsing:
Yep.. It's mad that the english version contains one more header. But new content should be copied automatically from English, right @Felienne? Or should I copy-paste it to all languages anyway?

Felienne:
Hi @MarleenGilsing!

In this situation (adding to the manual), Weblate cannot add the content in all languages, so you will have to do that manually. The tests (somewhat) explain what this issue is:

image

Can you copy the tests over?

Thanks!

PR #4736, opened 2023-11-09
[CHORE] Protect against misquery in DynamoDB layer

rix0rrr:
PR #4644 introduced a refactoring that tried using a nonexistent index. That query failed when run against the actual DynamoDB server, but succeeded when run against the local emulation layer.

Introduce protection into the database layer to check that the keys we use to query are exactly what is expected, nothing more and nothing less.

**How to test**

If the tests pass, nothing has functionally changed.


jpelay (approved):
Thanks a lot, @rix0rrr!

PR #4738, opened 2023-11-10
[FIX] Use library to validate emails

### LINKED ISSUE #4726: [BUG] Can't make an account with an edu address
I guess our email verification is a bit too strict!

image

This is where the validation happens:
https://github.com/hedyorg/hedy/blob/4c202a1466b41d09883d4fa14d64f8718371250b/utils.py#L179

No clue what this exactly does (it is very old code without any explanation, sadly)

Maybe we can use this:
https://emailregex.com/

It is also unreadable but at least it has documentation that we can put in a comment :)

jpelay:
**Description**

Adds the library `email-validator` to validate emails. I think using a library is better than using a regex, since emails are hard to validate using regexes only, and is better to just use a library that we know will work in every case.

**Fixes #4726 **


Felienne:
Seems to block wrong email indeed!

image

Felienne (approved):

Issue #4718, opened 2023-11-08
[UI idea] Dropdown arrows


hasan-sh:
**Idea incl level**
Switch the arrows when a dropdown is opened.
![image](https://github.com/hedyorg/hedy/assets/20051470/30e3bf1a-7a6d-4e28-9fff-8fc8a6e8b649)

**Additional context**
- downwards arrow when closed
- upwards arrow when opened
- Please use hyperscript.

Felienne:
I don't think I understand what arrows this issue refers to, but if you know @Annelein, go for it!

Annelein:
> I don't think I understand what arrows this issue refers to, but if you know @Annelein, go for it!

Yes I do, the arrows inside the buttons that open up a dropdown menu (level dropdown, commands dropdown, language dropdown, cheatsheet dropdown), are always pointed downwards. It would be clean to have them pointed upwards if the dropdown button is pressed and the dropdown is opened!

PR #4745, opened 2023-11-13
[UI] Dropdown arrows

Annelein:
See #4718 

Switching arrows when a dropdown is opened:

Screenshot 2023-11-13 at 10 42 47

Screenshot 2023-11-13 at 10 43 11


hasan-sh (commented):
these files will be generated automatically so no need to commit them each time

hasan-sh (approved):

PR #4747, opened 2023-11-13
[UI] Remove tabs in developer mode

### LINKED ISSUE #4719: [UI idea] remove tabs when in programmer's mode
The adventure tabs don't really make sense in programmer's mode?

**Idea**
image

Steps to fix this would be to change the template to show the tabs only when we are not in programmer's mode.

Annelein:
Fixes #4719

### DIFF
--- a/templates/incl/adventure-tabs.html
+++ b/templates/incl/adventure-tabs.html
@@ -1,5 +1,8 @@
       {# TABS #}
-      <div id="adventures">
+      <div id="adventures"
+      _="if localStorage.getItem('developers_mode') is true
+            toggle .hidden on #adventures
+          end">
         {% macro adventure_tab(tabname, title, selected, id, tab_class, index) %}
             <div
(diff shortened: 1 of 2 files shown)


hasan-sh:
also, @Felienne shouldn't we make this persistent; e.g., same mode after refreshing the page?

jpelay:
> also, @Felienne shouldn't we make this persistent; e.g., same mode after refreshing the page?

We can store it in localStorage for it I think.

Felienne:
> See #4719

Hi @Annelein! Tip: If you writes "fixes" instead of "See" Github will close the issue automatically when we merge the PR. :)

Annelein:
> > also, @Felienne shouldn't we make this persistent; e.g., same mode after refreshing the page?
> 
> We can store it in localStorage for it I think.

How do I access that in Hyperscript?

jpelay:
> How do I access that in Hyperscript? @jpelay @hasan-sh

Seems like you figure it out!

PR #4757, opened 2023-11-15
[UI] Public adventures view

### LINKED ISSUE #4604: [UI idea] Add search adventure by tag
This issue follows #4603. From the proposal:

_"Secondly, we need to make the content searchable by the meta data and present teachers with a solid interface to search and recommend content. We will engage the community in enriching the content."_

Steps to take:
* Create an interface to search adventures, and merge this with the existing customize class feature (https://hedy.org/for-teachers/customize-class/)
* For now we at least want to add a filter for languages (such that Spanish teachers only find Spanish content if they want to) (*)
* Add name credits at the end of an adventure, so that people can see who created it (we have to think of a non-intrusive way to do this) 

(*)  I don't know if we currently store the language of adventures in the database! We will have to look into that!

hasan-sh:
Fixes #4604 

**Info**
- Public adventures are listed as cards  
- Each card contains basic information about the adventure.
  - creator, level, clones (# of times it's created), last update date, and tags
- These are sorted by currenUser's adventures 

**Features and Todo**
- Teachers can navigate to their adventures by clicking on the title or edit button.
- Teachers can clone or view adventures of others
  - if they clone, a new copy of that adventure is created
  - the user gets a notification that cloning succeeded
  - the update is reflected by showing the newly cloned adventure only thus 
  - the other teacher's adventure that's cloned now, won't be visible.
- Teachers can use 3 filters: 
  - Search 
  - Language 
  - Tags 
  - [ ] Country?
  - [ ] User?
  - [ ] Level?
- [ ] add tests
- [ ] allow filtering through url


Felienne:
I love the tile UI @hasan-sh, looks very good!

image

Felienne:
A few random thoughts:

Should we add "public" as a tickmark to the adventure overview:

image

(and maybe also show which ones are cloned?)

Felienne:
If I click clone, the tile changes, which I found a bit confusing:

image

What does the bin mean, will it delete the adventure or the clone? The error message does not help me grasp the meaning
image

I think we can do without the bin button here? If people want to remove the adventure, they can do so from their adventure overview? 

The same might be true for the add tags link, for which it is not clear if we are adding the tags to the original adventure or the close. 

Also it is a bit confusing that it leads to the same edit view as the title link. Maybe we can make one link at the bottom with something like "edit directly"?

hasan-sh:
> > A few random thoughts:
> > Should we add "public" as a tickmark to the adventure overview:
> > image
> > (and maybe also show which ones are cloned?)
> 
> Thought of that and put them in that table, but it got too crowded...

but the tables now have `cloned_from` and `cloned_times` (i.e. how many times is an adventure cloned!) for future ideas/usage.

Felienne:
> Thought of that and put them in that table, but it got too crowded...

Fair enough! We will think about a redesign of this page for the future and keep it in mind!
 
> but the tables now have `cloned_from` and `cloned_times` (i.e. how many times is an adventure cloned!) for future ideas/usage.

Oooowwww, good thinking!! We could add `cloned_times` to the tile view so teachers can see popular adventures?

hasan-sh:
> If I click clone, the tile changes, which I found a bit confusing:
> 
> image
> What does the bin mean, will it delete the adventure or the clone? The error message does not help me grasp the meaning image
> 
> I think we can do without the bin button here? If people want to remove the adventure, they can do so from their adventure overview?
> 
When a user clones an adventure, we create a brand new one for that user. I did this because we weren't sure whether cloned adventures should be syncronized. I think they shouldn't.


> The same might be true for the add tags link, for which it is not clear if we are adding the tags to the original adventure or the close.
the main reason why i added this is to entice users to add tags since it's a new feature and great for data analysis later on!!
> 
> Also it is a bit confusing that it leads to the same edit view as the title link. Maybe we can make one link at the bottom with something like "edit directly"?

This was really temporary! I could add a button instead of the delete btn, but perhaps we want to change and add more things to this public adventures view; e.g., add adventures to classes for one or multiple levels.

hasan-sh:
> Fair enough! We will think about a redesign of this page for the future and keep it in mind!

INDEED! For instance, a card that has some info icon when clicked, it would show more info and/or direct to different places; same for classes!

hasan-sh:
> Oooowwww, good thinking!! We could add `cloned_times` to the tile view so teachers can see popular adventures?

Yess sure! Shall i add it as a filter? or just sort adventures by how many times they're cloned in the backend?

Felienne:
> > Oooowwww, good thinking!! We could add `cloned_times` to the tile view so teachers can see popular adventures?
> 
> Yess sure! Shall i add it as a filter? or just sort adventures by how many times they're cloned in the backend?

Let's for now just put it as info on the tiles? We can do more if we have thought about it a bit more :)

Felienne:
> When a user clones an adventure, we create a brand new one for that user. I did this because we weren't sure whether cloned adventures should be syncronized. I think they shouldn't. 

I agree!

> So, the card changes because the adventure now belongs to the current user!

Ah I see. I understand the model now but I am not sure that makes sense for a user. Maybe we should have a modal that says: adventure cloned, do you want to go to edit? And then yes brings you to the edit page and no brings you back to search (closes the modal). 
  
> This was really temporary! I could add a button instead of the delete btn, but perhaps we want to change and add more things to this public adventures view; e.g., add adventures to classes for one or multiple levels.

Ok, never mind my comments about the bin then. Indeed, there could be a button with more options (later)

templates/public-adventures.html
```diff
@@ -0,0 +1,62 @@
...
+        justify-around bg-blue-400 shadow-md rounded-lg p-4 mb-4"
+        _="on load set $filtered to []">
+        <div>{{_('apply_filters')}}</div>
+        <div id="search" class="flex-1">
+            <input id="search_adventure"
+                type="search"
+                placeholder="{{_('search')}}"
+                class="border text-black border-gray-500 rounded h-8 px-2 w-full"
+                _="on keyup debounced at 500ms set searchTerm to my value 
+                    hedyApp.applyFilter(searchTerm, 'search', $filtered)">
```
jpelay, line 17: Can we apply the same UI in this filter as the one we use to filter programs in the /explore and /programs page? We are trying to be a bit more consistent with our UX.

templates/public-adventures.html
```diff
@@ -0,0 +1,62 @@
...
+                {{_('no_adventures_yet')}}
+                <button class="green-btn mt-4" id="create_adventure_button"
+                    onclick='hedyApp.create_adventure({{_('adventure_prompt')|default(None)|tojson}})'
+                    {% if second_teacher and role=='viewer' %}disabled{% endif %}>{{_('create_adventure')}}</button>
+            </div>
+        {%endif%}
+        {% for adventure in adventures %}
+            {{ render_partial('htmx-adventure-card.html', user=user, adventure=adventure) }}
+        {% endfor %}
+    </div>
```
jpelay, line 60: Maybe the same can be said about these cards? I like them! But might be worth it to add a UI similar to the one we use for programs? You can find that code here: https://github.com/hedyorg/hedy/blob/pub-adventures/templates/macros/public-programs.html

templates/htmx-adventure-card.html
```diff
@@ -0,0 +1,44 @@
+<div class="adventure grid bg-white shadow-md rounded-lg transition-shadow hover:shadow-lg outline-blue-400 hover:outline hover:outline-2"
+    data-lang="{{adventure.language or current_language().lang}}"
+    _="on removeMe remove me">
+    <div class="py-4 px-6">
+        {% if adventure["creator"] == user["username"] %}
+        <h2 class="name underline cursor-pointer"
+            _="on click window.open('/for-teachers/customize-adventure/{{adventure.id}}', '_self')">{{adventure.name}}
```
jpelay, line 7: Regarding our discussion yesterday I think this is the kind of stuff where Hyperscript shines. Simple, short scripts that don't manipulate state.

website/database.py
```diff
@@ -585,6 +585,9 @@ def batch_get_adventures(self, adventure_ids):
         keys = {id: {"id": id} for id in adventure_ids}
         return ADVENTURES.batch_get(keys) if keys else {}
 
+    def get_public_adventures(self):
```
jpelay, line 588: Let's not forget to add this table to prod and alpha when this gets deployed

jpelay:
Hi! Just left a few comments regarding the code :D I like the way this is looking so far. Great idea!

templates/public-adventures.html
hasan-sh: I don't like tables and there there's tables. In the public adventures view, we wanna make new design, enhance and generalize it incrementally.

website/database.py
hasan-sh: This is the adventures table which exists on prod, right? Or does that have to be changed when we change its schema? Because I did include a new index to the table!

templates/public-adventures.html
jpelay: Yes, I understand. But the card themselves, which is the file I linked, are just a div.

hasan-sh: I don't really understand, but we will surely make changes on this page moving forward, so we can discuss it more later ;)

jpelay (approved):
We decided to merge this to main and use the styles for replacing the old styles!

jpelay:
@hasan-sh Do you know why so many Cypress tests are failing in this PR? A bunch of stuff that doesn't even has to do with your changes, so maybe there was an error when merging?

Felienne:
Yeah hopefully the new Cypress Cloud access will help us to figure out what is going on!

PR #4759, opened 2023-11-16
[UI idea] Tags dropdown

### LINKED ISSUE #4744: [UI idea] add dropdown to show existing tags
related to #4688 

- [ ] show a dropdown from which the teacher can select existing tags.

hasan-sh:
Fixes #4744 
**Structure**
- currently, the dropdown displays all tags that:
  - if used by adventures, at least one adventure must be public,
    - This ensures that if a teacher uses a tag in private adventures only, this tag won't be visible to other teachers.
  - are not used by any adventure 
    - (what tag would that be? e.g., a tag is created but then untagged from all adventures that used it)

**How to test?**  
- got to `/customize-adventure/[id]` view
- try adding/removing tags

**Archive**
- the choices we had:
   - get a tag iff all adventures that use it are public
   - get a tag iff one or more adventures that use it are public
   > what would happen if a tag isn't used by any adventure?
  - hence the choice!
  - any other thoughts are welcome.


Felienne:
> the choices we had:
> * get a tag iff all adventures that use it are public
> * get a tag iff one or more adventures that use it are public
> 
> > what would happen if a tag isn't used by any adventure?
> 
> * hence the choice!
> * any other thoughts are welcome.

Thanks for explaining your considerations @hasan-sh! It is a bit hard to see what choice you did pick, can you elaborate a bit? I think I would lean to showing all tags  iff one or more adventures that use it are public, and if you select a tag that does not exist for your level/language, simply show: "no adventures found"?

Felienne:
I don't think I know how to test this, I don't seem to be able to get to the search page from the UI in this branch:

image

hasan-sh:
> It is a bit hard to see what choice you did pick, can you elaborate a bit? I think I would lean to showing all tags iff one or more adventures that use it are public,

Currently, i show all tags regardless of the adventures that use them. I'll edit it to show tags that have at least one adventure public.

Also, if a tag isn't used by any adventure, I'll consider it as a public tag; so that it can be reused!

Felienne:
Ah got it now, this is about showing tags for existing adventures! (Somehow I thought we were talking about *searching* with tags)

Looks good!

image

Felienne:
> Currently, i show all tags regardless of the adventures that use them.

Now that I understand what we are doing, I think this is actually fine. I don't see a reason to not show all tags!

templates/htmx-tags-list.html
```diff
@@ -32,5 +33,5 @@
             value="{{tag}}"
             > -->
     {% endfor %}
-    <li id="no-tags" class='border-2 rounded-lg border-green-400 p-2 px-5 mx-2 mb-2 {% if tags|length > 0 %}hidden{% endif %}'>{{_('no_tags')}}</li>
+    <!-- <li id="no-tags" class='border-2 rounded-lg border-green-400 p-2 px-5 mx-2 mb-2 {% if tags|length > 0 %}hidden{% endif %}'>{{_('no_tags')}}</li> -->
```
jpelay, line 36: We probably want to remove this comment! And the one above it too

jpelay:
I think the UI looks pretty sleek! However, I'd like if for the dropdown you did the same as when you've already added the tags, because when there are many tags it overflows:

image

templates/htmx-tags-dropdown.html
```diff
@@ -0,0 +1,12 @@
+<!-- Requires #search_tags_input to exist -->
+<div id="tags_dropdown" class="dropdown-menu dropdown-blue z-20 min-w-full !border-gray-400"
+    style="display: none; opacity: 0; border: none; top: 100%; padding: 0px !important; margin: 0px; margin-top: 0.25rem;"
```
jpelay, line 3: I wonder if by this point isn't it better to make this into a class? Or use tailwind?

Felienne: Yes, I would prefer that too, also to make it easier in the future to have a more similar layout in other places.

Felienne: Apart from that, I think we are good to go!

hasan-sh: i agree, maybe in a separate issue since there are multiple dropdowns

Felienne: Fair enough! Can you open an issue so we don't forget?

jpelay:
The dropdown looks perfect now! Great work, Hasan!
image


I have a question about functionality: what happens after there are dozens, or even  hundreds of public tags created. All of those will be shown in the dropdown?

hasan-sh:
> The dropdown looks perfect now! Great work, Hasan! image
> 
> I have a question about functionality: what happens after there are dozens, or even hundreds of public tags created. All of those will be shown in the dropdown?

Good point and that's correct, all public tags are listed. We could probably add a limit later on! I think it's good to have this published asap, so that teachers begin tagging adventures!

Felienne:
> Good point and that's correct, all public tags are listed. We could probably add a limit later on! I think it's good to have this published asap, so that teachers begin tagging adventures!

Yes! About that, I added this to the newsletter board so @SabinaChita has it on her radar to explain to teachers.

jpelay:
> Good point and that's correct, all public tags are listed. We could probably add a limit later on! I think it's good to have this published asap, so that teachers begin tagging adventures!

Awesome, and thanks for the reply! Do you think we need to open an issue so we don't forget? I just approved so it will go to Alpha soon :smile:

Felienne:
> > Good point and that's correct, all public tags are listed. We could probably add a limit later on! I think it's good to have this published asap, so that teachers begin tagging adventures!
> 
> Awesome, and thanks for the reply! Do you think we need to open an issue so we don't forget? I just approved so it will go to Alpha soon 😄

Yes, let's do that! And I guess we also still need a table on beta?

jpelay:
> Yes, let's do that! And I guess we also still need a table on beta?

I think so, I created it only for Alpha. What is beta? I think I've never accessed it.

PR #4763, opened 2023-11-16
[CONTENT] Replacing the old teacher statistics section from all languages

Felienne:
Redo of #4733

Depends-On: #4735


Felienne:
Hi @jpelay!

As you can see, Marleen is adding this to the teacher manual, so maybe the optimization of this could be moved forward in time a bit? I worry teachers will start using it while it is not really... battle ready?

jpelay:
> Hi @jpelay!
> 
> As you can see, Marleen is adding this to the teacher manual, so maybe the optimization of this could be moved forward in time a bit? I worry teachers will start using it while it is not really... battle ready?

Understood!!!

Felienne:
> > Hi @jpelay!
> > As you can see, Marleen is adding this to the teacher manual, so maybe the optimization of this could be moved forward in time a bit? I worry teachers will start using it while it is not really... battle ready?
> 
> Understood!!!

You're the best!! (We will wait a bit with deploying anyway because of CodeMirror I guess)

rix0rrr (approved):

PR #4766, opened 2023-11-16
[FIX] Proper marking of flat compound statements while debugging

### LINKED ISSUE #4725: [BUG] Debugger not coloring correctly flat ifs
**Describe the bug**
Sometimes when coloring flat statatements that spans two lines, the debugger doesn't color them correctly.

In this level 5 code the conditonal part (`if favorite_color in pretty_colors `) is not colored 
```
pretty_colors is green, yellow
favorite_color is ask 'What is your favorite color?'
if favorite_color in pretty_colors print 'pretty!'
else print 'meh'
```

This a note to self, will work on this tomorrow

jpelay:
**Description**

Changes the regex that was used to group the different parts of the flat compound statements to be able to properly color nested flat statements. Also changes the way the line is computed.

A nested flat statement are `if` with `repeat` inside or viceversa. You cannot place an `if` inside a `if`, but you can place a 
`repeat` inside a `repeat`.

Even though we say these statements are flat, it's possible to add newlines between them. And this is where the problem lied. When these statements where split between lines, the code did not account for beginning and end of statements properly.

**Fixes #4725 **

**How to test**

I added a couple of tests!


Felienne (approved):
Fixes the problem!

image

(hope the broken Cypress test will be fixed with an update form main)

tests/cypress/e2e/hedy_page/debugger.cy.js
```diff
@@ -452,7 +451,7 @@ describe('Test editor box functionality', () => {
     cy.focused().type("repeat 1 times if x is x print 'a'\nelse print 'b'");
 
     cy.get('#debug_button').click();
-    cy.wait('@parse');
+    cy.wait('@parse')
```
jpelay, line 454: I removed a semicolon and they magically worked again 😐

Issue #4501, opened 2023-09-16
[WEBLATE] Install plugin for line length


Felienne:
So I have the answer as to why this keeps happening: the way Weblate and gettext calculate the length of the string is fundamentally different, one uses byte lenght and the other character length, there is an [issue for that](https://github.com/WeblateOrg/weblate/issues/6350) on the Weblate repository that is marked as wontfix, so no matter what width we choose, there will always be a difference in the way the lines are wrapped.
  
            So I have the answer as to why this keeps happening: the way Weblate and gettext calculate the length of the string is fundamentally different, one uses byte lenght and the other character length, there is an [issue for that](https://github.com/WeblateOrg/weblate/issues/6350) on the Weblate repository that is marked as wontfix, so no matter what width we choose, there will always be a difference in the way the lines are wrapped. 

The way I see we can fix this is by installing a [plug-in to Weblate](https://docs.weblate.org/en/latest/admin/addons.html#customize-gettext-output) and set the option to not wrap lines ([apparently someone had problems](https://github.com/WeblateOrg/weblate/issues/6350) with that too, but I'm not sure if they messed up something or if its actually an issue, still they marked it as wontfix), but as it concerns to this particular PR I don't think there's a way to prevent the changes done to the line spacing, at least in the state that main is right now, perhaps after applying the plug-in and if the lines aren't wrapped I can apply the same option here and there won't be all of those differences.

What would you like to do? Can I install the plug-in to Weblate or are there other considerations?

_Originally posted by @jpelay in https://github.com/hedyorg/hedy/pull/4492#discussion_r1327250459_

Felienne:
Hi @Mark-Giesen!

I am not sure this is easy to do, but since you are working on Weblate related things at the moment, is this something you could take a look at? (if you don't have time, I think we will close this since no one really has time to dive in and the issue is small)

Mark-Giesen:
> I am not sure this is easy to do, but since you are working on Weblate related things at the moment, is this something you could take a look at? (if you don't have time, I think we will close this since no one really has time to dive in and the issue is small)

Installing the add-on and setting it to --no-wrap is easy. I'm not sure if this is what is meant here en will solve the problem? Maybe some live discussing and testing would be an idea? Not the coming 2 weeks though.

Felienne:
[...]

PR #4767, opened 2023-11-16
[CHORE] Unwrap Babel translations files

jpelay:
**Description**

A while ago we opened #4501 to add an add-on to Weblate so it wouldn't wrap files, we also changed the documentation to specify a `--no-wrap` option when extracting the labels. However, we are still running into problems where it seems like the files are actually wrapped. 

Looking into the git blame for the files, it seems like some lines haven't been changed in months, so maybe we actually need to change them manually by unwrapping all of them, and then send them to Weblate and see what happens?


Felienne:
Hi @jpelay!

I am looking into the addons in Weblate, and it seems we only have no-wrapping for yamls!

image

As far as I can see, there is no addon for po files outputting?

Mark-Giesen:
@Felienne, you're looking at Adventures, which is a YAML component. Text is the PO component and this one does have a PO add-on set to no-wrapping. I'm not sure if unwrapping is triggered and if so when. We could try changing anything in a file and see if the complete file is unwrapped. If not, we could change all wrapped files by deleting one /n and see if Weblate does the rest.

image

Felienne:
> @Felienne, you're looking at Adventures, which is a YAML component. Text is the PO component and this one does have a PO add-on set to no-wrapping. I'm not sure if unwrapping is triggered and if so when. We could try changing anything in a file and see if the complete file is unwrapped. If not, we could change all wrapped files by deleting one /n and see if Weblate does the rest.
> 
> image

Ah thanks @Mark-Giesen!! So it is really weird that the lines are now not wrapped in main.

Could it be that pre-commit does that? I am not sure whether that only works on py files?

Mark-Giesen:
The PO add-on is set to no-wrapping, but I seem to remember I set it to "wrap on new-lines". This is the --no-wrap behaviour. Maybe someone changed it, or I remember incorrectly.

Felienne:
> > Ah thanks @Mark-Giesen!! So it is really weird that the lines are now not wrapped in main.
> 
> They ARE wrapped, as I understand @jpelay. Lines that are old (not changed recently) are still wrapped. I assume these were wrapped in the passed and will only be unwrapped on a change.

Ah yeah sorry I meant weird that they are *wrapped*. So probably this PR will fix the issues (but we might want to wait until @Annelein's change is merged?)

jpelay:
> Hi @jpelay!
> 
> I have merged all weblate PR's so we can do this again now (assuming you did it with a script and not manually haha)

Done!

Felienne (approved):
Hail mary that this will work and not break things 🤞!

PR #4768, opened 2023-11-16
[LANGUAGE, FIX] Translates runtime exceptions

### LINKED ISSUE #4676: [LOCAL] Run-time messages are always English!
Working on #4675 made me realize our run-time error are not localized:

image

Now that we have added a language parameter in some places (esp. in the color) this is not so hard to fix!

jpelay:
**Description**

Adds translations labels for runtime exceptions, specifically for when a wrong value is supplied to the command. Also, adds labels for the data type of suggestions: one for numbers, and one for color.

**Fixes #4676**

**How to test**


Felienne:
Love this change, thanks @jpelay! Can't believe I never thought of this before (I guess it was one small message, and then another, and then another etc...)

hedy.py
```diff
@@ -186,6 +186,22 @@ def _translate_index_error(code, list_name):
         """)
 
 
+def translate_value_error(command, value, suggestion_type):
```
Felienne, line 189: We might want to call this translate_runtime_error? I think that is a bit more clear

Felienne (approved):

Felienne:
ow no.... there are now issues in tests because of my prior work on the int/sleep stuff. Can you fix them to your new system? Otherwise it is good to go!

PR #4777, opened 2023-11-21
[UI] Improvements to Live Statistics Page

jpelay:
**Description**

The Live Statistics page is very slow and its UI can be improved.

With respect to the speed we are going to show errors per level, and also we will cache the errors in the database. This way we won't have to grab so much data from the database.

With respect to the UI. We are going to merge the Overview of Adventures view with this page, as well as improving the graphs that are shown to the teachers.


jpelay:
Already integrated the adventures overview table to the Live stats page :smile: 
image

jpelay:
Superseded by #5130

PR #4783, opened 2023-11-21
[CHORE] Extract, update and compile Babel files on merge to main

### LINKED ISSUE #4774: [CHORE] add pybabel extract on each PR and commit po(t) files to main
In recent PRs, often stuff changes in the po files in weird ways (related to other, recently merged PRs). See for example: https://github.com/hedyorg/hedy/pull/4711#discussion_r1399202257

My suspicion is that happens because people forget to run extract and then merge an outdated pot file. Then on a next PR, where someone *does* run extract, the keys disappear.

We don't run pybabel extract on each PR, nor in the js -> ts runner, see:
https://github.com/hedyorg/hedy/blob/307a3f32fb7411f67b588b3c48f7b2c645ea32ee/.github/workflows/update-javascript-on-main.yml#L47

In fact, the reason that we do pybabel compile is for the front-end, not to keep the po files in sync! 

I think we should add the extract scripts here (and then maybe also rename the script so it is more clear that that is what it does). Alternatively we could write a second script that deals with the babel stuff, which is a bit more explicit.

jpelay:
**Description**

Adds 2 more command to the Babel GitHub action. Worth noting that in this commands adds the option `--no-wrap` that isn't specified on the wiki to avoid merge problems

**Fixes #4774 **


Felienne:
> **Description**
> 
> Adds 2 more command to the Babel GitHub action. Worth noting that in this commands adds the option `--no-wrap` that isn't specified on the wiki to avoid merge problems
> 
> **Fixes #4774 **

I must admit I lost track of the path forward a bit with the wrapping/ not wrapping, sorry!

Let's document here. Is it this?

* We update the wiki now to avoid immediate problems
* We unwrap all files as done in #4767
* We update the Wiki again to not wrap?

hasan-sh:
On the one hand, with wrapping it's easier to read in the yml files. With no wrapping on the other hand is more structured since you don't split lines such as:
> "The name {variable} needs to be set before you can use it on the right-"
"hand side of the {is} command."

The previous one may be read as after right- there's a new line, which isn't the case! Hence a potential important detail we might need to document is that devs should not rely on whether the keys have new lines or not. For instance, in the survey pr, annelein tried at first to split questions on \n (which made sense), but that wouldn't work unless the \n is embeded within the translation key. Annelein changed it now to have separate questions for the survey, but I'm mentioning it for future reference.

hasan-sh:
Also whatever we choose, existing PRs must rerun with the chosen approach; e.g., I now run it with wrapping (so i don't supply --no-wrap).

jpelay:
> * We update the wiki now to avoid immediate problems
> * We unwrap all files as done in [[CHORE] Unwrap Babel translations files #4767](https://github.com/hedyorg/hedy/pull/4767)
> * We update the Wiki again to not wrap?

I think we should always unwrap files, because that was what was causing many merge conflicts between us and Weblate. If we unwrap the files and don't tell people to unwrap them, won't that cause their po files to be wrapped and then that onto itslef cause more conflicts?

Felienne:
> > * We update the wiki now to avoid immediate problems
> > * We unwrap all files as done in [[CHORE] Unwrap Babel translations files #4767](https://github.com/hedyorg/hedy/pull/4767)
> > * We update the Wiki again to not wrap?
> 
> I think we should always unwrap files, because that was what was causing many merge conflicts between us and Weblate. If we unwrap the files and don't tell people to unwrap them, won't that cause their po files to be wrapped and then that onto itslef cause more conflicts?

Yes, totally agree, but we have now recently removed no-wrap for the wiki, right. So I just want to remember that we need to put it back after we have unwrapped.

Felienne (approved):

PR #4786, opened 2023-11-22
[LANGUAGE] improve error message for unquoted string with comma

### LINKED ISSUE #3967: [BUG] confusing error message in level 4 for unmatched quotes
**Describe the bug**

In level 4 if you miss a closing quote:

```
print 'Hello world
```

you get a helpful error that says:

```
Be careful. If you `ask` or `print` something, the text should start and finish with a quotation mark.
You forgot that for the text 'Hello world.
```

However, if the unterminated string contains a comma, then the error message is confusing:

```
The code you entered is not valid Hedy code. There is a mistake on line 1, at position 13.
You typed a comma, but that is not allowed.
```

Level 4:

```
print 'Hello, world
```

(https://hedycode.com/hedy/6978271c818f42d9b6021f670ee47e2b/view)

It seems to me that the error message should be the same as in the first case, where there's no comma. If you agree, I'm willing to try to fix this. Or this there a fundamental reason why we can't handle the two cases the same?

Felienne:
Fixes #3967

**How to test**

Check the error from  #3967 is improved (I have also added tests)


jpelay (commented):
Hi Felienne! I like that we are working on our error messages a bit!

I tested the PR, and it does indeed solve the issue, however, maybe we should account for this case too? It's a similar type of error, although, granted, the syntax highlighter already hints that this is not correct code by not coloring the string purple.


image

Felienne:
> Hi Felienne! I like that we are working on our error messages a bit!
> 
> I tested the PR, and it does indeed solve the issue, however, maybe we should account for this case too? It's a similar type of error, although, granted, the syntax highlighter already hints that this is not correct code by not coloring the string purple.
> 
> image

Ah nice idea! Let me see if I can incorporate that into this PR!

Felienne:
Done!

jpelay (approved):
Works great!

PR #4711, opened 2023-11-03
[UI] Dynamic Survey View - Ask teacher for class info #4612

Annelein:
Fixes #4612 

Survey view:
- Takes a survey_id, a description and questions
- 'Remind me later' sets 'skip' in survey to todays date, so you could modify the survey to pop up tomorrow/next week
- 'Don't ask again' sets 'skip' in survey to True, so you could choose to never show the survey again
- Option to only show unanswered questions

Survey table:
- Survey_id is: `survey_name + '_' + response_id` (for ex for the class survey: 'class_*class_id*')
- Responses
- Skip option

Screenshot 2023-11-16 at 15 50 06

PR #4787, opened 2023-11-23
[FIX] Remove indexes from surveys table

jpelay:
**Description**

Removes indexes from the SURVEY table, because it was causing problems when creating the table on Alpha, specially when updating the `responses` field. I think this is due to the fact that indexes can't be arrays.


rix0rrr:
Hi Jesus,

Can you explain a bit more what the problem was? What was the error you were seeing, and what was the operation this happened in? 

Is this table new, or is it old?

rix0rrr:
Aha here I have it:

```
2023-11-23T09:18:46.011400+00:00 app[web.1]: botocore.exceptions.ParamValidationError: Parameter validation failed:
2023-11-23T09:18:46.011400+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 1), value: 1, type: <class 'int'>, valid types: <class 'str'>
2023-11-23T09:18:46.011400+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 2), value: 2, type: <class 'int'>, valid types: <class 'str'>
2023-11-23T09:18:46.011401+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 3), value: 3, type: <class 'int'>, valid types: <class 'str'>
2023-11-23T09:18:46.011401+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 4), value: 4, type: <class 'int'>, valid types: <class 'str'>
```

rix0rrr:
The issue might be that in Dynamo the index field is declared as being of type `string`, while the  data sent into the record is of type `int` ?

But you are right, according to [this page](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SecondaryIndexes.html) the indexed field has to be a `string`, `int` or `binary`, not an array.

hasan-sh:
> The issue might be that in Dynamo the index field is declared as being of type `string`, while the data sent into the record is of type `int` ?
> 
> But you are right, according to [this page](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SecondaryIndexes.html) the indexed field has to be a `string`, `int` or `binary`, not an array.

True that. The indexes that were added in https://github.com/hedyorg/hedy/pull/4711/ aren't really used (see [here](https://github.com/hedyorg/hedy/blob/main/website/database.py#L676-L689)) so removing them won't cause any issue.

hasan-sh:
@jpelay the cypress tests are fixed by https://github.com/hedyorg/hedy/pull/4790

jpelay:
> Aha here I have it:
> 
> ```
> 2023-11-23T09:18:46.011400+00:00 app[web.1]: botocore.exceptions.ParamValidationError: Parameter validation failed:
> 2023-11-23T09:18:46.011400+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 1), value: 1, type: <class 'int'>, valid types: <class 'str'>
> 2023-11-23T09:18:46.011400+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 2), value: 2, type: <class 'int'>, valid types: <class 'str'>
> 2023-11-23T09:18:46.011401+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 3), value: 3, type: <class 'int'>, valid types: <class 'str'>
> 2023-11-23T09:18:46.011401+00:00 app[web.1]: Invalid type for parameter AttributeUpdates.responses.Value.M (key: 4), value: 4, type: <class 'int'>, valid types: <class 'str'>
> ```

Yes that was the problem I faced! I made the field str since the documentation of the table said `str []`, but I read on the Dynamo documentation that you can't do array fields as indexes. So I removed the indexes altogether, in any case they're not used so is better to remove them!

jpelay:
> @jpelay the cypress tests are fixed by #4790

Thanks! I merged main so the tests hopefully will pass. If you want, approve this PR so we can start testing this feature on Alpha. 

Btw, I also created the Tags table and tested a bit on Alpha too, if you want you can also test a bit there.

Felienne:
It seems a new issue arose now in the Cypress tests related to the second teacher feature:

image

Can you take a look @hasan-sh? Thanks!

Felienne:
(otherwise I think this can be merged?)

Felienne (approved):

PR #4790, opened 2023-11-23
[FIX] hide survey modal whenever navigating to a class

hasan-sh:
Fixes the issues of not finding an element because of hte survey modal.


Felienne (approved):
Thanks!! 

Hope this fixes the issues in the open Weblate and survey PR!

PR #4791, opened 2023-11-23
[CONTENT] New turtle content for level 1-15

MarleenGilsing:
New content for level 1-5 and level 8-9.


MarleenGilsing:
@Felienne Hmm... Very angry tests and an error message I dont understand. Would you be able to take a look before next week?

Felienne:
> @Felienne Hmm... Very angry tests and an error message I dont understand. Would you be able to take a look before next week?

Weird! I will take a look, I am sure we can fix this before next week.

Felienne:
Hi @MarleenGilsing!

Ok, the error message is a bit cryptic, but the problem is you added the new adventure to hedy_content in level 6:

image

But not in the yaml file:

image

Therefore a key error was giving! I have fixed this now, but if you see this error again, now you know what it might be :)

Felienne:
I played around with some HTML and markdown but I could not find a solution :( I think the markdown renderer somehow has a feature that automatically inserts a newline but I don't know where that code is and how it works. Maybe @TiBiBa knows?

TiBiBa:
> I played around with some HTML and markdown to solve ##4792 but I could not find a solution :( I think the markdown renderer somehow has a feature that automatically inserts a newline but I don't know where that code is and how it works. Maybe @TiBiBa knows?

If I remember correctly the markdowns are rendered directly into HTML. As each image as  a `width='100'` attribute it will always take the whole row (and force the next image on a new line). I think it would be easiest to place the images into a container that enforces a row layout like this:

```
<div class="w-full flex flex-row gap-2">
  // Place the images here
</div>
```

However, this makes the markdown files more complex and adds the need for contributors to know HTML before they can contribute to our content. But, as the HTML image tags already add some complexity as well I think this is the best solution. In the long run it would be best to separate the images from the markdown files and include them dynamically somehow. As these are language-independent they shouldn't be here as it will be confusing for translators and we will store the same identical tag for each language.

Felienne:
Thanks @TiBiBa!

This indeed does the jobs:

image

It does make the markdown a bit more complex (so we can consider if it is worth it?) but we can always add a note in the Wiki explaining translators how to deal with this!

MarleenGilsing:
@Felienne (or maybe @hasan-sh or @jpelay can help me in case Felienne is too busy in Australia)

I'm having some trouble with the tests again.... It now tells me that two of my code snippets are wrong, which they are not in my view. 

It says that level 10 example 2 is wrong because you have to use quotation marks in the list, but this is only from level 12 and up, not level 10. So I don't understand what it wants... 

The other example is example 2 in level 13. It says I can't use define. I don't know why not though... 

Do you think the tests are mad it is because it's a second example? Eventhough we use them more often in other adventures?

If someone could take a look and maybe merge it that would be great, because the longer we wait the more conflicts appear with the fast translators we have ;) Thank you!

Felienne:
Hi @MarleenGilsing !

This time it was not us haha, you forgot a `_` in the code, and then Hedy cannot run it. You can also see this if you try the examples:

image

Although I will admit the error for define is not great:

image

(I will make a separate issue for that!)

.github/workflows/update-js-and-babel-on-main.yml
```diff
@@ -42,8 +42,10 @@ jobs:
 
     # This is necessary for the TypeScript to contain translations
     # from Gettex.
-    - name: Compiling Babel translations
+    - name: Update and compile Babel translations
       run: |
+        pybabel extract -F babel.cfg -o messages.pot . --no-location --sort-output
```
Felienne, line 47:
```suggestion
```

PR #4795, opened 2023-11-23
Made a startcodes adventure and tried to incorporate it on the slides.

### LINKED ISSUE #4794: [BUG] I'd like to be able to use startcodes in slides that aren't in hedy_content
**Describe the bug**
When making slides I can only incorporate codes that are start_codes in the adventures. I would like to be able to use more different codes, so I came up with the idea to make a mock-adventure, that is only there for startcodes. This adventure should not be on the website, so it should not be in hedy_content.py. But, the problem is, that as long as the adventure is not in hedy_content the slides will show this image:

image

Felienne said this should only take one line of code to fix. Can somebody do that for me please? :)

I case you don't understand what I mean, I made a draft PR of the mock-adventure and a level 19 slide that would use the startcode of that adventure. #4795 

image

MarleenGilsing:
Adds new start codes for slides only

Also fixes #4794


MarleenGilsing:
I have made content upto level 7 now. This seems ready to go @Felienne 

I haven't been able to check it with the local host since it kept breaking and saying 'no such adventure' sorry. But I think it's mistake free now.

Felienne:
Thanks for this hard work @MarleenGilsing! I am merging your changes into the new and improved way of showing code on slides (that is just code, a lot easier!) Will let you know when that is ready so you can continue the work in the new system.

PR #4797, opened 2023-11-23
[FIX, TESTING] Fixes failing teacher test

jpelay:
**Description**

The first test in the file `second_teacher_invites` was failing because the survey was getting on the way. This PR fixes it by creating a new class before all the tests are executed, and since this new class doesn't have any students we guarantee that the survey is not shown. This approach should be applied to all the other tests, I think, but this one is a bit more urgent since it's not letting us merge PRs


Felienne (approved):
Seems to fix the flaky test!

Felienne:
Hmmm, this PR did the job but then again the test failed on main, so it must be flaky? Hope @hasan-sh can offer a bit of help since it is related to the second teacher feature.

PR #4798, opened 2023-11-23
[FIX] Check for the type of indexed fields in DynamoDB

rix0rrr:
Make sure that during local testing of a new indexing strategy, we don't accidentally depend on types that cannot actually be indexed in a real DynamoDB table.

**How to test**

Nothing to test, there is no new functionality.


jpelay (approved):
I really like that when we have a problem with Dynamo you always come up with a new test so it doesn't happen again!

Felienne (approved):

PR #4799, opened 2023-11-23
🧹 Add a build tool

rix0rrr:
We are accumulating quite a lot of build scripts that are all written in bash, and need to be implemented in a just-so order, and you need to know exactly when you need to run them. It might be time to start looking into moving to an actual tool to manage the build for us.

These come in a lot of flavors with various capabilities. I've looked around a bunch, and I think [DoIt](https://pydoit.org) looks quite nice.

Here are the features that I think make this tool quite nice:

- It's written in Python, so comes in automatically as Python dependency during `pip install`.
- The configuration file is written in Python as well, so we don't have to learn a new configuration language just for this tool.
- Tasks have dependencies between them, so if there are multiple that need to run, they automatically run in the right order.
- Tasks can (optionally) declare files that are inputs to the task itself, and if the input files have not changed the task is skipped. This makes it safe to just run the build toold and trust that you're not waiting too long for nothing.

I've started converting some build scripts over to this new tool to show how it works. You can see it in action by going into the virtual env and running (must be in the root directory, unfortunately):

```
$ doit
$ doit run typescript
$ doit help
```

The `.doit.db.db` that I've ignored is a file it uses internally to track changes to dependencies.

I've paused here because I'd like to solicit feedback.  Does the tool look okay? Is this understandable? Do we prefer this to bash, or to other build tools?


dodo.py
```diff
@@ -0,0 +1,118 @@
...
+        title=lambda _: 'Generate static Babel content',
+        actions=[
+            ['build-tools/heroku/generate-static-babel-content'],
+        ],
+        targets=['static_babel_content.json'],
+        uptodate=[babel_version_unchanged],
+    )
+
+
+def task_typescript():
```
jpelay, line 71: Wow I really like this!!!

jpelay:
WOW Rico, I think this is a magnificent idea! 

I personally don't love bash and indeed is quite difficult to be aware of the bunch of bash scripts we use for configuration. So this is great, it also seems like is very easy to work with! With this I imagine we'd also need to change GitHub actions a bit?

rix0rrr:
> With this I imagine we'd also need to change GitHub actions a bit?

Yes, I would say that we move everything from bash into here, and then everywhere we need to invoke some scripts we change them to invoke `doit run <whatever>` instead.

dodo.py
```diff
@@ -0,0 +1,325 @@
+##
+# This file contains various build steps used to make sure the data the Hedy application
+# uses is up-to-date.
+#
+# Functions that start with the name `task_` create a "task", which runs one or more
+# "actions" (shell commands or Python functions). Tasks can have file dependencies and
+# file outputs, as well as other task dependencies. If none of the files that a task
+# depends on have changed since the last time it ran, the task will be skipped. That's
+# good for speed!
```
hasan-sh, line 9: this sounds promising! How does it do that and does it do it with tasks as well? Or it's just with files and tasks

dodo.py
```diff
@@ -0,0 +1,325 @@
...
+        file_dep=[
+            *glob('templates/**/*.html'),
+            *glob('main/**/*.md'),
+            *glob('content/**/*.md'),
+            *glob('static/js/*.ts'),
+            script,
+        ],
+        title=lambda _: 'Generate Tailwind CSS',
+        actions=[
+            [script],
```
hasan-sh, line 61: so these files would still run with bash? I personally don't mind bash as a build-tool at all.

hasan-sh: I'm wondering since this PR aims at using `doit` as the primary build-tool. But I'm guessing you kept this file just because we may need to run it locally; e.g., watching changes!

dodo.py
```diff
@@ -0,0 +1,325 @@
...
+
+            # Then bundle JavaScript into a single bundle
+            ['npx', 'esbuild', '__tmp__/static/js/index.js',
+             '--bundle', '--sourcemap', '--minify', '--target=es2017',
+             '--global-name=hedyApp', '--platform=browser',
+             '--outfile=static/js/appbundle.js'],
+
+            # Delete tempdir
+            ['rm', '-rf', '__tmp__'],
+        ],
```
hasan-sh, line 164: I personally don't like this syntax. I think having this command as a string and splitting it into multiple lines (if needed) would be more readable. And in the case of including variables, we could simply use string interpolation!

build-tools/heroku/generate-typescript
```diff
@@ -1,73 +1,6 @@
...
-# plain old JavaScript. `esbuild` does not do that, which is why have TSC do
-# this job first.
-#
-# We can also run in `--watch` mode, in which case we bypass the TypeScript
-# compiler and only run esbuild. This is because the TypeScript compiler is
-# rather slow (think 5-10 seconds to do a compilation), and we want to
-# optimize for iteration speed in that case. That does mean that the generated
-# JavaScript is slightly different between those cases--however, the final
-# JavaScript will always be generated on deploy using the official generation
-# method.
```
hasan-sh, line 29: Ahhh is this why you kept the bash files even though we could run them with `doit`?

dodo.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def task_deploy():
+    """Commands to run at deploy time on Heroku.
+
+    This groups other commands.
+    """
+    return dict(
+        actions=None,
+        task_dep=['frontend', 'backend']
```
hasan-sh, line 277: so this means that the `deploy` task would run both of the `frontend` and `backend` tasks? That's actually very nice for organization purposes!!

dodo.py
rix0rrr: Just with files. It does it by calculating and storing the `md5` hash of every file. By doing that, it can check if files have changed, and skip the task if they haven't.

dodo.py
```diff
@@ -0,0 +1,325 @@
...
+# Functions that start with the name `task_` create a "task", which runs one or more
+# "actions" (shell commands or Python functions). Tasks can have file dependencies and
+# file outputs, as well as other task dependencies. If none of the files that a task
+# depends on have changed since the last time it ran, the task will be skipped. That's
+# good for speed!
+#
+# Keep the following in mind:
+# - You can specify commands as a string `'mkdir dir'` or as an array `['mkdir', 'dir']`.
+#   The second case doesn't allow shell features, but is safer in case of variables.
+# - `file_dep` requires a list of source files, but you can use `glob('*.yaml')`
```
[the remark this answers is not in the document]
rix0rrr: Every task looks like this:

```
               ┌──────────────┐                
               │              │                
input ────────▶│    Script    │────────▶ output
               │              │                
               └──────────────┘                
```

`file_dep` should at least be the set of `input` files. That way, if any of the input files change, you know that the `output` needs to be regenerated.

But what if you change the script itself? The thing that turns `input` into `output`? 

If you *don't* put the script itself into `file_dep` as well, `doit` will think that `output` is still up-to-date! Because none of the `input` files have changed, so according to the list of instructions `output` doesn't need to be regenerated.

You probably do want the task to run again if you change the script though, because probably the `output` would change. That's why you should add the scripts to `file_dep` as well.


If you change the script, you probably want to run the task again, because

dodo.py
rix0rrr: > And in the case of including variables, we could simply use string interpolation!

Yeah but here's the thing: you need to escape special characters inside those strings, otherwise they get pass uninterpreted to the shell and the command might do something else than you want.

A trivial example that will probably only fail:

```py

def copy_file(file1, file2):
  run_command(f'cp {file1} {file2}')  # Seems easy enough

# This is fine
copy_file('one.txt', 'two.txt')  # cp one.txt two.txt

# This is not
copy_file('file with spaces.txt', 'two.txt')   # cp file with spaces.txt two.txt
# ^^^ oops! Now the cp command gets 4 arguments instead of 2!
```

Not to mention special characters like `&` and `>` and `;` etc, that can make the shell do weird things.

```py
copy_file('file1.txt', 'file2.txt; rm -rf /')    # cp file1.txt file2.txt; rm -rf / 
                                        # ^^^^ the shell will happily execute this
```

I guess for the domain of a build tool, where we don't really handle untrusted strings, it matters less. But I've learned to be hyper-allergic to strings for shell commands, for this reason.

build-tools/heroku/generate-typescript
rix0rrr: Oh, no. This functionality will actually be gone. I don't know how many people use it...

dodo.py
rix0rrr: That's what I was thinking, yes. 

On deploy on a Heroku machine we would run the `deploy` task, and that would just do "everything" that's necessary to get both the backend and frontend in order.

dodo.py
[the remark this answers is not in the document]
hasan-sh: Right, that makes sense. Shouldn't we then also include the `.doit.db`? Otherwise all tasks will run no matter whether the inputs/outputs changed?

dodo.py
hasan-sh: > * `doit` will take care of dependencies: run tasks that depend on each other in the right order, won't run them twice, and won't run them if it's not necessary

Wanted to investigate whether it does that. So, i created this dummy function:
```python
# Create a simple task
def task_copy_file():
    return dict(
      title=lambda _: "copy file into another",
      actions=['cp "dev_database.json" "test here.txt"'],
      file_dep=['dev_database.json'],
      targets=['test here.txt']
      )
```

Then ran: `doit copy_file`. Notably, it did two things:
1. it ran the task that you created: `devdb` and `copy_file`. Probably because the file dep in copy_file is in the target of devdb!
2. i ran it again after a few minutes, and as you explained, it actually did not rerun the tasks. Really nice!!

Later, i removed `.doit.db`, and reran it, boths tasks were fired. So, i do believe we need to include it to git, otherwise all tasks will be executed!?

rix0rrr: > Probably because the file dep in copy_file is in the target of devdb!

Yes! Exactly, it traces task dependencies through declared file dependencies. In this case, that's probably not what we want to happen, so that might be something we need to be careful with...

> `.doit.db` [...] we need to include it to git, otherwise all tasks will be executed!?

I've been thinking about this a little, and I think it's safer to leave the database out of git for now.

* First, the database is a binary file, and git doesn't deal well with binary files usually
* Second, especially because it's a binary file, and many PRs will make changes that cause changes to this file, we will nearly always have conflicts on this file when trying to merge any PR. That will get annoying very quickly.
* If we check this file in, it will also go to the server, and the server will also skip build steps based on this file. That MAY be fine, but it may also become the cause of hard-to-diagnose bugs in the future, if any bugs slip in (if we make mistakes in dependency declarations on the tasks, for example).

The only reason to check in the database file would be to save running tasks across machines. I.e., I run the task, and then you don't have to anymore. That's nice, but it's not the end of the world if you have run the tasks *once*, and can skip them after.

All that is to say, I think the benefit is not massive, and the annoyances and risks are real, so I don't think checking in the database file would be worth it.

build-tools/heroku/generate-typescript
jpelay: I use it :c

dodo.py
```diff
@@ -0,0 +1,325 @@
...
+        ],
+    )
+
+
+def task_frontend():
+    """Run all tasks necessary to prepare the frontend."""
+    return dict(
+        actions=None,
+        task_dep=[
+            'tailwind',
```
jpelay, line 264: Maybe we could add the lezer parsers here? They're also part of the front-end, and also since this is called on deploy we wouldn't call that task there, depending on people checking in the automatically generated files.

rix0rrr:
> When I execute a task it also generates this files that are tracked by git: .doit.db.bak, .doit.db.dat, .doit.db.dir perhaps we should also add them to .gitignore

Oh, I hadn't seen those. I just saw `.doit.db` I think. Yes, we should probably gitignore `.doit.db*`.

> Why does the tailwind task also call these:

Yeah, good question. When I do:

```
$ doit info tailwind

tailwind

Build Tailwind.

status     : run
 * The following file dependencies have changed:
    - static/js/syntaxModesRules.ts
    - static/js/modal.ts
    - static/js/auth.ts
    - templates/customize-class/partial-sortable-adventures.html
    - templates/quiz/partial-review-quiz.html
    - templates/printable/certificate.html
    - static/js/teachers.ts
    - static/js/parsons.ts
    -  ....
```

It's showing that the Tailwind job depends on all the `.ts` files. Presumably because any of the `.ts` files could include references to Tailwind CSS class names, and so those must be found and not be stripped from the final CSS bundle.

Because that list is `*.ts`, it includes `static/js/client-messages.ts`, which is the output of `generate client messages`. 

So as far as the current dependency graph is aware: regenerating `client-messages.ts` *could* potentially change the list of CSS classes referenced in the code, and so it must be rerun before generating the Tailwind CSS file.

We could break the dependency by making sure that `client-messages.ts` is NOT in the `file_dep` list for the Tailwind task.

> Is there a possibility to run these tasks on watch mode? I found that feature really useful when working on the front-end

Excellent question. The docs say that [doit has a "watch" mode](https://pydoit.org/cmd-other.html#auto-watch) but on my machine it doesn't appear when I do `doit help`.

> Maybe we could add the lezer parsers here? They're also part of the front-end, and also since this is called on deploy we wouldn't call that task there, depending on people checking in the automatically generated files.

I was under the assumption that the lezer parsers were being generated as part of the TypeScript build. But if they aren't, yeah, we should definitely add a dependency somewhere. Probably to the TypeScript task, as the lezer parsers generate `.ts` files which still need to be compiled to `.js` (during the Typescript task).

---------------

@jpelay, in the interest of gaining some experience with this tool, would you mind taking a stab at these yourself? 😇

jpelay:
> @jpelay, in the interest of gaining some experience with this tool, would you mind taking a stab at these yourself? 😇

Yes! I actually wanna try those suggestions!

rix0rrr:
I think I got `watch` figured out. I added a plugin for the `watch` command, so now `doit watch typescript` or just `doit watch` should work.

Fwiw, we control the [watch plugin code](https://github.com/rix0rrr/doit_watch) (I forked it from the original `auto` plugin, which doesn't work anymore but some other people helpfully left PRs to make it work again ☺️)

jtwaleson:
> I've paused here because I'd like to solicit feedback. Does the tool look okay? Is this understandable? Do we prefer this to bash, or to other build tools?

Thanks! Indeed it's become a bit much, and I like have a better organized set of tools. I don't think it's very high priority but definitely not complaining for getting something nice! I have no hands-on experience with doit but from what I've seen it's is a good middle position between bash/makefiles and full blown build tools like bazel. Sounds like the right approach!

jpelay:
Hi @rix0rrr! I did the improvements we discussed, but I don't understand why the cypress tests keep failing, they work correctly on my computer (it works on my machine!)

hasan-sh:
> Hi @rix0rrr! I did the improvements we discussed, but I don't understand why the cypress tests keep failing, they work correctly on my computer (it works on my machine!)

Probably something went wrong with merging main into this branch since i've made some changes after this was created. For instance, the current version of `cypresstests.yml` is different from the [one in main](https://github.com/hedyorg/hedy/blob/main/.github/workflows/cypresstests.yml). Another crucial change was in the cypress config, but i see the one in this branch has the required changes!

hasan-sh:
> Hi @rix0rrr! I did the improvements we discussed, but I don't understand why the cypress tests keep failing, they work correctly on my computer (it works on my machine!)

I believe that it has to do with the bundle files. I've seen this error before and adding the following: https://github.com/hedyorg/hedy/blob/main/build-tools/heroku/generate-typescript#L61-L68 helped. Perhaps make these tasks verbose, so that we can inspect the process!

rix0rrr:
> I believe that it has to do with the bundle files. I've seen this error before and adding the following: https://github.com/hedyorg/hedy/blob/main/build-tools/heroku/generate-typescript#L61-L68 helped. Perhaps make these tasks verbose, so that we can inspect the process!

We already run `doit deploy` and `doit frontend` both, both of which should regenerate the JavaScript bundle.

rix0rrr:
>  (it works on my machine!)

The build server starts from an empty clone every time, but your machine probably doesn't. So one way to try and reproduce this is to start from an empty clone yourself as well. 

You can run:

```shell
git reset --hard
git clean -fdqx .
```

To make your local working copy pristine and the same as from an empty clone, but you'll lose any untracked files you have lying around. You can also make a fresh new clone to a new directory if you'd rather keep your current working copy intact.

After that, run through the commands in the workflow file one by one. That should produce the exact same results as on the build server:

https://github.com/hedyorg/hedy/blob/9af04472d785a81cafd742a6cad1bba26ae489a6/.github/workflows/cypresstests.yml#L26-L55

rix0rrr:
Here is the stack trace. Is `tailwind.ts` new? I think I'm seeing that file for the first time.

```
tailwind.ts:8 Uncaught TypeError: (0 , mS.initTE) is not a function
    at tailwind.ts:8:1
    at appbundle.js:1:778
    at index.ts:26:1
    at appbundle.js:1:778
    at index.ts:28:10
    at index.ts:28:10
```

hasan-sh:
> Is `tailwind.ts` new?

Yes, relatively. It's a tailwind plugin that I added recently. Not sure why it's not a function here. But it requires two things: npm install (which you do i think) and compiling tailwind as well, since some settings in the tailwind.config file must be configured before this can work!

rix0rrr:
I think I found the issue. We are caught in the middle between CommonJS (the traditional method of importing JavaScript modules introduced by Node, since JavaScript started out without a module system) and ES Modules (a standardized method of defining and importing modules). 

Unfortunately there is some incompatibility between CJS and ESM: it is NOT possible to import an ESM module from a CJS module. 

`tw-elements` is ESM-only, so it's not possible to import from a CJS module. The TypeScript compilation step is currently configured to produce CJS files, and then the `esbuild` step does something wrong when it tries to resolve the CJS import of an ESM file (would produce a runtime error if you ran it directly in Node, but `esbuild` tries to do something useful anyway, and apparently produces an output file that is incorrect)

rix0rrr:
I've updated TypeScript from version 4 to version 5, and fiddled with the `tsconfig.json` settings to compile `TypeScript → ESM JavaScript`, and then use `esbuild` to bundle `ESM JavaScript → CJS JavaScript`.

The salient point here is that that `esbuild` now properly resolves the ESM-only import `ts-elements`, because the input JavaScript files are also in ESM format.

jpelay:
> I've updated TypeScript from version 4 to version 5, and fiddled with the `tsconfig.json` settings to compile `TypeScript → ESM JavaScript`, and then use `esbuild` to bundle `ESM JavaScript → CJS JavaScript`.
> 
> The salient point here is that that `esbuild` now properly resolves the ESM-only import `ts-elements`, because the input JavaScript files are also in ESM format.

Thank you for fixing the issue, now there are some issues with the lezer files that I'm going to look at!

rix0rrr:
> Why have the `module` set to ES6 though?

It needs to be some form of `ES`. I picked ES6 because I don't like targeting too new stuff. I like old and reliable! 😁

jpelay:
Let's hope that setting `moduleResolution` to `node` fixes the issue!

rix0rrr:
The problem seems to be here:

image

I don't think the build was designed for this. 

I'm honestly not even sure how this works at all, given that this requires `.js` files but in the source are only `.ts` files 🤔

rix0rrr:
So the pipeline for Cypress is that:

- Cypress uses WebPack to bundle the `.js` files into a single file that can be loaded into the Electron instance
- WebPack uses Babel to load both `.js` and `.ts` files, with some default behavior
- Babel then uses TypeScript as soon as it finds a `.ts` file, with some default behavior

The very first error was ` TS5053: Option 'sourceMap' cannot be specified with option 'inlineSourceMap'.` because Cypress' use of WebPack dictates the use of `devtool: "inline-source-map"`, which ultimately gets passed to TypeScript and  conflicts there with the `sourceMap: true` in `tsconfig.json` (you can't have both `sourceMap: true` and `inlineSourceMap: true`). 

I managed to fiddle with the config enough to disable that, but then the next error becomes:

```
Module not found: Error: Can't resolve '../../../../../static/js/lezer-parsers/level1-parser' in '/Users/huijbers/Dev/hedy2/tests/cypress/e2e/tools/lezer'
```

And I don't know how to change this. I've been messing around with Cypress wrapping WebPack wrapping Babel wrapping TypeScript trying to pass configs around to maybe use an alternative `tsconfig.webpack.json` file or something, but I can't get it to work.

I've been spending 2 hours on this already, I'm giving up for now.

rix0rrr:
Ah, with `moduleResolution: "node16"` we're running into the issue that the author of `@codemirror/state` and the TypeScript team disagree on where the `"types"` declaration on `package.json` should be if an `exports` key is present.

* The TypeScript team thinks that it should be under `exports`: https://github.com/microsoft/TypeScript/issues/49160#issuecomment-1137482639
* The author of `@codemirror/state` thinks the TypeScript team is being silly and their new rules are too strict: https://github.com/codemirror/state/pull/3#issuecomment-1406726887

This leaves us stuck in the middle, unable to move to the `node16` `moduleResolution` strategy.

jpelay:
> It also looks like you're using the Cypress test framework to run JavaScript unit tests that don't interact with the browser at all.
> 
> I guess that works, but it's also not necessary. If you want to run a suite of JavaScript tests, you can use a JavaScript test framework like [Jest](https://jestjs.io/) to run some tests before bundling even happens, and they don't need to run in the browser either: they can run in Node.

I made it that way, mainly because we already had Cypress and it worked fine, also since Mocha is already included in Cypress I didn't see the need. However, in light of these problems maybe we need to change the tests and use Jest or other framework, I'm totally up for it! So if you want we can disable the Lezer tests for now, we open an issue about replacing them with Jest or other similar framework and merge this! What do you think?

rix0rrr:
Sounds good to me! I'm just a tiny tad confused why all these problems are suddenly cropping up on this branch, and not the `main` one...

hasan-sh:
> Sounds good to me! I'm just a tiny tad confused why all these problems are suddenly cropping up on this branch, and not the `main` one...

I believe it's because in main we run a different command. Namely, we don't compile ts first and then run esbuild, as `task_typescript` now does: https://github.com/hedyorg/hedy/blob/20c8d4ff5ae4c00ba3d44f7649979819860f0020/dodo.py#L160-L163

I'm not sure what happens in the `generate-typescript` bash file, but I do remember running esbuild on the index.ts file directly whenever there's a `--tests` flag, and also remember that you removed that flag from the cypresstests.yml file, assuming that only generating bundle files would solve the cypress problem. However, we somehow overlooked the fact that even if the flag wasn't passed, the generate-typescript file does does the same thing. https://github.com/hedyorg/hedy/blob/main/build-tools/heroku/generate-typescript#L73

I was also following your progress not understanding why this is happening, but this is the only and main difference that would make sense on why!

rix0rrr:
Ready to ship! 🎉

hasan-sh (approved):
Amazingly done :100:

PR #4801, opened 2023-11-24
[FIX] Setting up view program page editor

### LINKED ISSUE #4796: [BUG] Code isn't visible in program view
**Describe the bug**
- go to programs (`/explore` view),
- try to view any program,
- the code isn't appended to the editor and thus you cannot run it

**screenshot**
image

**machine and browser**
- on  Linux
- using chrome, firefox

jpelay:
**Description**

Sets up the editor of the view program page correctly. This is done by adding a new `code` parameter to the initialization of the page to set the editor contents. Also initializes the syntax highlighter.

**Fixes #4796**

**How to test**

Go to the view page of a program and it should be shown correctly.


.github/workflows/update-js-and-babel-on-main.yml
```diff
@@ -42,8 +42,10 @@ jobs:
 
     # This is necessary for the TypeScript to contain translations
     # from Gettex.
-    - name: Compiling Babel translations
+    - name: Update and compile Babel translations
       run: |
+        pybabel extract -F babel.cfg -o messages.pot . --no-location --sort-output
```
Felienne, line 47: These are now removed from main due to issues, must be removed here too (they did not automatically because we force-pushed main)

messages.pot
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-11-23 12:10-0400\n"
+"POT-Creation-Date: 2023-11-23 22:22+0000\n"
```
Felienne, line 11: I guess all of these can be removed too (to avoid issues)

Felienne (approved):
Thanks!!

PR #4805, opened 2023-11-24
[FIX] Improve reliability of second teacher Cypress test

rix0rrr:
I tried improving the reliability of the cypress tests for second teachers, because I selfishly want my own PR merged and it blocked on this 😈.

I made the following changes:

### Class selection

The tests used to create a new class, but then always clicked into the *first* class they could find on the teachers page instead of the class they created. Depending on what names get generated, this was not necessarily the fresh class we just generated, but could be any class left by a previous test and there's no knowing what state that class is in. Use `navigateToClass(className)`.

### Should vs contains

Replace `should()` with `contains()` in most places. `contains()` is a query, meaning it *looks and waits* for elements containing that string. On the other hand, `should()` is an assertion: on the elements your query is already focused on, assert that they have this text. The difference comes up when timing comes into the picture. Here's an example:

Let's say we have an element with text `teacher2`, and after we click a button `teacher3` is going to be added to the list. However, this may take a few milliseconds.

```ts
cy.get('.teacher_link')                   // <--- (Q)
  .should('include.text', 'teacher3');    // <--- (A)
```

In this example, the query `Q` is "find me all elements with class `teacher_link`". If this executes *after* `teacher3` has been added to the list, then that query will match two elements, and the assertion `A` can see that, indeed, one of them has the text `teacher3`.

If, on the other hand, this executes *before* `teacher3` has been added to the list, then the query is still satisified. It can still find at least one element with the class `teacher_link`, and it will move on to the assertion. Even if the `teacher3` element appears a millisecond later, it's too late: the query has stopped waiting.

If we write the cypress statement like this:

```ts
cy.get('.teacher_link')     
   .contains('teacher3');     
```

Then now *both* statements are part of the query! Cypress will look and wait for an element that is BOTH a `.teacher_link` as well as has text `teacher3`. There isn't one, until there is. What won't happen, is that Cypress will settle on the wrong set of elements too early.

### DOM navigation

Replace some JavaScript navigating the invite table with a selector involving `.parent()`: we find the row containing the teacher we're interested in, then use `.parent().find()` to find other elements *in the same row*.

To finish the assertion, I've needed to move the test up, so that we revoke teacher3's invitation before we accept teacher2's invitation (otherwise at the end the table is altogether gone and the `cy.get()` fails).

**How to test**

GitHub will test for us!


rix0rrr:
The irony! 

I guess with anything build system related any PR title needs to be prefixed with *"Try to ..."*

image

tests/cypress/e2e/for-teacher_page/class_page/second_teachers_invites.cy.js
```diff
@@ -1,83 +1,78 @@
...
   before(() => {
     loginForTeacher();
-    createClass()
+    className = createClass()
   });
 
   it(`Invites ${secondTeachers.length} second teachers: by username`, () => {
     loginForTeacher();
-    cy.get(".view_class").first().click();
+    navigateToClass(className);
```
jpelay, line 20: Right! I forgot that we had this function. I also selfishly wanted this test to work for my PR to be merged hahaha

tests/cypress/e2e/for-teacher_page/class_page/second_teachers_invites.cy.js
```diff
@@ -1,83 +1,81 @@
...
+    loginForTeacher(secondTeachers[0]);
+    goToProfilePage();
+    cy.get("#messages #join").click();
 
-  it("Reads all second teachers", () => {
-    loginForTeacher()
-    cy.get(".view_class").first().click();        
-    cy.get("#second_teachers_container .username_cell").should("include.text", secondTeachers[0])
+    // Give the Ajax request that gets sent as a result of the click enough time to complete
+    cy.wait(500);
```
jpelay, line 72: What if instead of adding a hardcoded wait we do

```
cy.intercept('/class/join/*').as('join')
// test code
cy.wait('@join')
```

rix0rrr: Dunno, never used it. I like it better, but not behind my machine right now.

do you want to make the change?

rix0rrr: (And if you do, feel free to ship it 🤪🥸)

jpelay: Yes, I'll do the change!

jpelay (approved):
Thanks for fixing the tests @rix0rrr!

jpelay:
What is wrong with mergify?!

jpelay:
Hurra!

hasan-sh:
Thanks for the changes and detailed explanation @rix0rrr.  I do think that this test file looks cleaner ;)

> The tests used to create a new class, but then always clicked into the _first_ class they could find on the teachers page instead of the class they created. Depending on what names get generated, this was not necessarily the fresh class we just generated, but could be any class left by a previous test and there's no knowing what state that class is in. Use `navigateToClass(className)`

Agreed!

> Replace `should()` with `contains()` in most places. `contains()` is a query, meaning it _looks and waits_ for elements containing that string. On the other hand, `should()` is an assertion: on the elements your query is already focused on, assert that they have this text. The difference comes up when timing comes into the picture.

According to the documentation of [Cypress](https://docs.cypress.io/api/commands/should), the `should` command retries the assertion until it passes or the time outs. When creating these tests, I did read the documentation and therefore used it as such. [Here's another example](https://docs.cypress.io/guides/core-concepts/retry-ability#Commands-Queries-and-Assertions) that explains how queries and assertions are linked together and are retried, automatically.


> Even if the `teacher3` element appears a millisecond later, it's too late: the query has stopped waiting.

As a result, if the assertion fails, the query chain will be retried and thus if `teacher3` appears later, it will pass. Of  course this should be within the timeout threshold which is set to `4000ms`. 


So, i think this is also why the tests used to pass for most PRs. I do doubt the reliablility of how Cypress handles syncronous code and probably have to research it more in depth. And I cannot really see how using the command `contains` is more reliable than `should`.  Sometimes if you just rebuild the cypress tests, the piece that yielded errors, wouldn't anymore! This happend several times and I'm not sure if it's a syncronous problem or something else, perhaps cypress caches the code or something like that. Could it be that this github action is indeed caching the views or not retrieving the latest of the app bundles?

PR #4808, opened 2023-11-25
[FIX] eliminate hash collisions in test caching

jtwaleson:
**Description**

In a recent PR Felienne noticed something weird: a test that should fail actually succeeded. https://github.com/hedyorg/hedy/pull/4663/commits/aa3f4d9fb798f86ee9a19a24532198db99351003

After debugging we figured out that:
- we expect some tests to succeed
- in some tests we expect to raise an exception

When the same snippet and level is used across tests, the test runner would use a cached result from another run, reporting the wrong outcome.

Example:

- Test 1: Snippet A - Level 3 - expect successful execution -> this runs fine
- Test 2: Snippet A - Level 3 - expect an error  -> this should fail because we don't get an exception BUT it does not raise an error because it re-uses the cached result from Test 1.

In this PR we include not just the snippet and level but ALL of the arguments to the single_level_tester method. This way, if we expect an exception, it will result in a different hash.

**How to test**

* See if all the tests still run fine -> https://github.com/hedyorg/hedy/actions/runs/6988954244/job/19017110194
* See if caching still improves the test run by a factor of about 3x -> https://github.com/hedyorg/hedy/actions/runs/6989091374/job/19017406242

**Checklist**

- [ ] Links to an existing issue or discussion -> no, this PR describes the issue that was discussed on Discord


Felienne (approved):
Thanks for the extensive detective work!!

Issue #4646, opened 2023-10-19
💻 Realtime updates


hasan-sh:
**Idea incl level**
Currently, whenever the user makes an action, we mostly "reload" the page or redirect them. However, this isn't the best UX. For instance, removing a class, student and an adventure isn't reflected in the UI of all parties.  

**Additional context**
This is problematic when different parties make changes simultaniously; e.g., after allowing second teachers to make changes (#4644). Use socketio or something similar.

TiBiBa:
I have quite some experiencing with implementing Soketi websockets in a project. Please let me know if you need any help!

https://docs.soketi.app

Felienne:
Hi @hasan-sh!

I know we have now done this in some places (creating adventures) but I am not sure where else. It is totally ok if you leave this one, but can you make a list here of places where it needs to be done still, to ease handover?

Felienne:
Hi @Annelein!

Can you pick this up for the /admin page?

Felienne:
> #5631

Awesome, thanks!!

> I can however look into creating a list of pages for which it would be nice to implement autosave?

Love your proactive spirit, thanks! Sounds like a great idea to make this list, we can then prioritize in a next meeting.

PR #4812, opened 2023-11-27
[BUG] enable inviting second teachers to multiple classes at the same time

### LINKED ISSUE #4804: [BUG] A user cannot be invited as student/second teacher to more than one class at a time
**Describe the bug**

If a teacher is invited to one class, they cannot be invited to any other class anymore as long as the invite is pending.

Reproduce as follows:

* Create a class, invite `teacher2`
* Create another class, invite `teacher2` to that as well

You get an error saying that `teacher2` is already invited to the class (but it isn't, that's another class!)

This behavior may not occur a lot in practice, but makes it harder to write reliable tests because tests cannot be made independent, and state left behind from a past test in the database can interfere with future tests.

EDIT: Jesus noted that this may actually happen in practice: one teacher may be creating a bunch of classes at once, and then inviting a second teacher to all of them in one go. With the current behavior, the second teacher would need to be present to accept each invite before the next invite can be sent.

hasan-sh:
Fixes #4804 
**Problem**
1. The main issue spawns from having the _username_ as the partion_key for the invitations table, thereby preventing adding multiple invitations for the same user. 
2. Also, even if this worked, users currently only see one invitation at the time; due to restricting messages to be 0 or 1. _This is now changed as well._
3. There are a couple of things in the code that are inconsistent; e.g., https://github.com/hedyorg/hedy/blob/main/website/database.py#L737-L738 would remove _an invitation_ of the username, while the function name implies deleting an invitation to some class!

**Solution**
- create a unique, partition_key `username#class_id` in the INVITATIONS table.
- with two independent indexes: `username` and `class_id`
- allow second teachers (perhaps students later on as well!) to accept more than one invitation at the same time.
  - The reason we have messages in the session is mainly to display a number indicator in the profile dropdown menu. Logically, this should happen whenever teachers invite other users and these invitees do not have to reload the page to see the indicator, but now they do!
  - this should be fixed when we have realtime updates #4646!

**How to test?**
Follow the steps that Rico listed in #4804.


website/database.py
```diff
@@ -45,7 +45,8 @@
 # - tags_id (str): id of tags that describe this adventure.
 ADVENTURES = dynamo.Table(storage, "adventures", "id", indexes=[dynamo.Index("creator"), dynamo.Index("public")])
 INVITATIONS = dynamo.Table(
-    storage, "class_invitations", partition_key="username", indexes=[dynamo.Index("class_id")]
+    storage, "class_invitations", partition_key="username#class_id",
```
rix0rrr, line 48: This is a good idea! Unfortunately, DynamoDB does not allow changing the key of an existing table. If you want to do this, you'll have to create a new table (`invitations2` or whatever).

You then have a choice:

* We fully switch over to the new table and effectively invalidate all existing invitations
* We are a more complicated multi-step deployment if we want to make sure that old invitations still work while we are cutting over to the new table

If we do the multi-step deployment to keep all invitations alive:

* Create a new table with the desired keys (`invitations2` or whatever)
* Update the code to **write** to `invitations2`, but still **read** from both `invitations` and `invitations2`.
* Deploy
* Copy over all old invitations from `invitations` to `invitations2` and empty the old table, or wait a while until all old invitations have expired. 
* Remove all references to `invitations` from the code
* Deploy
* Delete the `invitations` table

By the way, a partition key of `username#class_id` with 2 indexes will totally work, but you can also do it like this:

```
partition_key: username
sort_key: class_id
```

If you do that, you already have the index on `username` (it's just the main table) and you only need an additional index on `class_id`.

website/database.py
```diff
@@ -728,16 +732,19 @@ def delete_class(self, Class):
     def resolve_class_link(self, link_id):
         return CLASSES.get({"link": link_id})
 
-    def get_username_invite(self, username):
-        return INVITATIONS.get({"username": username}) or None
+    def get_user_class_invite(self, username, class_id):
+        return INVITATIONS.get({"username#class_id": f"{username}#{class_id}"}) or None
+
+    def add_class_invite(self, key, data):
```
rix0rrr, line 738:
```suggestion
    def add_class_invite(self, username, class_id, data):
```

If you do it like this, you can keep the concern that the `username` and `class_id` field have to be concatenated together using a `#` inside this class. That way, the rest of the code doesn't need to know that.

You could also grab those fields out of `data`, which is presumably a whole row, and then you don't need `key` at all:

```py
    def add_class_invite(self, data):
        data['username#class_id'] = data['username'] + '#' + data['class_id']
        # Put the entire record in the database
        INVITATIONS.put(data)
```

website/database.py
```diff
@@ -728,16 +732,19 @@ def delete_class(self, Class):
...
 
-    def add_class_invite(self, data):
-        INVITATIONS.put(data)
+    def remove_user_class_invite(self, username, class_id):
+        return INVITATIONS.delete({"username#class_id": f"{username}#{class_id}"})
 
-    def remove_class_invite(self, username):
-        INVITATIONS.delete({"username": username})
+    def get_user_invitations(self, username):
+        return INVITATIONS.get_many({"username": username}) or []
```
rix0rrr, line 745: Doesn't `get_many` already return an empty list by itself?

website/classes.py
```diff
@@ -141,19 +141,19 @@ def join_class_id(self, user, class_id):
...
 
         # We only want to remove the invite if the user joins the class with an actual pending invite
-        invite = self.db.get_username_invite(username)
+        invite = self.db.get_user_class_invite(username, class_id)
 
-        if invite and invite.get("class_id") == class_id:
+        if invite:
             invited_as = invite.get("invited_as")
-            if invited_as == gettext("second_teacher"):
+            if invited_as == "second_teacher":
```
rix0rrr, line 148: 👍

website/database.py
hasan-sh: Yeah, it returns a ResultPage class which, if iterated over, we iterate over its records; a list indeed! Will change it, thanks!

website/database.py
Felienne: Hi both! 

There are currently no open invitations in the database, so I think we can save ourselves the hassle and migrate directly. I suspect invitations for students are not so often used, most teachers add kids to their class directly, or they are accepted immediately and then deleted from the table.

rix0rrr: Yep, confirmed, there are `0` rows in the table right now. Good shout @hasan-sh. We can just make a new table in Dynamo and change the table name that the `Table` variable points to, but we don't have to add a new variable if we're only going to do a one-step deployment. I can do that immediately after we settle on a table name and a key schema. 

## Name

Currently the table is called `class_invitations`... how about `invitations` ?

## Key schema

> Does this mean we could mitigate multi-step deployment and still use the same table?

Unfortunately, no. Adding a sort key is still changing the key schema, so we're still going to have to make a new table.

Using the table plus one index, as opposed to the table plus two indexes, just saves you a copy of the data.

rix0rrr (approved):
I created the new tables, so this is good to go.

PR #4814, opened 2023-11-27
[UI] Locked levels with tooltip

### LINKED ISSUE #4741: [UI idea] Consider dropdown
In the old UI, it was quite clear for kids that a level was locked and why (with a hover over the greyed out button):

image

image

Now, I find it less clear that levels are gone and why:

image

Can we make this clearer again (I would not know how?) 

Also, if we leave it live this, then the string we used ("Your teacher...") is not needed anymore I think?

Annelein:
Fixes #4741 

Let me know if this is what you had in mind! 

Screenshot 2023-11-28 at 15 27 59


translations/en/LC_MESSAGES/messages.po
```diff
@@ -1646,6 +1643,9 @@ msgstr "Hedy - View adventure"
 msgid "token_invalid"
 msgstr "Your token is invalid."
 
+msgid "tooltip_level_locked"
+msgstr "This level is locked"
```
hasan-sh, line 1647: To be clearer, change it to "Your teacher disabled this level"

templates/level-page.html
```diff
@@ -24,12 +24,18 @@
...
+              {% if dropdown_level == level %}bg-white{% elif dropdown_level not in customizations['available_levels'] %}bg-gray-400 tooltip{% else %}hover:bg-blue-100 bg-blue-200c{% endif %}"
+              {% if dropdown_level == level or dropdown_level not in customizations['available_levels']%}disabled{% endif %}
               style="width: 100%;"
               _="on click hedyApp.goToLevel({{dropdown_level}})">
+              <div class='{% if dropdown_level in customizations['available_levels'] %}hidden{% endif %}
+              text-center opacity-0 hover:opacity-100 duration-300 absolute inset-0 z-10 items-top'>
+                <span class='tooltiptext text-xs font-medium absolute rounded shadow-lg' style="margin-left: -50px; bottom: 130%; color:black; background:lightsteelblue;">
+                  {{ _('tooltip_level_locked') }}
+                </span>
+              </div>
```
hasan-sh, line 38: This is definitly the proper way of working with Hyperscript. Amazing! 

Do add some padding to the tooltip and center it as well.

hasan-sh (changes requested):

templates/level-page.html
Annelein: Is it better like this?

hasan-sh: See this:

image

Also, why is it squeezed within the dropdown? Make it free ;))

Annelein: Ah good catch, thanks. I've been trying, can you help me out on how to make it free?

hasan-sh: Alright, will text you on discord

jpelay (commented):
Hi @TiBiBa!! Sorry! I forgot about applying the CodeMirror change here. 

Can this editor grow larger than 352px? If so I think we need to change a few things
image

PR #4817, opened 2023-11-28
[LANGUAGE] improved error productions for non decimals

### LINKED ISSUE #4670: [LANGUAGE] Hedy doesn't fully support non-positional numeral systems
The [Hedy grammar](https://github.com/hedyorg/hedy/blob/main/grammars/terminals.lark#L4-L7) permits code points that participate in many kinds of numeral systems, but Hedy appears to only really support base-10 positional numeral systems (composed of sequences of code points in General Category `Decimal_Number`). When a user tries to use another numeral system that is permitted by the Hedy grammar, such as Ethiopic, Brahmi, or Roman numerals, they get error messages like the following:

<p>image</p>

<p>image</p>

There's a couple things that could be done here, in order from easiest to hardest.

1. Limit the grammar to only accept sequences of `Numeric_Type=Decimal` (or `General_Category=Decimal_Number`, same thing) code points.
2. Simply improve the error messages for programs that contain these unsupported numeral systems.
3. Limit the grammar to accept either sequences of `Numeric_Type=Decimal` or **just one** code point with an integral `Numeric_Value` property.
4. Add support for computing the numeric value of all of these currently unsupported numeral systems. This is a *lot* of work.

akseron:
Changes in error production and regex to support better syntax error messages for non decimal numbers. Currently regex allows non decimals, however they are unfunctional and provide complex parse errors. Fixes #4670.


Felienne:
Hi @akseron!

Just to check, is this a fix for https://github.com/hedyorg/hedy/issues/4670?

Also, I see it is said to draft, do you plan more work? I can help with the merging and failing tests if you want!

Felienne:
Hi @akseron! 

I updated the broken test. I am not sure why the parser gets confused with the new rule, but it is ok for me, the previous location was also wrong haha, and we need to aim to not give a ParseError anyway in such a situation.

So I am good with shipping this, If you are also, set it to Ready for Review and I will approve!

akseron:
Oh great thank you! I was already planning to ask about changing the error_location since it seemed hard coded but it's great that you already fixed it haha. It was indeed a fix to 4670 so this should do it. 

A last question I had about it was that during the last Hedy contributors meeting, I was told I had to refactor the code but I am not certain if that is necessary here. I will set it to Ready to Review in case it is not 👍

Felienne:
> Oh great thank you! I was already planning to ask about changing the error_location since it seemed hard coded but it's great that you already fixed it haha. It was indeed a fix to 4670 so this should do it.

Perfect! Can you add that to the original description? If you add "fixes #4670" then the issue will close automatically!

> A last question I had about it was that during the last Hedy contributors meeting, I was told I had to refactor the code but I am not certain if that is necessary here. 

Ah I wam not there so I don't know who said that and what refactoring they asked for (just "refactoring" is a but vague) Maybe you can ask who said that for their input?

jpelay:
Hi  @akseron! You don't need to refactor your code! But you do need to do a rebase of your branch, or perhaps force pushed it back to a previous state and then do your changes again. This is because we ran into some issues with main last weekend and we forced pushed main back, and that's what's causing a lot of unwanted changes in your PR

Felienne (approved):
Looks good to me! Thanks for the hard work @akseron!

PR #4818, opened 2023-11-28
[BUG] tags dropdown items get their content width only

### LINKED ISSUE #4816: [BUG] Make the tags dropdown items take content width only
**Describe the bug**
After merging #4759, on apha the items look like this:

image

However, they should be: 

image

hasan-sh:
Fixes #4816


jpelay (approved):
Nice work, @hasan-sh! Thanks.

Issue #4001, opened 2023-02-02
[UI idea] Add start page to Parsons, explain how they work


Felienne:
**EDIT**: this issue was repurposed. Instead of removing the Quiz start page, we'll restyle it to fit better with the rest of the content.

In addition to that, we also need a new Parsons start page that explains how the puzzles work (not entirely obvious if you're coming into it without context).

------------------

The quiz was once made by a student and feels a bit different from the rest of the UI now. Could we skip the first step, and just show the 1st question when someone visits the tab (like with the Parson's Puzzles?)

![image](https://user-images.githubusercontent.com/1003685/216275501-79722844-8a30-43cb-99b2-ad61d0fdb82d.png)

jpelay:
> Hi @jpelay!
> 
> Do you think you can pick this up? It should be relatively simple and would help a lot!

Yes Miss Felienne!

rix0rrr:
I'm redoing the quizzes. I'll take this.

rix0rrr:
I'll repurpose this issue, that saves me closing this one and creating a new one :)

rix0rrr:
I'll repurpose this issue, that saves me closing this one and opening a new one 😈

PR #4819, opened 2023-11-28
[BUG] Allow users to navigate between exercises using header

### LINKED ISSUE #4441: [BUG] Parsons puzzle: second exercise is unreachable
image
I can't click the second exercise and there seems to be no other button to get me to the second exercise?

hasan-sh:
Fixes #4441 

Enhanced the overall ui of exercises.  
- Better header;  
  - Clear background and font weight on the currently selected exercise 
  - Also, users now can click to navigate through existing exercises 
- Next button doesn't show currenlty. 
  - Hiding it always is fine, since navigation is possible using the headers. 

**How to test**
- see #4441 or simply view the puzzles..

### DIFF
--- a/static/css/additional.css
+++ b/static/css/additional.css
@@ -196,3 +196,8 @@ body[dir='rtl'] #editor .ace_gutter-layer > * {
   bottom: 1px;
   right: 3px;
 }
+
+.current {
+  font-weight: bold;
+  background-color: #edf2f7;
+}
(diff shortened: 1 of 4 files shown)


Felienne:
Nice, the switching now works:

image

I am wondering though, how will kids know there is a second puzzle? The UI now does not really invite to click the 2 I am afraid. Maybe we should make the "advance" button that now goes to the quiz, go to the next puzzle?

We once had an issue coining the idea to make the quiz and Parsons more similar (#4001) but it was closed without us pursuing it further. I think we should do something like that still, make the parsons look more like the quiz with a clear advance step (we can still merge this, of course!)

Felienne:
(added the original one to the project board for the meeting, but if I am needed to discuss, feel free to skip it till next time)

PR #4825, opened 2023-11-29
[TESTING] Adds tests for unquoted strings in loop body

### LINKED ISSUE #4355: [BUG] quote issues in body give ParseError rather than Unquoted
```
svar = ask 'Vad är 5 plus 5?'
if svar is 10
    print 'Bra jobbat!"
    print 'Svaret var faktiskt ' svar
```

Compare
image

with
image

Felienne:
I was adding tests to fix #4355 but it turned out the error is already fixed! (I suspect by #4786?)

So let's just add some tests.

**How to test**
The tests pass :)


jpelay (approved):
Love when an issue has already been solved haha!

Issue #4133, opened 2023-03-16
[BUG] Better error messages with the Clear command


MarleenGilsing:
![image](https://user-images.githubusercontent.com/80678586/225599477-4cb0ad8d-39ee-451f-9c8d-91fa9b5081f2.png)
At the moment you'll get this message after you've run a good code with the clear command. Maybe we can alter this message to only appear if the clear command wasn't used. 

I would also like to request a error message to warn kids if they have used a clear command without a sleep command, becaue it looks like the code is not working.

Felienne:
Ah good catch @MarleenGilsing!! Should not be hard to fix.

Felienne:
Hi @MarleenGilsing!

> At the moment you'll get this message after you've run a good code with the clear command. Maybe we can alter this message to only appear if the clear command wasn't used.

This is fixed in #4826, so you will no longer get a warning message when the output is empty when you have used sleep. 

> I would also like to request a error message to warn kids if they have used a clear command without a sleep command, becaue it looks like the code is not working.

This, I can do too, but... is it always the case that clear is not useful without sleep? You can do this for example ((which I have seen kids do this, the program below is actually a program from a learner in my class).

```
status is ask 'How are you?'
clear
print 'so you are ' status
status is ask 'Question 2: are you an alien?'
clear
print 'You are an alien? You said...:' status
```

Or this program, where a turtle is stuck! It is a bit of a useless program but kids had a lot of fun with it :D 
```
for counter in range 1 to 10
    forward 100
    clear
```

So what exactly do we want to warn for? I can make it however you think makes sense, but we have to think about it a bit.

And maybe this issue is also helped with the (new) debugger? If kids are confused, they can see in the debugger what happens:
<img width="1418" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/9955aff5-a119-4c24-8157-757ed4ff8efd">

ok and one final idea... now that I am considering this!

In the for loop, we actually have built in a tiny sleep, because otherwise it would print 10 right away for such a program:

```
for counter in range 1 to 10
    print counter
print 'Ready or not. Here I come!'
```

We can do the same for clear, where there is always a little sleep so output is also shown.

MarleenGilsing:
You're right, clear can be useful without sleep! I didn't think of that.

I like the idea of the built-in tiny sleep! This way the students will see their output appear and disappear almost immediately, which will hopefully make them realize their mistake.

PR #4826, opened 2023-11-29
[LANGUAGE] improve clear errors

Felienne:
Wip for [#4133](https://github.com/hedyorg/hedy/issues/4133)

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -2851,7 +2851,7 @@ def get_parser(level, lang="en", keep_all_tokens=False, skip_faulty=False):
     return lark
 
 
-ParseResult = namedtuple('ParseResult', ['code', 'source_map', 'has_turtle', 'has_pygame', 'commands'])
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -562,11 +562,15 @@ def parse():
 
             if transpile_result.has_turtle:
                 response['has_turtle'] = True
+
+            if transpile_result.has_clear:
+                response['has_clear'] = True
         except Exception:
             pass
 
         with querylog.log_time('detect_sleep'):
```
Felienne, line 571: @rix0rrr I don't think I understand what is happening here, and why, do you?

Firstly, why the exception? an "in" can't really fail, can it?

Secondly what does the logging do here exactly? Why do we need to log the time that this sleep checking occurs (if I understand correctly)

rix0rrr: I think the `except` is there for safety. This used to be a function call.

The `log_time` is there to measure how long it takes to do this, in case we need to investigate slowness. With the new code shape, probably not necessary anymore either.

jpelay (approved):
It works!

PR #4828, opened 2023-11-30
[UI] Make keyword toggle also switcher language of the editor

### LINKED ISSUE #4420: [FEATURE] Show translation switcher for some users
In https://github.com/hedyorg/hedy/pull/3327 we made it so that only the admin sees the translation option in the editor:

image

But @Mark-Giesen would like to use it as a teacher also. I am open to the idea but the reason we removed it was that it is a bit much to show for student users. I would prefer an option in your profile to enable this (maybe only visible for teacher users) 

What do you all think?

Felienne:
Fixes #4420 

**Some background**

I want to describe a bit of background here because this is such a lovely example of a set of very reasonable decisions leading to bad design!

When we introduced the keyword languages, we added a switcher on the top of the screen, to switch all code examples in the adventures. So far so good. But we also switched start codes in the editor and translated them (since they have {}, the translation does not mean *parsing* but means grabbing from the yaml file and replacing the keywords). This happens when you have not written code yet. If you do, it does not load the startcode and then it does not translate.

But... Since we seems to translate code in the editor... people thought it could also translate their own code. It actually took me a long time to figure out why people (like @Mark-Giesen) thought that that is what the button would do! 

This PR makes the one button translate in the editor too. 

**Limitations**

The error handling in the translation code is currently quite limited. I tried to expand it in https://github.com/hedyorg/hedy/pull/4828/commits/e08375dbcb463d2759e437547fda26d71ddc6470 but it will involve extensive refactoring to do it nicely, so I prefer to do that separately.

I was also planning to Rename the top switcher so that it does not show the translation emoji, but the current keyword language (so: en or nl), but this leads to some design issues (https://github.com/hedyorg/hedy/pull/4828#issuecomment-1833151233) that I also want to address separately.

**Changes**

* Makes the top switcher translate the program too
* Removes the switcher in the editor for the admin 
* Gives an error message in the regular place instead of a shortlived modal
* Show the top switcher, also when people are logged in, so they can use the button to switch it -> was already the case!

**How to test**
Set Hedy to any non-En language and write code in that language (f.e. slaap 20 in Dutch, level 4). Use the toggle and observe the keywords being translated.
Create a broken program with a parse error and see the error message appear below:


Felienne:
We are getting there!

One issue though is that the keyword language is the code of the language in latin letters. Does not look pretty for, for example, Arabic:

image

Felienne:
Hi @jpelay!

I think this all works as expected (basically I just merged 2 functions) but I would like a sanity check on my typescript :)

jpelay (commented):
> Hi @jpelay!
> 
> I think this all works as expected (basically I just merged 2 functions) but I would like a sanity check on my typescript :)

So cool to have you working on the front-end too :smile: The typescript code looks fine! I have a comment not related to the front-end but to the translation process:

image

If the program has a syntax error, it doesn't get translated, and no message is shown. Perhaps we can translated the commands that are properly written or show a message? What do you think?

Felienne:
> So cool to have you working on the front-end too 😄 The typescript code looks fine! 

Thanks! I actually really likes writing a bit of typescript, I maybe want to do more of that going forward!

> I have a comment not related to the front-end but to the translation process.
> If the program has a syntax error, it doesn't get translated, and no message is shown. Perhaps we can translated the commands that are properly written or show a message? What do you think?

That is a great catch! I saw the front-end message appear so I assumed it would work always but it turned out it only worked for ParseErrors but not for error productions that are caught on the way! Turned out... (plot twist!) there was an old comment from me documenting this issue:

image (https://github.com/hedyorg/hedy/blob/a8e7bf00b08d62dba0ad317feb4f66cb25269a88/hedy_translation.py#L95)

Fixed that now!

Felienne:
A few issues remain (note to self):

* Showing of None in the switcher button
* Latin view of in the switcher button
* Position of error box

Felienne:
So the error now works, but it shown on top:

image

And not in the regular error message place. I don't like that, I don't think it makes sense. Will try to see if my typescript skills reach wide enough to fix that :D

jpelay (approved):
Everything works great now!!!

Issue #4580, opened 2023-10-02
[UI idea] Improve grid layout and stats performance


Felienne:
Last semester, two student projects ran that implemented something to give teachers a better overview of their students work:

1. The grid:
<img width="1402" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/82f22700-c6dc-46cf-b748-14db799be71a">

2. The live stats:
<img width="1453" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/2fd70e61-40f5-4e1f-9c06-d5ce6e3b4b95">

Both features add value for teachers, which is why we added them, but they also have a number of issues, I am documenting them here, we can surely split them into separate issues.

* The lower left panel more of less offers the same overview as the grid, minus the tick mark option, but they look really different. I propose to move the style from stats to the grid, and then to delete this view from the stats page. I have considered merging it all into one view, but upon further reflection, I think they represent different types of actions, so keeping them separate is better, if the look the same. We can always merge it back in if the rest of the issues have been fixed.
* The live stats uses a selector for levels, the grid (and customization page) use a dropdown, let's make the stats page into a dropdown too
* The performance of the Class Errors is horrible, for my class this results in an error often, or in a few mins (!) wait time. At the very least, we should add a front-end element showing that stuff is loading, but I think a lot of optimizations can be made
* There is weird stuff going on with a database table for errors that I don't understand, we must document that a lot better (although I suspect it can be removed all together since it seems to be used only to show stuff in the front-end?) 
* The Class Errors have an action to dismiss, but that never really works for me (could also be performance?)

Felienne:
I made a few notes about slow lines here @jpelay, maybe that helps! https://github.com/hedyorg/hedy/tree/pretty-grid

I will also continue to figure out what is happening a bit more!

MarleenGilsing:
I'd like to add that in the grid overview the eyes get in the way of checking off the boxes. I propose that all the colums get the same width to solve this. 

![image](https://github.com/hedyorg/hedy/assets/80678586/7eaf9edc-a3d4-47a8-a143-cab5721d26fb)

PR #4834, opened 2023-11-30
[UI, FIX] Fixes placing of checkbox and eye icon on grid

jpelay:
**Description**

Changes the table elements to be each a table layout of 3 columns, where the first column is empty, for alignment, the second one is the checkbox and the third one is the eye. This fixes the overflowing behavior that the elements of this table presented.

Worth noting that this page will be moved from here to the dashboard, but we need this fix shipped because it's very annoying!

**Fixes [this comment](https://github.com/hedyorg/hedy/issues/4580#issuecomment-1803884643) on #4580**

**How to test**

* Make sure that some students have adventures done for a particular class.
* Go to the class and click on Overview of programs per adventure.
* Zoom-in and out and see that the elements do not overflow or mess with each other

### DIFF
--- a/templates/customize-grid/partial-grid-levels.html
+++ b/templates/customize-grid/partial-grid-levels.html
@@ -20,17 +20,27 @@
             {% set adventure_id = adventure_names[class_adventures[level][adventure_index]] %}
             {% set student_adventure_id = student ~ '-' ~ adventure_id ~ '-' ~ level %}
             {% if checkbox is not none %}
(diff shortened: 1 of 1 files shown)


Felienne:
I wanted to review this, but I don't think I know how to get the issue in view? For me it looks fine in main?

image

Felienne (approved):
Ah yes, got it now!

image

This PRi indeed fixes the issue.

PR #4840, opened 2023-12-01
[UI] Duplicate second teachers in class

### LINKED ISSUE #4653: [FEAT] (Re)invite/add existing students when duplicating a class!
The following function needs to be updated https://github.com/hedyorg/hedy/blob/ccc2f1f1d67d600d3ff8bed8b7e57305d573d5c1/website/classes.py#L185-L228 

Currently, duplicating a class only duplicates its customization. However, the class is empty and the teacher needs to invite (the same) students all over again. This also applies for second teachers; see #4644

Annelein:
Fixes #4653 

Before the input for the name of the duplicated class:

Screenshot 2023-12-01 at 20 03 47


Felienne:
Hi @Annelein!

This seems to work, but I sm not sure I love the UX? (We don't have a style guide now, so this is totally not on you!) Two things come to mind:

* The colourful modals are great for kids (who they were designed for) but I feel they are a bit weird for teachers. Thinking of it, that is also true for the survey! 
* Should this be a popup? I can also imagine we make a toggle with "copy all teachers" next to the copy button, which feels a bit less disruptive. Or a dropbown with options (copy with teachers, copy without teachers)

Let's discuss this in the next meeting?

Annelein:
@Felienne Maybe it would be fine to just add this feature now and I can create an issue: 'New teacher modals instead of using kids modals'. 

Because not only the survey but also deleting a class, creating a class, customizing a class, all use the kids modals. So it would be nice to create a whole new look for all of those and I also prefer the idea of having a toggle somewhere instead of the popup. Let me know what you think!

static/js/teachers.ts
```diff
@@ -54,18 +54,43 @@ export function rename_class(id: string, class_name_prompt: string) {
...
-export function duplicate_class(id: string, prompt: string, defaultValue: string = '') {
+export function duplicate_class(id: string, second_teacher_prompt: string, prompt: string, defaultValue: string = '') {
+  modal.confirm(second_teacher_prompt, function () {
+    callDuplicateClass(id, prompt, true, defaultValue);
+  }, function () {
+    callDuplicateClass(id, prompt, false, defaultValue);
+  });
+}
+
+function callDuplicateClass(id: string, prompt: string, second_teacher: boolean, defaultValue: string = '') {
```
hasan-sh, line 65:
```suggestion
function apiDuplicateClass(id: string, prompt: string, second_teacher: boolean, defaultValue: string = '') {
```
Also, don't forget to change it where it's being called. I do like that you use camel case variables :)) I want to convert all variable/function names in TS to camel case. But that's in the future, and great to start from now!

templates/for-teachers.html
```diff
@@ -31,7 +31,7 @@ <h2>{{_('my_classes')}}</h2>
                   <td class="px-4 py-2"><a href="for-teachers/class/{{class.id}}" class="view_class" data-cy="view_class_link">{{class.name|e}}</a></td>
                   <td class="username_cell">{{class.teacher}}</td>
                   <td class="text-center p-2">{{class.students|length}}</td>
-                  <td class="text-center p-2"><a class="no-underline cursor-pointer" onclick='hedyApp.duplicate_class("{{class.id}}", {{_('class_name_prompt')|default(None)|tojson}}, {% if class.teacher != username %}"{{class.name}}"{% else %}" "{% endif %})'><span class="fas fa-copy"></span></a></td>
+                  <td class="text-center p-2"><a class="no-underline cursor-pointer" onclick='hedyApp.duplicate_class("{{class.id}}", {{_('second_teacher_copy_prompt')|default(None)|tojson}}, {{_('class_name_prompt')|default(None)|tojson}}, {% if class.teacher != username %}"{{class.name}}"{% else %}" "{% endif %})'><span class="fas fa-copy"></span></a></td>
```
hasan-sh, line 34: if there's no second teachers, i still see the question about copying second teachers!!

hasan-sh (changes requested):
Great work overall;))

templates/for-teachers.html
Annelein: Good catch thanks!

hasan-sh (approved):

PR #4843, opened 2023-12-04
[LANGUAGE] add test and exception

Felienne:
WIP for remainder of #4133


Felienne:
Leaving this for now, probably the little pause suffices!

PR #4846, opened 2023-12-04
[UI] Localizing level keywords

### LINKED ISSUE #4570: [CONTENT] on "My programs" page, localization does not work for Adventure names
When is "my programs" page, adventure names show their literal values without translations such as `{if} & {else}`
This happens at least for Turkish

adventure name shown on saved program
Screenshot (4)

and adventure names in the dropdown list
Screenshot (5)

**Expected text**
Localization process should include these parts of the page and translate them. `{if} & {else}` would then be `eğer & değilse`

Annelein:
Fixes #4570

How to test: Go to the My Programs page, click 'Adventures', the command adventures are now localized.


Felienne:
Ah I see! You have translated the names but left the curlies around them:

image

image

I did not see that from print since print in Dutch is also print haha. 

You see now how useful the How to test is :)

I think I would prefer we remove the curlies too? They don't look pretty and they are used to indicate places where localization can happen internally.

hedy_content.py
```diff
@@ -506,8 +506,15 @@ def __init__(self, language):
...
 
-    def get_adventure_names(self):
-        return {aid: adv['name'] for aid, adv in self.file.get('adventures', {}).items()}
+    def get_adventure_names(self, keyword_lang):
+        adventure_names = {aid: adv['name'] for aid, adv in deep_translate_keywords(
+            self.file.get('adventures'), keyword_lang).items()}
+        for key, value in adventure_names.items():
+            if 'command' in key:
+                words = value.split()
+                adventure_names[key] = ' '.join(['{' + word + '}' if not word.isdigit()
```
Felienne, line 515: Because we already have the curlies in the string, you can actually do this in an easier way, by applying a dictionary to the string directly! I don't know the syntax by heart but it occurs in a number of places in the code. Let mw know if you need help!

Felienne: I think it is this function: https://github.com/hedyorg/hedy/blob/be7fd1e970fc050d584ad91dbc156f1e8728197e/safe_format.py#L4

Felienne (changes requested):

Annelein:
@Felienne sorry I didn't add a test! I wanted to add a screenshot but I can't screenshot the Adventure options on the MyProgram page... So forgot to write a how to test haha.

hedy_content.py
Annelein: I don't think there are curlies in the string yet right? Because it retrieves 'Slaap' from here? Or maybe I don't really understand what you mean?

Screenshot 2023-12-05 at 09 29 48

Felienne:
> @Felienne I added the curlies because I thought it would be nice to indicate keyword adventures. Because like its done here:
> 
> Screenshot 2023-12-05 at 09 19 39
> But I understand now the curlies are for where localization can happen internally. Let me know if you still think I should add something as an indication.

That is a great question, should we indicate it in any way? I would think it is fine to not do it, if you want to (and if it is possible) we could add a subtle link line in the dropdown? But I can totally live without it!

Annelein:
@Felienne I removed the curlies, so I think it is ready!

Felienne (approved):
Works perfectly now!

image

PR #4848, opened 2023-12-04
[LANGUAGE] improved error productions for repeat

akseron:
Changes in error production to support better syntax error messages for the repeat command. Currently deprecated repeat commands give a generic parse error from level 9 onwards.


akseron:
[IMPORTANT] added support in english for error production regarding 'is ask'
Cannot add support for hedy as it is right now, lark only detects _IS with a space before it because hedy the keywords only contain 'is' with a space before and after. It is most likely better to avoid using _SPACE in the keywords and delegate it to lark levels for better customisability. As a band-aid measure I added a _STANDALONE_IS to the english keywords meaning only the english version of hedy currently supports an error message.

PR #4849, opened 2023-12-04
[UI] Highlight variables

### LINKED ISSUE #4853: [UI idea] Highlight variables
One feature that could really help students is the highlighting of variables. Now that we've merged CodeMirror in, we can start working on this.

When I was first starting to work on the CodeMirror grammars, I wanted to parse variables directly on the grammars, but at the first few levels we really have no way to distinguish between text and variables. The way to move this forward then, is by using the Tree lezer generates and iterating trhough it.

jpelay:
**Description**

In the first few levels is really hard to differentiate between text and variables, this makes highlighting them directly after parsing very difficult. In this PR I'm exploring if highlighting them manually is possible: first we generate the tree, then we iterate over this tree, we enter the `Assign`, `Ask` and `AssignList` nodes and highlight the first Text node. Then we'd need to find these same strings over the other commands where the variables can be used and highlight them as well.

This will cause some overhead, and maybe be a little slow. I'm checking!

**Fixes #4853**


jpelay:
We've got advances!
image

jpelay:
Not so bad!

image

jpelay:
Undefined vars are not highlighted, in the future we can perhaps add a sort of linting to Hedy
image

jpelay:
Hi, @Felienne! This is pretty much done!!, but there are a couple things that I want to ask?
1. Do you want to highlight function names with a different color?
2. What should we do with functions parameters? Do you want to highlight them with the same color as other variables? For  the moment I don't even highlight them because managing the scopes will need a bit more work :sweat_smile:

Can you try this and tell me what you think?

Felienne:
Hi @jpelay!

I love that we are pushing this forward!!!!! Long desired feature.

And yes a linter would also be nice!! Maybe we can even make the undefined ones have a squiggly in the future (although my editor also does this gray as you show it)

One thing that occurred to me is that if we do it in the editor, we also have to do it in the explanation biocks, or it will be confusing:

image

To answer your questions on functions:
I would say yes for clarity for function names and parameters to both have different colors, but... now that I look at code it is very very colorful maybe a bit overwhelming?
image

Maybe the variables should be white and the plain text strings yellow (since in high levels we don't really have the mixed strings anymore) That opens up a bit of space maybe to color parameters differently?

Felienne:
ok and now I am thinking of a stretch goal, just for fun... different color themes?!

jpelay:
> And yes a linter would also be nice!! Maybe we can even make the undefined ones have a squiggly in the future (although my editor also does this gray as you show it)

I can do that, although at least until level 12 it'd be difficult to know what's Text and what's an undefined_var, at least in comparisons and assignments, on prints and asks is totally doable.

> One thing that occurred to me is that if we do it in the editor, we also have to do it in the explanation blocks, or it will be confusing

You are correct! I think I should first replace the example blocks with CodeMirror before moving forward with this.

> now that I look at code it is very very colorful maybe a bit overwhelming?

That's true! Perhpaps yellow is too bright, maybe a light blue could do it? Something like this:
image

> Maybe the variables should be white and the plain text strings yellow (since in high levels we don't really have the mixed strings anymore) That opens up a bit of space maybe to color parameters differently?

This is also a possibility, I will try both and let you know

Felienne:
> You are correct! I think I should first replace the example blocks with CodeMirror before moving forward with this.

Sounds good!

En for the colors' let's just experiment a bit! We might also ask @AnneliesVlaar as user/expert for her input?

AnneliesVlaar:
Very cool to see the highlighted variables!

Some thoughts from my side:

Why do strings have different colors in the low levels? Initially white and then they turn purple with quotation marks. I would suggest to keep strings the same colour independent of the syntax. Then there is visual feedback if the syntax is not right. When variables are yellow, undefined variables are white and strings purple, then even in the low levels you can distinguis strings from undefined variables. 

define and call are now blue/green, but **all** the other commands are pink (print, loops, conditional statements, operators, range, random, input, etc.). Therefor define and call can also be pink, this leaves room on the color palette to give the function name a different color than variables.

jpelay:
This PR has a bunch of commits that aren't in main, so closing this in favor of #4957

PR #4859, opened 2023-12-05
[UI] Fix highlighting of return command and list accesses

### LINKED ISSUE #4856: [BUG] Return not being highlighted
This happens because I messed up some variable names
image

### LINKED ISSUE #4857: [BUG] Assignment with at not being highlighted if variable is used
If I do `list at 1 = 3` it works, but not with Text

image

jpelay:
**Description**

* Fixes the incorrect name that was used to get the `return` value from the highlighting json
* Adds the `Int` and `Number` rule to ListAccess when appropriate. Functionally this does nothing since they were still being highlighted, but now the trees we generate don't have any error nodes.
* Allows use of list accesses on left hand side of assignments

**Fixes #4857 and fixes #4856 **

**How to test**

Go to level 12 and write this program:

```
define sum with n1, n2
    return n1 + n2
```
Check that return is being highlighted as a keyword

Now go to level 16 and type:
```
print l at [1]
```
The closing bracket should be highlighted

Go to level 12 and type:

```
list at number = 1
```
The variables and the at should be highlighted


Felienne (approved):
Works!

image

image

image

Issue #4678, opened 2023-10-27
🧹 Document database schema in database.py


hasan-sh:
**Description**
Currently, we there's some comments describing certain fields, but not all. Also, how do we know which fields a table hold if a some PR adds a field that's needed?

**Idea**
We should have an organized schema where all fields are described and can be read in their respective table.  So, create a catalog for every table of our database. For instance, in one of the following tools:
- https://www.dbvis.com
- https://dbdiagram.io/home 
- or simply do it with markdown!

Felienne:
I think you mean "document" instead of refactor? I think this is a excellent idea! If we go down this road, we should also add a new point to the PR checklist, asking if there are DB changes, and checking if the docs are updated too!

Felienne:
Not a small issue @hasan-sh but something you can pick up in between other things (one table at a time maybe?)

rix0rrr:
We can add the types to the `Table`, which can do things like automatically generate a Markdown report and validate records as they are being inserted into the table.

See a previous attempt for this here: https://github.com/hedyorg/hedy/pull/4862

PR #4862, opened 2023-12-06
[FEAT] Check field types in database emulation

rix0rrr:
There is currently a problem with a table that has an index on a certain field, but is trying to insert values of a type not matching that index. It fails in Dynamo, which requires that all values used as index keys are of a single declared type (either string, number or binary).

However, the local emulation layer doesn't catch that, because it doesn't know anything about types that need to be checked.

Add a required `types` argument to `Table`, which must declare at least the key fields (of the table itself and of the indexes), and can optionally declare the types of other fields. It will be an error to insert values of the wrong type into the database.

This may look like an easy feature, but it has some edge cases you need to be aware of:

* First of all, it relies on users declaring the indexes in code; they can choose NOT to declare the indexes in code, and there may still be an error after deploying.
* Second of all: just because you declare a type on a fields saying what may be inserted into a table going forward, this can not make any guarantees on data already in the table. Don't use this feature to make any assumptions on data you read out of the table, only use it to sanity-check new data going into it!

We may decide this change is not worth it... (for example, it would not have caught the current problem, as the index was there in Dynamo but not declared in code, so this feature wouldn't have caught it).


rix0rrr:
Let's see how badly this breaks on the Cypress tests... :)

rix0rrr:
Not sure anybody really wants or needs this, so I'll be closing it.

website/database.py
```diff
@@ -18,29 +47,93 @@
...
+
+USERS = dynamo.Table(storage, 'users', 'username',
+                     types=only_in_dev({
+                         'username': str,
+                         'email': str,
+                         'epoch': int,
+                         'created': int,
+                         'classes': Optional(SetOf(str)),
+                         'keyword_language': Optional(str),
+                         'password': str,
```
jpelay, line 69: I like this!!! I think this makes a lot of sense. More typing annotations on our DB would be a huge win. I think the work now would be to adapt the definitions of all the tables. I can handle that!

rix0rrr: 🙏 that would be great, thanks! You probably also have a better view into what the currently expected data contract is

Felienne:
I am picking up #4678 so it would be nice if we can merge this 🙏

jpelay:
@rix0rrr Do you know why when I run the server it always run on production mode? I realized that the type checking always passed, and it was because `DEBUG_MODE` is set to `False` does it also happen to you?

jpelay:
Ohh I think I get what's happening: `only_in_dev` is being called before `DEBUG_MODE` is set to `True`

utils.py
```diff
@@ -88,7 +88,7 @@ def times():
     return int(round(time.time()))
 
 
-DEBUG_MODE = False
+DEBUG_MODE = not os.getenv('NO_DEBUG_MODE')
```
jpelay, line 91: Fixed it for now, but maybe this is not the cleanest way?

rix0rrr:
> Ohh I think I get what's happening: `only_in_dev` is being called before `DEBUG_MODE` is set to `True`

Oh shit, yes that makes sense.

utils.py
rix0rrr: Doesn't this always set it to `True` ?

jpelay: I don't think so, it's what we are already doing: https://github.com/hedyorg/hedy/blob/6e7573d8710d90079385614e617f0581d7bde732/app.py#L2908-L2912 . Just that this get executed before

rix0rrr: I think the better solution is probably to control the initialization order of all these things better, by making the tables class members of Database and initializing it when the app is created

jpelay: Yeah, that makes sense too.

website/dynamo.py
```diff
@@ -1566,3 +1647,139 @@ def make_predicate(obj):
...
+        self.inner = Validator.ensure_all(inner)
+
+    def is_valid(self, value):
+        if not isinstance(value, dict):
+            return False
+        first_type = list(self.inner.keys())[0]
+        if isinstance(first_type, InstanceOf):
+            return all(first_type.is_valid(k) and self.inner.get(first_type).is_valid(v) for k, v in value.items())
+        else:
+            return all(validator.is_valid(value.get(key)) for key, validator in self.inner.items())
```
jpelay, line 1782: Rico, what do you think about this? I changed that you now can specify `str` in the left-hand side of the validator dictionary. This way we can be a little more generic in our type checks, like this one:

```
'sorted_adventures': RecordOf({
    str: ListOf(RecordOf({
        'name': str,
        'from_teacher': bool
}))
```

It's important to note that you can only use `str` and no other type, since that would go against Dynamo naming rules either way. This is just a way to relax the type checking a little bit without needing to use `dict`.

jpelay:
Seems like adding an element to a set is giving errors, for example adding a student to a class gives this:

```
    CLASSES.update({"id": class_id}, {"students": dynamo.DynamoAddToStringSet(student_id)})
  File "/home/capybara/hedy/website/querylog.py", line 211, in wrapped
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/capybara/hedy/website/dynamo.py", line 547, in update
    self._validate_types(updates, full=False)
  File "/home/capybara/hedy/website/dynamo.py", line 695, in _validate_types
    raise ValueError(f'In {data}, value of {field} should be {validator} (got {value})')
ValueError: In {'students': Add(('user1',))}, value of students should be optional set of instance of <class 'str'> (got Add(('user1',)))
```

website/dynamo.py
rix0rrr: A "record" (a set of known fields, with known and potentially different types for every value) is a different kind of type than a "map" (can be any field, but the types are usually the same).

Examples:

```py
person = { 
  "name": "Bertus",
  "age": "67"
}
  
favorite_restaurant_in_city = {
  "Oegstgeest": "Suzie's & Van Aken",
   "Amsterdam": "Ron's Gastropub",
}
```

In Python, you might use a `dict` to represent both of these, but the type-of-type is very different! 

They are:

```py
person :: { "name" -> str, "age" -> int }

# but
favorite_restaurant_in_city :: { str - > str }
```

So I wouldn't overload `RecordOf` in this way, because it looks like it will just get confusing. Also, it would allow you to construct this... now what does that mean?

```py
RecordOf({
  str: str,
  "age": int
})
```

I would make a `MapOf(str, str)` or `DictOf(str, str)` or something.

But you have to explain the differences very clearly, because I can see if even you are confused about it, there is a good chance other people will also not realize the difference.

rix0rrr:
> Seems like adding an element to a set is giving errors, for example adding a student to a class gives this:

You're right, I made a mistake. Fixed it and added a test.

website/dynamo.py
jpelay: > I would make a MapOf(str, str) or DictOf(str, str) or something.

Hmm I think this would make sense! In essence what I'm trying to do is relax the checking a bit, so another name for that type of validation makes a ton of sense.

> So I wouldn't overload RecordOf in this way, because it looks like it will just get confusing. Also, it would allow you to construct this... now what does that mean?
> ```python
> RecordOf({
>  str: str,
>  "age": int
> })
> ```

I thought of that! If you use `str` in the left hand side, then you can't use a named field (and all other classes aren't allowed)

https://github.com/hedyorg/hedy/blob/dd20d554e9d2485dfbc038b2d3ba59fd94f49a5a/website/dynamo.py#L1677-L1680

> But you have to explain the differences very clearly, because I can see if even you are confused about it, there is a good chance other people will also not realize the difference

Yes, I need to add something to the wiki and to the comments.

jpelay:
@rix0rrr I think I'm finished tweaking the PR! Added types for all of the tables and also added two new type validators: `Or` and `DictOf` as you suggested. What do you think?

rix0rrr:
❤️  Love it! Instead of `Or()`, consider perhaps naming it `Either()` or `OneOf()`, but this is great stuff!

Still not sure about the initialization of `only_in_dev()`... but if you feel comfortable with it, I feel good about it too!

Feel free to ship this at any point (since I'm the author, you need to be the shipper 😉 )

jpelay:
> ❤️ Love it! Instead of `Or()`, consider perhaps naming it `Either()` or `OneOf()`, but this is great stuff!

Damn, you're good at coming out with names! 

> Still not sure about the initialization of `only_in_dev()`... but if you feel comfortable with it, I feel good about it too!

I think it's best to tackle that in a new PR, so ship this one and then take your suggestion of explicitly initializing the tables in a new PR. @Felienne  wants this PR ready because it'll help her with the live documentation of the tables.

> Feel free to ship this at any point (since I'm the author, you need to be the shipper 😉 )

Understood 🫡

rix0rrr:
Thanks for taking this across the finish line, and fixing my gigantic fuckup of the code not actually doing anything 🫣🫠😆

jpelay:
> Thanks for taking this across the finish line, and fixing my gigantic fuckup of the code not actually doing anything 🫣🫠😆

Hahaha the core idea was so cool already! Thanks a lot for that!!

jpelay (approved):
Let's go! And hope nothing breaks !

PR #4868, opened 2023-12-07
[FIX] Use string index in surveys

jpelay:
**Description**

Casts the index of the survey fields to be a string, it was failing on alpha because we were passing a number.

**How to test**

It should work normally, but the real test will be on Alpha


hasan-sh (approved):

jpelay:
I tested in Alpha and this seems to be working properly!

PR #4867, opened 2023-12-07
[LANGUAGE] Check translations

Felienne:
More extensive testing of the translation part (which is now a lot more exposed since #4828)

PR #4875, opened 2023-12-08
[LANGUAGE] Allow for translations of erroneous programs

Felienne:
(This is a redo of #4867 which contained some commits now already on main)

Since https://github.com/hedyorg/hedy/pull/4828 translation features are more prominently visible, so I figured we need to do a bit better. This actually changes some conceptual assumptions, so let's document here:

* Programs that have a ParseError or contain Placeholders, or have indentation issues, still cannot be translated (since they can't be parsed and then we don't have a tree to work wtih)
* All other errors however create a parsetree with error nodes, and thus... can just be translated keeping the error nodes, so that is what we now do!

**How to test**

Try a program with an error production, f.e. in German, level 7:
`repeat 3 times 'Hedy macht Spaß!'`

Main shows an error:
image

This PR translates the code:
image


Felienne:
(oops closed by mistake!)

jpelay (approved):
It works nicely! This is very cool!!

PR #4876, opened 2023-12-08
[LANGUAGE] Remove invalid space

Felienne:
Removes the "invalid space" exception since it is no longer needed, the editor removes spaces when needed.


Felienne:
Oops! This warning was in use just not shown!

PR #4880, opened 2023-12-09
[FIX] Add warning back in

### LINKED ISSUE #4869: [BUG] InvalidSpaceWarning not being shown in the editor
**Describe the bug**

The editor is not showing the warning for `InvalidSpace` and maybe for other warnings that come from the server?:

image

Felienne:
Fixes #4869 

As far as I can see, @ToniSkulj removed this warning in #4320, I suspect by accident, but if not, let me know :) 

See:
image

**How to test**
Observe that code with a space at the start runs but also throws an error:

image


Felienne:
Would be a good idea maybe to add a Cypress tests for this @jpelay?

jpelay (approved):
Thank you!! This was probably an oversight by us.

Felienne:
> Thank you!! This was probably an oversight by us.

Np! Do we want to add an issue to make a test so we don't forget?

jpelay:
> > Thank you!! This was probably an oversight by us.
> 
> Np! Do we want to add an issue to make a test so we don't forget?

Yeah, Im on it!

PR #4881, opened 2023-12-09
[LANGUAGE] Warning unused var

### LINKED ISSUE #2394: [LANGUAGE] Add warning for unused variables
My students often create a variable that they don't use. For example:
```
name is Fatih
print 'my name is Fatih'
```
It would be nice to raise a Warning for each variable that was created but not used.

This could also help a lot in level 5 with if/else. I see a lot of students doing this:
```
name is print 'what is your name?'
if naam is gvr print 'ok' else print 'ai'
```
This code always prints 'ai' because the check (`naam is gvr`) is always false. It can be hard for a student to discover why. Raising a warning would inform the student that they've not used the variable `name` in the if-statement.

### LINKED ISSUE #4884: [BUG] More broken fix faulty tests
Hi @ToniSkulj!

I am afraid I am wrecking more havoc on your tests, sorry! See: #4881 

I commented out the broken tests for now. Could you take a look 🙏?

(I suspect that #4879 too might have something to do with the skip faulty code, I took a look at what is going wrong but it does seem to raise the right exception, it looks like it is just caught somewhere in your code)

Felienne:
Fixes #2394 (finally!!) by checking if declared variables are actually used, also fixes #4884 that was introduced here.

Depends-On: #4880

**How to test**

I have added a test for the exception. To see it in action on the front-end, run code that defines but not uses a variable, and observe you get a warning but the code still runs:

image

Note that this PR needs #4880 to show the warning in the front-end!


jpelay (approved):
I tested it and this works great! I wonder if in the future we'd want to show the user several warnings or errors at the same time? Or do you think this would overwhelm them?

Felienne:
> I tested it and this works great! I wonder if in the future we'd want to show the user several warnings or errors at the same time? Or do you think this would overwhelm them?

Ow great question! I have to think about that a bit more!

PR #4838, opened 2023-12-01
[CHORE] move exceptions into isvalid

Felienne:
wip for #4836

PR #4882, opened 2023-12-09
[BUG] Fix broken tests

### LINKED ISSUE #4863: [BUG] New exception system creates issues in skip faulty
Hi @ToniSkulj!

In #4838 I have made a change to how we process invalid code, but it breaks two of your tests (see: https://github.com/hedyorg/hedy/pull/4838/commits/cf1b87fca76889ccc3a20d492f433c58db983557) I dod not know how to fix those, could you maybe take a look?

I suspect that now we throw an exception when we find the first error (rather than processing the remainder of the tree also) might require a bit of a change to your handling of code with multiple errors? Hope you manage to make it back in order, thanks!

ToniSkulj:
Fixes #4863.

After #4838, the processing of the string was done after setting the input string on the source map.
This led to 2 tests failing.

This PR first uses process_input_string before setting the input on the source map. Fixing the failing tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,8 +149,9 @@ logs.txt
 .local
 node_modules
 
-# Ignore things that start with underscores
+# Ignore things that start with underscores (except __init__.py)
 _*
+!__init__.py
 # Ignore pickled versions of YAML files
 *.yaml.pickle
 dev_database.json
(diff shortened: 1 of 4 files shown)


Felienne (approved):
Thanks for the quick fix @ToniSkulj!!

#4881 has similar issues, I hope this fixes those too!

PR #4883, opened 2023-12-11
[UI] Dropdown class

Annelein:
Fixes #4759 

How to test: test the dropdown classes
@hasan-sh is this what you meant?


Felienne:
Hi @Annelein!

In the interest of cleaning up PRs, do we want to keep this open?

PR #4891, opened 2023-12-12
🧪 update programs and add parser filer

Felienne:
PR for @akseron to test parse errors.

How to use this:

* Run list_public_parse_errors.py
* See programs with a parse error:

for example this one:

image

* Be inspired for new error productions :)


Felienne:
Sadly there are not so many (I guess because people don't share erroneous programs...) I will dive into non public programs too but I can;t share these publicly of course

akseron (approved):

PR #4892, opened 2023-12-12
[UI] no pop up for adventure name

### LINKED ISSUE #4889: [UX] No pop up for adventure name
This issue follows #4888 a bit and #4885. When you make a new class, you also get a popup:

image

But then when you are at the adventure page, there is a name field also:

image

We should make the name an edit box so that we can skip the popup. We can generate a simple name (AdventureX) or a cute funny name as many sites do now (HappyPandaAdventure).

Annelein:
Fixes #4889 

When creating a new adventure, it automatically gets the name 'AdventureX' and so on:
Screenshot 2023-12-12 at 11 35 56

And there is no popup but you automatically get redirected to customize class with an autofocus on the name so you can edit it:
Screenshot 2023-12-12 at 11 36 26


templates/for-teachers.html
```diff
@@ -96,10 +96,12 @@ <h2 class="m-0"> {{_('my_adventures')}} </h2>
...
-          <button class="green-btn mt-4" id="create_adventure_button"
-          onclick='hedyApp.create_adventure({{_('adventure_prompt')|default(None)|tojson}})'
-          {% if second_teacher and role=='viewer' %}disabled{% endif %}>{{_('create_adventure')}}</button>
-          <a href="/public-adventures" class="white-btn mt-4">{{_('public_adventures')}}</a>
+        <button id="create_adventure_button" class="green-btn mt-4" 
+        hx-trigger="click"
+        hx-get="/for-teachers/create-adventure"
+        _="on htmx:afterRequest if detail.xhr.response
+        window.open('/for-teachers/customize-adventure/' + detail.xhr.response, '_self')"
+        data-cy="edit-link"{% if second_teacher and role == 'viewer' %}disabled{% endif %}>{{_('create_adventure')}}</button>
```
hasan-sh, line 104:
```suggestion
        <button id="create_adventure_button" class="green-btn mt-4" 
            hx-trigger="click"
            hx-get="/for-teachers/create-adventure"
            hx-swap="none"
            _="on htmx:afterRequest if detail.xhr.response
                window.open('/for-teachers/customize-adventure/' + detail.xhr.response, '_self')"
        data-cy="edit-link"{% if second_teacher and role == 'viewer' %} disabled{% endif %}>{{_('create_adventure')}}</button>
```

website/for_teachers.py
```diff
@@ -971,32 +974,26 @@ def parse_preview_adventure(self):
             return gettext("something_went_wrong_keyword_parsing"), 400
         return {"code": code}, 200
 
-    @route("/create_adventure", methods=["POST"])
+    @route("/create-adventure", methods=["GET"])
```
hasan-sh, line 977: Please keep this as a POST method since it indicates what this function actually does. You can still return whatever you want (the id in this case!)

hasan-sh (changes requested):
Amaing work overall :)

hasan-sh (approved):

PR #4894, opened 2023-12-12
[TESTS] Upgrade and enhance Cypress tests

hasan-sh:
**Describe the bug**
Cypress tests have been a great pain over the past few months due to the amount of new features we are introducing. Currently, we use the Cypress github actions that's maintained by the Cypress team to run the tests, this will stay the same but the used action will be upgraded to the latest (v6). Lately, the tests fail due to some vague reasons. I say vague because they pass locally but not on github actions. 

After a lot of digging and researching, I _believe the problem is_ that Cypress isn't accessing the latest of the bundle files that we create. Thanks to Felienne, Cypress dashboard gives so many different possibilities to debug why tests fail. For instance, the following ss informs us that `initTE` is not a function or the fact that we cannot read the property `initialize` from a variable that's `undefined`, in the login page. All of this indicates something wrong in the `appbundle.js` file.

Screenshot from 2023-12-12 10-07-33

Another reason why the build files may be the problem is that sometimes we would rebuild the tests and they would pass, other times we would have to add a wait command. Also, in such cases, usually pushing the bundle files (i.e., `appbundle.js[.map]`) would make the tests pass on Cypress CI as well!

-------------- 

It's worth mentioning that we're not the only ones having such issues, but many people have reported similar ones in Cypress and Cypress github-action github repositories. And what I described above may not entirely/always solve the problem, as the [documentation lists several possible reasons.](https://docs.cypress.io/faq/questions/using-cypress-faq#Why-do-my-Cypress-tests-pass-locally-but-not-in-CI).


.github/workflows/cypresstests.yml
```diff
@@ -40,18 +40,25 @@ jobs:
       # This replaces the json file, do it before starting the server
       run: ./feed_dev_database.sh
 
+    - name: Bundle files
+      # Generates js and css files.
+      run: |
+        build-tools/heroku/tailwind/generate-development-css
+        build-tools/heroku/generate-typescript --tests
```
rix0rrr, line 47: Can you add one commit where you remove this flag, to prove that this behavior is what is making the difference?

```suggestion
        build-tools/heroku/generate-typescript
```

Felienne: Let's see what breaks if we commit this :)

.github/workflows/cypresstests.yml
```diff
@@ -40,18 +40,25 @@ jobs:
       # This replaces the json file, do it before starting the server
       run: ./feed_dev_database.sh
 
+    - name: Bundle files
+      # Generates js and css files.
+      run: |
+        build-tools/heroku/tailwind/generate-development-css
+        build-tools/heroku/generate-typescript
```
rix0rrr, line 47: I think that the addition of this command here is the load bearing part!

rix0rrr (approved):

PR #4898, opened 2023-12-12
[FIX] Return token in Lezer parser only when it applies

### LINKED ISSUE #4772: [BUG] `to` keyword not being highlighted correctly in Spanish
Note to self
image

jpelay:
**Description**

On Spanish, Bulgarian, and possibly other languages, there was an error when highlighting `for` commands, because in these languages the word for `in` and `at` is the same. This PR fixes that and also fixes it in other possible occurrences of the same problem by only returning the token if the parser can shift the state were the token be accepted, for example take into consideration the following code of level 10:

```
para animal en animales
    imprimir animal
```
The English version of this program is:

```
for animal in animales
    print animal
```

When the parser reaches the token `en`, the stack already has: `[For, Text]`. Then it iterates through every keyword relevant for this level in the highlighting.json dictionary and asks: Does `en` matches the regex for the current keyword? If it does, it then asks, can the parser shifts if I accept this token? If the answer is yes, it is accepted, if it ins't it continues to the next token, and finishes when it finds one that it can accept or it returns -1, in the case than none is accepted. 

The problem was in this last step: we didn't ask if the parser would be in a good state if we accepted the token, resulting in accepting the first token that was found.

**Fixes #4772**

**How to test**

Go to level 12, set the language to Spanish and write:

```
para nombre en nombres
    imprimir nombre ' pertenece a ' casas en aleatorio
    imprimir nombre ' es mejor en ' asignaturas en aleatorio
    imprimir nombre ' su gran miedo es ' miedos en aleatorio
```


jpelay:
This break nested ifs :(

image

jpelay:
I think for this to move forward I need to fix the grammar a bit, and handle indentation properly.

jpelay:
> This break nested ifs :(
> 
> image

Solved it!!! @Felienne this is ready for review :D
image

jpelay:
@Felienne This PR is ready now! I also added some tests for levels 5, 6 and 7, please take a look whenever you want

Felienne (approved):
Works!

image

PR #4901, opened 2023-12-13
[BUG] prevent sutdents being in multiple classes

hasan-sh:
- prevent students from joining more than 1 class
- translation keys

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -1,22 +1,3 @@
-# Translations template for PROJECT.
-# Copyright (C) 2023 ORGANIZATION
-# This file is distributed under the same license as the PROJECT project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2023.
-#
-#, fuzzy
-msgid ""
-msgstr ""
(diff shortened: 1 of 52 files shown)


Felienne (changes requested):
Hi @hasan-sh!

I can't test this locally since it has the broken css files, and because of the conflicts I can't update. Can you fix the updates so I can test?

Felienne:
Hi @hasan-sh!

I guess the easiest quick fix here (what I usually do in this type of situations where all po files have the same due to a new key) is to accept all changes from main with --theirs and then run pybabel extract again.

Felienne:
Ah and it seems that you used the no header option on this pr originally too: 
image

So if you run the codes again do include the header.

let me know if you need further help!

Felienne:
Hi @hasan-sh!

Trying with an invite indeed gives the right error, but joining though a link shows:

image

Felienne (approved):
Hi @hasan-sh!

The po issues have been solved in main, so this can go in!

PR #4903, opened 2023-12-13
[LANGUAGE] Adds music to the language

Felienne:
**How to test**

Paste this code in level 1 and be amazed!! 🤣 

```
play C4
play D4
play E4
play C4
play C4
play D4
play E4
```

TODO/Notes to self:

- [x] Syntax highlighting
- [ ] Outline a few adventures for the content developers to work on
- [x] Should we also support integer values, like this: `play 21` so that we can do nice calculations with sounds?
- [ ] Should we also support (at a certain level) passing in duration: `play C4 0.1` to make nicer music? Maybe in level 6 dealing with numbers of 12 dealing with floats (a bit late... I fear)
- [ ] Should we support playing "chords" aka multiple notes at once? `play C4 E4 F4` (note.js supports this easily, just drop the sleep from the Python) That would also require changes to level 3, then we should support multiple arguments with a list_access (like print) not just one
- [ ] From level 12, assignments needs quotation (see the test we add in [this commit](https://github.com/hedyorg/hedy/commit/e4df3a67e924633de51280800b3666ed50da6364#diff-10a085743beaa14a629a100cc542db6bc7799eb73bc7aad87561234882e2ef2c): `n = 'C4'` I am not sure I like that, it feels like a note should be a literal in their own right. We can change the parser to support both?)
- [ ] What are note names in other languages? Arabic and Chinese come to mind, but maybe even other Latin languages use different conventions (I have a vague memory of Flemmish Dutch using Do-Re-Me rather than C-D-E?!) This is an argument in favor of (localized) integers 😅) -> I have asked @alaaeddin-swidan who is looking into this for Arabic!
- [ ] Fix parser error for var named `player` (fixed poorly in https://github.com/hedyorg/hedy/pull/4903/commits/fccbef834f79da50616ee4bf653c48105442ed11)
- [x] Stop the "too long" message for song programs:
image


highlighting/definition.py
```diff
@@ -25,6 +25,7 @@
     "def",
     "return",
     "print",
+    "play",
```
jpelay, line 28: You don't need to edit these files anymore, since they are related to Ace, and after #4924 is merged, we wouldn't need to keep mainting this files.

highlighting/highlighting.json
```diff
@@ -48,6 +48,15 @@
                     "next": "value",
                     "unicode": true
                 },
+                {
+                    "regex": "(^ *)(__play__)",
```
jpelay, line 52: Same here!

highlighting/lezer-grammars/level1.grammar
```diff
@@ -1,8 +1,9 @@
 @top Program { eol* (Command eol+)* Command? }
 Command {
-    Print | Ask | Echo | Turtle | ErrorInvalid
+    Print | Ask | Echo | Play | Turtle | ErrorInvalid
```
jpelay, line 3: This is great!  Now, we only need to edit `tokens.ts`. Since we can change the keywords at run time, we can't define the keywords directly in the grammar, therefore the need for a external specializer.

Now after compiling this parser, go to `tokens.ts` and modify the import for level 1 too add the play keyword. After this, go to the `keywordToToken` dictionary and add the keyword in the `extend` field. And at last, go to `cm-editor.ts` and edit the `hedyStyleTags` var to add the new keyword.

jpelay (commented):
Hi Felienne!  Bellow  I explain a bit about how you can make the highlighting work!

highlighting/lezer-grammars/level1.grammar
Felienne: Thank you!!

Felienne: Success!!

image

I found all of it except for the hedyStyleTags, so thanks a lot for that pointer!!

jpelay:
>  What are note names in other languages? Arabic and Chinese come to mind, but maybe even other Latin languages use different conventions (I have a vague memory of Flemmish Dutch using Do-Re-Me rather than C-D-E?!) This is an argument in favor of (localized) integers 😅)

In Spanish we also use Do-Re-Mi (mi rather than me)! I remember being taught that in elementary school haha

jpelay (approved):
I LOVE this so muuuch!!!!! Also love the inclusion of numbers! Made a simple for with a print a play and it played the notes one by and one, and I thought that was pretty cool. Let's approve this and then tackle the other points in the bullet list

PR #4904, opened 2023-12-13
[CONTENT] Update babel to 2.14

### LINKED ISSUE #4761: [CONTENT] Update babel
We keep adding languages that our Babel version does not support (today: Papiamentu) which exist in Weblate but not in babel:

image

We are running old versions of babel, maybe updating it will add some new locales?

We run 2.11 but Babel is at 2.13.1: https://babel.pocoo.org/en/latest/changelog.html#version-2-13-1

And Flask-babel is at 4.0.0 now: https://python-babel.github.io/flask-babel/ (not the cause of this problem but we are very far behind, maybe we should update that too?)

If the problem is solved, we can remove these overrides:

https://github.com/hedyorg/hedy/commit/f5df62dcad86c7b0b13d15402b2a3dd766b1b37d

Felienne:
Fixes #4761 

Sadly Tagalog/ᜆᜄᜎᜓᜄ᜔ (ISO: tl) remains unsupported for now, so I left that override in place, meaning that all babel-translated content (like UI elements) will remain in English for now. 

Our own yaml content like adventures and keywords can be shown properly.

**How to test**

Navigate to the site in Setswana or Papiamentu and observe all still works: http://127.0.0.1:8080/hedy?language=tn#default

Also observe that the one word I translated (Home) is now shown:

image


jpelay:
Hi, I must be doing something wrong here, because after a `pip install` it breaks the page

image

Felienne:
Should work again now @jpelay! Somehow main was missing all css but that has been solved.

hedy_content.py
```diff
@@ -18,14 +18,10 @@
...
                           'kmr': 'ku_TR',
-                          'tn': 'en',
-                          'pap': 'en',
                           'tl': 'en'}
 
 # For the non-existing language manually overwrite the display language to make sure it is displayed correctly
 CUSTOM_LANGUAGE_TRANSLATIONS = {'kmr': 'Kurdî (Tirkiye)',
-                                'tn': 'Setswana',
-                                'pap': 'Papiamentu',
                                 'tl': 'ᜆᜄᜎᜓᜄ᜔'}
```
jpelay, line 25: Sad that these languages, even having large speaking communities, are not officially supported.

jpelay (approved):
It works now!!!! Thanks! And thanks @rix0rrr for the Babel PR that made this support possible

PR #4905, opened 2023-12-14
[BUG] fix adventure public index

hasan-sh:
DynamoDB doesn't support boolean indexes. As a result, we have to convert these to numbers.


rix0rrr (approved):
👍

Issue #4906, opened 2023-12-14
[CONTENT] Remove start_codes, load code in slides differently


Felienne:
### Discussed in https://github.com/hedyorg/hedy/discussions/4837

<div type='discussions-op-text'>

<sup>Originally posted by **Felienne** December  1, 2023</sup>
All adventures currently have a field start_code, see:

https://github.com/hedyorg/hedy/blob/5c49bdce3e3007948fe10d145c00990838c2f574/content/adventures/en.yaml#L30

This is a bit of legacy content, as I have said before, I only realised after a long while that we are building a system for teaching programming, not a system for kids to learn or explore programming alone. These start codes are very old, when we (I) thought that every level would just have a start code and kids would append and explore alone. But our preferred workflow for kids now is to read the exercise, take the startcode from the exercise, and change that!

So, there are a number of reasons to get rid of these start codes now:

* They are not useful anymore; the workflow is based on exercises (see above)
* They are adding a lot of translation efforts! Since we have to translate all of them separately. `'# place your code here'` occurs 73 times in en.yaml and has to be translated each time!

So, let's remove them from the content files!</div>

PR #4908, opened 2023-12-14
[CONTENT] Load slides from yaml

### LINKED ISSUE #4794: [BUG] I'd like to be able to use startcodes in slides that aren't in hedy_content

Felienne:
Addressed the first part of #4906 by loading code into the slides, rather than loading start codes. I again have learned a valuable lesson here haha. I though "ah I will just half-ass the slides a bit by repurposing something existing, and I will improve it later" and then of course slides are expanded and translated and it gets harder and harder to still improve. I keep forgetting this is not my tiny hobby project anymore :) 

**How to test**
Go to the slides and see that code is loaded from the yaml files.

This also fixes #4794 since it is no longer needed to look up slides code.


rix0rrr (approved):

Felienne:
Hi @MarleenGilsing!

I have changed the code so that slides can simply use code instead of links to start_code!! You can see how it works in this PR, or in main because it has already been merged in. Let me know if you need further help!

PR #4913, opened 2023-12-17
[CONTENT] Remove start codes

### LINKED ISSUE #4906: [CONTENT] Remove start_codes, load code in slides differently

Felienne:
Fixes #4906 by removing all start_codes and associated test code and validation.

**How to test**
Verify that everything works without using start_codes


Felienne:
Dropping the current Weblate stats from the readme here because I want to see the impact on the percentages after merge :) 

image

Felienne:
Hi @SabinaChita!

This is a small change in content, but it will have quite an impact on teaching, I suspect! Here are some ideas to place in the newsletter, feel free to adapt where needed, the headers are more for you that to be placed in the newsletter like this! The screenshots are also for you, I can imagine we will make other ones for the newxletter.

**How it used to work**
All levels used to have a start_code, if you go to a new adventure, code would be shown in the editor:

image

**How it works now**
There are no more start_codes, so if you go to a new tab now, where you have never created something, the editor box will be empty:

image

(will finish the rest later!)
**Why**

**Potential impact on teaching**

tests/cypress/e2e/hedy_page/run_code_button.cy.js
```diff
@@ -4,6 +4,7 @@ describe('Is able to run code', () => {
     it('Passes', () => {
       goToHedyPage();
 
+      cy.get('#editor').type("print Hallo!'\n");
```
rix0rrr, line 7: I've seen this not work well in the past. In one of my PRs I changed this to something that seems to work better:

```suggestion
      cy.get('.cm-content').click();
      cy.focused().type("print Hallo!'\n");
```

tests/cypress/e2e/hedy_page/keyword_language.cy.js
```diff
@@ -1,11 +1,15 @@
-import {goToHedyPage} from "../tools/navigation/nav";
+import {goToHedyPageWithEnKeywords} from "../tools/navigation/nav";
 
 describe('when the user changes their language to Arabic', () => {
   beforeEach(() => {
-    goToHedyPage();
-
+    goToHedyPageWithEnKeywords();
+    cy.get('#editor').type("print Hallo!'\n");
```
Felienne, line 6:
```suggestion
    cy.get('.cm-content').click();
    cy.get('#editor').type("print Hallo!'\n");
```

tests/cypress/e2e/hedy_page/run_code_button.cy.js
Felienne: Ah great idea! I added the same change to the other Cypress test I updated!

rix0rrr (approved):

PR #4914, opened 2023-12-17
[FIX] Disable autoload and autosave of programs for slides

rix0rrr:
We used to only disable autoload, by checking an attribute of `"raw"` on the turtle canvas.

Change that into a JavaScript initialization property for clarity, and also use it to disable autosave.

**How to test**

Log in and go to the slides page, edit the code in the slides and click **Run**. 

Go your **My programs** page. Observe that the program is not there.


Felienne:
HI @rix0rrr!

I can't seem to reproduce the behaviour on the current main that this PR is trying to fix, if I run code from the slides in main, it also does not show up in my programs. 

So I don't know how to test this (of course I can verify nothing else goes wrong?)

Felienne (approved):
Yes this indeed does solve the error shown!

PR #4915, opened 2023-12-17
[FIX] Remove fixed height for editor

### LINKED ISSUE #4895: [BUG] Code in slides not rendered, also achievements mustn't show
**Describe the bug**
The code is not show in editor unless you run it, see the following ss:

image

Although the code is not shown here, it is visible within the Speaker view: 

image

**Achievements** should not be granted when viewing the slides:

image

rix0rrr:
Fixes #4895

We used to set fixed pixel heights for the (CodeMirror) editor; one of the things we would do is measure the height of the container during editor initialization, and then force the editor to have that height.

On the slides page, where we would load a lot of pages in `<iframe>`s that would start out hidden, the calculated height would sometimes be measured as `0px`, which would then be copied onto the editor. The end result would be that the editor would appear empty on some pages.

Instead, in this PR: remove all height calculations. We will set a height on the grid cells that contain the editor and output box, and add the CSS declarations necessary to make the actual editor automatically resize to take up available remaining space, after any fixed elements in the same cell are taken away (like the error and success popups, the "handed in" warning box, etc).

This makes it no longer necessary to recalculate the height of the editor every time one of those popups is shown or hidden.

For the "developer mode" toggle, we introduce a new attribute: `data-devmodeheight="22rem,36rem"`, and the toggle will switch between the two sizes in the attribute. This avoids having to duplicate element sizes between HTML and JavaScript.

**How to test**

Verify this solves #4895 by examining the slides of level 1 and seeing that all code blocks are always shown.


rix0rrr:
Ah good call. Let me fiddle some more then.

hasan-sh (commented):
Switching to dev mode works.

I also noticed the issue that Jesus pointed out. In addition, we have the following:

Screenshot from 2023-12-18 14-10-08

The iframe is not taking the full height. It's more noticable here:

Screenshot from 2023-12-18 16-42-47

and
Screenshot from 2023-12-18 16-48-19



I think this both can be solved in a separate PR? Shall we create an issue for this?

rix0rrr:
Yeah, the layout of the slides itself is out of scope for this PR. Good attention to detail though!

rix0rrr:
@jpelay @hasan-sh I've changed the method to limit the height of the editor; instead of having a growing editor inside a container that we limit using Tailwind, I've added some styles on the `.cm-xxx` elements themselves in a way I found recommended on the CodeMirror forums.

I don't have an easy way to check how this looks on Windows. Can one of you check if this fixes the scrollbar issue?

hasan-sh:
> @jpelay @hasan-sh I've changed the method to limit the height of the editor; instead of having a growing editor inside a container that we limit using Tailwind, I've added some styles on the `.cm-xxx` elements themselves in a way I found recommended on the CodeMirror forums.

Great! 



> I don't have an easy way to check how this looks on Windows. Can one of you check if this fixes the scrollbar issue?

I use and it works on linux so I don't have Windows either. Hopefully Jesus does!

rix0rrr:
Linux is also fine. Just another platform than Mac -- Mac really de-emphasizes the scroll bars by default which makes it hard to see when they're wrong.

Felienne:
Great! If all is well you may approve this @hasan-sh!

hasan-sh (approved):
:+1:

jpelay:
Works great!!! Just tested this on windows and it works properly there too on Chrome and Firefox

PR #4917, opened 2023-12-17
[LANG] No warning for unused function

### LINKED ISSUE #4911: [BUG] Unused variable warning when using functions
**Describe the bug**

When defining a function, even if later is being called, the transpiler returns the Unused variable warning.

I tried this in the level 16

```
define sum with n1, n2
    return n1 + n2

print call sum with 1, 2
```
Screenshot_20231216-072540

Felienne:
Fixes #4911. Turns out, there were a few open issues in the unused variable warning! Calculations were not properly caught, and built-in names (like sum) were not detected. This fixes all open issues (I hope!)

**How to test** 

I added a few new tests


Felienne:
Ah good catch!  I would prefer to do it in a separate PR because this PR already has quite some changes all over hedy.py so I suspect it can lead to conflicts? Can you make an issue? Thanks!!

jpelay (approved):

jpelay:
> Ah good catch! I would prefer to do it in a separate PR because this PR already has quite some changes all over hedy.py so I suspect it can lead to conflicts? Can you make an issue? Thanks!!

Opened #4921! :smile:

PR #4924, opened 2023-12-18
[UI] Switch read only editors to code mirror

### LINKED ISSUE #4442: [UI idea] Change the puzzle interface to make dragging easier
As I'm working on making more puzzles I notice how it's quite a hassle to drag the lines of code to the input screen. Especially if the codes are getting longer. 

Is it maybe possible to put the lines where you have to fill in your answers next to the place where they start? (Like in the mock-up I made below)
image

This would make it much easier to drag the answers, but it might mean we need an extra button (or two) to get the code from the "answer place" to the input screen. I do think it might be a nice touch if the kids could see the lines from the puzzelcode into real code (so without the seperation between the lines.

### LINKED ISSUE #4935: [BUG] Cheatsheets code is not being replaced by editors
So apparently when we switched to HTMX it made the cheatsheets code to stop being replaced. 
image

jpelay:
**Description**

We are currently using CodeMirror in the main editor, and Ace in the read-only editors, this situation is not ideal since we need to maintain two wildly different highlighting systems, therefore this PR switches all the editors to use CodeMirror.

This is the current state of the replacement:

* [x] Example editors in the level page
* [x] Editors in the quiz's questions
* [x] Parsons
* [x] Explore page editors
* [x] Create your own adventure page's editors
* [x] Common mistakes page of the teacher's manual
* [x] Editors in the Cheatsheet page (fixes #4935)

**About Parsons**

The parsons exercises where deeply coupled with Ace, and the dragging code was difficult to understand. In order to switch these editors to CodeMirror it was necessary a redo of most of the code, therefore, this PR also fixes #4442 because we are now using the Sortable library, making the dragging easier. This is the parson interface now:
image

Now students can simply shuffle the lines in place, instead of dragging them from one place to the other.


jpelay:
Should I remove all of the code related to Ace too?

static/js/parsons.ts
```diff
@@ -113,3 +160,16 @@ function fisherYatesShuffle<A>(xs: A[]) {
         xs[i] = h;
     }
 }
+
+export function initializeParsons() {
```
Felienne, line 164: Really nice clean up of the Parson's code!

templates/incl/editor-and-output.html
```diff
@@ -264,73 +264,3 @@
 {% else %}
   <div _="on load hedyApp.toggleDevelopersMode('load', true)"></div>
 {% endif %}
-<script>
-// FIXME: Stop using inline JavaScript
```
Felienne, line 268: Haha, partly addressed with this PR? ;)

Felienne (changes requested):
Looks great @jpelay!

Just a small thing, a one line editor gets a scroll bar now (macOS, Chrome):

image

While this is not the case in the current main:
image

Since it is the first code block in the first level, it looks a bit sloppy if we leve it like this, so can you see if you can fix that?

And I'd also like a sanity check on the ts from @hasan-sh because I only skimmed it!

jpelay:
> Looks great @jpelay!
> 
> Just a small thing, a one line editor gets a scroll bar now (macOS, Chrome):
> 
> image
> While this is not the case in the current main: image


Good catch! I'll fix it!

Felienne (approved):
Looks great now!

Felienne:
@jpelay Do you think we need to open an issue to remind ourselves to remove Ace code? (f.e. the highlighter code that I changed in #4903)

jpelay:
> @jpelay Do you think we need to open an issue to remind ourselves to remove Ace code? (f.e. the highlighter code that I changed in #4903)

Yes, opened #4941 for that matter!

PR #4928, opened 2023-12-19
[BUG] fix cypress memory overusage

hasan-sh:
Cypress tests fail due to memory problems. The added flags do two things:
- does not store any test in memory within github actions
- and uses a new memory management recormmended by cypress for Chromium-based browsers.

This will hopefully solve the memory problem we are having lately.


Felienne (approved):
Hope this solves the issue!

PR #4930, opened 2023-12-19
[CHORE] Check .po files for accidental wrapping

rix0rrr:
Translations in the `.po` files should never be wrapped. When we extract them, we configure `--no-wrap`, and we also configure Weblate to never wrap them.

However, for a reason we don't understand, wrapping still occurred in https://github.com/hedyorg/hedy/pull/4878.

Add a test to detect this so we don't run into the problem again.


Felienne:
Perfect, this fails :)

Felienne:
Cool, let's see if this passes now that #4929 is merged!

Felienne:
> Cool, let's see if this passes now that #4929 is merged!

Ah there are some strings that have a (wanted) newline like email content:

image

And I see now some strings are also empty, which I guess we'd like to catch too in a test?

image

(but only in English, not in other languages, because Weblate will copy it over)

Felienne (approved):
Thanks! This will help us to understand why Weblate messes up stuff despite the no-wrap setting!

PR #4942, opened 2024-01-02
🧹 Rename `start_code` to `editor_contents`

### LINKED ISSUE #4922: [CHORE] Remove start_code useages
#4913 and #4920 removed start_code from content (and needed to be merged to reopen Weblate) but the code still has a few references:

There are a number of other places where `start_code` is still referenced. Are you interested in removing these as well, or do you just want the current PR shipped?

```
tools/rewrite-content-yaml.py
52:    "start_code",

website/frontend_types.py
86:    start_code: str
103:            start_code='',  # Teacher adventures don't seem to have this

content/adventures/adventures.schema.json
43:        "start_code": { "type": "string" },

static/js/app.ts
247:      adventure.start_code = programFromLs.code;
556:          adventure.start_code = code;
1928:    theGlobalEditor.contents = adventure.start_code;
2027:  const programChanged = theGlobalEditor.contents !== adventure.start_code;
2037:  const programSuspiciouslyShrunk = wasSavedBefore && theGlobalEditor.contents.length < adventure.start_code.length * su [...]
2086:    adventure.start_code = code;
2093:    adventure.start_code = code;

static/js/types.ts
38:  start_code: string;

app.py
165:                start_code=adventure['levels'][level].get('start_code', ""))
182:    Mutates the adventures in-place, by setting the 'save_name', 'start_code'
208:        adventure.start_code = program.code
1224:                start_code=loaded_program.code,
1391:                start_code=loaded_program.code,
1515:                                               start_code=code,
```

_Originally posted by @rix0rrr in https://github.com/hedyorg/hedy/issues/4920#issuecomment-1860786538_

rix0rrr:
We're not using `start_code` anymore for Adventures, but there is still a field called `start_code` in the back-end to front-end communication.

Rename that field to `editor_contents` to more clearly indicate its purpose.

Finally remove `start_code` from a couple of places where it explicitly meant the `start_code` in the YAML.

This change also made `editor_contents` have a default value so it doesn't need to be passed when instantiating `Adventure()`. That requires making the fields after it have defaults as well, or moving the field around. Because there was an instantiating of `Adventure('start', 'start', 'start', 'start', code, False, False)`, changing the field order was too dangerous. That code probably shouldn't have been allowed in the first place, so also add a decorator to make it required to pass all fields as keyword arguments (instantiation will fail if positional arguments are used).

Fixes #4922

**How to test**

This PR did not make any functional changes. Everything should still work as before.


Felienne:
Hola @jpelay! I thought it'd be good if you review this, because knowing that the start_code field is renamed is probably going to be useful knowledge going forward :)

website/frontend_types.py
```diff
@@ -77,15 +99,16 @@ def from_program(program: Program):
...
 @dataclass
 class Adventure:
     short_name: str
     name: str
     text: str
     save_name: str
-    start_code: str
-    is_teacher_adventure: bool
-    is_command_adventure: bool
+    editor_contents: str = field(default='')
```
jpelay, line 109: Nice fix!

jpelay (approved):
Thanks Rico!!

PR #4944, opened 2024-01-03
[LANGUAGE] multiple improved error productions

akseron:
Changes in error production to support better syntax error messages for the following instances:

- repeat on level 8 does not give a deprecated error

- repeat errors has multiple instances where it does not give custom errors

- 'is ask' without a prior variable being used is a commonly mistaken as a command without a custom error message

- ask does not give a proper error message from level 4 onwards, which it did in level 2 and 3

- common user errors with 'in range' command


akseron:
[IMPORTANT] added support in english for error production regarding 'is ask'
Cannot add support for hedy as it is right now, lark only detects _IS with a space before it because hedy the keywords only contain 'is' with a space before and after. It is most likely better to avoid using _SPACE in the keywords and delegate it to lark levels for better customisability. As a band-aid measure I added a _STANDALONE_IS to the english keywords meaning only the english version of hedy currently supports an error message.

akseron:
> [IMPORTANT] added support in english for error production regarding 'is ask' Cannot add support for hedy as it is right now, lark only detects _IS with a space before it because hedy the keywords only contain 'is' with a space before and after. It is most likely better to avoid using _SPACE in the keywords and delegate it to lark levels for better customisability. As a band-aid measure I added a _STANDALONE_IS to the english keywords meaning only the english version of hedy currently supports an error message.

possible other solution: in a hedy function itself the code can contain a regex where any line beginning with 'is ask ' will raise an error. here we do have to be careful that this does not override other error messages in earlier code.

akseron:
suggestion: add an error production called deprecated_commands, which per level is adjusted according to which commands are deprecated from there onwards (some deleted in case they do not work in the level), using this we could keep using deprecated errors throughout levels in a neat manner, adjusting it where need be. this might be especially useful considering users might like to try out commands from previous levels, while jumping around in later levels out of curiosity.

akseron:
error_ask_dep_2 support for later levels is not possible because it disallows using the 'ask' as a variable instead of a command. which one has priority over the other?

Felienne:
This looks amazing @akseron!!

I am adding a few screenshots here, maybe @SabinaChita wants to add them to a newsletter or update.

Current main:
image

This PR:
image

-----

tests/test_level/test_level_11.py
```diff
@@ -81,6 +81,30 @@ def test_for_loop_with_string_var_gives_type_error(self):
             extra_check_function=lambda c: c.exception.arguments['line_number'] == 3,
             exception=hedy.exceptions.InvalidArgumentTypeException)
 
+    def test_for_loop_without_in_var_gives_error(self):
+        code = "for no_in range 1 to 5"
```
Felienne, line 85: Clever naming!

Felienne (approved):
Thanks for the hard work @akseron !!!

PR #4945, opened 2024-01-03
[LANGUAGE] Undefined function warning

### LINKED ISSUE #4921: [BUG] Warning for undefined function
Original comment from https://github.com/hedyorg/hedy/pull/4917#pullrequestreview-1787114869

> It does indeed solve the issue! Now, while testing, I found that you can use a function that you have not defined inside a calculation:
image
When you just call it, it returns a Python error:
image

Felienne:
Fixes #4921 

**How to test**

I have added new tests. You can also observe the error message in the UI:

image

Also when using a calculation as parameter:

image

**Checklist**


jpelay (approved):
Tested and works great!!!!

PR #4946, opened 2024-01-03
[WIP] add meta in a few more (existing) places

Felienne:
WIP for #3388.

Note to self: I also want to remove the =100 that is not so pretty, but it will require a lot of tiny changes, so I might do it in a separate PR.

### DIFF
--- a/hedy.py
+++ b/hedy.py
@@ -1398,20 +1398,6 @@ class ConvertToPython(Transformer):
         elif ConvertToPython.is_quoted(name):
             return f"{name}"
 
-    def make_f_string(self, args):
-        argument_string = ''
-        for argument in args:
-            if self.is_variable(argument):
-                # variables are placed in {} in the f string
-                argument_string += "{" + escape_var(argument) + "}"
-            else:
-                # strings are written regularly
-                # however we no longer need the enclosing quotes in the f-string
-                # the quotes are only left on the argument to check if they are there.
-                argument_string += argument.replace("'", '')
-
-        return f"print(f'{argument_string}')"
-
     def get_fresh_var(self, name):
         while self.is_variable(name):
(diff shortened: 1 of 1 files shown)


Felienne:
> Note to self: I also want to remove the =100 that is not so pretty, but it will require a lot of tiny changes, so I might do it in a separate PR.

I give up! Changing the default parameter gives very weird errors in skipped code that I can't figure out.

jpelay:
> I give up! Changing the default parameter gives very weird errors in skipped code that I can't figure out.

What errors are you getting? Maybe I can be a little help if they're related to the source mapper, since I've dabbled with that one.

Felienne:
> update f-string location

If I remove the default = 100 value (as I do here: https://github.com/hedyorg/hedy/pull/4946/commits/68cc221df2ce29148fc4dde4ed4ee50789b571e1) tests start to fail.

My goal is to remove the =100 everywhere and to nicely pass in the meta.line where needed. If you want to take a stab at it, go for it! If this explanation is not clear enough, we can merge this PR and then I can create a branch in the failing start if you want!

jpelay:
> > update f-string location
> 
> If I remove the default = 100 value (as I do here: [68cc221](https://github.com/hedyorg/hedy/commit/68cc221df2ce29148fc4dde4ed4ee50789b571e1)) tests start to fail.
> 
> My goal is to remove the =100 everywhere and to nicely pass in the meta.line where needed. If you want to take a stab at it, go for it! If this explanation is not clear enough, we can merge this PR and then I can create a branch in the failing start if you want!

I think that explanation is clear enough for me to start seeing what happens, but I think is better if  we merge this and then I'll take a look, so we can get this shipped!

jpelay (approved):

PR #4947, opened 2024-01-03
🖊️ Fix non-english leading space

### LINKED ISSUE #4925: [BUG] Leading spaces work in English but not French, Spanish, Portuguese, presumably others.
**Describe the bug**

Go to https://www.hedycode.com/hedy#ask_command, and this works...

```
  print Hello!
ask What is your name?
echo hello
```

It still works if you change to French after having made that change, because the program is still in English.

But switching to French before starting the lesson, and adding the spaces, gets this result...

image

**NOTE:** I had to go back to the home page and Clear Site Data and go back to the lesson in between after being in English to reproduce the issue reliably.

I checked Spanish and Portuguese, they have the same issue.

https://www.hedycode.com/hedy#ask_command

```
  affiche Salut !
demande Quel est ton nom ?
réponds salut
```

- Linux
- Chrome Version 120.0.6099.71

(I'm watching Felienne's GOTO 2022 video and she mentions kid's being able to make that specific mistake. Hedy looks great!)

Felienne:
Fixes #4925 I am still not sure why this did not also cause an issue in English! I might dive in a bit more but at least this addresses the problem

**How to test**

I added a test. You can also try the UI:

image

**Checklist**


jpelay (commented):
It works with everything but if elses! Do you think it's a good idea to fix this in this PR?
image

Felienne:
> It works with everything but if elses! Do you think it's a good idea to fix this in this PR? image

Good catch, but that does not work now either, this is the current main:

image

So maybe we can slove this is a separate issue/PR?

jpelay (approved):
Great! Let's do it that way.

PR #4952, opened 2024-01-04
[BUG FIX] Fix embedded editor when in read only mode

TiBiBa:
This PR fixes a bug where there would be whitespace add the bottom of the embedded editor when on readonly and disabled run mode. This because we calculated the height of the editor based on the height of the run button. But if it's not there, no need to take this space!

The following test link should no longer have whitespace at the bottom:
http://127.0.0.1:8080/embedded/1/?lang=nl&run=false&readOnly=true&program=cHJpbnQgZGl0IGlzIGVlbiB0ZXN0IQplY2hvIHdhYXJvbSBrbG9wdCBkaXQgbmlldD8KdGVzdCB0ZXN0CnRlc3QgdGVzdA%3D%3D

### DIFF
--- a/templates/embedded-editor.html
+++ b/templates/embedded-editor.html
@@ -11,8 +11,8 @@
     </head>
     <body class="flex flex-col gap-2 h-screen" dir="{{ g.dir }}">
         <div id="editor-area" class="relative grid grid-cols-1 {% if run %}md:grid-cols-2{% endif %} md:grid-flow-row-dense gap-2">
-            <div class="flex flex-col order-1 relative" id="code_editor" style="height: calc(100vh - 4em);">
-                <div id="editor" style="background: #272822; max-height: calc(100vh - 4em); font-size:0.95em; color: white" blurred="false" class="h-full w-full flex-1 text-lg rounded min-h-0 overflow-hidden fixed-editor-height" {% if readOnly %}data-readonly="true"{% endif %}></div>
+            <div class="h-full flex flex-col order-1 relative" id="code_editor" {% if run %}style="height: calc(100vh - 4em);"{% endif %}>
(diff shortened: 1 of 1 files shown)


jpelay (commented):
Hi Timon! We totally forgot about trying the read only mode as well! Good catch!

Also, do you think it can be possible to extend the gutter all the way to the bottom? The same way as in the the other editors.
This PR:
image

The main editor:
image

TiBiBa:
> Hi Timon! We totally forgot about trying the read only mode as well! Good catch!
> 
> Also, do you think it can be possible to extend the gutter all the way to the bottom? The same way as in the the other editors. This PR: image
> 
> The main editor: image

I'm unable to re-produce this issue. See the following screenshot:
Scherm­afbeelding 2024-01-04 om 16 45 31

Do you have any clue what might cause this? The editor is already set to full height, I assume CodeMirror handles the gutter height?

jpelay:
> > Hi Timon! We totally forgot about trying the read only mode as well! Good catch!
> > Also, do you think it can be possible to extend the gutter all the way to the bottom? The same way as in the the other editors. This PR: image
> > The main editor: image
> 
> I'm unable to re-produce this issue. See the following screenshot: Scherm­afbeelding 2024-01-04 om 16 45 31
> 
> Do you have any clue what might cause this? The editor is already set to full height, I assume CodeMirror handles the gutter height?

It works perfectly when you have the run button, but not in the case of read only editors, try the link you put in the How to test section.

> I assume CodeMirror handles the gutter height?

You have to set the minimum-height of the parent I think, otherwise it just takes up to the maximum line.

jpelay (approved):
Sorry! I didn't generate the tailwind first! It works fine now

PR #4957, opened 2024-01-04
🖊️💻 Syntax highlighting for variables

### LINKED ISSUE #4853: [UI idea] Highlight variables

jpelay:
Adds syntax highlighting for variables in both read only and main editors. It works by checking for variables definition in `is`,`ask`, etc and then highlighting the places where those names appear in the program. This is done in every update of the document, but since the size of our programs is small, this doesn't cause such a big overhead.

Fixes #4853 

**How to test**

* Go to the Hedy page and see that the variables are highlighted


Felienne:
Owwwww looking great!!

image

Felienne:
Let me know when it is ready to be merged!

jpelay:
> Let me know when it is ready to be merged!

These are the things that I want to do before merging this:

* Different color for function names.
* Scope the variables that are inside a function, so variables defined inside them won't be colored in the main program.

And I'd like to discuss the following before continuing:

* Change the yellow color of the variables, since, as you said it can be a bit overwhelming in higher levels when
* Annelise suggested that we change the color of text in the earlier levels to also be purple, like the one enclosed in quotes. What do you think?

jpelay:
Hi, @Felienne! This is ready for review 😄 I changed the colors to be subtler. 

image

Felienne:
Love it!!!!

image


> * Annelise suggested that we change the color of text in the earlier levels to also be purple, like the one enclosed in quotes. What do you think?

I think she is right, it should be the same color (not sure purple is the right color to be honest!) Shall we make an issue for that?

Felienne (approved):

PR #4958, opened 2024-01-05
🧹 Keep NPM dependencies updated using doit

rix0rrr:
Currently, the checked-in Tailwind file has been generated with an older version of Tailwind than the version of Tailwind we actually say we depend on. You can tell, because the version number is right there in the generated file.

The reason is because the person generating that CSS file (me) had forgotten to run `npm ci` before running Tailwind. This will now cause the next PR to have unnecessary diffs in their generated CSS file if they accidentally have different versions of Tailwind installed for whatever reason (either they did run `npm ci`, or the last time they ran it Tailwind was at a different version than my version).

Add the running of `npm ci` to the `doit` script, so that this can't happen anymore: it will be run automatically for you if necessary.

**How to test**

Run `doit run frontend` and observe that NPM dependencies are installed.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -33,6 +33,36 @@ python3 = sys.executable
 DOIT_CONFIG = {'default_tasks': ['deploy']}
 
 
+def task_npm():
+    """Install NPM dependencies.
+
+    If the `package-lock.json` file changes, we will automatically re-install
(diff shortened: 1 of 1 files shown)


jpelay (approved):
Works!

PR #4970, opened 2024-01-08
🪲 only check threshold if quiz is enabled

### LINKED ISSUE #4937: [BUG] if quiz is hidden, the threshold shouldn't be effective
**Describe the bug**
If we hide the quizes and still have the threshold value set to some value, the students should be able to go to next levels without the threshold condition.

Here the class customization restricts on having 20 as a minimum score:
image

But the student cannot access further levels because of that threshold
image

hasan-sh:
Fixes #4937 

Now students can see all levels if the teacher disabled the quiz feature. Otherwise, the quiz threshold and their scores will determine whether they can see it or not.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1345,15 +1345,18 @@ def index(level, program_id):
     # - But, if there is a quiz threshold we have to check again if the user has reached it
 
     if 'level_thresholds' in customizations:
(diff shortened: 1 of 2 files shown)


jpelay (approved):
Works great, thanks Hasan!

PR #4973, opened 2024-01-09
🚚 add `doit run devserver` command

rix0rrr:
While updating the [Hedy development wiki](https://github.com/hedyorg/hedy/wiki/Hedy-Development-Process) to match the newest ways to run scripts using `doit`, I noticed that there are still some elaborate instructions around running scripts and setting environment variables etc in there.

Those can be encoded using `doit` as well, so that running a local copy of the server can be as simple as:

```
doit run devserver
```

This will take care of recompiling Babel files when necessary, so that those instructions can be removed. It sets the environment variables that some parts of the code require (`BASE_URL` and `ADMIN_USER`) so that those instructions can be removed as well.

**How to test**

Run:

```
doit run devdb frontend devserver
```

Observe that Babel files are regenerated, before the dev server is eventually run. Run `npm run cypress` in a different terminal and observe that all tests succeed.

(`devserver` purposely does not depend on `frontend`, so that people can run a copy of the `devserver` without having Node installed)

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -275,6 +275,7 @@ def task_extract():
         ],
     )
 
+
 def task_devserver():
     """Run a copy of the development server.
 
--- a/dodo.py
+++ b/dodo.py
@@ -25,6 +25,8 @@ from os import path
 from glob import glob
 import sys
 
+from doit.tools import LongRunning
+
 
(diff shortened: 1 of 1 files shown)


hasan-sh (approved):
Needed and works as expected!

PR #4974, opened 2024-01-09
🖊️  Improved error productions for 'is ask'

akseron:
* Produces an error message for incorrect use of 'is ask'. A common made error is that Hedy users forget to add a variable before 'is ask', resulting in a parse errror:
image
* No existing issue as far as I am aware.

**How to test**
* Go to http://0.0.0.0:8080/
* Run this code on level 2 and 3:
`is ask hello`
* Receive the following error message:
image

### DIFF
--- a/translations/ar/LC_MESSAGES/messages.po
+++ b/translations/ar/LC_MESSAGES/messages.po
@@ -2205,6 +2205,3 @@ msgstr "Your program"
 #~ msgid "tag_deleted"
 #~ msgstr "لقد تم حذف العلامة"
 
-#~ msgid "public_adventures"
-#~ msgstr "المغامرات العامة"
-
--- a/translations/bg/LC_MESSAGES/messages.po
+++ b/translations/bg/LC_MESSAGES/messages.po
@@ -2492,6 +2492,4 @@ msgstr "Your program"
 #~ msgid "no_tags"
 #~ msgstr "No tags yet."
 
-#~ msgid "public_adventures"
-#~ msgstr "Public adventures"
 
--- a/translations/bn/LC_MESSAGES/messages.po
+++ b/translations/bn/LC_MESSAGES/messages.po
@@ -2596,6 +2596,4 @@ msgstr "Your program"
 #~ msgid "no_tags"
 #~ msgstr "No tags yet।"
 
-#~ msgid "public_adventures"
-#~ msgstr "Public adventures"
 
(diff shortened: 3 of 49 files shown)


Felienne:
Ow sorry @akseron! One key still occurs double, you will have to remove it my hand (the error message tells you what key and where) If you don't know how to do it, I can do it Monday.

akseron:
> Ow sorry @akseron! One key still occurs double, you will have to remove it my hand (the error message tells you what key and where) If you don't know how to do it, I can do it Monday.

There seems to be some message id's  in messages.pot and error-messages.txt which do not contain any information. These keep coming back when compiling pybabel translations. The error message is specifically caused by 'public_adventures'. Should we remove these from messages.pot and error-messages.txt?
image

Felienne:
> > Ow sorry @akseron! One key still occurs double, you will have to remove it my hand (the error message tells you what key and where) If you don't know how to do it, I can do it Monday.
> 
> There seems to be some message id's in messages.pot and error-messages.txt which do not contain any information. These keep coming back when compiling pybabel translations. The error message is specifically caused by 'public_adventures'. Should we remove these from messages.pot and error-messages.txt? image
> 
> This also happened to be the reason for the error message in my other pull request.

I think they are new keys that @hasan-sh added, I am not sure if they should be kept or not, I hope he can chime in (if not I will take a look after the weekend)

hasan-sh:
> I think they are new keys that @hasan-sh added, I am not sure if they should be kept or not, I hope he can chime in.

One or two will be removed, but I guess it's better to let me handle them in my PRs regarding public adventures!

Felienne:
> > I think they are new keys that @hasan-sh added, I am not sure if they should be kept or not, I hope he can chime in.
> 
> One or two will be removed, but I guess it's better to let me handle them in my PRs regarding public adventures!

@akseron so that means you can remove both (you will have to do that by hand or with a bit of careful scripting)

akseron:
> > > I think they are new keys that @hasan-sh added, I am not sure if they should be kept or not, I hope he can chime in.
> > 
> > 
> > One or two will be removed, but I guess it's better to let me handle them in my PRs regarding public adventures!
> 
> @akseron so that means you can remove both (you will have to do that by hand or with a bit of careful scripting)

The latest commit should have fixed it! The same has been done for https://github.com/hedyorg/hedy/pull/4992 and they should both be ready for merging :smile:

PR #4978, opened 2024-01-09
🪲 Don't execute npm task on Heroku

jpelay:
Recently we found out that the deploy to Alpha action was failing. After some digging, we found out that it first started here: https://github.com/hedyorg/hedy/actions/runs/7450720163, our guess is that after merging https://github.com/hedyorg/hedy/pull/4958 somehow the `npm ci` that happens on GitHub interferes with the npm install that happens on Heroku.

This PR checks for the environment variable 'ON_HEROKU' which is set to one on the Heroku environment (maybe we should add it too to the production environment, but I think I don't have access to that one). If the variable is present it means we are on Heroku and therefore don't execute the task.

How to test:

Let's approve and see if the deployment action is executed properly.


rix0rrr (approved):

Felienne:
> (maybe we should add it too to the production environment, but I think I don't have access to that one)

Is that needed? If so I can do it (but I think @rix0rrr can do that too!)

PR #4982, opened 2024-01-10
💻 use the newly created cards in explore page

### LINKED ISSUE #4844: [UI idea] Switch public programs cards to new UI
On #4757 we are adding a new UI element: cards for public adventures, these cards differ from an already present design: the design of cards for explore programs. During discussion on our biweekly Hedy meeting we agreed that we couldn't have different designs for what is essentially the same thing, and therefore we agreed to move the existing cards over to the new design.

hasan-sh:
Fixes #4844

image

The cards now are the tiles that were created for the public adventures. You may notice three different colors. These indicate the status of a program and can be one of the following:
- yellow -> favorite program
- red -> erroneous program
- blue -> normal state
> NB: Hovering over the cards give a small shadow and outline whose color is the same as the program's status! E.g., the ss, I hovered over a favorite program!

**How to test?**
Navigate to the explore page and see the new cards.


jpelay (changes requested):
Hi, Hasan! Looks great in the Explore page :D

image

Can you also change the cards in the public page as well: http://127.0.0.1:8080/user/teacher1? There we are still showing the old style:

image

And since we are replacing the old way entirely could you also delete the file that has that code, `macros/public-programs.html`, and all of its references? thanks a lot 😄

hasan-sh:
> Can you also change the cards in the public page as well: http://127.0.0.1:8080/user/teacher1? There we are still showing the old style:

Yes we should indeed. Will remove the old code as well:)

hasan-sh:
> Can you also change the cards in the public page as well: http://127.0.0.1:8080/user/teacher1? There we are still showing the old style:

Yes we should indeed. Will remove the old code as well:)

jpelay (approved):
Nice work!

PR #4983, opened 2024-01-11
📜 Removes an old header

MarleenGilsing:
Removes an old header in level 6 maths adventure


Felienne (approved):
Can be merged now weblate is!

PR #4990, opened 2024-01-11
💻 adventure tabs and editor in public adventures page

### LINKED ISSUE #4954: 💻 show public adventures as tabs
**Idea incl level**
- each adventure that's filtered should be shown in a tab with its code for its different levels
- the tile could be shown on the right side to give more info regarding that adventure
- [ ] add tests

**Additional context**

hasan-sh:
Fixes #4954 
# Features
Initially the page shows all available public adventures.
image

There are four filters that teachers can utilize in order to find adventures quickly. These filters are:
1. search input: finds an adventure whose name matches the search input
2. level (default=1): gets adventures of a specific level. Only one item can be selected.
3. language: gets adventures of a specific language. Only one item can be selected.
4. tags: gets adventures that have at least one of the selected tags. Multiple items can be selected.

Another interesting and important feature is that the URL always updates wrt the filters. For instance, if you filter by English, the URL will be something like: public-adventures?level=1&**lang=en**&tag=&search=. The same applied to all available filters. Additionally, this means that you can share a url with someone and they would see the exact same adventures you've filtered. So, if you share `/public-adventures?level=1&lang=&tag=print&search=parr` with another teacher, they will see exactly what you see:
image

# **How to test?**
Go to `/public-adventures` and start applying some filters, run some code of any adventure you like, and perhaps clone one that's created by some other user. 

# **Technicality (for devs)**

Whenever a filter is applied, we send a request to python and expect two things:
- `html`: the template of the new filtered adventures, to replace the current adventure tabs with. 
- `js`: some properties that are needed for js; e.g., to initialize the editor.

The html is simply a string that we replace the innerHTML of the target element that includes the editor and tabs. And since this is a string, we need to manually initialize the JS code; e.g., run the `initializeHighlightedCodeBlocks` or initialize the editor with the selected level.

The reason why I didn't use htmx here is due to the fact that the created template that has the editor and tabs (i.e., in `public-adventures-body.html`) does not issue a rerender on the js side. Most probably, this behavior is because htmx only replaces a target element by whatever the server returns and does not mind the `js` property that we pass to the template. As a result, the editor and tabs become just views with no interactivity.

Another matter is the usage of Tailwind Elements. I used this for two reasons: it uses Tailwind (which we also do) and it has a plenty of already styled elements. The problem with TE is that they don't support RTL currently, [however, they did mention that they're working on it](https://github.com/mdbootstrap/TW-Elements/issues/2264#issuecomment-1899791394). So, let's just keep using our custom select that I created, until they fix that issue or we need some of the beautiful components they provide.


Felienne:
Loooooooove how extensive this PR is @hasan-sh!!!

Before we merge just one question, do we want to move the "Technicality" information in this PR elsewhere? To the docs or a discussion? Or is it fine if it just lives here so we can find it when needed?

hasan-sh:
> Loooooooove how extensive this PR is @hasan-sh!!!

Glad to hear that, Felienne :)))


> Before we merge just one question, do we want to move the "Technicality" information in this PR elsewhere? To the docs or a discussion? Or is it fine if it just lives here so we can find it when needed?

I thought here just for that reason indeed, also to discuss it in the contributors meeting, but I thought it's not suitable yesterday!

Felienne (approved):

Issue #4962, opened 2024-01-05
🎶 📜 Outline a few music adventures 


Felienne:
I will sketch some so @MarleenGilsing and @SabinaChita can expand them

PR #4991, opened 2024-01-11
📜  Adds music adventure level 1 to 14 in English

MarleenGilsing:
Start of #4962 

Can be merged once the music notes bugs are resolved.

Depends-On: #5035


content/adventures/en.yaml
```diff
@@ -2531,6 +2531,304 @@ adventures:
...
+                    play G4
+                    play A4
+                    play B4
+                    play C5
+                    ```
+            2:
+                story_text: |-
+                    Finish the songs! We have started the codes for 3 melodies.
+                    As you can see, you can also use the `{sleep}` command to add a little pause in the song.
+                example_code: |
```
Felienne, line 2566: Hi @MarleenGilsing!

I have added a story_text here, because if you have more example codes than story_texts they will not show. I have also removed 1 song, since 3 is a bit much and the block becomes long. Once we have shared adventures (very soon!!) we can add as much as we want :)

content/adventures/en.yaml
```diff
@@ -2531,6 +2531,304 @@ adventures:
...
+                    The example code creates a random melody, but it's very short and not many notes are used.
+                    Add more notes to the list and create a longer melody by copying the last line a couple more times.
+                example_code: |
+                    ```
+                    notes {is} A4, B4, C4
+                    {play} notes {at} {random}
+                    {play} notes {at} {random}
+                    {play} notes {at} {random}
+                    ```
+            4:
```
Felienne, line 2616: Owww I love this karaoke machine idea!!!!

Felienne:
Great work @MarleenGilsing!

I have made changes to the music codes until level 5, these now all work, level 6 not yet, still working on it!!

You can continue adding music adventures tomorrow for the rest of the levels as I am improving the rest of the code,

Felienne:
Hi @MarleenGilsing!

Awesome work, just 2 remarks:

**Level 9**
I think the code might be wrong? 
http://localhost:8080/hedy/9#music A second if is missing I think but I did not know how to improve it?

**Level 6**
It works but it does not really explain why you would use a number. Could we do something like this?
```
play_number = ask 'Say a starting number'
play play_number
play_number = play_number + 1
play play_number
play_number = play_number + 1
play play_number
```

or if you want to go this road, we can also do:
```
play_number = ask 'Say a random number'
play play_number
play play_number + 1
play play_number + 2
```

That currently does not work but I can make it work :)

MarleenGilsing:
Hi @Felienne, 

The level 9 code didn't work properly indeed, but now it does :) 

Your example code for level 6 was better, so I changed it, I also added that it prints the number, so the students might get a better understanding of which sound belongs to which number. One small issue I noticed: 
image
After the play command the variable doesn't turn light blue yet. Should I make a seperate issue for this, or is this already in the works. 

But this PR is ready to go 👍

Felienne (approved):

PR #4999, opened 2024-01-16
🧹 Add script to inventorize templates

rix0rrr:
There currently are a lot of templates and it's not entirely obvious how they relate to each other.

This PR adds a script that makes an overview, so a cleanup operation can be safer and more targeted.

**How to test**

On the command line, run:

```
# Mac
python3 tools/analyze-templates.py | pbcopy

# Linux
python3 tools/analyze-templates.py | xclip -selection clipboard

# Windows
python3 tools/analyze-templates.py | clip
```

Then paste the result into an online GraphViz visualizer, like for example [Edotor](https://edotor.net). The result looks something like this:

image


jpelay (approved):
Works great! Now we can use this to analyze our templates dependencies, which certainly are quite convoluted right now,.

PR #5001, opened 2024-01-18
📜  Removes extra line

MarleenGilsing:
Found an unnecessary line and removed it in all languages.


MarleenGilsing:
Don't know it the weblate lock is necessary, because i didn't change anything to the structure of the file, only one line of content. But I did change multiple languages in this PR so labeled it to be sure.

Felienne (approved):
oeeewww this might clash with the change to clear, but let's see....!

PR #5006, opened 2024-01-18
💻 If no classes or adventures, show info

### LINKED ISSUE #4645: [UI idea] No classes or adventures
**Idea incl level**
Add a message when there's no classes or adventures in the `/for-teachers` page.

**Additional context**

Annelein:
Fixes #4645 

**How to test**
Go to the `/for-teachers` page and remove all classes and adventures. The class info or adventure info is now visible. If you add a class or adventure, the info is not visible. 

Screenshot 2024-01-18 at 12 07 30
Screenshot 2024-01-18 at 12 07 40

### DIFF
--- a/templates/for-teachers.html
+++ b/templates/for-teachers.html
@@ -21,7 +21,7 @@
             </svg>
           </button>
         </div>
-        <div class="border border-gray-400 px-4 rounded-lg w-full lg:w-3/4 p-4 text-center hidden" id="classes_info">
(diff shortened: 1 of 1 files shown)


Annelein:
@hasan-sh I though this would be a nice solution for #4645 , what do you think?

hasan-sh (approved):
Amazinggg, this is a solution to that indeed.

Issue #4701, opened 2023-11-02
[BUG] Rename the teacher manual section in the yaml files (if possible)


MarleenGilsing:
![image](https://github.com/hedyorg/hedy/assets/80678586/8390cd78-c438-491e-8f64-6b6fa891249e)

Maybe a weird request, but the teacher manual is currently called 'sections', which is a very vague name. If this doesn't cause too much trouble, could it be changed to 'teacher_manual' or something like that? If not, you can just close this issue, because not a lot of people will be bothered by it.

Felienne:
This should be possible, but:

* weblate needs to be closed
* it also needs to be changed in code, so it is a little bit involved

PR #5014, opened 2024-01-18
📜 rename the teacher manual section in the yaml files #4701

rmagedon97:
Fixes issue #4701 

**How to test**

I have changed the KEY "section" to teachers-manual" in all. YAML files, which are in content/pages/.yaml.
This should just provide a better naming for the KEY. The change is internal, and it should not affect the website.
To verify everything is working, navigate with a teacher account to the teacher manual. If there is text, it all works.

**Checklist**
Done? Check if you have it all in place using this list: (mark with x if done.)


Felienne (changes requested):
Hi @rmagedon97!

The change seems to work well, thanks!!!

However, you seem to also change the order of the sections, start-sections used to come after teacher-guide/sections but now comes before. Is that intentional?

It is not a problem, but it makes it a lot harder to see the changes that you did, so I would prefer it if you can revert that order change

Felienne:
To continue my note above, if I look at the changes now, it shows this:

image

Which is not really the important part of the change.

rmagedon97:
Hello @Felienne, Thank you for your comments.
I think I have filled in all that was missing in the description.

About the sections, because I couldn't fix the issue from the first try, the pre-commit hook did an auto-fix, which changed their order. This should have also been fixed now.

Let me know if I'm missing something else.

Felienne:
Ah, my bad, the order was changed by the rewriter not by you! I should have know this, sorry!

Felienne (approved):
All good @rmagedon97! Congrats on getting your first PR approved!!

PR #5019, opened 2024-01-22
📜 Update public adventures string

Felienne:
Hi @hasan-sh!

I just noticed this string is still missing: 

image

But rather than making an issue I though I can just fix it right away :) 

I think you may have added it at one point, but @akseron's changes set it ba k to empty. Can you approve if this looks ok?


hasan-sh (approved):
This is great

Issue #5020, opened 2024-01-22
📜 Adventures misses a title


Felienne:
**Describe the issue**

Similar to #5019, the adventures title is also missing: 
<img width="186" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/028da7cd-1f4c-4f86-8985-d5bf8996dcc1">

(https://hedy-alpha.herokuapp.com/public-adventures)

Maybe a good idea to check all related strings? And maybe to slab a test on the po files that English should never have "" as value?

hasan-sh:
> Similar to #5019, the adventures title is also missing:

I'll add and remove translation keys related to the public adventures in https://github.com/hedyorg/hedy/pull/4990

PR #5022, opened 2024-01-23
🧪 Test for empty strings in English po file

Felienne:
Fixes the testing part of #5020

**How to test**
The tests should pass when the content has been fixed 

**Checklist**


Felienne:
Hi @hasan-sh!

Even with your new strings merged, there are still a few missing. I don't think this is a desired state, but maybe these strings are not used anymore? Potentially @jpelay or @Mark-Giesen can chip in? I do think we want this test to prevent issues in the future, but we need to make sure all strings are accounted for.

hasan-sh:
> Hi @hasan-sh!
> 
> Even with your new strings merged, there are still a few missing. I don't think this is a desired state, but maybe these strings are not used anymore? Potentially @jpelay or @Mark-Giesen can chip in? I do think we want this test to prevent issues in the future, but we need to make sure all strings are accounted for.

Just added an extra check for the following case:

image

hasan-sh:
Also i noticed an undesired behavior with pybabel update. In my other PR, after i extract the keys, some keys are removed from the `messages.pot` because they are unused. However, the update command still appends them to the end of each po file. See this  https://github.com/hedyorg/hedy/pull/5026/commits/7e358efff4732a502cf429ae32d912f0cc6b6950

Felienne:
> Also i noticed an undesired behavior with pybabel update. In my other PR, after i extract the keys, some keys are removed from the `messages.pot` because they are unused. However, the update command still appends them to the end of each po file. See this [7e358ef](https://github.com/hedyorg/hedy/commit/7e358efff4732a502cf429ae32d912f0cc6b6950)

I am not sure this is undesired. The goal of this behaviour is that if we remove a string and later place it back (and that does happen), the translation is not lost. The string is commented out at the end of the file so it will not interfere with anything. 

Maybe this behaviour can be changed with a setting on Weblate but I think it is not too problematic. Is it causing you trouble?

Felienne:
> > Hi @hasan-sh!
> > Even with your new strings merged, there are still a few missing. I don't think this is a desired state, but maybe these strings are not used anymore? Potentially @jpelay or @Mark-Giesen can chip in? I do think we want this test to prevent issues in the future, but we need to make sure all strings are accounted for.
> 
> Just added an extra check for the following case:
> 
> image

Thanks for the fix!! Happy you are cleaning up my half assed work 😅

hasan-sh:
> Maybe this behaviour can be changed with a setting on Weblate but I think it is not too problematic. Is it causing you trouble?

No, it doesn't! Now i understand and it makes sense indeed, thanks for explaining.

rix0rrr (approved):

PR #5026, opened 2024-01-23
💻 Enhance customize-adventure page

### LINKED ISSUE #4632: [UI idea] Adventure page enhancements
**Idea incl level**
Update the `/for-teachers/customize-adventure` page to show which classes are affected by this said adventure. 

**Additional context**
This helps teachers better decide how and whether to tweak adventures.

hasan-sh:
Fixes #4632 

**What?**
There are two main features introduced in this PR. 
- show all classes that use the current adventure in one or more levels.
- replace the plain text editor by a WYSIWYG editor CKEditor.

**How to test?**
go to the customize-adventure page and use the editor and see the classes (if any).

**Notes**
The editor is the main change, thus. See the following: 

image

As can be seen, the toolbar has different plugins that teachers can utilize in order to enhance their experience with writing/creating custom adventures. For instance, teachers can make text bold by any of the obvious button in the toolbar. Alternatively, they could press CTRL-B to do so. Similarly, they can add ordered and unordered lists. Or create different headings for better visualization and attracting students' attention. 

Perhaps we could add a small paragraph that explains such features. But we certainly need to explain that the `code` button makes a keyword, and the code block button creates a pre tag that would be rendered as a code that students can use. The pre tags aren't highlighted still, so further investigation could be done on how to do so. However, i don't think it's really needed since the editor really shows that it's a piece of code. And as Felienne noted that teachers can always preview what they have written so far, with syntax highlighting since that functionality is just HTML and would use codemirror.


static/js/adventure.ts
```diff
@@ -0,0 +1,94 @@
...
+
+function initializeEditor(language: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+        if ($editor) {
+            $editor.destroy();
+        }
+
+        ClassicEditor
+            .create(editorContainer, {
+                codeBlock: {
```
jpelay, line 26: Maybe in the future we could investigate integrations between CKEditor and CodeMirror. I found this plugin (https://ckeditor.com/cke4/addon/codemirror), but is quite old, but it suggest that perhaps such a thing could be possible with newer versions of CodeMirror and CK

static/js/adventure.ts
```diff
@@ -0,0 +1,94 @@
...
+            })
+            .catch(error => {
+                console.error(error);
+                reject(error);
+            });
+    });
+}
+
+
+// Update the editor with new data
```
jpelay, line 48: Will you use this code in the future?

static/js/custom-window.ts
```diff
@@ -0,0 +1,5 @@
+import ClassicEditor from "@ckeditor/ckeditor5-build-classic";
+
+export interface CustomWindow extends Window {
```
jpelay, line 3: I think perhaps we could give this interface a more descriptive name, because it's not clear right now the purpose of it. If it's just meant to be used to hold the ckEditor, then perhaps `CKWindow` could be a fine option? What do you think?

static_babel_content.json
```diff
@@ -66,7 +66,7 @@
         "DO": "Rep\u00fablica Dominicana",
         "DZ": "\u0627\u0644\u062c\u0632\u0627\u0626\u0631",
         "EC": "Ecuador",
-        "EE": "Eesti",
+        "EE": "Estonia",
```
jpelay, line 69: Maybe this was an unintended change?

jpelay (changes requested):
Hello Hasan! I think this is great, and looks amazing. I just had some comments that you can see in the comments bellow :D

Also, I think the default explanation looks a bit broken:

image

static/js/adventure.ts
hasan-sh: Yeah, it's quite old indeed! Let's see how it goes!

static/js/custom-window.ts
hasan-sh: No, it's meant to be used overall in TS. This due to the fact that TS will complain when we try to attach variables to the window object! So, custom window sounded logical to me. I don't mind renaming it if you have other suggestions! Maybe i could add a comment in that file to explain its purpose?

static_babel_content.json
hasan-sh: Not sure how that changed took place, maybe due to the extract function? But will undo it, thanks:))

hasan-sh:
> Also, I think the default explanation looks a bit broken:

by default you should click on the exclamation mark, but I think it's better to show it then since it wasn't clear for you, how would it be for users /:

static/js/adventure.ts
hasan-sh: this was a trail to highlight code, but will remove it for now and can always retrieve it back:)

Felienne (approved):
Loooove this new look @hasan-sh! And thanks for the review @jpelay!

static/js/adventure.ts
Felienne: Do we need an issue for this so we don't forget?

Felienne (changes requested):
Ow, we still need to fix the default explanation, sorry!

hasan-sh:
> Ow, we still need to fix the default explanation, sorry!

Updated! I still want to add the possiblity to create links, so wanna add an extra button to the toolbar for that!

hasan-sh:
> Ping me when that's ready so I can approve the PR :D

@jpelay I want to do that later, perhaps when we all use it and have further feedback:)

hasan-sh:
> we need a way to convert the teacher adventure's to this, otherwise they're also shown in a funky way.

Yes that's very true! 



> I have a further comment, and I don't know if this was if I add a code section, then I have no way to get out of the code section, look:

If you just pressed 3 enters hahah that's the escape. I thought of adding that to the explanation as well. Also, i think i'll create an issue to explain this new editor in teacher manual.

jpelay (approved):
Great work!

PR #5030, opened 2024-01-24
🪲 Adds input field for filtering programs

### LINKED ISSUE #4346: [BUG] Filtering students programs returns My Programs Page
Reported by Zenior on the Discord page:

> If I choose one of my students from the classlist to see their programms, this seems to work, but as soon as I select a level (15 in my case), this screen changes into MY programms.

This happens when you try to filter by adventure too.

jpelay:
Adds an input field in the filter from, so the user can actually filter the program's in the desired user, and not only their own, also changes the value of the label for no programs to be more generic.

Fixes #4346

**How to test**

* Go to a class and select a student page
* Use the filter
* The programs shown should be the student's


Felienne (approved):
Thanks for fixing this long-standing and annoying bug!!

Issue #4396, opened 2023-07-25
Add previous page token in pagination


jpelay:
Right now in the places that have pagination, our links only point forward to the next page, but never to the previous. I think this is very important for a pagination feature, however I'm not entirely sure how to do it so I'm pinging @rix0rrr in case he has any ideas! I tried playing with the `/programs` page and add two buttons: one that goes forward that has the next token, and one that goes back that has the current token, and will become the previous token, but it doesn't work since the token is assigned after filtering! So if we're splitting a page in half in Python, the current and previous token are the same!

rix0rrr:
I have to think about this a little.

My initial thought was: "but you have a *back* button in your browser, is that not good enough? 😬"

To cheat, you could send the current URL into the next page, and use that URL for the `< Previous` button.

PR #5043, opened 2024-01-29
💻 Sorting my program adventures per level and alphabetically

### LINKED ISSUE #5027: 💻My programs adventures sort alphabetically
When I go to the 'my programs' page. There is a drop down menu to search for programs from a specific adventure. The adventures however are not sorted alphabetically. It would be really helpful if it was sorted, to easier find the right adventure. 

afbeelding

Annelein:
Fixes #5027 

**How to test**

Go to My Programs and select a random level, then look at the adventure options. They are now sorted alphabetically and only the adventures for the selected level are shown.


Felienne:
Ah great work @Annelein. Just a small thing that can be improved. In levels, we get a list of all levels:

image

But in adventures, the list is empty:

image

Until we select a level and then it works:

image

Can we either:

* Start by filtering on level 1 (I think this is what we do for the new public adventures search, right @hasan-sh?, so that is consistent) 
* Only show adventures that exist in the list of programs? That is maybe even nicer and more compact?

Annelein:
> * Only show adventures that exist in the list of programs? That is maybe even nicer and more compact?

I like this one! Would it be the same for the levels?
So also only show levels that have adventures in the list of programs?

Felienne:
> > * Only show adventures that exist in the list of programs? That is maybe even nicer and more compact?
> 
> I like this one! Would it be the same for the levels? So also only show levels that have adventures in the list of programs?

Yes, why not both? It makes no sense to show levels that are ot present in the list of programs.

hasan-sh:
> * Start by filtering on level 1 (I think this is what we do for the new public adventures search, right @hasan-sh?, so that is consistent)

Exactly!

Annelein:
@Felienne what do you think? Is this what you had in mind?

Felienne:
> @Felienne what do you think? Is this what you had in mind?

Almost!

It works perfectly with the levels, showing only the ones I have:

image


And then if I select one level, I only get adventures for that level:

image

This is exactly what I had in mind!

However, for the adventures, I get no list, in the beginning:

image

Can we make it so that it also shows all adventures? Then if I pick an adventure, it shows all levels in the dropdown?

Hope that makes sense! (you can also do this in a separate PR if that is easier?)

Felienne (changes requested):
See comment above

Annelein:
Allright @Felienne, it works as we would want it to but for these 2 exceptions:

1. I just found the 'Next page' button, which appears if you have more then 10 programs. Only thing is, once pressed, there is no 'Previous page' button.

Screenshot 2024-02-05 at 14 13 39

- Should I add the 'Previous page' button?
- Should I add 'Programs 1 -10' as text or something, in between those buttons, because it might be confusing that in the search boxes, all the levels and programs are visible, but only 10 are listed. _(I thought my code wasn't working because I didn't see the 'Next page' button for quite some time)._ Hope you understand what I mean?
- Also to make it more clear that only 10 programs are listed, move these buttons and text to the top instead of below?

2. The adventure: 'Introduction' has an adventure_name: 'default'. So it looks like this: `'Introduction':'default'`. Only thing is, if you search for the 'Introduction' adventures, no programs are found (already does this in main). Not sure why it is `'Introduction':'default'` and how I can create a workaround?

Felienne:
> Allright @Felienne, it works as we would want it to

Indeed, I checked and it works great now!

> but I stumbled upon these issues:
> 
> 1. I thought my code wasn't working because a lot of programs were missing but it appeared that only 10 programs are listed. And so I found the 'Next page' button. But if you click it, you can't go back to the previous page.
> 
> Screenshot 2024-02-05 at 14 13 39
> * Should I add the 'Previous page' button?
> * Should I add 'Programs 1 -10' as text or something, in between those buttons, so it is more clear that only 10 programs are listed?

Yes, these are great ideas! We can either do that in this PR, but since this is an existing issue not related to this PR, I am also very ok with doing it separately.

And apparently, the back button is not so easy to do well, so (according to @jpelay and @rix0rrr) we should do this using going back with the back button in the browser (see #4396)

Felienne:
> 2. If you would search on the Introduction adventure, no results were given because of this code in `filtered_programs_for_user` in `database.py`:
> 
> ```
> if adventure == 'default' and program.get('adventure_name') != '':
>    continue
> if adventure != 'default' and program.get('adventure_name') != adventure:
>     continue
> ```

Just a tip for future reference, you can also permalink to parts of the code, that way yiu don't have to copy-paste and people can navigate to the code with a click (and you get a preview) It works like this:

image

And then the link would show like this:

https://github.com/hedyorg/hedy/blob/04ccac1d6959088f0f147dffb028528407096a06/website/database.py#L234

> * Can I just remove the part where it continues if it is an Introduction adventure? Or is there a reason why it is needed somewhere else?

Felienne:
Looking at the git blame, it looks like this was added in the autosave feature (https://github.com/hedyorg/hedy/commit/827754b20de052893bb13edf170339b2a85620b4) so I am not sure if it makes sense when collecting saved programs. Maybe we decided that this intro program was never interesting to show?

Tagging @rix0rrr so he can chime in.

rix0rrr:
> And apparently, the back button is not so easy to do well, so (according to @jpelay and @rix0rrr) we should do this using going back with the back button in the browser (see https://github.com/hedyorg/hedy/issues/4396)

I have an implementation of a "previous page" button in a PR where I reimagine the teacher adventure explorer... but that never got pushed because Hasan took the task ultimately.

The idea though is to keep a stack of "next page" tokens:

```
http://hedy/browse? next=J1tifiBuYj5aKXtiWjdUWElxcjEjeFQ & prev=KylMOl9JVm4xRUh1RC8idlNDZA , KVIyPCxkNiNfdmMzWShudg
```

For the "next page" link, we add the *current* `next` token to the stack of `prev` tokens, and for the "previous page" link we do the reverse.

(There is probably also a solution in which we can iterate the "current page" token in reverse in Dynamo, but that requires some more database work on my part)

> Filtering on `adventure_name` and `default`

I noticed empty `adventure_name`s in the database, which would never show up in a filter if we required the field to always have a value. I think what I had assumed here is the following equivalence:

"The `adventure_name` in the database being empty is exactly equivalent being for the `default` adventure (and *only* empty fields are equivalent to the default adventure)"

Perhaps that's wrong, and we just have some records in the database whose `adventure_name` is empty, and we *also* have adventures in the database whose `adventure_name='default'`.

In that case, the following code might be better than what we have today:

```python
# Use the name found in the database, but treat it as 'default' in case it's missing
program_adventure = program.get('adventure_name', 'default')
if adventure != program_adventure:
    continue
```

Annelein:
As discussed in the meeting: I reverted the if statements back, this is the new issue I created for that: #5072
There is an existing issue for the 'Previous page' button: #4396

@Felienne can you approve so this can be merged?

Felienne:
A cypress test is now failing :(

Felienne (approved):

PR #5045, opened 2024-01-30
💻 Copying my program code

### LINKED ISSUE #5037: 💻 Add "copy code" to my programs page
If you want to reuse code, you have to go to myprograms and then select a program, open it and copy the code. We could add a "copy code" button to the programs list (this idea comes from @AnneliesVlaar, maybe she wants to expand?)

Annelein:
Fixes #5037 

**How to test**
Go to My Programs and run the 'Copy Code' button.
It copies all the code, not just the code previewed. 

Only thing, if the 'Share' button is pressed, the buttons get stretched oddly @hasan-sh could you maybe help me out with that?

Screenshot 2024-01-29 at 20 34 41

@Felienne is this how you had it in mind?


hasan-sh:
> Only thing, if the 'Share' button is pressed, the buttons get stretched oddly @hasan-sh could you maybe help me out with that?

Hi Annelein, great you picked this up! This behavior is caused by having a full width property on the `public_button_container_{i}`, specifically when other buttons are added which happens after you share a program!

hasan-sh (commented):
I think  the submit button supposed to be on the right side.
image


Also, I think we should move some of the buttons to the right as well. Perhaps we could move the delete button since it's not very smooth to have open and then delete next to each other. 

Another thing is that I'm not sure about the copy code button.. it seems rather convoluted to have it in white and all other backgrounds. I believe the best is to create a copy button like:
image


This can then be used anywhere we like! What do you think @Felienne ?

Felienne:
> I think the submit button supposed to be on the right side. image
> 
> Also, I think we should move some of the buttons to the right as well. Perhaps we could move the delete button since it's not very smooth to have open and then delete next to each other.
> 
> Another thing is that I'm not sure about the copy code button.. it seems rather convoluted to have it in white and all other backgrounds. I believe the best is to create a copy button like: image
> 
> This can then be used anywhere we like! What do you think @Felienne ?

Let's discuss tomorrow in the meeting or in the hackathon!

Annelein:
As discussed in the meeting: I will contact Gisel for the designing part, but for now the functionality can be added. 

Screenshot 2024-02-06 at 17 54 02

@Felienne @hasan-sh can one of you approve?

gisellandrade:
Hey, Jesus told me about this  

I worked on this, like hiding the second actions and just showing the main ones like submit and open. Deleting, copying and share would be hidden in that menu. Also, I thought about changing the share button for a drop down that says private or public. Just like i'm showing on the images. Let me know what do you think about this 

Frame 7
Frame 6

Annelein:
@gisellandrade thank you for the suggestions!! Just the copy code button is missing in it, what would you suggest for that?

And I think I will leave it to the others to make a decision, should we create a new issue for it and discuss it there?

Felienne (approved):
Let's merge this! I made an issue for the redesign of the buttons!

PR #5049, opened 2024-01-30
📜Micro-bit code

rmagedon97:
Work in progress for #4586


Felienne:
Thanks for the partial work @rmagedon97! Really useful to see how to get code on the micro:bit.

To get you going in the transpiler I have added a parameter to the transpiler, so that we can output different code if we microbit variable is set to True.

Some steps for you to do now:

* Update test test (I left a comment)
* Then change print() in the higher levels, if you do not know how to, maybe @akseron can help
* Add a "send to microbit" button to the front-end, if you don't know how to, ask @Annelein or @hasan-sh and connect the button to the transpiler

Steps later on:
* Think of other micro:bit features that we can add to Hedy. Maybe something with the LED? That will require new code in Hedy, but at least you can make a few suggestions and look up the corresponding micro:bit code

tests/test_level/test_level_04.py
```diff
@@ -35,6 +35,22 @@ def test_print_single_quoted_text(self):
...
+            from microbit import *
+            while True:
+                display.scroll('a')""")
+
+        self.multi_level_tester(
+            code=code,
+            translate=False,
+            skip_faulty=False,
+            expected=expected,
+            max_level=11,  # teodor, set this to 17, see the test fail and moke sure that it then does not fail anymore!
```
Felienne, line 50: We need to remove this comment, and also do we need to make it into 17 and update the code (either in this PR or in a separate one)

templates/incl/editor-and-output.html
```diff
@@ -144,6 +144,7 @@
         </div>
         <div id="saveFilesContainer" class="hidden">
           <button id="saveFiles" class="green-btn" onclick="hedyApp.saveMachineFiles();">💾</button>
+          <button id="saveMicrobit" class="green-btn" onclick="hedyApp.saveMachineFiles();">⚡</button>
```
Felienne, line 147: We could remove this button but merge the PR, so that we don't run into merge conflicts but also don't bother current users?

tests/Tester.py
```diff
@@ -278,7 +282,12 @@ def single_level_tester(
                     # <- use this to run tests locally with unittest
                     skipped_commands = ['ask', 'input', 'clear', 'play']
                     if not any(x for x in skipped_commands if x in all_commands):
-                        self.assertTrue(self.validate_Python_code(result))
+                        if microbit:
```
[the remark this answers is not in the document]
rmagedon97: Hello @Felienne, you are exactly right this was just for me to test if it works i update it to just skip if it is microbit.

templates/incl/editor-and-output.html
rmagedon97: i'll remove the button now and will push

Felienne (approved):
This can be merged!

PR #5050, opened 2024-01-30
🪲 Fix highlighting of function calls inside expressions

### LINKED ISSUE #5038: 🪲 Functions call not being highlighted properly
When a function is called inside a print, is not being parsed correctly by lezer when there's a newline after the print, or there are other expressions. Instead of parsing it as a Call(call, Text, with, Arguments) its just Expression(text), Expression(text) etc

image

jpelay:
There was an issue where calling functions inside `print` or similar commands wouldn't work because it wasn't included in the `Expression` rule. Therefore this code:

```
print call salute with name ' so pretty' 
```

was parsed as

```
Program(Command(Print(print,Expression(Text),Expression(Text),Expression(Text),Expression(Text),Expression(String))))
```

As you can see, instead of parsing the call as a single command, it parses it as a bunch of single Text instances. It looks like this:

image

To fix this, I made the following changes:

* Including `call` and `with` in the `specialize` type of keywords, instead of `extends`, essetianlly giving them highest priority by parsing them as keywords not mattering where they're located in the code.
* Adding a precedence `comma` for the list of arguments, so the parser doesn't get confused

Resulting in the previous code being parsed correctly:

```
Program(Command(Print(print,Expression(Call(call,Text,with,Arguments(Expression(Text)))),Expression(String))))
```
image

Fixes #5038

**How to test**

* Paste this code from level 12 to 17.
```
print call salute with name ' so pretty' 
```


Felienne (approved):
Works!!

image

Issue #5054, opened 2024-02-01
🪲 Problems with new slides code loading


Felienne:
f.e. level 6 (https://hedy.org/slides/6#/4/0/0)

<img width="1236" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/e0da0e72-443d-4142-bc56-fd793b1e3b0e">

From Discord:
```
De meeste slides gaan goed (level 4,5 bekeken), maar op level 6, slide 4 en 5 gaat het fout (een error scherm) (ik probeer een screendump toe te voegen, maar discord werkt me tegen)

Ook een errorscherm in Level 14, slide 4 (de rest t/m 13 lijkt goed)
```

Also, level 0 slides do not show code, from Discord:

```
Ik begin morgen met twee nieuwe groepen programmeren met Hedy. Ik zie dat in de 'introduction slides' alle IDE-vensters leeg zijn.
```

Felienne:
Thanks for the detailed feedback https://github.com/LegoPink! There was a specific issue when / appeared in code and your list of broken slides helped me find it quickly! 

Issue should be fixed later today!

PR #5057, opened 2024-02-01
🪲 Fix slides

Felienne:
Fixes both issues described in #5054


hasan-sh (approved):
Amazingggg, everything works again:
image

image

Issue #4994, opened 2024-01-13
🧪 Test code in slides 


Felienne:
In https://github.com/hedyorg/hedy/pull/4979 keywords were translated, not a real issue since @JanDommerholt noticed himself, but that should have been tested!

Since #4906 we load code directly from the yaml, so we should also add that to the test infra!

Felienne:
Done!

PR #5058, opened 2024-02-01
🧪 Test code in slides

Felienne:
WIP for #4994


MarleenGilsing:
the pre-commit might be a bit too enthousiastic if it comes to adding

rix0rrr (approved):

PR #5060, opened 2024-02-01
💻 customize class page enhancement

### LINKED ISSUE #3976: [INTERFACE] Customize class page still needs some love

hasan-sh:
Fixes #3976 
**How to test?**
- go to the customize-class page
- see the new layout, and try it out!


Felienne:
This looks a lot better!! 

A small issue is with the opening dates, if you click the button it appears below:

image


But if your screen is small you might not see the change (only the arrow switches direction). This is a generic design issue (that also happened with the "add students" button, so we can discuss it separately!

Felienne (approved):

PR #5071, opened 2024-02-06
🪲 FIx not-in command missing from level 8 and up

### LINKED ISSUE #3934: [BUG/CONTENT] issues with "not in"
**Describe the bug**

We recently added `not in` (see https://github.com/hedyorg/hedy/pull/3739) but there are a few open issues.
1. "not in" doesn't work in level 9: It works in Level 5, but not in 9: 

```
j = yes, Yes
a = ask "answer?"
if a not in j
    print "ok"
else
    print "no"
```

2. Syntax highlighting of "not in" doesn't work in any level.

boryanagoncharenko:
- Fixes #3934
- Add support for the not-in-list command starting from level 8
- Add syntax highlighting for the not-in-list command starting from level 5

**How to test**

The language change is covered with unit tests. Ensure that the tests check all cases including static type checking.
* Run Hedy locally and execute the following snippets for the specified levels. In all cases, the output of the program should be 'not in' and the syntax 'not in' should be highlighted in pink in the editor:
```
# level 5
colors is green, yellow
choice is red
if choice not in colors print 'not in'
``` 

```
# level 8
colors is green, yellow
choice is red
if choice not in colors
  print 'not in'
```

```
# level 12
colors is 'green', 'yellow'
choice is 'red'
if choice not in colors
  print 'not in'
```

```
# level 16
colors is ['green', 'yellow']
choice is 'red'
if choice not in colors
  print 'not in'
```


highlighting/lezer-grammars/level10.grammar
```diff
@@ -46,8 +46,8 @@ Else { elses+ }
 
 Condition { 
     EqualityCheck { Text is+ (String | Expression | pressed+) } |
-    InListCheck { Text ins+  Text } |
-    NotInListCheck { Text not_in+  Text }
+    InListCheck { Text ins+ Text } |
+    NotInListCheck { Text not+ ins+ Text }
```
jpelay, line 50: Hi Boryana! So cool that you're looking at the highlighter a little bit! If you have any comments with respect to that, you can let me know. 

With respect to this change, @Felienne  and I have discussed before wether having a keyword for `not` and for `in`, but it turns out that chinese actually uses just one symbol for it, so this would fail in that language. Therefore, we think is better to keep keywords under a single name, even if they have several words, since many of the languages already have spaces for many of the keywords! Does this make sense to you? What do you think?

PR #5103, opened 2024-02-08
Implement rate limiter on reset password

### LINKED ISSUE #5102: 💻 Implement a request rate limiter
**Description**
Some routes automatically sent an e-mail. Either to ourselves, or to a user (for example for the 'reset password'). This introduces risks as there is no limit on the amount of times a user can make this request. We should implement either a username, e-mail and/or ip-address rate limiter.

https://flask-limiter.readthedocs.io/en/stable/

TiBiBa:
This PR fixes #5102.


requirements.txt
```diff
@@ -28,10 +28,11 @@ turtlethread>=0.0.6
 pygame==2.1.2
 pre-commit==2.20.0
 babel==2.14.0
-jinja-partials==0.1.1
+jinja-partials==0.1.1clea
```
Felienne, line 31: I suspect this is a typo @TiBiBa? Did you mean just to leave it at 0.1.1 or is there another version needed?

TiBiBa: I think this is a typo, will look into this. We should also decide on a nice test case, but that might be easy to fix on Alpha.

Felienne:
Here is another old left over PR from the hackathon that could be either closed or finished?

utils.py
```diff
@@ -22,6 +22,15 @@
 from ruamel import yaml
 import commonmark
 
+from flask_limiter import Limiter
+from flask_limiter.util import get_remote_address
+
+# Implement the rate limiter
+limiter = Limiter(
+    get_remote_address,
+    storage_uri="memory://",
```
jpelay, line 31: I managed to get the limiter working, but for us to make this really work on production we'd have to use a persistent storage solution, like memcached, redis or MongoDB. Now, all of these solutions come at a cost, since as far as I can tell, we'd need to pay a cloud provider for this (e.g; [Memcached Cloud](https://devcenter.heroku.com/articles/memcachedcloud)), with varying costs. @TiBiBa Did you have something in mind when opening this PR about this, do you know a way to fire a memcached server on Heroky without incurring on costs?

PR #5115, opened 2024-02-09
🪲 🖊️  repeat body error in level 9 and over

### LINKED ISSUE #4827: 🪲 🖊️ missing body error in level 9 and over
This program:

```
repeat 5 times
```

gives a nice error in level 8, but not above:

image

Should not be hard to fix, since we have a rule and also an error message already, but since it involves error productions I am assigning @akseron for now :)

akseron:
Fixes #4827 

**How to test**
Attempt to use the following code in level 9 and higher:
`repeat 5 times`


akseron:
Appears to be more difficult to implement than previously expected. 

An error keeps popping up sporadically for test 'test_repeat_nested_in_repeat', sometimes it seems to work and sometimes it does not which is unexpected. The test itself seems to work properly, while the error message is that the 'error_repeat_no_command' error production from level7-Additions.lark is triggered. This is weird as it should only be triggerable after the repeat command. 

This is probably? due to interference from _if_less_command in level 7. The rule seems to influence how repeat works, but trying to add repeat error rules on top of it does not work.

akseron:
image
image
This is the same code for both instances. In one instance the code runs properly, in the other the repeat error production is not detected and it does not. The two instances fluctuate like a coin flip every time you run the Hedy app file.

Felienne:
> This is the same code for both instances. In one instance the code runs properly, in the other the repeat error production is not detected and it does not. The two instances fluctuate like a coin flip every time you run the Hedy app file.

Hi @akseron! I have not dived into this specific issue but if it is like a coin flip, it is a grammar issue where there are two options and both fit. So you probably need to change the grammar to indicate preference. Maybe @boryanagoncharenko can help out because she created the if_less commands if I remember correctly.

Felienne:
(if you don't have time @boryanagoncharenko I can also take a peek but then it will probably be after the voorjaarsvakantie)

boryanagoncharenko:
This is indeed grammar ambiguity. The `error_repeat_no_command` grammar rule by definition is part of the `repeat` grammar rule which makes it extra hard to differentiate the two rules when the `repeat` is nested in an `if` statement. The only way I was able to make this work is by giving the `repeat` rule higher priority (this is done via the '.1' suffix, see https://lark-parser.readthedocs.io/en/latest/grammar.html#id1). However, we have not used rule priority so far in hedy and my guess is that there is a simpler way of achieving the same result. So @Felienne probably you should have look.

grammars/level9-Additions.lark
```diff
@@ -1,3 +1,3 @@
 // only remove repeat error commands
 
-command:+= repeat-=repeat | error_repeat_no_command | error_repeat_no_print | error_repeat_no_times
+command:+= repeat | error_repeat_no_command-=repeat | error_repeat_no_command | error_repeat_no_print | error_repeat_no_times
```
Felienne, line 3: Following what @boryanagoncharenko just added, you can also give priorities like this, in the |. If we have X | Y | Z, then X will be tried first, if that does not fit, it will go to Y etc. But if *both* match, then it is a coin toss. 

You can try to exploit this by rewriting rules a bit, but that can be complex. So maybe try the .1 trick first? If not I can probably fix it after the voorjaarsvakantie!

Felienne (dismissed):

boryanagoncharenko:
Here is the proof that rule priority has not been used in Hedy so far. Adding the suffix '.1' entered the source map and broke some debugger tests. I added a temporary fix for now in case we decide to stick to this solution.

akseron:
Wow, what a nice feature, that indeed solves it nicely! I suspect we might even be able to leave out the rules in level 9, resulting in both added functionality of the two error rules which are not removed, and less code. I will also look into the alternatives which Felienne suggested to see whether playing around with the rules might also be another solution, hopefully resolving the ambiguity at its core.

akseron:
I will leave this as it is, because my current knowledge of how lark works does not allow for me to find a solution. I am mostly confused about the fact that in level 9, there is a need to both add AND remove the repeat command. I do not know how it works, but somehow the parser won't work without it, and I feel as though it might be a hint towards solving the ambiguity through error productions. Do you any of you two have a clue as to the reason behind this?

Felienne:
> I am mostly confused about the fact that in level 9, there is a need to both add AND remove the repeat command.

Ahhh there I can help! We do that to. change the order and thus the presedence! But you can totally leave this as is, you have done enough work, and I can finish this after the break!

boryanagoncharenko:
Excellent remark. It makes me admit that I shamelessly took a shortcut in the grammar of level 9.

We could just remove the error_repeat_no_print | error_repeat_no_times rules. However, it seems that the merging functionality does not allow to only subtract rules (see https://github.com/hedyorg/hedy/blob/fb3e9fb17f0ddb8d76fa544b3a9dbf685f452d5f/hedy.py#L2780). There must be addition for any of the other operators to work. @Felienne would having the -= and > operators work independently of the += operator be something nice to have?

Felienne:
> @Felienne would having the -= and > operators work independently of the += operator be something nice to have?

Yes! Do you want to pick this up?

And can we merge this PR as it is or do we want to make the merger changes in this one?

boryanagoncharenko:
If this is ok, I would like to clean up a bit the commits in this PR, so that it looks a bit more presentable.
And I will create a separate issue for the grammar merging operators change since it may inflate the issue too much.

Felienne:
> If this is ok, I would like to clean up a bit the commits in this PR, so that it looks a bit more presentable.
> 
> And I will create a separate issue for the grammar merging operators change since it may inflate the issue too much.

Cleaning up is not really needed because we will squash merge on acceptance!

boryanagoncharenko:
> > If this is ok, I would like to clean up a bit the commits in this PR, so that it looks a bit more presentable.
> > And I will create a separate issue for the grammar merging operators change since it may inflate the issue too much.
> 
> Cleaning up is not really needed because we will squash merge on acceptance!

I actually thought I did not add all the test I had in mind, but I just checked and I did add them. So we are good for a review!

grammars/level8-Additions.lark
```diff
@@ -1,6 +1,6 @@
 command:+= error_repeat_dep_8 | error_ifelse | error_ifpressed_missing_else | ifpressed ifpressed_elses? | ifs elses?-= error_invalid_space | ifelse
 
-repeat: _REPEAT (INT | var_access) _TIMES _SPACE? _EOL (_SPACE command) (_EOL _SPACE command)* _EOL? _END_BLOCK
+repeat.1: _REPEAT (INT | var_access) _TIMES _SPACE? _EOL (_SPACE command) (_EOL _SPACE command)* _EOL? _END_BLOCK
```
jpelay, line 3: I think adding this priority is fine, sometimes Hedy's grammar is difficult to disambiguate, and leveraging tricks like this to make it work is fine.

jpelay (approved):
Hi @akseron  ad @boryanagoncharenko! I think the addition of the precedence rule in this case is fine, and everything works so I think this is ready to go!

Felienne:
> And I will create a separate issue for the grammar merging operators change since it may inflate the issue too much.

Thanks!

Felienne (approved):

PR #5116, opened 2024-02-09
💻 Offline Hedy for Windows

rix0rrr:
This adds a build for a standalone Hedy package on Windows.

The end result will be a directory with an executable and some data files hidden in a directory called `_internal`. It will use the dev database in JSON format.

The build is triggered with the (hidden) doit task `doit run _offline`; this also adds a GitHub Workflows script that performs the build.


jpelay (approved):
Let's goooo!

Issue #4765, opened 2023-11-16
[BUG] Debugger doesn't work with pressit command


jpelay:
This is a know issue and limitation of our current debugger. Now that I finished my project, maybe we can keep improving on this aspect of the debugger.

Our first step should be looking if we can  fix the integration with PyGame and the debugger. If this isn't possible, I think we could add a message to the user with information about the debugger not working this feature.

jpelay:
Maybe this is a problem that @juliabolt would like to tackle in to familiarize with the debugger?

juliabolt:
yes I can look into it :)

jpelay:
Now that we replaced the PyGame code with our own implementation on #5117, we can start looking into this now.

AnneliesVlaar:
presit functionallity completely removed

PR #5117, opened 2024-02-09
🧹 Refactor Pygame Code

ToniSkulj:
PR for removing the use of Pygame within Hedy.
We still want to keep the functionality of "if button pressed".


Felienne:
Hi @jpelay and @ToniSkulj!

I have updates all level tests, except for one that I did not know how to deal with (I commented that out in https://github.com/hedyorg/hedy/pull/5117/commits/0061806df62fa378fc10c7b86b416f6efb7920eb) 

I am not sure if more it needed at the moment, so I will leave this to you two for now!

ToniSkulj:
Hi @Felienne!
Thank you for adding the tests! I fixed the level 8 test, all tests should be passing now.

The only case that remains is when the user uses elifs without an else. Currently, it does not do anything.
I will keep you updated when I find something to solve that.

Felienne:
> Hi @Felienne! Thank you for adding the tests! I fixed the level 8 test, all tests should be passing now.
> 
> The only case that remains is when the user uses elifs without an else. Currently, it does not do anything. I will keep you updated when I find something to solve that.

Awesome, thanks so much for seeing this through! If you have an idea for the elif/else issue, do keep us posted. Otherwise we can always throw an exception and say: you can't use elif in pressed without else.

jpelay (approved):
This is such a great work!!! Thank you SO MUCH for this! This is a great PR and everything worked, but a couple of things that I fixed myself.

There's an issue still with the debugger, but there's an issue opened fort that already on #4765!

PR #5133, opened 2024-02-12
🪲 Favourite misspelling

### LINKED ISSUE #5131: 🪲 Favourite program undefined error at public profile
https://github.com/hedyorg/hedy/blob/48c28f9a9190809a17d303341d31122ba8861521/templates/public-page.html#L135

incorrect spelling: adventure=favorite_program

should be: adventure=favourite_program

Screenshot 2024-02-12 at 10 22 36

Annelein:
Fixes #5131

### DIFF
--- a/templates/public-page.html
+++ b/templates/public-page.html
@@ -132,7 +132,7 @@
             <div class="border rounded-lg border-yellow-500 mb-4 py-2 w-full">
                 <h3 class="text-center z-10 font-bold pb-0 mb-0 text-2xl">{{_('favorite_program')}}</h3>
                 <div class="flex flex-wrap mx-auto justify-center gap-4 turn-pre-into-ace" id="user_favourite_program">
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -2540,9 +2540,9 @@ def public_user_page(username):
         sorted_adventure_programs = hedy_content.Adventures(
             g.lang).get_sorted_adventure_programs(all_programs, adventure_names)
 
-        favourite_program = None
+        favorite_program = None
         if 'favourite_program' in user_public_info and user_public_info['favourite_program']:
-            favourite_program = DATABASE.program_by_id(
+            favorite_program = DATABASE.program_by_id(
                 user_public_info['favourite_program'])
```
Annelein, line 2546: I'm not sure if it makes sense haha @hasan-sh because now in one function there are different translations

hasan-sh: It's a variable that's also saved as favourite, terrible indeed! It would stil be part of what we could change later on.

hasan-sh (approved):

PR #5134, opened 2024-02-12
🧹 Update python to 3.12

Felienne:
Update Python

**How to test**
Verify that nothing is broken


rix0rrr:
For PyCharm support, I found [this thread](https://youtrack.jetbrains.com/issue/PY-61694/Packaging-support-for-Python-3.12). They're saying the following should work:

```
venv/bin/python -m ensurepip --upgrade

# or in your case
venv-12/bin/python -m ensurepip --upgrade
```

rix0rrr:
For the installation problems, older `pygame`s are not guaranteed to work with newer Python versions, but looks like the newest one (`2.5.2`) should work:

https://github.com/pygame/pygame/releases/tag/2.5.2

rix0rrr:
<3 would love to ship this.

Upgrade instructions are here: https://github.com/hedyorg/hedy/wiki/Hedy-Development-Process#updating-python-versions

jpelay (approved):

PR #5136, opened 2024-02-12
🪲 When editing code while ask prompt is open, editor now stays focused

### LINKED ISSUE #4742: [BUG] When editing the code while ask prompt is open, the editor loses focus
**Describe the bug**

Consider this video:

When an ask prompt is open, and user decides to edit the code, by clicking on the editor, the ask prompt correctly closes and the editor gets focus. But annoyingly, after a single keystroke the editor loses focus again, and the user has to click it.

https://github.com/hedyorg/hedy/assets/45865185/ef97d32f-344c-488b-8a47-59e946b379fb

This happens both with CodeMirror and Ace, so it's likely not a problem with the editor itself.

Annelein:
Fixes #4742

**This is the first time that I'm working with the editor so it could be that this line is necessary, but as far as I could test, it seems to solve the issue without it and not run into issues**

**How to test**
Go to level 1 > Adventure 'ask' > Copy the example code into the editor > Run program > Edit the code > It now stays focused

https://github.com/hedyorg/hedy/assets/48122190/6b429ba4-6cc9-4295-b4dc-94f5949277ba


Felienne:
Maybe best if @jpelay checks this? (It seems to be working for me!)

jpelay (approved):
Love this!!!! It works great!! This bug was really annoying

PR #5142, opened 2024-02-13
🚚 redirect to hedy.org

rix0rrr:
Google is still indexing old domains for Hedy (`hedy-beta.herokuapp.com` and `hedycode.com`).

When Hedy is loaded from any of the old domains, do a redirect to the modern domain (`hedy.org`).

**How to test**

Locally, add `localhost` (or `127.0.0.1`, whichever one you use) to the list of redirected hosts, then load Hedy locally and observe that you get redirected to https://hedy.org/. Don't commit those changes 😉


jpelay (approved):
Works!

PR #5143, opened 2024-02-13
🪲 Prevent failure page after logout

rix0rrr:
We try to keep the user on the same page after they log out, to not disturb their place on the website. This leads to the experience that when you log out and you are on a page that was only visible if you are logged in, you will see an error message.

Add a fix that the first page load that fails due to permissions after a logout is turned into a redirect to the front page. That way, if you were on an unprivileged page you stay where you are, but if you were on a privileged page you will not see an error.

**How to test**

- Log in as a teacher
- Go to a page that's only for teachers
- Log out
- You are redirected to the front page
- Log in as a teacher
- Go to the `/hedy` page to edit a program.
- Log out
- You are still on the code page.


Felienne (approved):
Awww this is nicer indeed!

PR #5145, opened 2024-02-13
💻 Limit tags to 10

### LINKED ISSUE #4815: 💻  Limit the number of tags shown in the tags dropdown.
**Idea incl level**
Check this comment https://github.com/hedyorg/hedy/pull/4759#issuecomment-1828083043

Annelein:
Fixes #4815 

Limiting the public tags to only 10 tags


hasan-sh (changes requested):
If there are more than 10 tags, the rest won't be suggested to the user and thus duplicate tags will be and is a present problem. Therefare i said it might be a bit difficult.. so with the current implementation, we need to limit the number of elements in the front end only!

Annelein:
@hasan-sh is this what you mean?

hasan-sh (commented):
Two things: the loop.index starts from 1, so it should be <= 10 or < 11. Another thing is that it still doesn't work when there are more than 9 and you search for them. That's because they aren't rendered at all.

The only solution i can think of is moving the logic of hidding or showing an item to TS to have more control. If you need further help with that, let's chat on discord.

PR #5146, opened 2024-02-13
🚚 register users' actions

### LINKED ISSUE #5137: Log front-end features
As discussed in the hackathon, @hasan-sh is working on it with @rania74!

hasan-sh:
Fixes #5137 

This is a basic service that's aimed to give us more clarity on what's important to users and whether certain features are used or not. This service has two main layers:

**Front-end**
- listen on events of certain features (e.g., clicks on buttons or dropdowns) and register them in localStorage
- if the user is logged in, we send all registered events to the server every 5 minutes (change it to 3s while testing!)
- if user is not logged in, we just gather these events until they are.

**Back-end**
- requires a logged in user
- get these events from frontend 
- log them into s3-logger in prod, or a csv file locally.

**How to test?**
- do some activities such as creating a class or an adventure, debugging the code, or any other event that's fired by the specified in ELEMENT_TO_TRACK within the tracking.ts file.
- see the events recorded in the `tracking_data.csv` file.


config.py
```diff
@@ -37,6 +37,13 @@
         'postfix': ('-' + dyno if dyno else '') + '-' + str(os.getpid()),
         'region': 'eu-west-1'
     },
+    's3-parse-tracking-logs': {
+        'bucket': 'hedy-parse-tracking-logs',
```
rix0rrr, line 41: "parse" means having to do with parsing and executing Hedy programs. I don't think this has to do with tracking of just parsed programs, so I would avoid using the word "parse" here. Call it "s3-tracking-logs" everywhere instead.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+let lastActiveTime = Date.now();
+let changesSent = false;
+
+
+export function initializeTracking() {
+    document.addEventListener("DOMContentLoaded", documentLoaded);
+}
+
+
+function documentLoaded() {
```
rix0rrr, line 98: It looks like we won't track user activity if they spend less than 5 minutes on the page and then navigate away. I know that if they return, we will get another chance to log those events in the future because we held on to them in localStorage, so it might not be all bad... but there is a limit to the size of localStorage as well. Have a look in `app.ts`, where in the program autosaving we make an attempt to send off pending data to the server when the user closes the window.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
+import { postJson } from "./comm";
+import { isLoggedIn } from "./utils";
+
+// const WAITING_TIME = 5 * 60 * 1000; // 5min in milliseconds
+const WAITING_TIME = 3000; // 3s for testing
+const ELEMENT_TO_TRACK = [
```
rix0rrr, line 6: To make this list easier to manage, what you could also do use use another data attribute on the HTML. For example:

```
<button data-record>
```

Then, in your onClick handler, you can test for the presence of that attribute. You can even use the attribute value to pass in parameters for the tracking, like the name under which the even should be recorded (you could still default to using the `name` or `id` if the attribute doesn't have a value)

```
<button data-record="run-click">
```

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+    // Explore page buttons
+    "explore_page_adventure",
+    "explore_page_level"
+];
+const CLICK_COUNTS = "clickCounts";
+const LAST_ACTIVE = "lastActiveTime";
+const INTERVAL_KEY = "interval";
+
+
+let clickCounts: any = [];
```
rix0rrr, line 88: I don't think you need to the global variable of `clickCounts`. Looks like all places where this is used, you read it from local storage directly (or you COULD read it from global storage directly). If you store the same information in multiple places, you only run the risk of the copies getting out of sync if and when the code gets modified later, so I would recommend against it.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+    lastActiveTime = handleLocalStorage(LAST_ACTIVE, Date.now());
+    
+    if (isLoggedIn()) {
+        removeTrackingInterval(); // Possibly removing lingering ones.
+        setTrackingInterval(); // Resume with fresh timer
+    }
+}
+
+function removeTrackingInterval() {
+    const storedData = localStorage.getItem(INTERVAL_KEY);
```
rix0rrr, line 115: I'm pretty sure that timers will not survive a page reload. So you can store the handle of the timer in a global variable, no need to store it in `localStorage`.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export function initializeTracking() {
+    document.addEventListener("DOMContentLoaded", documentLoaded);
+}
+
+
+function documentLoaded() {
+    // attach events
+    document.addEventListener('click', trackEvent);
+    document.addEventListener('change', trackEvent);
```
rix0rrr, line 101: I really like how you're using a single document-level listener to catch all events! Very clever and efficient.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+            
+            // console.log(target, clickCounts)
+            // console.log(`Event: ${event.type}, Element ID or Name: ${elementIdOrName}, Click Count: ${clickCounts}`);
+            handleUserActivity(clickCounts);
+        } 
+    }
+}
+
+// Function to handle user activity
+function handleUserActivity(clickCounts: any) {
```
rix0rrr, line 167: Please take a couple of moments to define a type for the data you're storing, so you don't have to use `any`. The compiler will be able to help you better, and it will be more clear to other people later what information you're storing and why:

```
interface ClickEvent {
  time: number,
  id: string,
  page: string,
  extra: any,
}
```

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+    lastActiveTime = handleLocalStorage(LAST_ACTIVE, Date.now());
+    clickCounts = handleLocalStorage(CLICK_COUNTS, clickCounts);
+    changesSent = false;
+}
+
+// Retrieve or set items in local storage.
+function handleLocalStorage(item: string, value: any = undefined) {
```
rix0rrr, line 174: This function is doing too much, and the word "handle" is a bit vague. It seems to be doing triple double duty:

- Loading from localStorage
- Writing to localStorage
- Knowing the types of keys it's reading and writing.

The last bit first: if the value you're storing is a number, `JSON.parse()` and `parseInt()` will do the same thing, so I think you can simplify and always say you `JSON.parse()` whatever is in `localStorage` and always `JSON.stringify()` before storing. 

As for the double duty: two different functions for reading and writing are simple and easy to understand:

```ts
function readLocalStorage(key: string, defaultValue: any): any;
function writeLocalStorage(key: string, value: any);

// Extra neat
function readLocalStorage<A>(key: string, defaultValue: A): A;
```

If you really want a single function both for reading and writing, I think it's more clear to have that behavior depend on if the user passes in a value or not, so it sort of behaves like an overloaded function. If they supply a `value` argument, then they want to write. If they don't supply a `value` argument, then they want to read. The difference now depends on what the user passes in AND what is already in `localStorage`, and it makes it hard to reason about what the function is doing.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+            value = JSON.parse(retrievedItem);
+        } else if (item === LAST_ACTIVE) {
+            value = parseInt(retrievedItem);
+        }
+    }
+    return value;
+}
+
+// Function to check user activity and send request if inactive for 5 minutes
+async function checkUserActivity() {
```
rix0rrr, line 190: I like that you're batching data up, and not sending individual events! We're doing this to be efficient about our network requests and CPU usage. Any amount of batching on a reasonable time scale will do just fine. I don't think we need to wait for the user definitely not having done anything for 5 minutes; we can send every 5 minutes regardless of when the last activity was. We will get the batching benefit either way -- and by waiting for 5 minutes of inactivity, there is a greater chance that this condition will never be satisfied, because kids are constantly doing something. If a user never sits idle for 5 minutes, we will never get any click tracking for them!

Similarly, I don't think we need to reset our interval when users switch away from the browser window. We already have a test to not send an empty array, so if they switch away and stop using us, we won't send any network requests. And a timer running in the background at an interval of 5 minutes is not going to appreciably affect the CPU usage or battery of the kid's computer.

static/js/tracking.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Function to send request to the server
+async function sendRequestToServer() {
+    try {
+        const data = handleLocalStorage(CLICK_COUNTS)
+        if (data.length) {
+            // console.log('Sending request to server...');
+            await postJson('/tracking', data);
+            handleUserActivity([]);
```
rix0rrr, line 213: Because this function is `async`, there is a race condition here. The following sequence of events can happen:

1. We read the current set of clicks from LocalStorage
2. We send them off to the server (this is async, so other code can run in the mean time)
3. User interacts with the page, adding a new event to the array
4. The response comes back, this function resumes
5. We clear the LocalStorage array

We've now dropped the additional user event on step 3.

To fix this, we can store the length of the array that we've sent off to the server, and when we go to clear the array do `array.splice(0, amountOfRecordsWeSent)`, so that if new ones got added in the mean time they don't get dropped.

Do we need `changesSent`? When we send changes, we reset the array to empty, so the fact whether the array is empty or non-empty can by itself be a signal for whether there are things to send or not.

static/js/utils.ts
```diff
@@ -0,0 +1,7 @@
+
+export function isLoggedIn() {
+    if (document.body.dataset["loggedIn"]) {
+        return parseInt(document.body.dataset["loggedIn"])
+    }
+    return false;
```
rix0rrr, line 6: The return type of this function ends up being `number | boolean`. That's not a huge problem because both can be tested for truthiness, but you probably intend to just be returning a `boolean`. Test the data attribute by doing:

```
document.body.dataset.loggedIn === '1'
```

website/aws_helpers.py
```diff
@@ -23,15 +23,15 @@ def s3_querylog_transmitter_from_env():
     return make_s3_transmitter(config.config["s3-query-logs"])
 
 
-def s3_parselog_transmitter_from_env():
+def s3_parselog_transmitter_from_env(config_key="s3-parse-logs"):
```
rix0rrr, line 26: This was called "parselog_transmitter" because it was only used for logs that have to do with parsing. If we're going to make it more generic and have it be about logs that don't have to do with parsing, you can rename it more generically to `s3_log_transmitter_from_env` or something.

website/s3_logger.py
rix0rrr, line 1: Unfortunately, you can't do it like this. The reason is that the S3 Logger uses a global variable, `S3_LOG_QUEUE`, which is shared by all `S3ParseLogger`s, and then will all write to the same S3 bucket. Since you want to write to 2 different S3 buckets, you cannot reuse these classes. You could try to untangle this, but given that you want your default logging to be different than the current `NullLogger` as well, I think the most efficient solution would be to copy/paste this file to a new logger file (`activity_logger.py` ?) that you use for activity logging. You can have the default logger be to write to a file, and we can leave the refactor for later.

The standard S3 ParserLog transmitter writes JSON files, so I would recommend writing a JSON file in your local logger as well. Otherwise you might write scripts to process files in this way, but on the actual server they would look different.

Also -- do we want this CSV file to be created on every developer's machine, and also on any teacher that uses the new Offline Hedy?  Or was this just for you, to test? You might want to gate the use of the `CsvFileLogger` behind an environment variable so people can use it to test the tracking code whenever they want, but that it's not on by default.

rix0rrr (commented):
Hey! I like the approach! It's simple and effective!

I have one specific concern around the writing to S3: I'm pretty sure that the way you're repurposing `s3_logger.py` is going to interfere with the parse logging, because there is a global variable in that file that controles where the files get written, which gets overwritten any time anyone calls `logger_from_env()`. Probably copy/pasting to a new file and editing will be not the best solution, but the most expeditious one 😬.

Another concern, won't break anything but worth thinking about: I'm not sure if we'll get as much data as we want with the current sending regime. Batching is fantastic, but by *also* waiting for inactivity I feel there's a risk that we may never receive anything. 

All the rest of my comments are around style or readability, you can take those or leave them as you want.

Finally, as a general comment, "tracking" sounds a bit ominous 🙃. Maybe try to find an alternative word like "usage" or "interaction" or "activity".

website/s3_logger.py
hasan-sh: Probably it's better to refactor it now. I could simply add the `S3_LOG_QUEUE` to the constructor and initialize each time we use it. And good point about the env var and it was for testing indeed. Will add that too.

activity_data.json
rix0rrr, line 1: Did you mean to commit this?

Maybe remove this file and add it to `.gitignore`.

website/s3_logger.py
```diff
@@ -1,38 +1,70 @@
 import logging
+import json
+import os
 
 from . import aws_helpers, log_queue
 
 logger = logging.getLogger(__name__)
 
 
+should_log_activity = os.getenv("LOG_USER_ACTIVITY")
```
rix0rrr, line 10: If you're making `s3_logger` generic, then this should probably go into `user_activity.py`, because it deals with that.

website/s3_logger.py
```diff
@@ -1,38 +1,70 @@
...
+
+
 class NullLogger:
     """A logger that doesn't actually do anything."""
 
+    def __init__(self, file_path="activity_data.json", **kwargs):
+        self.is_tracking = kwargs.get("tracking")
+        self.file_path = file_path
```
rix0rrr, line 18: `NullLogger` is intended to mean "have the classes, but don't actually log".

If you change the `NullLogger` to log to a file, it's not actually a `NullLogger` anymore. Wouldn't you want to add a `FileLogger` instead?

Then, depending on `should_log_activity` (or potentially `utils.is_heroku()`) you can decide to construct one of `NullLogger`, `FileLogger` or `S3Logger` at startup. 

Here is how this was intended to work:

image

The idea is that we have an interface (which of course doesn't existing in Python because it's all duck-typed), which is just an object with a `log()` method, and then we'd construct one of 3 objects that satisfy that interface: a `NullLogger` which doesn't do anything, a `FileLogger` which writes to a file, or an `S3Logger` which writes to S3 by means of a log queue that sits in the background and periodically writes to S3 (because again we want to batch up little writes when we do network calls).

rix0rrr: Of course, we could also have done the following:

image

There is a `Logger` object that is concrete and that every caller interacts with, but the transmission strategy varies (either do nothing, write to a file, or send to S3).

The designs don't differ that much, the only difference is that in the 2nd we have a place to put some common logic between the 3 loggers, if we need to. I don't know that we need this, so that's why so far in my head we would just have different Loggers themselves.

hasan-sh: I made some changes @rix0rrr! 
- created a FileLogger on top the NullLogger.
- moved code logic appropriately.

website/user_activity.py
```diff
@@ -0,0 +1,63 @@
...
+from website import s3_logger
+
+cookie_name = config["session"]["cookie_name"]
+invite_length = config["session"]["invite_length"] * 60
+
+
+# Add LOG_USER_ACTIVITY=True to your env. to log the data for testing
+write_to_file = os.getenv("LOG_USER_ACTIVITY")
+logger = s3_logger.S3Logger(name="activity", config_key="s3-activity-logs",
+                            write_to_file=write_to_file,)
```
rix0rrr, line 21: How about the following:

```
if utils.is_herokuo(): 
    logger = s3_logger.S3Logger(name="activity", config_key="s3-activity-logs")
elif write_to_file:
  logger = s3_logger.FileLogger(filename="user_activity.json")
else:
  logger = s3_logger.NullLogger()
```

?

hasan-sh: Don't mind, it's just more code everywhere we use it

rix0rrr (approved):
👍 thanks!

PR #5154, opened 2024-02-15
💻 Create adventure within class

### LINKED ISSUE #4633: [UI idea]  Create adventure from class page
**Idea incl level**
Allow adding an adventure from within `/for-teachers/class/[id]` page. 

**Additional context**
Perhaps with a `create adventure` button.

### LINKED ISSUE #5182: 💻 Add adventure to class
It is now only possible to add an adventure to a class through the class but not through the adventure itself, so add a box underneath 'levels':

Screenshot 2024-02-27 at 15 06 40

Annelein:
Fixes #4633 and fixes #5182
Create an adventure inside a class, which is automatically added to the class(es) and level(s) selected.

https://github.com/hedyorg/hedy/assets/48122190/c8c2f819-4a4f-400e-9d00-a698a9d3128c


hasan-sh (commented):
This is great, but i think the idea is that when you click on this button, the teacher would expect that adventure to be added to the current class. So, if this adventure is for level 1 and 4, it'd be added to the customization class for these levels, otherwise the teacher would still  have to go to their class, customization page, and add the adventure for the required levels!

Annelein:
> This is great, but i think the idea is that when you click on this button, the teacher would expect that adventure to be added to the current class. So, if this adventure is for level 1 and 4, it'd be added to the customization class for these levels, otherwise the teacher would still have to go to their class, customization page, and add the adventure for the required levels!

I think I missed that in the meeting but I will add it!

Annelein:
> I think I missed that in the meeting but I will add it!

This needs to wait for #5026 to be merged

Felienne:
@hasan-sh #5026 is in, so this can be merged, I think? Can you approve?

Annelein:
> @hasan-sh #5026 is in, so this can be merged, I think? Can you approve?

I need to add something to it now that #5026 is merged! Will do it now

Annelein:
@hasan-sh I think it's ready!

hasan-sh (changes requested):
There's three things we need to adjust:
- when i create an adventure from within a class, i do see the class selected in the classes list, however, i don't see the class card at the bottom: 
image
- if an adventure is available in 1, 3, I remove an adventure from level 1, then i decide to update the adventure (e.g., make it public), the adventure is re-added to all levels. This shouldn't be the case. Let me know if you need help on how to design this.
- whenever you add an adventure to class' levels, use the function i created that makes sure the order of puzzles and quizes are always at last; https://github.com/hedyorg/hedy/blob/main/website/for_teachers.py#L398


I'll check the code in the meanwhile!

Annelein:
> * when i create an adventure from within a class, i do see the class selected in the classes list, however, i don't see the class card at the bottom:
>   image

It is available after reload... Can you help me out on how to show it straight away?

Annelein:
> * if an adventure is available in 1, 3, I remove an adventure from level 1, then i decide to update the adventure (e.g., make it public), the adventure is re-added to all levels. This shouldn't be the case. Let me know if you need help on how to design this.

could you maybe add a screenshot video, it's not really clear what goes wrong

hasan-sh:
> > * if an adventure is available in 1, 3, I remove an adventure from level 1, then i decide to update the adventure (e.g., make it public), the adventure is re-added to all levels. This shouldn't be the case. Let me know if you need help on how to design this.
> 
> could you maybe add a screenshot video, it's not really clear what goes wrong

[Screencast from 14-03-24 10:52:55.webm](https://github.com/hedyorg/hedy/assets/20051470/a144d301-036a-42cc-9948-d284f003e9e6)

website/for_teachers.py
```diff
@@ -985,9 +985,11 @@ def view_adventure(self, user, adventure_id):
             current_page="for-teachers",
         )
 
-    @route("/customize-adventure/<adventure_id>", methods=["GET"])
+    @route("/customize-adventure/<adventure_id>/<current_page>", methods=["GET"])
```
hasan-sh, line 988: i don't understand why you need current_page?

website/for_teachers.py
```diff
@@ -1140,9 +1199,13 @@ def create_adventure(self, user):
             "date": utils.timems(),
             "creator": user["username"],
             "name": name,
+            "classes": class_id,
```
hasan-sh, line 1202: the classes field you use should be a list of IDs, you change it later if i wanna add the current adventure to other classes, but since it indicates a list, make sure it is initially.

website/for_teachers.py
```diff
@@ -1140,9 +1199,13 @@ def create_adventure(self, user):
...
             "level": 1,
             "content": "",
             "language": g.lang,
         }
         self.db.store_adventure(adventure)
+        if class_id != '0':
```
hasan-sh, line 1208: instead of this, either
- pass the `class_id` as a parameter and then simply check if it's passed to this endpoint
- or make it optional and then simply check `if class_id: ..`

this way you don't have to pass do:
`hx-post="/for-teachers/create-adventure/0"`

hasan-sh: but rather `hx-post="/for-teachers/create-adventure"` if no class id needed, or`hx-post="/for-teachers/create-adventure?class_id={{class.id}}"` or `hx-post="/for-teachers/create-adventure/{{class_id}}"` if needed!

website/for_teachers.py
hasan-sh: it doesn't make sense to have a url like the following:
`"/for-teachers/customize-adventure/view/{{ adventure.id }}/for-teachers"` (for-teachers is repeated twice!!)

Annelein: It's for the go back button, wether to go back to the for-teacher page or to the class page

hasan-sh: If you're in the customize-adventure page, you can simply go back to the for-teafchers page. If it's necessary, we should design this functionality differently; e.g., save it somewhere and go back accrodingly (in session preferably!)

Annelein: How else should I solve it? I don't understand how it would be possible to use session for it?

Annelein: solved!

website/for_teachers.py
```diff
@@ -1113,7 +1112,10 @@ def update_adventure(self, user):
         for class_id in body["classes"]:
             if class_id != '0':
                 for level in body["levels"]:
-                    self.add_adventure_to_class_level(user, class_id, body["id"], level, False)
+                    if level not in current_levels and class_id in current_classes:
+                        self.add_adventure_to_class_level(user, class_id, body["id"], level, False)
+                    elif class_id not in current_classes:
+                        self.add_adventure_to_class_level(user, class_id, body["id"], level, False)
```
hasan-sh, line 1118: why are they in if-elif statement? The body is the same!
```suggestion
                    if class_id not in current_classes or (level not in current_levels and class_id in current_classes):
                        self.add_adventure_to_class_level(user, class_id, body["id"], level, False)
```

website/for_teachers.py
```diff
@@ -1201,7 +1203,9 @@ def create_adventure(self, user, class_id=None):
             "name": name,
             "classes": [class_id],
             "level": 1,
+            "levels": "1",
```
hasan-sh, line 1206: levels is a list of level strings!
```suggestion
            "levels": ["1"],
```

hasan-sh:
I encountered the following error:

image

for-teachers.py (L: 1168)


I created an adventure. It's available in 1,2. I remove it from level 1 from customize-class. I update the levels, i get this error! Check if the adv does exist before removing it!

website/for_teachers.py
```diff
@@ -1123,26 +1145,71 @@ def parse_preview_adventure(self):
...
+        if not Class or (not utils.can_edit_class(user, Class) and not is_admin(user)):
+            return utils.error_page(error=404, ui_message=gettext("no_such_class"))
+
+        customizations, adventures, adventure_names, _, _ = self.get_class_info(
+            user, class_id)
+
+        teacher_adventures = list(self.db.get_teacher_adventures(user["username"]))
+        second_teacher_adventures = self.db.get_second_teacher_adventures([Class], user["username"])
+        teacher_adventures += second_teacher_adventures
+        is_teacher_adventure = self.is_adventure_from_teacher(adventure_id, teacher_adventures)
```
hasan-sh, line 1159: you don't need this code since we know that this is an actual teacher's adventure!

website/for_teachers.py
```diff
@@ -1123,26 +1145,71 @@ def parse_preview_adventure(self):
...
+        sorted_adventure = SortedAdventure(short_name=adventure_id,
+                                           long_name=adventure_names[adventure_id],
+                                           is_teacher_adventure=is_teacher_adventure,
+                                           is_command_adventure=adventure_id in hedy_content.KEYWORDS_ADVENTURES)
+
+        if not remove:
+            adventures[int(level)].append(sorted_adventure)
+        else:
+            adventures[int(level)].remove(sorted_adventure)
+        self.reorder_adventures(adventures[int(level)], from_sorted_adv_class=True)
```
hasan-sh, line 1179: why are you doing this step actually? Any specific reason? Because i don't think you need it!

website/for_teachers.py
```diff
@@ -1123,26 +1145,71 @@ def parse_preview_adventure(self):
             return gettext("something_went_wrong_keyword_parsing"), 400
         return {"code": code}, 200
 
-    @route("/create-adventure", methods=["POST"])
+    def add_adventure_to_class_level(self, user, class_id, adventure_id, level, remove):
```
hasan-sh, line 1148: just to make it clear what you're removing.

```suggestion
    def add_adventure_to_class_level(self, user, class_id, adventure_id, level, should_remove_adv):
```

hasan-sh (commented):
just a couple of things regarding the code!

website/for_teachers.py
Annelein: would `remove_adv` also be fine? to make it shorter

website/for_teachers.py
Annelein: it also happens in `for-teachers/add-adventure`, or what is the reason why it's needed there?

website/for_teachers.py
hasan-sh: Sure!

website/for_teachers.py
Annelein: You told me to use `self.reorder_adventures(adventures[int(level)], from_sorted_adv_class=True)`:

> * whenever you add an adventure to class' levels, use the function i created that makes sure the order of puzzles and quizes are always at last; https://github.com/hedyorg/hedy/blob/main/website/for_teachers.py#L398

Annelein: Or do you mean `adventures[int(level)].remove(sorted_adventure)`?

hasan-sh: i mean the adventures variable doesn't need to be updated it. This is because it's updated automatically from the customizations["sorted_adventures"] in the get_class_info function. Thus, you don't need to update it here!

website/for_teachers.py
hasan-sh: Yes, but that's a different thing. Your `add_adventure_to_class_level` function always gets teachers' adventures!

Annelein: Solved it!

Annelein:
@hasan-sh thanks for the thorough testing! helps a lot, solved all of your comments now hopefully

website/for_teachers.py
```diff
@@ -1123,26 +1145,66 @@ def parse_preview_adventure(self):
...
+            customizations['sorted_adventures'][level].remove(
+                {'name': adventure_id, 'from_teacher': is_teacher_adventure})
+        sorted_adventure = SortedAdventure(short_name=adventure_id,
+                                           long_name=adventure_names[adventure_id],
+                                           is_teacher_adventure=is_teacher_adventure,
+                                           is_command_adventure=adventure_id in hedy_content.KEYWORDS_ADVENTURES)
+
+        if not remove_adv:
+            adventures[int(level)].append(sorted_adventure)
+            self.reorder_adventures(adventures[int(level)], from_sorted_adv_class=True)
```
hasan-sh, line 1174:
```suggestion
        if not remove_adv:
```

website/for_teachers.py
```diff
@@ -1095,6 +1103,20 @@ def update_adventure(self, user):
...
             self.db.update_tag(tag["id"], {"tagged_in": tag["tagged_in"]})
 
+        if current_levels and current_classes:
+            for old_class in current_classes:
+                for level in current_levels:
+                    if old_class not in body["classes"] or level not in body["levels"]:
+                        self.add_adventure_to_class_level(user, old_class, body["id"], level, True)
+
+        for class_id in body["classes"]:
+            if class_id != '0':
```
hasan-sh, line 1113: i think this is from the previous version, right?

hasan-sh (changes requested):
With these, i think we should be good to goooo:)

Annelein:
@hasan-sh check!

hasan-sh (approved):
Let's gooo!

Issue #4888, opened 2023-12-12
[UX] Use autosave everywhere 


Felienne:
As we have discussed in #4823, I think we want to go to auto-save.

This will both make front-end development easier and also will feel more modern for users.

A few places where this can be considered:

**1. The lower half of the customization page**

<img width="1401" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/324b3a49-7b2f-4f6b-a475-a6e99f688860">

**2. Create an adventure**

<img width="1394" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/91f62e73-abcf-4968-b277-ee1d9445efa6">

**3. Update profile**

<img width="1229" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/f958b848-fd28-4fcf-a77f-56721da033ac">

**4. Update public profile**

<img width="1131" alt="image" src="https://github.com/hedyorg/hedy/assets/1003685/3d7ba7ea-f062-4e3f-9d84-2334b3e04882">

We can potentially split this into separate issues of course!

### Relevant (shortened) text from the discussion

**(Auto) Saving** 
I think I agree with @rix0rrr that an explicit save is not desired. I don't know why we went that way initially (because the sate of the customization page was a lot smaller then, so it probably made more sense) but people nowadays expect that things they do on a website are saved immediately (like Google Docs) and this is line we have been following recently. We never made this explicit but new features (like the reordering of tabs) show this **"auto save"** behaviour, while old parts use the button.

This is due to a lack of project management. When we introduced the "auto-save" with the reordering, we should probably have followed through for the rest of the page. And we never had the clear user stories as described above where "user clicks save" is an explicit point.

I propose opening an issue to use autosave in most places (maybe even for a new adventure, like a new file in Docs is also saved immediately) so that the front-end behaviour for users is more systematic.

_Originally posted by @Felienne in https://github.com/hedyorg/hedy/discussions/4823#discussioncomment-7707191_

PR #5156, opened 2024-02-15
💻 hide all quizes/parsons and show tabs in levels

### LINKED ISSUE #5141: 💻 level quiz or all quizes
**Idea incl level**
In the class customization page, we allow teachers to hide the quiz. It's not clear whether this means hide all quizes or the quiz for the selected level, teachers think it's the latter. Therefore, we need to tackle this issue.

@Felienne and I discussed this in the hackathon and an idea has popped up!
- add the quiz to the adventures tab for each level. This way if they remove it from level 2 for instance, other levels' quizes are still visible.
- the existing setting about hiding a quiz should be converted to be "hide all quizes"; label and functionality should be changed.

hasan-sh:
Fixes #5141 

**Key points**
- a quize/parson is added as a tab in the customize-class page. This makes it closer to how it'd look like for students.
- the current "hide quize/puzzle" switches will behave as "toggle all quizes/puzzles". Why? Consider the following:
  - the teacher decides to hide all quizes. 
  - they are removed from all levels,
  - they can of course still choose to show a quize in certain levels, while keeping the rest hidden.
  - however, what if they (for whatever reason) want to show all quizes but hide one or two?
  - this scenario is likely because until now the "hide quize" behaves as toggling all quizes, so it should still behave similarly.
  - thus it could be considered as resetting all quizes/puzzles.
- the order of puzzle and quiz tabs are always fixed to be always the last two tabs, respectively.

Something went missing when i added the key translations for https://github.com/hedyorg/hedy/pull/5026 (check https://github.com/hedyorg/hedy/pull/5026/commits/6ee2c4f3161c94a8e5c613b595af840569f83a56#diff-852046eb47f474139974edf5e69ed9758f9fde82e1432c49d772b89aef354b37), so I'm adding them here.

**How to test?**

- go the customize-class page
- add/remove a quiz/puzzle
- reset all puzzles/quizes to test that functionality.
   - _note_ that if you disable all quizes and you add one quiz in level 2, for instance, the disable checkbox will be updated since not all quizes are disabled now. (it needs a refresh for now for that to be reflected in the DOM, but this should be dealt with in #4888)
- log in as a student or preview as teacher to see  your changes from the perspective of students.
- verify that everything works as expected.


jpelay (changes requested):
Hi @hasan-sh! Sorry for not reviewing it earlier! Everything works great, and I think this is a great idea, however there's something we must change: if a class already has a customization object and I enter the customization page again, then my class won't have any quizes. Can we find a way to automatically add parsons and quizzes to the customization objects that hasn't been touched previous to this PR? Like a migration.

hasan-sh:
> Hi @hasan-sh! Sorry for not reviewing it earlier! Everything works great, and I think this is a great idea, however there's something we must change: if a class already has a customization object and I enter the customization page again, then my class won't have any quizes. Can we find a way to automatically add parsons and quizzes to the customization objects that hasn't been touched previous to this PR? Like a migration.

@jpelay thanks for the reminder, i had a TODO in the code, but apparently i missed it! Does your IDE notify you about todos!?

Now it should be ready to roll!

jpelay (approved):

jpelay:
> @jpelay thanks for the reminder, i had a TODO in the code, but apparently i missed it! Does your IDE notify you about todos!?
> 
> Now it should be ready to roll!

Actually no haha! I often write those down in my notebook or in the calender. If I do it in the code I tend to forget.

PR #5159, opened 2024-02-17
🚚 advertise Offline Hedy in frontpage README

rix0rrr:


jpelay (approved):
Awesome! Btw, I tried the release again in my desktop computer, and the local network is running everything fine.

PR #5160, opened 2024-02-19
💻 My programs redesign

### LINKED ISSUE #5112: 💻 Redesign program page
I worked on this, like hiding the second actions and just showing the main ones like submit and open. Deleting, copying and share would be hidden in that menu. Also, I thought about changing the share button for a drop down that says private or public. Just like i'm showing on the images. 

Frame 7
Frame 6

_Originally posted by @gisellandrade in https://github.com/hedyorg/hedy/issues/5045#issuecomment-1934625987_

Annelein:
Fixes #5112 
Screenshot 2024-02-19 at 15 08 11
Screenshot 2024-02-19 at 15 09 10


Annelein:
@hasan-sh if making a program private or public, the page doesn't get updated automatically, how do I do this?

hasan-sh:
> @hasan-sh if making a program private or public, the page doesn't get updated automatically, how do I do this?

The page doesn't have to be updated nor should it reload. The best solution (although hard) is to return the program from the backend and then update it via htmx in the frontend. This means you'd have to create a seperate component for the program box; which is good as well. This is very similar to the approach that I've been following with htmx. For example, take a look at `htmx-tags-dropdown-item.html` and how it updates the list of tags if it's clicked; via `hx-target="#tags-list"`

hasan-sh (changes requested):
Glad you fixed the unittests;))

I left some suggestions and one comment. So, since we now converted the whole program into its own component, we should name the file accordingly `htmx-program.html` ;) 

Also, you still need to adapt the usage in https://github.com/hedyorg/hedy/blob/main/static/js/app.ts#L1953-L1956 and https://github.com/hedyorg/hedy/blob/main/static/js/app.ts#L2109. Let me know if you need any help or clarification!

Annelein:
> Also, you still need to adapt the usage in https://github.com/hedyorg/hedy/blob/main/static/js/app.ts#L1953-L1956
Let me know if this is the correct way?

website/programs.py
```diff
@@ -223,25 +223,30 @@ def share_unshare_program(self, user, program_id, loop_index):
...
+            result = True
+        else:
+            result = False
+        return jinja_partials.render_partial('htmx-program.html',
                                              program=program,
                                              adventure_names=adventure_names,
                                              public_profile=public_profile,
-                                             loop_index=loop_index)
+                                             loop_index=loop_index,
+                                             second_teachers_programs=result)
```
hasan-sh, line 249:
```suggestion
                                             loop_index=loop_index,
                                             second_teachers_programs=second_teachers_programs)
```

website/programs.py
```diff
@@ -223,25 +223,30 @@ def share_unshare_program(self, user, program_id, loop_index):
...
         keyword_lang = g.keyword_lang
         adventure_names = hedy_content.Adventures(g.lang).get_adventure_names(keyword_lang)
         program["date"] = utils.delta_timestamp(program["date"])
         program["preview_code"] = "\n".join(program["code"].split("\n")[:4])
         program["number_lines"] = program["code"].count('\n') + 1
-        return jinja_partials.render_partial('htmx-program-buttons.html',
+        if second_teachers_programs is True:
+            result = True
+        else:
+            result = False
```
hasan-sh, line 243:
```suggestion
```

static/js/app.ts
```diff
@@ -1951,8 +1949,7 @@ function initializeShareProgramButtons() {
         }
 
         const response = await postJsonWithAchievements('/programs/share', {
```
hasan-sh, line 1951: I don't see an endpoint for this request! I think the simplest is to make the loop_index and second_teachers_programs optional; by adding another route to the same function
```@app.route('/share', methods=['POST'], defaults={'loop_index': 0, 'second_teachers_programs': False })```

**Review by hasan-sh — CHANGES_REQUESTED** [review]

File: website/programs.py (hunk 2 above)
Comment (Annelein) [reply website/programs.py:243]: @hasan-sh if I remove this, the dropdown dissapears after toggeling private/public

Comment (hasan-sh) [reply website/programs.py]: yeah i see the issue.. cuz the second_teachers_programs is given as a string, and need to be checked for indeed. However, your if statement won't solve it since second_teachers_programs is never True. I'd do the following for now:

```second_teachers_programs = second_teachers_programs == 'True'```

File: static/js/app.ts (hunk 3 above)
Comment (hasan-sh) [reply static/js/app.ts:1951]: What i meant is:
```suggestion
        const response = await postJsonWithAchievements(`/programs/share/${saveInfo.id}`);
```

This way you still use your existing api endpoint!

website/programs.py
```diff
@@ -205,46 +207,43 @@ def save_program(self, user):
             "achievements": self.achievements.get_earned_achievements(),
         })
 
-    @route("/share", methods=["POST"])
+    @route('/share', methods=['POST'], defaults={'loop_index': 0, 'second_teachers_programs': False})
```
hasan-sh, line 210: Even if this endpoint is reached, it'll complain about several thinsg; missing arguement and then the fact that you're retrieving an id of none. To mitigate that,  see my last comment in app.ts!

hasan-sh (changes requested):

hasan-sh (approved):

PR #5162, opened 2024-02-19
 🪲Only show programs that has been modified in Overview of programs per adventure

### LINKED ISSUE #5161: 🪲 Only show saved programs in "Overview of programs per adventure" if the program has been modified
One of the complaints we got from teachers during the hackathon was that students would just copy and paste the code from the examples, and take that as a submission. Ideally, we would only show programs that has been modified to a certain point.

jpelay:
Modifies the way we choose if a program is suitable for having a tick mark in the Overview of programs per adventure table.

The process is basically this:

1. Get all of the code snippets from the yamls. For this I made a simple loop that kind of parses the markdown code, if it finds 3 consecutive backticks and then 3 more, the text found in-between is the code for that snippet.
2. For the teacher adventures, I use BeautifulSoup to parse the HTML code used to format the text in these adventures.
3. After getting all of the relevant code, and removing newlines, I compare the code from the student program with each one of the snippets. If the difference to any of the snippets is less than 10 characters, I consider the snippet not to be worthy of reviewing.

**Fixes #5161**

**How to test**

* Login as a teacher and go to Overview of program per adventure
* Login as a student in other window and copy-paste one of the example's code into the editor, and run the program.
* Check that the program isn't being shown for reviewing.
* Now in the student's window, modify the program before running it.
* Now the program should be available for reviewing.


hasan-sh (commented):
Hi @jpelay , great work!! I had the following error with a teacher adventure, but after resetting local db, it worked again. 

I just wanna share my thoughts on the code. All works fine and all for loops are plausible. Just the one that you go through each charecter within the code is a bit of a hassle! And if the code is long, you can imagine the time complexity of this nested for loop would grow exponentially! So, why not doing it with reg? 

Another suggestion is about how you decide whether there's a difference! Your decision is based on the charecter level and if there's 10 chars different, then you say they can modify. What if they just changed the line orders without adding anything new for instance? Do you know https://docs.python.org/3/library/difflib.html? Especially the sequenceMatcher! This way instead of relying on chars being changed from position, we rely on similarity score!

jpelay:
> I just wanna share my thoughts on the code. All works fine and all for loops are plausible. Just the one that you go through each charecter within the code is a bit of a hassle! And if the code is long, you can imagine the time complexity of this nested for loop would grow exponentially! So, why not doing it with reg?

Hi! I did it this way in case there was weird nesting inside the adventures, I did it this way to make sure we parse it properly.

> Another suggestion is about how you decide whether there's a difference! Your decision is based on the charecter level and if there's 10 chars different, then you say they can modify. What if they just changed the line orders without adding anything new for instance? Do you know https://docs.python.org/3/library/difflib.html? Especially the sequenceMatcher! This way instead of relying on chars being changed from position, we rely on similarity score!

I think this a worth suggestion!!! Although I think that changing the line orders might be a worth addition in the perspective of teachers, since I don't think the kids are going to press Alt + up arrow to change the order easily. @Felienne what do you think?

Felienne:
> I put an arbitrary value of 0.95 that I don't know if is too strict or not. Perhaps should we include a clause that says that if the original snippet had a underscore and the student filled them, it should be saved?

I like that idea, but I am not sure it is worth the hassle of implementing it. Shall we open an issue and discuss it separately? (we can ask a teacher too).

Felienne:
> > I put an arbitrary value of 0.95 that I don't know if is too strict or not. Perhaps should we include a clause that says that if the original snippet had a underscore and the student filled them, it should be saved?
> 
> I like that idea, but I am not sure it is worth the hassle of implementing it. Shall we open an issue and discuss it separately? (we can ask a teacher too).

We agreed in the meeting to try to implement the underscore check into this PR after all!

Felienne (changes requested):
Can be merged if we add the underscore check

website/statistics.py
```diff
@@ -267,10 +268,14 @@ def is_program_modified(self, program, full_adventures, teacher_adventures):
         can_save = True
         for snippet in adventure_snippets:
             seq_match = SequenceMatcher(None, snippet, student_code)
-            if seq_match.ratio() > 0.95:
+            # Allowing a difference of more than 10% or the student filled the placeholders
+            if seq_match.ratio() > 0.95 or (not self.has_placeholder(student_code) and self.has_placeholder(snippet)):
                 can_save = False
         return can_save
 
+    def has_placeholder(code):
```
hasan-sh, line 276: your forgot self!

website/statistics.py
```diff
@@ -267,10 +268,14 @@ def is_program_modified(self, program, full_adventures, teacher_adventures):
         can_save = True
         for snippet in adventure_snippets:
             seq_match = SequenceMatcher(None, snippet, student_code)
-            if seq_match.ratio() > 0.95:
+            # Allowing a difference of more than 10% or the student filled the placeholders
+            if seq_match.ratio() > 0.95 or (not self.has_placeholder(student_code) and self.has_placeholder(snippet)):
```
hasan-sh, line 272: i tested it with the restaurant adventure in level 1, and i can't seem the sutdent's progress still! The function is returning False for both snippets!

hasan-sh (changes requested):

website/statistics.py
jpelay: You're right!! It should work now!

hasan-sh (approved):

website/statistics.py
hasan-sh: Cool and it does:)

PR #5166, opened 2024-02-20
🪲 Insert cheatsheet code at the end

### LINKED ISSUE #4099: [UI idea] Cheat sheet should paste the code under your own work, not over it.
**Idea incl level**
If you use a code from the cheat sheet, your own code is removed. It would be better if the lines of code are pasted underneath your own code.

jpelay:
When selecting code from the cheat sheet, instead of simply deleting the user's existing code, now it is appended to the end of the user's code.

Fixes #4099

**How to test**

 * Go to Hedy's page and click the yellow button in one of the sections, and check that the code is appended to the end.


Felienne:
I think we might need to update the try button Cypress test for this to work?

image

(and there is another test failing but I don't know why that is!)

Felienne:
Hi @jpelay!

This works well! However, if you use it in a fresh tab, there is an empty line:

image

I think it would be a bit prettier to remove the one line?

Felienne (changes requested):
Let's remove the first line?

jpelay:
> Let's remove the first line?

Done!

Felienne (approved):

PR #5167, opened 2024-02-20
🪲 Translate keywords in the teacher manual

### LINKED ISSUE #4134: [UI idea] Localized keywords in the teacher manual
The teacher manual doesn't use localized keywords, only the English ones. It would be nice if the keywords could change to the preferred language setting of the teacher. I mainly mean in the common mistakes section. 

I've tried to just use the { } around the keywords, but that didn't do the trick.

jpelay:
In this PR I want to tackle the translation of keywords in the teacher manual. First, we need to allow the keywords to be translated using the curly brace system we use already in the adventures, that is simple enough to do and requires only the call to a function; after that we need to test the programs inside the yamls in the teacher manual, because when they do have curly braces, sometimes they're not correctly. And third, what do we do with the mess we already have?

For example in Spanish there's a mixture of several methods:

* Sometimes the keywords are directly translated
```
bebida = preguntar '¿Qué te gustaría beber?'
imprimir bebida
bebida = preguntar '¿Qué te gustaría beber?'
imprimir bebida
bebida = preguntar '¿Qué te gustaría beber?'
imprimir bebida
```
* Sometimes they have curly braces:
```
superhéroes = Spiderman, Batman, Iron Man
superhéroe = superhéroes {at} {random}
{if} superhéroe = Batman
    {print} '¡SOY BATMAN!'
```
* Sometimes they have intentional errors.
* Sometimes they have unintentional errors

Fixes #4134

**How to test**


jpelay:
Ok, so here's the sketch of a plan: maybe I can use a script to translate parse the correct programs, translate them to english and add the curly braces, but that leaves the question: what to do with those that have **intentional** errors? Perhaps first we need to add a flag to the yamls?

jpelay:
Depends on the fix made in #5175

jpelay:
Done! All languages done with the exception of zh_Hans. I'm not sure what to do with that one since Chinese speakers don't often use spaces, which makes the parser to give up. Not sure about an easy solution here.

boryanagoncharenko (commented):
Apologies for the super delayed review!
Great job! Really! There only a couple of things I noticed that we could improve:
- The keywords are not "highlighted" if they appear in headings. So, the command `echo` would appear correctly in a <p> but not if it is a <h4>: 
Screenshot 2024-07-22 at 11 16 56
- I noticed that the word `doens't` appears in many places. Perhaps we could change it with this PR?
- There are 2 words which I am sure are not used correctly: 'commando' and 'codes'. They change the meaning of the text. Commando is a military unit and code (programming code) can never be plural by definition. This `Codes using the if commando can get very long` is a hilarious example, which I think should be changed to something like: `Using the if command can make the code lines very long`.

jpelay:
> Apologies for the super delayed review! Great job! Really! There only a couple of things I noticed that we could improve:

Don't worry!

> * The keywords are not "highlighted" if they appear in headings. So, the command `echo` would appear correctly in a `<p>` but not if it is a `<h4>`:
> 
> Screenshot 2024-07-22 at 11 16 56

Thanks for noticing that! I fixed it by using a Jinja2 filter (`commonmark`)

> * I noticed that the word `doens't` appears in many places. Perhaps we could change it with this PR?

I tried replacing, let me know if that's ok.

> * There are 2 words which I am sure are not used correctly: 'commando' and 'codes'. They change the meaning of the text. Commando is a military unit and code (programming code) can never be plural by definition. This `Codes using the if commando can get very long` is a hilarious example, which I think should be changed to something like: `Using the if command can make the code lines very long`.

Omg! For sure, I replaced every mention of codes, and commando.

boryanagoncharenko (approved):

PR #5169, opened 2024-02-21
💻 Redesign go back button

### LINKED ISSUE #5155: 💻 Go back buttons
The "Go back" buttons are positioned all over the place and I feel like this button could look way more professional and consistent.
@gisellandrade could you look into this?
 
Naturally I would think of an arrow top left, because I think most "Go back" buttons are designed like that on other websites.
Or do we prefer one with text? @Felienne

'Go back to class' bottom right:
Screenshot 2024-02-15 at 13 44 06
'Go back to teacher page' bottom left:
Screenshot 2024-02-15 at 13 44 26
'Go back to class' top right:
Screenshot 2024-02-15 at 13 44 41

Annelein:
Fixes #5155 
Depends-On: #5156

Added a general "Go back" button to: `/customize-class` `/live-stats` `/customize-adventure` `/grid-overview` and `/for-teachers/class`

Screenshot 2024-02-29 at 14 34 15


hasan-sh (changes requested):
Hopefull it'll be used, but i want to validate that by giving the btn a generic class (since you give different IDs in different pages!) and add it to the https://github.com/hedyorg/hedy/blob/main/static/js/user-activity.ts#L6. 

Further, I have just two remarks:
- There's one page that still needs to be updated: `/for-teachers/create-accounts/{class_id}`
- for Arabic (or RTL languages), we need to horizontally flip the arrow!
image

hasan-sh (commented):
All good @Annelein, just need to fix the tests!

Annelein:
@hasan-sh the tests are depending on that pull request, or is there a way around?

Annelein:
@hasan-sh if you approve it can be merged :)

hasan-sh (approved):
All good!

Issue #5013, opened 2024-01-18
💻Make a error messages for when [ ] are forgotten around lists


MarleenGilsing:
In level 16 (and up) the square brackets are introduced to use around lists. If you don't use them you'll get this error message:
![image](https://github.com/hedyorg/hedy/assets/80678586/220b1bf8-1d5c-415a-8bc1-68c29e432a92)

Can we make an error message that tells you that you forgot the brackets instead of telling you not to use a comma?

brielle5810:
Hi! Could I give this a try?

MarleenGilsing:
@brielle5810 Sure!

akseron:
> @akseron will probably not pick this up, maybe one for @boryanagoncharenko?

Oh pardon, I thought the other commenter was already trying out the issue 😅 I will take a look at it in the weekend if @boryanagoncharenko does not mind.

boryanagoncharenko:
@akseron of course! I am assigning the issue to you. Feel free to assign it back to me in case you want to get rid of it in the future :)

PR #5177, opened 2024-02-26
🖊️ 💻 Display error message when brackets are forgotten around lists

akseron:
Fixes https://github.com/hedyorg/hedy/issues/5013

How to test
Attempt to use the following code in level 16/17:
```
animals = 'chicken', 'horse', 'cow'
print animals[random]
```


Felienne (approved):
Nice work @akseron. I just changed the error message a bit because there are other brackets too.

Felienne (approved):

PR #5178, opened 2024-02-26
🖊️ Add support for calculations with play command from level 6 #5044

### LINKED ISSUE #5044: 🖊️ allow for calculations in play
**From level 6 on**
It would be cool to be able to do:
```
play_number = ask 'Say a random number'
play play_number
play play_number + 1
play play_number + 2
```

boryanagoncharenko:
Fixes #5044

**How to test**
- Check if the added unit tests cover correctly the added logic
- Start hedy locally and navigate to level 6. Execute the following command and ensure the three tones are played:
```
play_number = 34
play play_number
play play_number + 1
play play_number + 2
```


jpelay (changes requested):
Everything works great in the transpiler and I found no issues! However, the highlighter doesn't yet highlight the expressions correctly:

image

I think we just need to add Expression, and list access to the grammars and it will work.

boryanagoncharenko:
> I think we just need to add Expression, and list access to the grammars and it will work.

Thanks! I completely missed highlighting. It is now added with tests.

jpelay (approved):
Awesome!!! This is ready to ship 🚀

PR #5179, opened 2024-02-26
💻 Allow teacher adventures to use 2 columns

### LINKED ISSUE #5021: 💻 Allow teacher adventures to use 2 columns
**Idea incl level**
The can now only use one:

image

I am not sure how we should allow teachers to do that? Give them two columns to edit separately? Maybe we can brainstorm in the hackathon?

Annelein:
Fixes #5021 

How to test:
- Create an adventure as a teacher with more than 750 characters of explanation (I used 750 because that is about the amount that the editor is not visible on a laptop because there is so much text)
- Go to customize class 
- Add the adventure
- Click 'Preview'
- Go to the added aventure


Felienne (changes requested):
Nice work @Annelein!

We do break in a random place now in the middle of a word:

image

Could we break at the end of a line (maybe not so easy since there can be a lot of characters ending a line?) or at least at the first space after the 750 maximum?

Annelein:
@Felienne good one, done!

Felienne (approved):

PR #5185, opened 2024-02-27
🪲 Allow for variables to start and finish with underscore

### LINKED ISSUE #5184: 🪲 Variables in Hedy can't start or end in underscore
If a variable starts with an underscore or ends in an underscore, then we will pick it as being an incomplete program. For example, this is level 2:

image

image

I think perhaps this should be allowed?

jpelay:
Changes the method of checking for underscores in the transpiler to use a regex, which allows us to exactly match underscores that are not followed or preceded by anything other than a space.

Fixes #5184 
**How to test**
I added a test for this behavior, you can also copy paste this code on level 4:
```
var_ is Hedy
 _var is Hedy
print var_ ' ' _var
```


jpelay:
The tests that are failing, do it because they add underscores without spaces around  them, like in this case:  

```
{print} _Escape from the haunted house!_
{print} _There are 3 doors in front of you..._
choice {is} {ask} _Which door do you choose?_
{print} _You picked door ..._ choice
monsters {is} a zombie, a vampire, NOTHING YOUVE ESCAPED
{print} _You see..._
{sleep}
{print} monsters {at} {random}
```

I personally think that it'd be better if underscores are not next to actual words, and I'm willing to change the examples in the pages, but maybe you guys don't feel that way? What do you think?

Felienne:
> The tests that are failing, do it because they add underscores without spaces around them, like in this case:
> 
> ```
> {print} _Escape from the haunted house!_
> {print} _There are 3 doors in front of you..._
> choice {is} {ask} _Which door do you choose?_
> {print} _You picked door ..._ choice
> monsters {is} a zombie, a vampire, NOTHING YOUVE ESCAPED
> {print} _You see..._
> {sleep}
> {print} monsters {at} {random}
> ```
> 
> I personally think that it'd be better if underscores are not next to actual words, and I'm willing to change the examples in the pages, but maybe you guys don't feel that way? What do you think?

Hi @jpelay! I think the reason we do this is that it is more clear that way that the quotes should go against the text.

If we do:

> ```
> {print} _ Escape from the haunted house! _
> etc
> ```

Kids will enter


> ```
> {print} ' Escape from the haunted house! '
> etc
> ```

And that looks weird, making them think they did it wrongly (I can be convinced this is not a real big problem, but that is the reason for it)

jpelay:
> And that looks weird, making them think they did it wrongly (I can be convinced this is not a real big problem, but that is the reason for it)

That makes sense! Then I think we can close this one? I'll open an issue to properly highlight the cases where underscores sit before or after a variable, because right now I think we're not handling them properly.

Felienne:
Yes, let's close this one!

PR #5186, opened 2024-02-27
🖊️ Allow -= and > operators to be used independently when merging grammars

### LINKED ISSUE #5176: 🖊️ When merging grammars support > and -= operators independently of the += operator
**Language idea incl level**
Currently, when grammars are merged, it is necessary to use the += operator for any custom operator to work. This means that, for example, we cannot use -= independently of +=. As a result, sometimes a rule is removed and re-added, only to allow other rules to be removed. For example, in level 9, the grammar file looks like this:
```
command:+= repeat | error_repeat_no_command-=repeat | error_repeat_no_command | error_repeat_no_print | error_repeat_no_times
```
While the only necessary change is:
```
command:-=error_repeat_no_print | error_repeat_no_times
```

boryanagoncharenko:
Fixes #5176

Merging grammar rules was only possible if the += operator was present, which means that the -= and > operators could not be used independently of +=. With this change, all operators could be used independently of each other.

Because the > operator now does not need the context of the += operator, I had to change it to a symbol which does not appear in the lark language (e.g. ->). Please consider whether there is a better candidate than `>>`, for example `>=`?

**How to test**

- The introduced merging changes are covered with unit tests
- All existing tests should pass


hedy.py
```diff
@@ -2791,53 +2791,55 @@ def merge_grammars(grammar_text_1, grammar_text_2):
...
 
-    return new_rule, deletables
+def merge_rules_operator(prev_definition, new_definition, name, complete_line):
+    op_to_arg = get_operator_to_argument(new_definition)
+
+    add_arg = op_to_arg.get(ADD_GRAMMAR_MERGE_OP, '')
+    remove_arg = op_to_arg.get(REMOVE_GRAMMAR_MERGE_OP, '')
+    last_arg = op_to_arg.get(LAST_GRAMMAR_MERGE_OP, '')
+    remaining_commands = get_remaining_rules(prev_definition, remove_arg, last_arg)
+    ordered_commands = split_rule(remaining_commands, add_arg, last_arg)
```
jpelay, line 2807: A lot better now!

jpelay (approved):
Hi, Boryana! Awesome job as always!!

PR #5188, opened 2024-02-28
🪲 Correctly filter teacher adventures in programs page

jpelay:
Adds the teacher adventure's names to the `adventure_names` dictionary, preventing a bug that was causing the `/programs`  page to crash. Also adds a bit of defensive programming around the use of that dictionary, in case in the future we have a problem in the content or in the adventures from the DB.

**How to test**
* Log in as a student that has a program saved which comes from a teacher adventure and go to My programs page. Check that the page works correctly and doesn't crash.

### DIFF
--- a/hedy_content.py
+++ b/hedy_content.py
@@ -531,9 +531,9 @@ class Adventures(StructuredDataFile):
         sort = {}
         for program in programs_by_level:
             if program['level'] in sort:
-                sort[program['level']].append(adventure_names.get(program['adventure_name']))
+                sort[program['level']].append(adventure_names.get(program['adventure_name'], program['adventure_name']))
             else:
-                sort[program['level']] = [adventure_names.get(program['adventure_name'])]
+                sort[program['level']] = [adventure_names.get(program['adventure_name'], program['adventure_name'])]
         for level, adventures in sort.copy().items():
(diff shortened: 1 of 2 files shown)


Felienne:
Hi @Annelein!

Would it make sense for you to review this since you were recently working on the programs pages?

hedy_content.py
```diff
@@ -543,7 +543,7 @@ def get_sorted_adventure_programs(self, programs, adventure_names):
         programs_by_adventure = []
         for item in programs:
             programs_by_adventure.append(
-                {'adventure_name': adventure_names.get(item.get('adventure_name')),
+                {'adventure_name': adventure_names.get(item.get('adventure_name'), item.get('adventure_name')),
```
Annelein, line 546: What does this new line do?

Felienne: if you give a second argument to a get, it will return that second argument when nothing is found (see: https://365datascience.com/question/what-does-the-second-argument-of-the-get-method-stand-for/)

Annelein: Sweet good to know!

Annelein (approved):

Felienne:
Grrr this too has the flaky test, merged it in manually

PR #5195, opened 2024-02-29
📜 Content Cleanup upto level 8

MarleenGilsing:
Cleaning up all mistakes and imperfections I can find in the content. 

I did unfortunately have to throw away some translated content, if the content was outdated. I replaced it with the up-to-date English version. 

Next week I'll do the same for level 9 and up. This can be merged before that :)


Felienne (approved):
THanks!!

PR #5197, opened 2024-02-29
🪲 Fix old programs in program page, improve filtering teacher programs

jpelay:
Same thing happened to us back in #4417. Some old programs don't have the field `adventure_name` in them, and thus our systems fails. I also improve a bit the filtering of programs, by not using the program name, but rather the adventure name.

I also have a question, what to do with programs that might have an id that's neither in the DB or our own contents? Right now, Im resorting to just show the id, but perhaps we should skip them? What do you think?

image

**How to test**

Go to My Programs page, it should be working. 

If you want to see a program with a non-existing adventure linked, copy an existing `program` record in the dev_dabase.json and change the `adventure_name` field to something that's not present in the database.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1005,10 +1005,15 @@ def programs_page(user):
     all_programs = DATABASE.filtered_programs_for_user(from_user or username,
                                                        submitted=submitted,
                                                        pagination_token=page)
-
+    ids_to_fetch = []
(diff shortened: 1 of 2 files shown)


Felienne (approved):
Thanks!!

PR #5202, opened 2024-03-01
💻 variable list appears automatically open instead of hidden 

### LINKED ISSUE #5147: Open variable watches when there is a variable
Right now, when running Hedy code step-by-step, the label icon is not there, until there is a variable to show. Then, you need to click that icon, if you want to see the variables and their values. How would we actually want this?

- maybe we can have the icon there, with an empty table, when there are no variables yet?
- maybe, when there is a variable, we do not wait for the user to click on the icon, but we show the list of variables immediately? Then they can click the icon if they do not want to see the list (instead of the other way around)
- we can still change the color or something when a variable changes, to give it extra attention

_Originally posted by @juliabolt in https://github.com/hedyorg/hedy/issues/5092#issuecomment-1941636487_

juliabolt:
Fixes #5147 

**How to test**

* Go to level 5 of Hedy (or another level in which variables exist)
* Create a program with at least two variables, for example
```
naam is Hedy
leeftijd is 15
print naam is leeftijd jaar oud
leeftijd is 40
print naam is leeftijd jaar oud
```
* Run the program step-by-step
* Check that the variable list and label button do not appear when there are no variables yet, but when there are, the variable list is shown without the need to click the label button. 
* Check that the variable list is hidden after you click the label button, and reappears if you click the label button again
* Check that the variable list stays hidden after clicking the label button, when continuing to run the program

**Checklist**

- [ x] Links to an existing issue or discussion


Felienne:
> I see it says I changed 3 grammar/lark files, but I haven't, must be an old version or so... it's just the files in static/js/

This is caused by running the scripts, it should have been kept up to date by other people, not your fault they appear here!

Felienne (approved):
Nice, this works perfectly @juliabolt!

PR #5210, opened 2024-03-05
🧪 extra tests for the my programs

hasan-sh:
- **needed data-cy strings**
- **if name given**
- **delete an adventure**
- **needed functions for testing programs**
- **more tests + refactoring of the tests of my-programs**
- **hide delete/submit action if teacher is viewing secon-teachers' programs + correct logic**
- **add needed data-cy labels**
- **change short_name to be cutom adventure's name not id!**
- **more tests**


Felienne (approved):
Thanks!! Hope this prevents more issues with this page in the future

PR #5214, opened 2024-03-06
🧪 cypress tests for placeholders

hasan-sh:
- **test placeholder command**
- **inner unneeded placeholder**


hasan-sh:
The approach of detecting placeholders is infeasible with cypress. Perhaps we could research other methods to test this.

PR #5216, opened 2024-03-06
 💻 Adding a button for microbit

rmagedon97:
Adding a button that that transpiles the Hedy code to Micro-python code and then sends it to the user for downloading. 
The button is in class=hidden with id=saveMicrobitFilesContainer in templates/incl/editor-and-output.html
When the button is clicked sets the Microbit boolean to true and transpiles user code. The user code is saved in the Micro-bit folder. Then the saved code is converted to .hex and it is send to the user for downloading. At the end both files are deleted. 

Created @app.route for generating and downloading the files in app.py

**How to test**

Set the button to not be hidden. 
Click on the Run Micro bit button and a hex file should be downloaded automatically. Then the file can be uploaded [here](https://python.microbit.org/v/3) to check if the code is correct or directly saved to a Microbit via USB. 

**Checklist**


Felienne:
Hi @rmagedon97!

I made this into a draft since the description is not filled in. Feel free to set this to ready to review when everything is ready.

rmagedon97:
Hello @Felienne. 

Is this better. I have also made the button hidden. Maybe we can merge that and then i can start working on the feature flag?

Felienne (changes requested):
Thanks @rmagedon97!

We want to add this button intot he feature flag too (so only show the button if the feature flag is true). Otherise we will show the butoon now while there is hardly functionality. Can you add that?

Also you will have to fix the po conflicts, do you know how to (it is documented in the wiki)

rmagedon97:
Hello @Felienne, 
I have fixed the .PO conflicts. 
I thought as the button is hidden and doesn't appear anywhere will be fine but I'll try to add the button to the feature flag.

hasan-sh:
I'm looking at the implementation with Teodor, however, this PR will be quite hard to merge since it discards many recent changes, probably due to merge conflicts. The best solution @rmagedon97 is to copy the specific pieces you need and create a new branch&pr, as I'm afraid it'd be messy otherwise!

Felienne:
> I'm looking at the implementation with Teodor, however, this PR will be quite hard to merge since it discards many recent changes, probably due to merge conflicts. The best solution @rmagedon97 is to copy the specific pieces you need and create a new branch&pr, as I'm afraid it'd be messy otherwise!

Yes, cherry picking might be easiest? I think @rmagedon97 did that last week so hopefully it can work again!

rmagedon97:
I'm closing this one as #5249  contains the newest code.

PR #5225, opened 2024-03-07
🪲 Prevent stopped programs from executing

### LINKED ISSUE #5211: 🪲 Switching to another tab with an ask prompt open, causes the program to mix up
**Describe the bug**
To reproduce this bug do the following:

1. Student beging to runs program
2. Then student switches to another adventure, while previous program still has an ask propmpt open
3. When the student runs the new program, somehow skulpt still is processing the old ask prompt (handled internally as a promise), and then everything gets mixed up

jpelay:
In Skulpt there really isn't a way to stop the execution of a program. If the program uses things like prompts for the `input` function, or sleep, then those promises are still somewhere in the JavaScript handler. Therefore, when you execute a new program while another one is still waiting for a promise to be fulfilled (i.e: waiting for user input through an ask prompt), when the new program finishes execution, the old one will continue its path.

To fix this I added a new field to `sessionStorage` which will have an id for the program being executed, and I edited Skulpt to pass the id of the program. Therefore, when an old program is being executed, its id will be different than the current one and therefore will finish its execution.

Fixes #5211

**How to test**

* Go to level 14 and execute the TicTacToe adventure, fill the first ask.
* With an ask prompt still open, switch to the Quizmaster tab and execute the program.
* Only the Quizmaster program should be executed.


jpelay:
This is ready for review! I tried to solve to also bypass printing and exceptions, but ultimately was not able to! Remember to also review https://github.com/hedyorg/skulpt/pull/2!

Felienne:
> This is ready for review! I tried to solve to also bypass printing and exceptions, but ultimately was not able to! Remember to also review [hedyorg/skulpt#2](https://github.com/hedyorg/skulpt/pull/2)!

I will pick this PR up! Do I test the skulpt PR in the same way as this one? Should I pull both and test at once?

jpelay:
> > This is ready for review! I tried to solve to also bypass printing and exceptions, but ultimately was not able to! Remember to also review [hedyorg/skulpt#2](https://github.com/hedyorg/skulpt/pull/2)!
> 
> I will pick this PR up! Do I test the skulpt PR in the same way as this one? Should I pull both and test at once?

Only test this one! If this one works, approve both, since the files generated by Skulpt are also included here.

Felienne:
I tried the steps above (after running frontend), but I still get the ...'s in view:

image

jpelay:
> I tried the steps above (after running frontend and doing a hard refresh), but I still get the ...'s in view:
> 
> image

Yes, this happens because when there's an exception in the original program it fails. You have to answer the first ask of the tic tac toe program, with any number, and then switch to the other tab.

Felienne:
> > I tried the steps above (after running frontend and doing a hard refresh), but I still get the ...'s in view:
> > image
> 
> Yes, this happens because when there's an exception in the original program it fails. You have to answer the first ask of the tic tac toe program, with any number, and then switch to the other tab.

Ah got it! So we should just make sure to never throw exceptions anymore (which we want anyway) and it will be solved haha!

Felienne (approved):

PR #5249, opened 2024-03-11
💻 Adding a button for Microbit

rmagedon97:
Adding a button that that transpiles the Hedy code to Micro-python code and then sends it to the user for downloading.
The button is in class=hidden with id=saveMicrobitFilesContainer in templates/incl/editor-and-output.html
When the button is clicked sets the Microbit boolean to true and transpiles user code. The user code is saved in the Micro-bit folder. Then the saved code is converted to .hex and it is send to the user for downloading. At the end both files are deleted.

Created @app.route for generating and downloading the files in app.py

**How to test**

Set the button to not be hidden.
Click on the Run Micro bit button and a hex file should be downloaded automatically. Then the file can be uploaded [here](https://python.microbit.org/v/3) to check if the code is correct or directly saved to a Microbit via USB.

**Checklist**


hasan-sh (approved):
Looks good!!

rmagedon97:
> Looks good!!
Hello @hasan-sh, 
Thank you for approving.
It seems there is an issue with static/js/appbundle.js.map.

Should i fix the conflict and push again?

hasan-sh:
> > Looks good!!
> 
> Hello @hasan-sh, Thank you for approving. It seems there is an issue with static/js/appbundle.js.map.
> 
> Should i fix the conflict and push again?

Yes you can!

PR #5256, opened 2024-03-13
🪲 Escape internal var used in repeat

### LINKED ISSUE #5255: 🪲 Internal variable in repeat `i` should be escaped
**Describe the bug**
Internal variable `i` used in repeat leaks in the right top corner of the output window. The issue was already logged and addressed for level 7 #3676 but the issue persists in upper levels.

Level 8 and up
```
options is 1, 2, 3, 4, 5
repeat 5 times
  print options at random
```

Screenshot 2024-03-13 at 10 43 56

boryanagoncharenko:
Fixes #5255 
The internal variable `i` used in repeat statements is now escaped with the `__` prefix, so that it does not show in the output window.

**How to test**
- Run Hedy locally and run the following snippets for the specified levels. Ensure that the `i` variable does not appear in the blue variables list in the upper right corner of the output window:

Level 7:
```
options is 1, 2, 3, 4, 5
repeat 5 times print options at random
```
Level 8:
```
options is 1, 2, 3, 4, 5
repeat 5 times
  print options at random
```


jpelay (approved):
Works!
image

PR #5265, opened 2024-03-14
💻 For teachers redesign

Annelein:
[The design](https://www.figma.com/file/qGCM9e25GuNCuriJ3xNrAG/Teacher's-front-page?type=design&node-id=0-1&mode=design&t=sCgBWCel7cAwFTCS-0)


Felienne:
> @jpelay @hasan-sh @Felienne @gisellandrade this is what it would look like! What do you think?
> 
> (If you open the slides, the public content container stretches with it, @jpelay @hasan-sh do you know how to solve this?)
> 
> Screenshot 2024-03-19 at 16 53 39

I like this!!

hasan-sh:
> (If you open the slides, the public content container stretches with it, @jpelay @hasan-sh do you know how to solve this?)

Do you still need help? I like it from the picture too!

hasan-sh:
Hi @Annelein so instead of solving this problem, which we could, why don't we follow the same style of the other "view ..." buttons? So, just create a table and when you click on view slides, that table appears and the user is navigated to it.

Annelein:
> Hi @Annelein so instead of solving this problem, which we could, why don't we follow the same style of the other "view ..." buttons? So, just create a table and when you click on view slides, that table appears and the user is navigated to it.

Hmm we could, I'm not sure if I like it as much, because the other tables include teacher actions, and the slides not so much.   So its kind of something different I guess. But at the same time it would be consistent. @gisellandrade and @Felienne what do you think?

jpelay:
> Hmm we could, I'm not sure if I like it as much, because the other tables include teacher actions, and the slides not so much. So its kind of something different I guess. But at the same time it would be consistent. @gisellandrade and @Felienne what do you think?

I noticed that you display the slides inside the div where the view slides button is, what if you display them bellow like you do with the tables? You can keep the same styles as you do right now, but move them to their own spaces, so they don't modify the Slides' explanation div. Something like this!

image

Annelein:
> > Hmm we could, I'm not sure if I like it as much, because the other tables include teacher actions, and the slides not so much. So its kind of something different I guess. But at the same time it would be consistent. @gisellandrade and @Felienne what do you think?
> 
> I noticed that you display the slides inside the div where the view slides button is, what if you display them bellow like you do with the tables? You can keep the same styles as you do right now, but move them to their own spaces, so they don't modify the Slides' explanation div. Something like this!
> 
> image

Thats exactly what @hasan-sh suggested, I just implemented the design from @gisellandrade. So if you guys think this would look best, I will implement it!

jpelay:
> Thats exactly what @hasan-sh suggested, I just implemented the design from @gisellandrade. So if you guys think this would look best, I will implement it!

Great mind thinks alike!  haha What if we we keep the same idea of keeping the slides table in a separate the div, but make it the same width as the Slides explanation one?

jpelay:
image

What do you think?

Annelein:
@jpelay added it! So it's ready to be reviewed!

jpelay:
> @jpelay added it! So it's ready to be reviewed!

Awesome! Will review some other PRs and then come back to this one!

templates/for-teachers.html
```diff
@@ -10,90 +10,126 @@ <h2 id = "tutorial_title" class = "text-white" > </h2 >
...
+      <h2 class="m-0 basis-1/12"> {{_('classes')}} </h2>
+      <div class="text-black basis-10/12">
+      <p>{{ _('classes_info') }}</p>
+      </div>
+      <div class="flex flex-row gap-3 justify-end basis-1/12">
+        <button class="green-btn inline-flex items-center" id="create_class_button" onclick='hedyApp.create_class ({{_('class_name_prompt')|default(None)|tojson}})'
+        {% if second_teacher and role == 'viewer' %}disabled{% endif %}>{{_('create_class')}}</button>
+        {% if teacher_classes %}
+        <button class="white-btn inline-flex items-center gap-2" id="view_classes"
+        _="on load hedyApp.getForTeacherTable('classes_table') end
```
jpelay, line 25: Great usage of Hyperscript and javascript!

Felienne (approved):
Lovely work @Annelein!!! Such a big change for users!

PR #5269, opened 2024-03-14
🪲 Debug play

### LINKED ISSUE #5111: 🪲 Can't debug play
There's no possibility to debug the debug the `play` command currently, due to it being added to the `Sk` object  only in the `run` if.

jpelay:
Adds the `play` builtin function to both `run` and  `debug` paths, so now we're able to debug the play command. Also fixes the highlighting of variables within the play command, as well as fixing the highlighting of the lines containing the play command while debugging.

Fixes  #5111 

**How to test**
*  Automated tests should pass
* Go to level 2 and debug this code:

```
note is C1
play note
play C2
```

Also check that the `note` variable is colored blue.


Felienne:
Nice work @jpelay! Just one thing (my fault, of course haha!) In the variables list, chosen_note is also shown because we use it in hedy.py:

https://github.com/hedyorg/hedy/blob/8a2c0679c363afa3338db31fd45932993e9c6a5a/hedy.py#L1705

image

Can we escape it? I guess there is a list of to be escaped variables somewhere already?

jpelay:
> Can we escape it? I guess there is a list of to be escaped variables somewhere already?

Done! Indeed there's a list [here](https://github.com/hedyorg/hedy/blob/8a2c0679c363afa3338db31fd45932993e9c6a5a/static/js/debugging.ts#L120) in case you want to check it out!

Felienne (approved):

PR #5276, opened 2024-03-18
🪲 fix quiz/puzzles tabs not showed to students

hasan-sh:
**Problem**
Related to https://github.com/hedyorg/hedy/pull/5156

image

The feature actually works. However, it only works if the teacher has logged in and at least visited the customize-class page. This is because there's a migration step that I added which basically adds the quizzes/puzzles to the list of adventures if they were not hidden before. The problem that this PR solves is related to the scenario in which a student is logged in and interacts with a class while their teacher has not refreshed the customize-class page: Thus the migration was not done yet at this point, hence the students then would not see the quizzes/puzzles.

_Why didn't we figure it out earlier not during testing nor reviewing?_ This is mainly because **we tested it while relying on** the _previewing class_ feature. This feature, supposidely, would allow a teacher to exactly simulate what students see and how they interact with the related class. But apparently that's not the case, since the preview feature is simply coplying the customization class and "allowing the user to merely see that class". 

**How to test?** 
- in your `dev_database.json` file, remove the quiz/puzzle manually from at least the first level of the `sorted_adventures` of your class (e.g., CLASS1)
- then log in as a student and the quiz/puzzle should be readded and shown to you, iff. the quiz/puzzle was not disabled


app.py
```diff
@@ -1412,7 +1412,35 @@ def hour_of_code(level, program_id=None):
         ))
 
 
+def possibly_migrate_quizzes_parsons_tabs(customizations, parsons_hidden, quizzes_hidden):
```
jpelay, line 1415: I see that this code is repeated in the For Teachers Module, perhaps we can make this code a function in that module and import it here?

jpelay (changes requested):
Hello Hasan! Great work figuring this bug out, I didn't foresee this case when testing.  I tested and it indeed works fine, I just have a comment about some duplicated function, but otherwise this is good to go!

app.py
hasan-sh: Yes indeed, just reused that function in the foreteachers!

jpelay (approved):
Awesome! Great work, Hasan. And good detective skills figuring out the bug, that can be the most frustrating part of programming.

PR #5277, opened 2024-03-18
💻 Add sleep prompt until level 7

### LINKED ISSUE #4769: [UI idea] Show sleeping state
While working on #4677, two things occurred to me. One is, when a program is sleeping, this is not really clear to the user, for example, this program:

image

Kids might not know 10 means 10 seconds, and might misunderstand what is going on. We could show a bar in the execution box like the "wait" we have for pressit, if a sleep is used that is more than 1 second long?

Potentially we can show an option "discable this warning" so that kids can disable it if they understand sleep. I am not sure this is a valuable idea (maybe it's overkill?) but it could be useful?

Annelein:
Fixes #4769 

After level 7 'if' statements are possible and it gets a lot more complicated to run the sleep prompt in order.

https://github.com/hedyorg/hedy/assets/48122190/a5854391-3a69-472a-9c70-4145a8f2b3e1


jpelay (commented):
Hi!! Good job Annelein, it does work when I supply a number to the sleep command, but not a variable! It'd be nice if it also worked with them.

Annelein:
> Hi!! Good job Annelein, it does work when I supply a number to the sleep command, but not a variable! It'd be nice if it also worked with them.

Didn't think of it at all haha, thanks! Added it

jpelay:
Hi Annelein! Thanks for solving the variable issue. Now it does work with variables, but I tested it in further levels and found a few other bugs:

1. Only writing `sleep` does not show the modal. For example this program on level 2.
```
sleep
print hello
```
2. If I have 2 sleeps, only the first one will have a modal. For example this program on level 2:
```
print hello world
sleep 1
print after sleep 1
sleep 1
print after sleep 2
```
3. From level 8 onwards the sleep modal is shown even if the sleep instruction hasn't been reached yet, for example this program:

```
name = ask 'What is your name?'
if name is Hedy
    print 'Welcome Hedy'
    print 'You can play on your computer!'
    sleep 1
    print 'after sleep'
```

Annelein:
Points 1 and 2 are solved, for point 3, after level 7 it gets a lot more complicated to run the sleep prompt in order, so the prompt is only shown until level 7

jpelay (approved):
Great work  Annelein!! Thanks so much for addressing my comments.

PR #5278, opened 2024-03-18
💻 Programs count

### LINKED ISSUE #1275: [UI idea] Do not show confetti when running demo code

### LINKED ISSUE #5192: 💻 My programs page should also show program count
I went to a school yesterday, and the kids had some ideas!

Here is one:

Public profile shows the number of programs on the top:

image

Could we add it to my programs also?

### LINKED ISSUE #5193: 💻 change calculation and phrasing of number of programs in profile page
Here is another remark from yesterday's school, in your public profile, why is there "programs created" and "programs saved", what is the difference?

image

I suspect this is legacy from when we really had a save button! IN the current state of things, where a run is a save, what does this mean?

And even more general (this is my note, not from the kids...) what does programs created mean? Should we use the new system here that @jpelay made in #5162 so we count the actual changed adventure codes (and change the label into something like "number of adventures completed" or "number of adventures tried" ?

Annelein:
Fixes #5193 
Fixes #5192 
Fixes #1275

This fixes that only the programs which are modified are counted as (for now) programs saved.
[This](https://github.com/hedyorg/hedy/issues/5386) is the new issue for the design that shows this counter (where programs_saved will be changed):

**Also solved that only if a program is modified, the confetti button is shown:**

https://github.com/hedyorg/hedy/assets/48122190/dd0e6fe2-5484-4192-8f92-be01d1d5607e

How to test:
- Run a program with preview code, make sure it doesn't get added to my-programs
- Run a program with 'modified code' so not the preview code and make sure it does get added to my programs
- Remove a program from my programs and make sure the counter and the options in the forms get properly updated


Felienne:
Hi @Annelein! We are somewhat eager to get this PR out. What is the blocker, are we waiting for a design from GIsell or for a review or something else?

Annelein:
> Hi @Annelein! We are somewhat eager to get this PR out. What is the blocker, are we waiting for a design from GIsell or for a review or something else?

Everything is ready but already waiting for the design from Gisel for a long time...

Felienne:
> > Hi @Annelein! We are somewhat eager to get this PR out. What is the blocker, are we waiting for a design from GIsell or for a review or something else?
> 
> Everything is ready but already waiting for the design from Gisel for a long time...

Thanks! I will ping her again. Otherwise we can maybe split of the fix for counting the programs (https://github.com/hedyorg/hedy/issues/5193) from the new design?

Annelein:
> > > Hi @Annelein! We are somewhat eager to get this PR out. What is the blocker, are we waiting for a design from GIsell or for a review or something else?
> > 
> > 
> > Everything is ready but already waiting for the design from Gisel for a long time...
> 
> Thanks! I will ping her again. Otherwise we can maybe split of the fix for counting the programs (#5193) from the new design?

I will because the design is still not what we are looking for I think, see https://github.com/hedyorg/hedy/issues/5192#issuecomment-2042362957

Annelein:
@jpelay could you look at the code if there is something I'm doing wrong? Becaus `is_program_modified()` returns True for this program:

https://github.com/hedyorg/hedy/assets/48122190/d1042aed-b694-459d-b952-a4894cc0923a

Other than that, I think this is ready, apart from the design, that will be done in a new PR.

Annelein:
@Felienne this will just implemented the backend code, https://github.com/hedyorg/hedy/issues/5386 is a new issue for the design of it

jpelay (changes requested):
Hi Annelein! It seems like all program's are saved, even if I didn't modify them, see. It seems to happen with every program:

https://github.com/hedyorg/hedy/assets/45865185/e7e0e877-86b4-4735-9bd9-cd11e8eafd17

Second: if I'm not logged in, I get an error and can't execute a program.

Annelein:
> @jpelay could you look at the code if there is something I'm doing wrong? Becaus `is_program_modified()` returns True for this program:
> 
>  Screen.Recording.2024-04-08.at.13.59.01.mov 

@jpelay Sorry it probably wasn't clear enough that by this comment I meant that `is_program_modified()` returns True for most / maybe all programs, so I already was aware of that problem and I was wondering if you could help me and look at the code what I'm doing wrong.

jpelay:
> > @jpelay could you look at the code if there is something I'm doing wrong? Becaus `is_program_modified()` returns True for this program:
> > Screen.Recording.2024-04-08.at.13.59.01.mov
> 
> @jpelay Sorry it probably wasn't clear enough that by this comment I meant that `is_program_modified()` returns True for most / maybe all programs, so I already was aware of that problem and I was wondering if you could help me and look at the code what I'm doing wrong.

I see! Sorry I misunderstood, i thought you only meant that one program. Ill dive into this right away.

website/programs.py
```diff
@@ -69,6 +71,11 @@ def store_user_program(self,
             "adventure_name": adventure_name,
         }
 
+        full_adventures = hedy_content.Adventures("en").get_adventures(g.keyword_lang)
+        teacher_adventures = self.db.get_teacher_adventures(current_user()["username"])
+        is_modified = self.statistics.is_program_modified(updates, full_adventures, teacher_adventures)
```
jpelay, line 76: Hi @Annelein! This is where the problem lies. We are sending this `updates` object and `is_program_modified` expects a `program`, specifically with the fields: `adventure_name`, `code`, and `level`. Now, I see that `adventure_name` might be null on this function, which can cause issues in the function, so I'll say that is `adventure_name` is null we just return true and save the program.

Annelein: > Hi @Annelein! This is where the problem lies. We are sending this `updates` object and `is_program_modified` expects a `program`, specifically with the fields: `adventure_name`, `code`, and `level`. Now, I see that `adventure_name` might be null on this function, which can cause issues in the function, so I'll say that is `adventure_name` is null we just return true and save the program.

@jpelay but this already returns True if `adventure_name` is null, because then `if program['adventure_name'] == adventure["id"]:` does not happen right?

https://github.com/hedyorg/hedy/blob/dbc117abc336199be2bb87af45f6d8d0e36554a6/website/statistics.py#L267C1-L277C43

Annelein: And level and code will never be null, so I still don't really get why copied code is being saved as a program?

Annelein: what does this line do?         
```
adventure_info = full_adventures.get(program['adventure_name'], {})\
            .get('levels', {})\
            .get(program['level'], {})
```

because it becomes None / empty, while adventure_name and level are set in program:
`{'session': '8179ba4ca4e242e29959017837f4313b', 'date': 1712750025338, 'lang': 'en', 'version': 'DEV', 'level': 1, 'code': 'forward 100\nturn left\n', 'name': 'Turtle 1', 'username': 'teacher1', 'error': False, 'adventure_name': 'Turtle'} `

jpelay: This line get's the adventure information. It's a dictionary where each adventure is identified by its key in the yalm. I see in the example you posted that `adventure_name` has the value `Turtle` which is not the key of the adventure, but the name in English, the real key is: `turtle` that's why is returning empty.

jpelay:
@Annelein Just made a few changes! Hope that it works now

Annelein:
@jpelay works amazing thank you so much!!! Only thing left is that when a program is removed, the program count is not updated right away, pinging @hasan-sh if he could help me with implementing auto save?

jpelay (approved):
It works!!!! 🥳🎉🎊 Can you test a bit on alpha and check that everything works fine?

Annelein:
> It works!!!! 🥳🎉🎊 Can you test a bit on alpha and check that everything works fine?

I will!

PR #5292, opened 2024-03-21
🪲 Check if classes and levels exist in adv

### LINKED ISSUE #5289: 🪲 adding an adventure to classes
**Describe the bug**
When i create/clone an adventure and add it to an existing class, an error occurs!
image

@Annelein I thought i mentioned we should use `.get(..)` in case we're accessing new properties that may or may not exist?! I see you changed it in one place and likely forgot other places, but we need to do convert them all!

Annelein:
Fixes #5289


website/for_teachers.py
```diff
@@ -1075,10 +1075,10 @@ def update_adventure(self, user):
         elif current_adventure["creator"] != user["username"]:
             return gettext("unauthorized"), 403
         current_classes = {}
-        if body.get("classes"):
+        if body.get("classes") and current_adventure.get('classes'):
             current_classes = current_adventure["classes"]
```
hasan-sh, line 1079: Imagine `current_adventure` being an existing adventure. It most likely doesn't have classes, therefore the next line should be changed. 

```suggestion
        if body.get("classes"):
            current_classes = current_adventure.get('classes', [])
```

Same goes for the levels you made below and all other places you use classes as well!

hasan-sh (changes requested):

website/for_teachers.py
Annelein: These are the only places

hasan-sh (approved):
Great, will keep an eye and see it on prod.! Thanks:))

PR #5297, opened 2024-03-21
🪲 fix custom adventures' alignment

### LINKED ISSUE #5293: 🪲 all <br>'s should be replaced, not the first one only!
**Describe the bug**
https://github.com/hedyorg/hedy/blob/af093f0af02d02592cd265b3102237e59a6e0b45/static/js/teachers.ts#L248
Should add the g flag so that it replaces all occurrences of the `<br>` tags.

### LINKED ISSUE #5296: 🪲 custom adventure in two columns
**Describe the bug**
As reported by one of the teachers, their custom adventure is messed up when its content is large enough.
image

The probelm occurs due to how we split the content. Currently, we split the content and apply the commonmark filter to convert it to html. It becomes thus problemetic when we split the content while a tag (`pre` tag in this case) is not closed yet.

hasan-sh:
Fixes #5296 
Fixes #5293

**Description**
In this PR, I attempted to split the teachers' adventures into two columns optimally. This is done by splitting the tags rather than the text, as we currently do. Furthermore, I've set a maximum height on the code blocks to equal the editor's height. This is to avoid very long code blocks which would most likely smaller than other information within the adventure content.

**How to test?**
- create long adventures
- add them to your class
- preview the class and whether the adventure is split intow 2 columns, the best they could!

**ALSO** this PR should fix how currently live adventures, please try to find public adventures and test them locally. Here's some:
- https://hedy.org/public-adventures?level=5&lang=&tag=&search=lootjes
- https://hedy.org/public-adventures?level=5&lang=&tag=&search=Schildpad
- https://hedy.org/public-adventures?level=13&lang=&tag=&search=bonus


Felienne (changes requested):
Ah sorry I meant request changes not approve!

hasan-sh:
> I just tested it and it works well for new adventures, but not entirely correct yet for existing adventures (I forst edited an existing one)
> 
> image
> Can you take a look at that?

I believe it has to do with formatting adventure's code! Could you tell me what you did to reproduce this?

Felienne:
> > I just tested it and it works well for new adventures, but not entirely correct yet for existing adventures (I forst edited an existing one)
> > image
> > Can you take a look at that?
> 
> I believe it has to do with formatting adventure's code! Could you tell me what you did to reproduce this?

I selected one of my existing adventures (created before the automatic curly addition, so maybe that is related?), and added more lines of code to the code example with copy paste.

hasan-sh:
@Felienne it was from that change indeed as was another one that i fixed here as well. I'll keep on any possible errors of that PR. 

This should be fixed now!

Felienne:
> @Felienne it was from that change indeed as was another one that i fixed here as well. I'll keep on any possible errors of that PR.
> 
> This should be fixed now!

Yes, the newlines are fixed indeed: 

image

Although I still get the curlies. This is not an issue for this PR, but I will make a new one for @jpelay

hasan-sh:
> > @Felienne it was from that change indeed as was another one that i fixed here as well. I'll keep on any possible errors of that PR.
> > This should be fixed now!
> 
> Yes, the newlines are fixed indeed:
> 
> image
> Although I still get the curlies. This is not an issue for this PR, but I will make a new one for @jpelay

@jpelay the problem is that the formatted_content is only converting <br> tags in the front-end and when you save the adventure, therefore existing adventrues have this problem. Here i simply passed the content which may have curly braces for existing adventures. I'd do that step in the backend instead!

Felienne:
> > > @Felienne it was from that change indeed as was another one that i fixed here as well. I'll keep on any possible errors of that PR.
> > > This should be fixed now!
> > 
> > 
> > Yes, the newlines are fixed indeed:
> > Although I still get the curlies. This is not an issue for this PR, but I will make a new one for @jpelay
> 
> @jpelay the problem is that the formatted_content is only converting tags in the front-end AND when you save the adventure, therefore existing adventrues have this problem. Here i simply passed the content which may have curly braces for existing adventures [09d26c3](https://github.com/hedyorg/hedy/commit/09d26c3111b249e5d7f4933099cc7f348b1d9036). I'd do that step in the backend instead!

Thanks for the explanation and for the suggested fix. Will you do that in this PR?

hasan-sh:
> Will you do that in this PR?

I think we should discuss it first with @jpelay, i think i'd make more sense to take the code related to the adding curly braces automatically to the backend. I'm just wondering if there's a reason why this is in the frontend.

Felienne (approved):

PR #5309, opened 2024-03-25
💻 Language switcher text

### LINKED ISSUE #4835: [UI idea] change fa language icon into the used language
I already did this in: https://github.com/hedyorg/hedy/pull/4828/commits/9516f7ff40025c69632198bfc81a00d66a26d596

But it has two limitations:

1. https://github.com/hedyorg/hedy/pull/4828#issuecomment-1833151233 I think pybabel might have language abbreviations in the actual language? Someone must dive into that

2. Somehow it starts at None? Should not be so hard to fix but needs a bit of debugging.

Annelein:
Fixes #4835 

[Babel](https://www.unicode.org/cldr/charts/44/supplemental/territory_language_information.html) only lists language codes in Latin alphabet. I did some searching and I think it would be quite difficult to find the non Latin language codes. So for the non Latin languages, I thought the best solution would be to just state the _non Latin language_ in the switcher instead of the _non Latin language code._ 

**How to test**
1. Select a non latin language in your profile
2. Go to Hedy
3. Check if there is a language in the switcher (and it's not None)
4. Switch between English and the (non Latin) language

https://github.com/hedyorg/hedy/assets/48122190/73adcb1c-fefb-42f3-958d-3876dbf6eb7b


app.py
```diff
@@ -414,6 +415,9 @@ def setup_language():
     # front-end is fixed based on this value
     g.dir = static_babel_content.TEXT_DIRECTIONS.get(g.lang, 'ltr')
 
+    # True if it is a Latin alphabet, False if not
```
Felienne, line 418: Love this comment! This is exactly the type of thing that is very easy to understand/explain and superhard to get from the code!

Felienne (approved):

PR #5313, opened 2024-03-26
💻 Add adventure within customize class

### LINKED ISSUE #4084: [UI idea] Add adventures from customization page
Adding your own adventures is a bit cumbersome now, if you want to add a new one to your class you have to go all the way back to the teachers page (2x back) add it there and go back to your class (2 clicks forward)*.

Couldn't we do this in the customization page, with an extra button?

image

That could just bring up the author page and, upon saving, add the fresh adventure to the class right away? Is this doable?

* We used to have (I think?) the option to add the adventure to a class right away but that seems to be gone.

Annelein:
Fixes #4084 

**How to test:**
1. Go to `/customize-class`
2. Next to the Reset button (top right), click 'Create adventure'. 
3. Check that the Adventure is correctly added to the level and the class you were in.

https://github.com/hedyorg/hedy/assets/48122190/0a5cb4ee-d231-4328-99fa-d9a86e9b8072


hasan-sh:
I'd add this button next to the reset button, change the color of the reset button to be red and the new create adventure btn to be blue/green.

gisellandrade:
What do you think about this? 
image

gisellandrade:
We have discussed this in today's meeting and we agreed to place the button next to the reset button because if we place it next to the adventures, the button will move and change its position because it depends on the number of adventures there are. created in the class. So, it would look like this.
image

hasan-sh:
All good! Shall we remove the create adventure that's in the class overview?

Annelein:
> All good! Shall we remove the create adventure that's in the class overview?

It will be in #5130

hasan-sh (approved):

PR #5324, opened 2024-03-27
💻 Check for underscore in variable, remove it

### LINKED ISSUE #4749: [UI idea] Map reserved keywords back to original in variable view
**Idea incl level**
We map reserved keywords to unused words by prepending an underscore, which should be an internal thing but now leaks out in the watches screen:

image

Steps to fix this include:

* Creating a dictionary with original value (continue in this case) and remapped value (_continue) in Python
* Fetching the original value from that dictionary in the watches view (not sure if that will be done at the front-end or backend?)

Annelein:
Fixes #4749 

**How to test**
1. Add for example this code:
```
continue = ask 'Do you want to continue?'
if continue is yes
    sure = ask 'Are you sure?'
    if sure = yes
        print 'We will continue'
    else
        print 'You are not sure'
else
    print 'You do not want to conintinue'
```

2. none of the variables should have an underscore:
Screenshot 2024-03-27 at 14 13 04


static/js/debugging.ts
```diff
@@ -86,11 +86,12 @@ export function load_variables(variables: any) {
     for (const i in variables) {
       // Only append if the variable contains any data (and is not undefined)
       if (variables[i][1]) {
-        variableList.append(`<li style=color:${variables[i][2]}>${variables[i][0]}: ${variables[i][1]}</li>`);
+        const variableName = variables[i][0].replace('_', '');
```
boryanagoncharenko, line 89: I think we need to get rid of underscores if they appear in the beginning of the variable name. With the current approach, we remove the first underscore even when it is in the middle of the variable name, which is incorrect:
Screenshot 2024-03-28 at 13 07 54

boryanagoncharenko (changes requested):

Felienne:
Hi @Annelein!

I think the variable is still not showing correctly:

image

Also, can we remove the underscore in the yellow warning bar too?

Annelein:
> Hi @Annelein!
> 
> I think the variable is still not showing correctly:
> 
> image
> Also, can we remove the underscore in the yellow warning bar too?

Oh that's weird! It worked for me.. Will look into it.

Annelein:
@Felienne does it work now?

static/js/debugging.ts
```diff
@@ -86,11 +86,13 @@ export function load_variables(variables: any) {
     for (const i in variables) {
       // Only append if the variable contains any data (and is not undefined)
       if (variables[i][1]) {
-        variableList.append(`<li style=color:${variables[i][2]}>${variables[i][0]}: ${variables[i][1]}</li>`);
+        console.log("HAAI")
```
boryanagoncharenko, line 89: Perhaps this is committed by accident? Could you please remove this line? Otherwise, it looks great.

boryanagoncharenko (changes requested):

boryanagoncharenko (approved):

PR #5331, opened 2024-03-28
📜 New blackjack adventure

MarleenGilsing:
This PR should make the blackjack adventure actually doable for the students. :)

@Felienne (or other testers), would you mind playing the adventure yourself? I would love to know if my instructions are clear enough to make the adventure doable. It is level 17, so it's okay if it's a little hard. 

I'd also like you to give your opinion on the first adventure. It is still quite long (and therefore a little ugly). The alternative is that I could split part 1 into 2 parts, with part 1 only being drawing a card maybe doing the calculating function and part 2 adding the rest. Pro: It does make the adventure smaller and more 'behapbaar', Con: It's less linear (now we only add parts underneath, if we choose the first option we'll have to add parts in the middle etc'. I liked the fact that we just add more code, in each adventure, so I chose for this format. But let me know if you prefer the alternative. 

And as I told you on discord, the numbers for the cards are seen as string text, which isn;t weird, because they come from a list with other words like 'king'. This does cause the parser (if that is the right word) to add the numbers up asif they are words so 8+5 = 85 instead of 13. In the long and ugly code we have on the website currently, where I haven't used functions, this doesn't happen somehow. So this needs to be fixed before this PR can be merged. 

Also, I haven't copied it to other languages yet, nor translated it to Ducth yet, because I want it to be right in ENG first.


Felienne:
Some notes here too:

* fucntion -> function

I agree step 1 is very long, could we split it in two, where we first only program how to calculate cards? That can even be tested in isolation (I think in general trying to explain to kids how to test larger pieces of code is good!) If this remark makes no sense, leave it for now until we can have a meeting about it!

Felienne (approved):
Let's go!

PR #5332, opened 2024-03-28
🪲 fix programs page

hasan-sh:
The program overview currently throws an error. 
image

This PR fixes it and makes sure that we show a custom adventure's name and not id in the program block.

**How to test?**
- run a program/adventure 
- go to my programs
- it should work. 
- stop the server
- remove the `adventure_name` field from that program/adventure's in dev_database > programs > your program
- re-run the server
- my programs page should still run fine!


jpelay (approved):
Solves the issue!

PR #5333, opened 2024-03-28
📜 Hangman adventure

MarleenGilsing:
@Felienne I've made hangman the same way as the Blackjack adventure and split it up in 3 parts.

There is quite a lot of text, but without it I think it would be too hard to make this code. And it is almost the last level, so this is meant for the students who've made it this far anyway, so they could use a challenge. Could you (or someone else) test it to see if the instructions are clear enough to follow and create a game of hangman (especially part 1 and 2, you don't need to figure out how to draw tha gallow for part 3 haha, unless you feel like it) ? 

If it's doable, I will add this adventure to other languages as well and translate it to dutch.


Felienne:
Looks great @MarleenGilsing!! It works a lot better in steps!

A few things:

**Step 1**
* We wil have to be precise about how we indicate variable names, now sometimes you do it with quotes ('amount_letters') and sometimes without (game_over). Same for values like False (In the text you do not use quotes but in the text you do.) Shall we use `` , that makes them show like keywords?
* Do we even want to use 'False' and not False (as in Python) We don't support that now, but I think we should in fact do that!
* "We want player 1 to be able to choose the choose the answer. " -> choose and choose?

**Step 2**
* "*Paste your code " -> Is that * intentional?
* player 2 -> Do we want to use caps (Player 2)?

**Step 3**
Perfect, no notes!

Felienne (approved):
Looks great!

PR #5345, opened 2024-04-01
🧹 Reduce duplication in snippet testing

rix0rrr:
The snippet testing code had some duplication. In this code, we separate out how to find the interesting snippets in the YAML files. For every snippet, we remember where in a YAML file we found it (by keeping track of the keys we used to descend into dictionaries and the indexes we used to descend into lists).

That way, the "restore to English" code can become generic: since every snippet knows where in the YAML structure it comes from, we can automatically find the same snippet in the English language file and copy it over.

**How to test**

If the tests pass, everything works!


tests/test_snippets/snippet_tester.py
```diff
@@ -235,19 +212,112 @@ def do_snippet(self, snippet, yaml_locator=None):
...
 
-    def restore_snippet_to_english(self, snippet, yaml_locator):
+    def restore_snippet_to_english(self, snippet):
         # English file is always 'en.yaml' in the same dir
         en_file = path.join(path.dirname(snippet.filename), 'en.yaml')
 
         # Read English yaml file
         original_yaml = YamlFile.for_file(en_file)
-        original_loc = yaml_locator(snippet, original_yaml)
+        original_loc = locate_snippet_in_yaml(original_yaml, snippet, yaml_locator)
```
boryanagoncharenko, line 221: The `yaml_locator` does not need to be supplied any more

tests/test_snippets/snippet_tester.py
```diff
@@ -9,18 +9,17 @@
 - Snippets can hold on to the information where in the YAML file they were
```
boryanagoncharenko, line 9: This could be scratched now, right :)

boryanagoncharenko (changes requested):

tests/test_snippets/snippet_tester.py
```diff
@@ -30,165 +29,146 @@
...
+            if tag.name == 'pre' and tag.contents and tag.contents[0].contents]
 
 
 def filter_snippets(snippets, level=None, lang=None):
     if (lang or level) and os.getenv('CI'):
         raise RuntimeError('Whoops, it looks like you left a snippet filter in!')
 
     if lang:
-        snippets = [(name, snippet) for (name, snippet) in snippets if snippet.language[:2] == lang]
+        snippets = [s for s in snippets if x.language[:len(lang)] == lang]
```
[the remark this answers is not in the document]
rix0rrr: Oops!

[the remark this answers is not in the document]
rix0rrr: Tiens. I thought `pylint` would catch errors like these.

rix0rrr:
Good catches @boryanagoncharenko, thanks! I'm a bit surprised that the static analysis tools we run didn't catch these, since they're pretty obvious. 

I've confirmed that `pylint` doesn't catch them (grr). `pylance` could potentially do it, but I didn't find out how to run it.

We should probably find a static checker to catch trivial errors?

tests/test_snippets/snippet_tester.py
rix0rrr: Ha. Of course!

rix0rrr:
I've re-added `pylint` to the build checks, and confirmed that `pylint` would have caught those errors.

We used to have it, not sure why we removed it?

rix0rrr:
Shit, pylint was removed in this PR: https://github.com/hedyorg/hedy/pull/5134

@jpelay, do you remember why?

rix0rrr:
Doesn't seem to be easy to re-enable. My struggles will continue in here: https://github.com/hedyorg/hedy/pull/5377

In the mean time, this one should be good to go.

boryanagoncharenko (approved):

PR #5348, opened 2024-04-01
🧹 Update Flask and Jinja2 to the latest version

rix0rrr:
The advantage of upgrading will be that we'll have a new version of Jinja2, which we will be able to better template analysis on.

Some APIs have changed, and we've moved to vendoring `flask-commonmark` in since the original authors are not maintaining it anymore.

**How to test**

Everything should work the same.


rix0rrr:
> What is this Rico? Do you have any plans for that?

Perhaps. No guarantees yet 😅

jpelay (approved):

PR #5326, opened 2024-03-27
💻 format all old adventure contents with curlies

hasan-sh:
Fixes #5310 

**How to test?**
- have an existing adventure with curly braces
- try to see this adventure from different locations; /hedy, /public-adventures, etc.

PR #5350, opened 2024-04-01
🪲 Fixes adventure translation

### LINKED ISSUE #5328: 🪲 Translating keywords in teacher adventures causes the program to be translated in the database
If I log in as a student and change the language of the keywords, aftewards the adventure is always shown in the translated language in the UI

jpelay:
This PR actually solves two problems: first, that it wasn't possible for students to translate the keywords in the teacher adventures, due to the improper field being used to fetch the content of the database, and second, the local database was changed when translating the adventures.

This second issue was what caused this problem to be so hard to debug, since I didn't know if my fix actually worked, because the fields in the database were being changed. The actual problem here was that we were modifying a map from the database, and as @rix0rrr suggested, it was necessary to deep copy it so the fields weren't modified on our local copy.

Fixes #5328

**How to test**

* Create a new adventure, or save again `adventure1` logged in as teacher1
* Log in as student1 and select a keyword language other than English
* Check that the adventure can be translated back and forth and that the local database is not modifed by this.

### DIFF
--- a/app.py
+++ b/app.py
@@ -236,8 +236,12 @@ def load_customized_adventures(level, customizations, into_adventures):
 
     adventure_ids = {a['name'] for a in order_for_this_level if a['from_teacher']}
     teacher_adventure_map = DATABASE.batch_get_adventures(adventure_ids)
-    builtin_adventure_map = {a.short_name: a for a in into_adventures}
+    # Make a deepcopy if working locally, otherwise the local database values
+    # are by-reference and overwritten
+    if not os.getenv('NO_DEBUG_MODE'):
+        teacher_adventure_map = copy.deepcopy(teacher_adventure_map)
 
+    builtin_adventure_map = {a.short_name: a for a in into_adventures}
     # Replace `into_adventures`
     into_adventures[:] = []
     for a in order_for_this_level:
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -236,8 +236,12 @@ def load_customized_adventures(level, customizations, into_adventures):
 
     adventure_ids = {a['name'] for a in order_for_this_level if a['from_teacher']}
     teacher_adventure_map = DATABASE.batch_get_adventures(adventure_ids)
-    builtin_adventure_map = {a.short_name: a for a in into_adventures}
+    # Make a deepcopy if working locally, otherwise the local database values
+    # are by-reference and overwritten
+    if not os.getenv('NO_DEBUG_MODE'):
+        teacher_adventure_map = copy.deepcopy(teacher_adventure_map)
```
rix0rrr, line 242: It's great that this works, but the next step would be to make sure this but doesn't happen in the Dynamo layer. It should not be returning references to existing objects in the first place.

Otherwise we would need to deploy this fix everywhere that `batch_get_adventures()` is called, which is not ideal

rix0rrr: See here: https://github.com/hedyorg/hedy/pull/5353

jpelay:
Superseded by #5326 and #5353

PR #5353, opened 2024-04-02
🪲 Fix accidental object sharing in Dynamo layer

rix0rrr:
As discovered in #5350 by @jpelay, the Dynamo emulation layer has a bug that can lead to accidental data sharing between different requests, because the DB layer hands out references to long-lived objects. If those get mutated in-place by a reader, those mutations accidentally affect the stored in-memory copy of the database.

How I did this:

- Write the tests first to encode the contract that mutations to retrieved (or inserted) objects shouldn't persist to objects retrieved in other queries.
- Replace `copy.copy()` with `copy.deepcopy()` everywhere 😳 oops!

**How to test**

Follow the steps in #5350.

### DIFF
--- a/tests/test_dynamo.py
+++ b/tests/test_dynamo.py
@@ -123,6 +123,62 @@ class TestDynamoAbstraction(unittest.TestCase, Helpers):
         })
 
 
+    def test_no_memory_sharing_direct(self):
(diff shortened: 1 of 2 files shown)


jpelay (approved):
Ah yeah, this is a better fix than mine 😅 Thanks for the quick fix, and the pointer.

PR #5361, opened 2024-04-03
🪲 Embedded editor broken due to missing adventure name

TiBiBa:
This MR fixes an issue with the embedded editor due to a non-existing adventure object. Adding this check should fix it.

### DIFF
--- a/prefixes/normal.py
+++ b/prefixes/normal.py
@@ -48,6 +48,9 @@ def int(s):
 
 
 def convert_numerals(alphabet, number):
+    if number is None or number == '':
+        return ''
+
     numerals_dict_return = {
         'Latin': ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'],
         'Brahmi': ['𑁦', '𑁧', '𑁨', '𑁩', '𑁪', '𑁫', '𑁬', '𑁭', '𑁮', '𑁯'],
--- a/static/css/generated.css.min
+++ b/static/css/generated.css.min
@@ -1,0 +1,3 @@
+Need to install the following packages:
+minify@10.5.2
+Ok to proceed? (y) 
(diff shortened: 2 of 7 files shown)


Felienne:
HI @TiBiBa! I guess this needs to be deployed quickly? 😬 If @jpelay approves I will deploy!

TiBiBa:
> HI @TiBiBa! I guess this needs to be deployed quickly? 😬 If @jpelay approves I will deploy!

Yes, if we could deploy this today that would awesome! Thanks for the quick response, I just tested the fix and it seems to solve the issue. Will prioritize the embedded view testing to prevent this from happening again in the future.

Felienne:
Deploying now! And yes adding tests will make your own future life less stressful :D

TiBiBa:
> Deploying now! And yes adding tests will make your own future life less stressful :D

Unfortunately it is still not working, is there someway I can verify what code is running on production? The same error still seems to exist.

jpelay (approved):
Thanks for fixing it so quickly!

jpelay:
@Felienne I think we need to redeploy? Now that this is merged to main

TiBiBa:
> @Felienne I think we need to redeploy? Now that this is merged to main

Yes I think so too! @jpelay can you deploy or can only @Felienne do this?

jpelay:
> Yes I think so too! @jpelay can you deploy or can only @Felienne do this?

No, she's the only one who can deploy.

Felienne:
> > Deploying now! And yes adding tests will make your own future life less stressful :D
> 
> Unfortunately it is still not working, is there someway I can verify what code is running on production? The same error still seems to exist. **Edit:** The last commit is 1acc76. Should this be merged first or can you deploy this PR as well?

Sorry, my bad! Your merge just missed the deploy train because the tests were still running. Now it should be there, can you check again?

PR #5363, opened 2024-04-03
🪲 fix preview adventure

### LINKED ISSUE #5360: 🪲 preview adventure isn't always working!
**Describe the bug**
To reproduce the problem, create an adventure that's not for level 1. Now if you preview it now, you'll get a 404! 

Don't forget to add tests for both cases; i.e., when the adventure is used in level 1 or any other levels!

hasan-sh:
Fixes #5360

**How to test?**
- try previewing adventure1 and adventure3 with the teacher1 account
- you could also create adventures with different levels and test that too!


tests/cypress/e2e/for-teacher_page/general/preview-adventure.cy.js
```diff
@@ -0,0 +1,22 @@
+import {loginForTeacher} from '../../tools/login/login.js'
+
+describe('Is able to preview adventures', () => {
```
Felienne, line 3: Yes, let's add tests when stuff is broken!!

boryanagoncharenko (approved):
Works smoothly!

PR #5364, opened 2024-04-03
🧹 Remove ace death code

### LINKED ISSUE #4941: Remove Ace death code
After merging #4924 we know don't use Ace for any editor on Hedy, meaning that we won't be maintaining it going forward. Therefore, we're going to remove it, which entails removing:

* ace-editor.ts
* ace related styles
* ace plug-ins
* highlight files related to ace

We think it's still important to maintain the editor abstraction, so even if we remove ace, it doesn't mean we will be solely dependent on CodeMirror (although I do hope we don't have to change it again in the future!).

Annelein:
Fixes #4941


jpelay (approved):
Thanks!

PR #5365, opened 2024-04-04
🧪 Profile live statistics page

jpelay:
Adds a sort of profiling to this page by placing some timers in some key function where I think the slowdown happens. Why is it so hacky? I tried for several hours to use werkzeug's `ProfilerMiddleware` but I think there's a bug with the implementation, as several people have reported in some other tools: https://github.com/jazzband/django-debug-toolbar/issues/1875, https://github.com/jazzband/django-silk/issues/682

I thought that since we didn't need the full-fledged tool and didn't want to spend more time on this, this approach would suffice.

### DIFF
--- a/website/statistics.py
+++ b/website/statistics.py
@@ -14,10 +14,17 @@ import jinja_partials
 from website.flask_helpers import render_template
 from website import querylog
 from website.auth import is_admin, is_teacher, requires_admin, requires_login
-
+from timeit import default_timer as timer
 from .database import Database
 from .website_module import WebsiteModule, route
 from bs4 import BeautifulSoup
+
+import logging
+from logging_config import LOGGING_CONFIG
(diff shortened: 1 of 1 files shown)


Felienne (approved):
Thanks for diving in!! Do we want to make a note somewhere (an issue?) so that we don't forget to remove this code when we are done?

jpelay:
> Thanks for diving in!! Do we want to make a note somewhere (an issue?) so that we don't forget to remove this code when we are done?

I will remove them when I address the slowness so Ill  be touching this code very soon!

Felienne:
> > Thanks for diving in!! Do we want to make a note somewhere (an issue?) so that we don't forget to remove this code when we are done?
> 
> I will remove them when I address the slowness so Ill be touching this code very soon!

Yes, the profiling is deployed now so you can dive in when you want.

PR #5377, opened 2024-04-04
🚚 Re-enable pylint

rix0rrr:
Pylint has been disabled since the migration to Python 3.12. It's valuable to have as a backstop for simple programming mistakes, so this PR tries to re-enable it.

**How to test**

If the build passes, we did it!


jpelay:
> If the build passes, we did it!

I see that the build passed! Does that mean we can merge this?

jpelay (approved):

PR #5385, opened 2024-04-08
🪲 duplicate adventures

### LINKED ISSUE #5375: 🪲 duplicate adventures
**Describe the bug**
Something is off when you want to create an adventure. Currently we append an X to the adventure's name to avoid duplicates. This doesn't seem to work. 

How to reproduce? Create an adventure, you'd see it has the same name as other adventures. If not, create another adventure and you'd get a duplicate error if you try to save. 

This is very problemetic since the content of adventures would dissapear from the code page due to duplicate adventures.

[Screencast from 08-04-24 10:52:17.webm](https://github.com/hedyorg/hedy/assets/20051470/5dbbbcbb-43b7-4d0a-bc70-a505e8fd5026)

hasan-sh:
Fixes #5375

This PR changes how adventures are created. We don't need to "store" the adventures unless the teacher appllies some changes. This means whenever we redirect the teacher to the customize-adventure page, we simply initialize an adventrue, save in the current session, and if they apply a change, then we basically save it in our DB. 

Not only does this prevent the issue I'm fixing, but also reduce the number of trial adventures created by teachers. 

Moreover, the only downside is that teachers would have to write out the name themselves, however, this may not be a min-point afterall since giving default names is kinda encouraging creating more, and more will be created just because it becomes easy to do so!

**How to test?**
Create an adventure from the 
- for-teachers page
- class-overview page
- customize-class page


Felienne (approved):
Yes, works, thanks!

I am wondering (maybe for another issue?) if we can make all warnings look the same? Now an empty title is a lot more subtle that adding an adventure to 2 levels:

image

Maybe we should make the empty title warning in the same place?

PR #5389, opened 2024-04-09
🧪 🪲 Fix flaky public program test

jpelay:
Adds an intercept for the public profile call, which I think is the cause behind the flakiness.


Felienne (approved):
Hope this does the job!

PR #5391, opened 2024-04-10
🖊️ Add error for else statement without if statement #5099

### LINKED ISSUE #5099: 🪲 in level 5, error message should be improved
**Describe the bug**
a student handed in a program that has the following if statement:

```
if antwoord in hehe print 'nou.. ga maar dan'
sleep 2
print 'en wees voorzichtig..'
else print 'WEGWEZEN!'
```
The following is the error message:
image

boryanagoncharenko:
Fixes #5099
Add a custom error in levels 5-7 when there is an `else` statement when there is no `if` right before it.

**How to test**
Run the following snippets in levels 5-7 and check that the custom error is appropriate:
```
if answer is yes print 'great!'
print 'correct'
else print 'wrong'
```
```
else print 'wrong'
```


jpelay (approved):
Works!

Issue #4027, opened 2023-02-10
[TESTING] Refactor cypress tests


jpelay:
As stated by @rix0rrr in #3972 there are two ways in which our tests can be improved:

1. Test organization: Instead of having `define`s with a single `it` we can group different tests under a single `define`. Saving code and making our test suite more organized.
2.  Selecting HTML elements by ID: Cypress actually prevents against using `id` or other HTML attributes used in functionality for the tests.

Update: I will change tests bit by bit
- [x] admin page
- [x] explore program page
- [x] feedback
-> #5393 
- [x] for-teacher page
    - [x] class page -> https://github.com/hedyorg/hedy/pull/5530
    - [x] customize class 
    - [x] customize adventure -> https://github.com/hedyorg/hedy/pull/5683
    - [x] general -> #5592 
- [x] hedy page -> https://github.com/hedyorg/hedy/pull/5648
- [x] home page
- [x] login page
- [x] recover page 
- [x] signup page
- [x] subscribe page
-> #5647
- [x] lezer test
- [x] public adventures
- [x] public programs
- [x] teacher page
- [x] tools
-> https://github.com/hedyorg/hedy/pull/5683

Felienne:
One more task for @Annelein?

Felienne:
All done!

PR #5393, opened 2024-04-10
🧹 Refactor cypress tests (part 1)

Annelein:
Fixes the following tests for #4027 :

- [x] admin page
- [x] explore program page
- [x] feedback


Annelein:
@jpelay is this what it should look like for example? Then I can start going through all tests and implement these changes!

Annelein:
@jpelay @Felienne would it make sense to separate this issue into a few PR's? Otherwise it will be huge and then once in a while I can just change a few tests again.

jpelay:
>@jpelay @Felienne would it make sense to separate this issue into a few PR's? Otherwise it will be huge and then once in a while I can just change a few tests again.

Sure!

Annelein:
@jpelay then it's ready to be reviewed :)) Think it would be great if you could do it asap to avoid conflicts because it has changes in a lot of different places

Felienne (approved):
Looks great, thanks!!

PR #5394, opened 2024-04-10
🧹 Replace 403 by 401 in some occurences 

### LINKED ISSUE #3634: [CLEAN-UP] Replace 403 error (forbidden) by 401 (unauthorized)

Annelein:
Fixes #3634 

**How to test**
Not really possible I think, but all unit tests should pass of course.

### DIFF
--- a/website/auth_pages.py
+++ b/website/auth_pages.py
@@ -389,6 +389,7 @@ class AuthModule(WebsiteModule):
         token = self.db.get_token(body["token"])
         if not token or body["token"] != token.get("id") or body["username"] != token.get("username"):
             return gettext("token_invalid"), 403
+            # is this a 401?
 
         hashed = password_hash(body["password"], make_salt())
         self.db.update_user(body["username"], {"password": hashed})
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -130,7 +130,7 @@ class ForTeachersModule(WebsiteModule):
     @requires_login
     def get_class(self, user, class_id):
         if not is_teacher(user) and not is_admin(user):
-            return utils.error_page(error=403, ui_message=gettext("retrieve_class_error"))
+            return utils.error_page(error=401, ui_message=gettext("retrieve_class_error"))
         Class = self.db.get_class(class_id)
         if not Class or (not utils.can_edit_class(user, Class) and not is_admin(user)):
             return utils.error_page(error=404, ui_message=gettext("no_such_class"))
@@ -217,7 +217,7 @@ class ForTeachersModule(WebsiteModule):
     @requires_login
     def preview_class_as_teacher(self, user, class_id):
         if not is_teacher(user) and not is_admin(user):
-            return utils.error_page(error=403, ui_message=gettext("retrieve_class_error"))
+            return utils.error_page(error=401, ui_message=gettext("retrieve_class_error"))
         Class = self.db.get_class(class_id)
         if not Class or (not utils.can_edit_class(user, Class) and not is_admin(user)):
             return utils.error_page(error=404, ui_message=gettext("no_such_class"))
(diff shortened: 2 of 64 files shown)


Annelein:
@Felienne 
- for some 403's I wasn't sure so I commented on them, can you let me know if these should be 401's?
- some errors that I changed to 401's, have different error messages then, "unauthorized", for example: should these error messages be changed to "unauthorized"?

Felienne:
> @Felienne
> 
> * I wasn't sure about some 403's, so I commented on them, can you let me know if these should be 401's?
> * some errors that I changed to 401's, have different error messages then, "unauthorized", for example all occurrences in statistics.py. Should these error messages be changed to "unauthorized"?
> * I will of course fix the unittests after I know for sure that these are the correct error responses!

Pinging @TiBiBa (or otherwise maybe @rix0rrr or @jpelay here) I did not make this issue and I am not really an expert in this actually :)

website/auth_pages.py
```diff
@@ -389,6 +389,7 @@ def reset(self):
         token = self.db.get_token(body["token"])
         if not token or body["token"] != token.get("id") or body["username"] != token.get("username"):
```
jpelay, line 390: Yes I think this is a 401

website/for_teachers.py
```diff
@@ -920,6 +920,7 @@ def create_accounts(self, user, class_id):
         Class = self.db.get_class(class_id)
         if not Class or (not utils.can_edit_class(user, Class) and not is_admin(user)):
             return utils.error_page(error=403, ui_message=gettext("no_such_class"))
+        # should this be split up in 403 and 401?
```
jpelay, line 923: I think this actually a case of a 404 and 401. Because there's a possibility that there isn't a class, and also that this person can't edit this class.

website/for_teachers.py
```diff
@@ -1044,6 +1045,7 @@ def get_adventure_info(self, user, adventure_id,):
 
         if not adventure or adventure["creator"] != user["username"] and not is_teacher(user):
             return utils.error_page(error=403, ui_message=gettext("retrieve_adventure_error"))
+        # should this be split up in 403 and 401?
```
jpelay, line 1048: Same case here, this being a 404 and a 401

website/statistics.py
```diff
@@ -575,7 +575,8 @@ def refresh_live_stats(self, user, class_id):
         if student:
             class_students = class_.get("students", [])
             if student not in class_students:
-                return utils.error_page(error=403, ui_message=gettext('not_enrolled'))
+                return utils.error_page(error=401, ui_message=gettext('not_enrolled'))
```
jpelay, line 578: I think this one might remain as a 401 or maybe become a 404?

jpelay (commented):
Hi @Annelein I commented on each case that I saw a comment of yours, hope it helped!

jpelay:
> some errors that I changed to 401's, have different error messages then, "unauthorized", for example all occurrences in statistics.py. Should these error messages be changed to "unauthorized"?

I think that it's fine to have some more descriptive error messages than just a generic one.

PR #5397, opened 2024-04-11
🚚 Prevent Weblate Update workflow from overwriting existing PR

rix0rrr:
If an open Weblate/translations PR still exists, we shouldn't tell Weblate to push again.  Because we ended our previous push with a 'reset', Weblate would otherwise force-push new changes over the unmerged old changes, causing them to lost.

**How to test**

After merging:

* Reopen an old Weblate PR
* Run the "update weblate" workflow on-demand
* Observe that it errors out and doesn't overwrite the old PR with new changes.

### DIFF
--- a/.github/workflows/update-weblate.yml
+++ b/.github/workflows/update-weblate.yml
@@ -18,7 +18,8 @@ jobs:
     # new changes over the unmerged old changes, causing them to lost.
     - name: Check for existence of old Weblate PR
       run: |
-        if [[ "$(gh pr -R hedyorg/hedy list --search 'Translations update' --json number)" != "[]" ]]; then
+        prs=$(gh pr -R hedyorg/hedy list -q '.[] | select(.title | contains("Translations update"))' --json title)
+        if [[ "$prs" != "" ]]; then
           echo "Open Pull Request for Weblate!"
           gh pr -R hedyorg/hedy list --search 'Translations update'
           exit 1
--- a/.github/workflows/update-weblate.yml
+++ b/.github/workflows/update-weblate.yml
@@ -13,6 +13,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    # If an open Weblate/translations PR still exists, we shouldn't tell Weblate to push again.
+    # Because we ended our previous push with a 'reset', Weblate would otherwise force-push
+    # new changes over the unmerged old changes, causing them to lost.
+    - name: Check for existence of old Weblate PR
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-weblate.yml
```diff
@@ -13,6 +13,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    # If an open Weblate/translations PR still exists, we shouldn't tell Weblate to push again.
+    # Because we ended our previous push with a 'reset', Weblate would otherwise force-push
+    # new changes over the unmerged old changes, causing them to lost.
+    - name: Check for existence of old Weblate PR
+      run: |
+        if [[ "$(gh pr -R hedyorg/hedy list --search 'Translations update' --json number)" != "[]" ]]; then
```
boryanagoncharenko, line 21: Searching for 'Translations update' returns the current PR, which I think is not the indent.
Screenshot 2024-04-11 at 14 39 08

The issue is addressed if I add double quotes in the search parameter: `'"Translations update"'`

boryanagoncharenko (changes requested):

boryanagoncharenko:
We prolly also want to update the docs here: https://github.com/hedyorg/hedy/wiki/Hedy-Translation-Tutorial#fixing-weblate-merge-snippet-errors

.github/workflows/update-weblate.yml
```diff
@@ -13,6 +13,17 @@ jobs:
...
     steps:
+    # If an open Weblate/translations PR still exists, we shouldn't tell Weblate to push again.
+    # Because we ended our previous push with a 'reset', Weblate would otherwise force-push
+    # new changes over the unmerged old changes, causing them to lost.
+    - name: Check for existence of old Weblate PR
+      run: |
+        prs=$(gh pr -R hedyorg/hedy list -q '.[] | select(.title | contains("Translations update"))' --json title)
+        if [[ "$prs" != "" ]]; then
+          echo "Open Pull Request for Weblate!"
+          gh pr -R hedyorg/hedy list --search 'Translations update'
```
boryanagoncharenko, line 24: Would you like to change the filter here too? This filter is looser than the upper one, so it might lead to confusions if one is looking at the error output of the action.

rix0rrr: Mmmyeah. But the unfiltered one looks way nicer than output JSON... 🥺

boryanagoncharenko (approved):

Issue #5399, opened 2024-04-11
🪲 Programs get flagged as 'too long' too quickly


MarleenGilsing:
Programs with sleep commands in it will very quickly get the error 'it takes too long to load'. 
I think we built this error ourselves, to prevent students from using 'repeat 100000 times print x', but in the higher levels it shuts down perfectly fine programs. 

For example, the simon says code in level 16: 
```
colors = ['red', 'blue', 'green', 'yellow']
simon_sequence = ['a','b']
remove 'a' from simon_sequence
remove 'b' from simon_sequence

define add_random_color
    random_color is colors[random]
    add random_color to simon_sequence

define display_sequence with level
    for i in range 1 to level
        print simon_sequence[i]
        sleep 1
        clear

define get_player_sequence with level
    for i in range 1 to level
        color = ask 'what is color number ' i '?'
        add color to player_sequence

level = 1
game_over = 'False'
print 'Welcome to Simon Says!'
sleep 1
clear
while game_over == 'False'
    print 'Level ' level
    sleep 1
    clear
    player_sequence = ['a','b']
    remove 'a' from player_sequence
    remove 'b' from player_sequence
    call add_random_color
    call display_sequence with level
    call get_player_sequence with level
    if player_sequence == simon_sequence
        print "Correct! Moving on to the next level."
        sleep 1
        level = level + 1
    else
        print 'Incorrect! Game Over.'
        game_over = 'True'
```

You can only play this game for 1 level, because in level 2 you'll get the error message. Can we fix this?

Felienne:
Ah yes a teacher on Discord was also complaining about this (and rightly so!). I think we should just up the limit to a few minutes (2 or 3 minutes?) so that at one point programs are stopped, but not too early. 

With a bit more work, we can also make it program dependent but I don't feel that is needed per se.

Annelein:
@Felienne were you picking it up or am I?

Felienne:
> @Felienne were you picking it up or am I?

If you can do it, that would be nice!

PR #5401, opened 2024-04-11
 📜 Simon says adventure

MarleenGilsing:
It's a Simon Says adventure, with thanks to @AnneliesVlaar!

It works, but after level 2 you'll get an error that the code is too long: Issue #5399


Felienne (approved):

Felienne:
The too long issue is not closed yet but let's merge this to prevent conflicts!

PR #5413, opened 2024-04-15
💻 Show turtle with ask modal

### LINKED ISSUE #5396: 🪲 ask with turtle hides turtle screen
**Describe the bug**
If you have code that creates turtle output but also an ask, the turtle output disappears when asking:

image

Code:
```
repeat 10 times
    forward 100
    hoek is ask 'Hoe ver moet ik draaien?'
    repeat 5 times
        forward 10
        turn hoek
```

Annelein:
Fixes #5396 

**How to test**
1. Check this code:
```
repeat 10 times
    forward 100
    hoek is ask 'Hoe ver moet ik draaien?'
    repeat 5 times
        forward 10
        turn hoek
```
2. Check if other ask modals and turtle output still looks as it should.

https://github.com/hedyorg/hedy/assets/48122190/eaef727c-b280-4bba-993b-d3f368f45979

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-static/js/appbundle.js
-static/js/appbundle.js.map
-
 # C extensions
 *.so
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+static/js/appbundle.js
+static/js/appbundle.js.map
+
 # C extensions
 *.so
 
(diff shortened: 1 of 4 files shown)


boryanagoncharenko (commented):
When the ask modal shows, it has the focus. So if I start typing directly and then hit enter, everything works as expected. However, if I get the focus out of the ask modal, I cannot get it back in. So, when the ask modal opens, if I click anywhere on the screen, and then click again on the ask modal (or its Enter button), the modal never becomes active and I cannot type in it.

Annelein:
@boryanagoncharenko good catch thank you! Pushed a fix

boryanagoncharenko (approved):

PR #5414, opened 2024-04-15
🪲 Add support for numbers in return statements

### LINKED ISSUE #5170: Unsupported operand type(s) for +📜
In level 18 I run below program and it reacts with:

afbeelding

Why is this? It concerns the line: midden= laag+hoog

As far as I know laag and hoog are numbers.

---

def round(x):

  y= 0

  while x> 0:
    x= x-1
    y= y+1

  return y

laag= 1
hoog= 10

print('Neem een getal van 1...', hoog, ' in gedachten...')

while laag!= hoog:
  midden= laag+hoog
  midden= midden/2

  midden= round(midden)

  print('Is je getal lager (0) of hoger of gelijk aan (1) dan ', midden, '?')
  a= input('0/1: ')

  if a!= 0 en a!= 1:
    print('Ongeldige keuze!')
  else:
    if a= 0:
      hoog= midden

    if a= 1:
      laag= midden

print('Ik denk dat je getal ', hoog, ' is')

boryanagoncharenko:
Fixes #5170
With this PR, `return` statements try to convert their output to integers and floats and output strings if the number conversion fails. This change is required because the output of a function cannot be used in an arithmetic operation.

**How to test**
Navigate to level 14 and run the following program. Ensure that it completes without an error:
```
define round with x
    return x

r = call round with 1
print r + 1
```


jpelay (approved):

Jsgn3:
I tried the example above at level 14. It ended with this error:

--
Er ging iets fout bij het uitvoeren van het programma.

While running your program the command + received the values 1 and 1 which are not allowed. Try changing the values to be all text or all numbers.
--

boryanagoncharenko:
> I tried the example above at level 14. It ended with this error:
> 
> ... Er ging iets fout bij het uitvoeren van het programma.
> 
> While running your program the command + received the values 1 and 1 which are not allowed. Try changing the values to be all text or all numbers. ...
> 
> I ran Hedy at: https://www.hedycode.com/hedy/14#default

Thanks for looking into it!

Yes, the program will still fail when you try it at [https://www.hedycode.com](https://www.hedycode.com) and this is normal. The reason is that the code in this pull request has been merged to the main branch but has not been deployed to [https://www.hedycode.com](https://www.hedycode.com) yet. If you want to run this new version, you would have to run Hedy locally (we have great [docs](https://github.com/hedyorg/hedy/wiki/Hedy-Development-Process#running-hedy-to-make-changes)). Otherwise, you could wait a few days and the fix will become available at [https://www.hedycode.com](https://www.hedycode.com).

PR #5421, opened 2024-04-16
🧪 Test a few broken programs from the front-end

### LINKED ISSUE #4405: [TESTING] Test a few broken programs from the front-end
See #4403 

Can we add a few Cypress tests where we verify a broken program leads to the correct *front-end* error message/warning?

Annelein:
Fixes #4405 

**How to test**
Cypress tests should pass


tests/cypress/e2e/hedy_page/invalid_program.cy.js
```diff
@@ -0,0 +1,58 @@
...
+        const error_code = "prnt Hello world!"
+        const error_code2 = "print Hello world!\nak Hello world?"
+        const error_message = `prnt is not a Hedy level 1 command. Did you mean print?`;
+        const error_message2 = `ak is not a Hedy level 1 command. Did you mean ask?`;
+        goToHedyPage();
+
+        cy.get('#editor > .cm-editor > .cm-scroller > .cm-content').click();
+        cy.focused().type(error_code);
+
+        cy.get('#runit').click();
```
jpelay, line 14: Since the `#runit` button triggers a response from the server, I think it'd be wise to wrap it within an intercept, like we already do with the `#debug` button here: https://github.com/hedyorg/hedy/blob/179c72391ab0ddfa52cd6775fba9b9570d262155/tests/cypress/e2e/hedy_page/debugger.cy.js
It'd be something like this:

```suggestion
        cy.intercept('/parse').as('parse')
        cy.get('#runit').click();
        cy.wait('@parse')
```

What this will do is to wait for the server to respond in order to continue with the test. In the past we've had some flaky test due to not waiting properly.

tests/cypress/e2e/hedy_page/invalid_program.cy.js
```diff
@@ -0,0 +1,58 @@
+import {goToHedyLevel5Page, goToHedyPage} from "../tools/navigation/nav";
+
+describe('Error code gives correct error', () => {
+    it('Misspelled Keyword', () => {
+        const error_code = "prnt Hello world!"
+        const error_code2 = "print Hello world!\nak Hello world?"
```
jpelay, line 6: I think perhaps these could be 2 different tests? You can change the `it` for a describe and then have two it within that one. We do something like that in the lezer tests.

jpelay (changes requested):
Hi annelein, good work! I think this will be enough to test the error-box modal. I suggested a few changes with respect to waiting for the server to parse the errors.

Annelein:
@jpelay thanks! just added your suggestions

PR #5431, opened 2024-04-17
new approach in public-adventures

### LINKED ISSUE #5367: 🪲 Public adventures 'no adventures available'
**Describe the bug**
I sometimes get the message that there are no public adventures available on this level, while there are. When I get that message, I can't select another level and I'm stuck on this page. 

image

If you reload the page, everything is fine, but I had this happen a couple of times now. 

Windows, chrome

### LINKED ISSUE #5368: 💻Public adventures, select the language automatically?
**Idea incl level**
On the public adventures page you can choose the language, but it might be nice to set the language automatically to the preferred language the user has chosen in their profile. Obviously, not a very major point, but once we start getting more and more adventures, it might look neater to have only your own language presented to you at first. We can still keep the language button in case teachers want to check out other languages.

hasan-sh:
Fixes #5368 
Fixes #5367

This PR changes the functionality of how we retrieve public adventures. It also encapsulates changes related to creating indexes and filters. As a result, we create two additional tables: `public-adventure-filters` and `public-adventure-indexes`. 

The first table creates a combination of filters, creating field and value in each record. The field can, for instnace, be `lang_level` and the value `en_1`. These filters are updated regularly once public adventures are created. 

The second table creates the indexes so that we get the adventure_ids of all public adventures based on what users filter by. The two main fields are `field_value` and `date_adventure_id`. As you can expect, the `field_value` would hold the info that we retrieve from the filters table. For example, `level_lang_2_ar` where `level_lang` is the field and `2_ar` are the values, respectively. Then we can get the adventure ids of these adventures and list them to the user. 

Thereby, we can paginate the adventures and have much fewer problems with the overload and download restriction of AWS.

> NB: We need to run the scripts in the `migration_scripts` on alpha and beta.


website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
+
+        available_languages = set()
+        available_levels = set()
+        available_tags = set()
+
+        # Get all possible filters
+        lang_level_filters = self.db.get_public_adventures_filters({"field": "lang#level", })
+        for record in lang_level_filters:
+            lang, _level = record.get("value", "#").split("#")
+            available_levels.update([_level])
```
rix0rrr, line 52: You can do `.add()` if you want to add 1 element

website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
+            if level == _level:
+                available_languages.update([lang])
+        customizations = {"available_levels": available_levels}
+
+        tag_filters = self.db.get_public_adventures_filters({"field": "tag"})
```
rix0rrr, line 57: You could have gotten all of these in one go using a table scan. Saving roundtrips is a good idea.

website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
+            tags = [t for t in tags if t]
+
+        # Get indexes
+        level_adventure_ids = self.db.get_public_adventures_indexes({"field#value": f"level#{level}"})
+        lang_adventure_ids = self.db.get_public_adventures_indexes({"field#value": f"lang#{language}"})
+        tag_adventure_ids = []
+        for _t in tags:
+            tag_adventure_ids += self.db.get_public_adventures_indexes({"field#value": f"tag#{_t}"})
+
+        level_adventure_ids = [record["date#adventure_id"].split("#")[1] for record in level_adventure_ids]
```
rix0rrr, line 81: Similar here, I would leave the bashing of those records into a usable shape inside the `get_public_adventures_indexes` function.

website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
+            if toReset:
+                # then it's the current selected tags, so remove current's tag
+                tags = [_tag for _tag in toReset.split(",") if _tag != tag]
+            else:
+                tags = tag.split(",")
+            tags = [t for t in tags if t]
+
+        # Get indexes
+        level_adventure_ids = self.db.get_public_adventures_indexes({"field#value": f"level#{level}"})
+        lang_adventure_ids = self.db.get_public_adventures_indexes({"field#value": f"lang#{language}"})
```
rix0rrr, line 76: I would also do a combined index of `language + level`, so that you can filter on 2 conditions at the same time. 

The problem is that if you do it by combining 3 different indexes like this, you won't be able to paginate properly.

website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
+        for _t in tags:
+            tag_adventure_ids += self.db.get_public_adventures_indexes({"field#value": f"tag#{_t}"})
+
+        level_adventure_ids = [record["date#adventure_id"].split("#")[1] for record in level_adventure_ids]
+        lang_adventure_ids = [record["date#adventure_id"].split("#")[1] for record in lang_adventure_ids]
+        tag_adventure_ids = [record["date#adventure_id"].split("#")[1] for record in tag_adventure_ids]
+
+        # Filter out based on level
+        lang_adventure_ids = [_id for _id in lang_adventure_ids if _id in level_adventure_ids]
+        tag_adventure_ids = [_id for _id in tag_adventure_ids if _id in level_adventure_ids]
```
rix0rrr, line 87: If you make them sets of IDs, you can use set intersection. And if you use set intersection before splitting the date and the ID, you can afterwards still sort the final list (descending) to come up with a list with the newest program first.

But again, you won't be able to paginate if you do that. How big is your data?

website/public_adventures.py
```diff
@@ -31,67 +31,168 @@ def __init__(self, db: Database, achievements: Achievements):
...
-
     @route("/", methods=["GET"])
     @route("/filter", methods=["POST"])
     @requires_teacher
-    def filtering(self, user, index_page=False):
+    def filtering(self, user):
+        level = request.args["level"] if request.args.get("level") else "1"
+        language = request.args.get("lang", g.lang)
+        tag = request.args.get("tag", "")
+        search = request.form.get("search", request.args.get("search", ""))
```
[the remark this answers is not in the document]
hasan-sh: With GET indeed, but they can also be used in the POST verb as well!

website/database.py
```diff
@@ -608,6 +611,13 @@ def batch_get_adventures(self, adventure_ids):
         keys = {id: {"id": id} for id in adventure_ids}
         return ADVENTURES.batch_get(keys) if keys else {}
 
+    def get_public_adventures_filters(self, key):
+        return PUBLIC_ADVENTURES_FILTERS.get_all(key)
```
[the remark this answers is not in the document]
hasan-sh: The levels should never be filtered out alone, but with combination with the language/tags. I do think if you select level 1, you'd need to see languages/tags for that level, right? 

What you're saying is that based on the selected language, we show available levels/tags?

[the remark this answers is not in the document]
hasan-sh: Also, about scanning the table. This will have to be triggered whenever the user refreshes the page or applies some filters. I don't think it'd be heavey, but let's see its performance once it's done!

website/public_adventures.py
hasan-sh: How would you combine them?

rix0rrr: I was thinking `{ "field#value": "level_lang#5_es", "date#adventure_id": "12345#abcd" }`, so you can search for a combination of level + language in one query.

That's easier and quicker than querying 2 sets and intersecting them, plus you can paginate through a single query (and you can't paginate very well through multiple queries)

website/database.py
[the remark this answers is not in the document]
rix0rrr: I know, scanning a table is usually bad. But in this case, the table isn't too big I don't think.

> I do think if you select level 1, you'd need to see languages/tags for that level, right?

There's no arbitrary navigation hierarchy here though; there is a definite order. No real user is going to pick the level first, and only then see what language they can choose. They will begin with picking a language, because it will be *their* language.

It does make sense to filter the list of levels down when a language is picked, and perhaps the list of tags as well.

But I don't think it needs to be more complicated than that, and selecting language, then level, and having the list of tags filtered down to only that combination of language and level is possible, but overkill. So let's not design for it yet.

website/database.py
```diff
@@ -617,9 +666,11 @@ def delete_adventure(self, adventure_id):
     def store_adventure(self, adventure):
         """Store an adventure."""
         ADVENTURES.create(adventure)
+        self.update_public_adventure_filters_indexes(adventure)
```
rix0rrr, line 669: Only if and when we make the adventure public, right?

website/database.py
```diff
@@ -617,9 +666,11 @@ def delete_adventure(self, adventure_id):
...
 
     def update_adventure(self, adventure_id, adventure):
         ADVENTURES.update({"id": adventure_id}, adventure)
+        self.update_public_adventure_filters_indexes(adventure)
```
rix0rrr, line 673: Same here.

website/database.py
```diff
@@ -608,6 +611,52 @@ def batch_get_adventures(self, adventure_ids):
...
+            PUBLIC_ADVENTURES_INDEXES.put({"field_value": f"level_{level}", "date_adventure_id": f"{date}_{id}"})
+            value = f"{lang_value}#{level}"
+            PUBLIC_ADVENTURES_FILTERS.put({"field": "lang#level", "value": value})
+        tags = adventure.get("tags", [])
+        for tag in tags:
+            PUBLIC_ADVENTURES_INDEXES.put({"field_value": f"tag#{tag}", "date_adventure_id": f"{date}_{id}"})
+            PUBLIC_ADVENTURES_FILTERS.put({"field": "tag", "value": tag})
+
+    def remove_public_adventure_filters_indexes(self, field, value, adventure):
+        PUBLIC_ADVENTURES_FILTERS.delete({"field": field, "value": value})
```
rix0rrr, line 650: The way we're storing information, you can't ever delete from `public_adventure_filters`.

Think of the following sequence of events:

* Make an adventure in Spanish public → `filters.put({ "field": "lang", "value": "es" })`  # Adds a new row
* Make another adventure in Spanish public → `filters.put({ "field": "lang", "value": "es" })`  # Replaces the row
* Make one of the Spanish adventures private → `filters.delete({ "field": "lang", "value": "es" })`  # Deletes the row

Oops! We now no longer know that there are any Spanish adventures!

rix0rrr: I also feel like `remove_public_daventure_filters_indexes` should be private, and be called from this class and this class alone. That way, callers don't have to be aware of this (and also can't forget!)

```
_remove_public_adventure_filters_indexes
```

website/public_adventures.py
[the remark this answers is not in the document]
rix0rrr: But a search operation should never be a POST? 

- GET should be used for read-only operations
- POST should be used for mutating operations

Search is a read-only operation.

website/public_adventures.py
```diff
@@ -31,67 +31,166 @@ def __init__(self, db: Database, achievements: Achievements):
...
+    @requires_teacher
+    def filtering_old(self, user, index_page=False):
         index_page = request.method == "GET"
 
         level = int(request.args["level"]) if request.args.get("level") else 1
         language = request.args.get("lang", "")
         tag = request.args.get("tag", "")
         search = request.form.get("search", request.args.get("search", ""))
         if index_page or not self.adventures or not self.adventures.get(level):
+            print("\n\n\n GET IT MORE TIME!!\n\n", level, language, tag)
```
rix0rrr, line 193: Oops?

website/tags.py
```diff
@@ -57,6 +57,8 @@ def create(self, user, adventure_id, new_tag=None):
         if tag_name not in adventure_tags:
             adventure_tags.append(tag_name)
             self.db.update_adventure(adventure_id, {"tags": adventure_tags})
+            if db_adventure.get("public"):
+                self.db.update_public_adventure_filters_indexes(db_adventure)
```
rix0rrr, line 61: Do this inside the database class.

website/tags.py
```diff
@@ -105,5 +107,6 @@ def delete_from_adventure(self, user, tag, adventure_id):
         adventure_tags = db_adventure.get("tags", [])
         adventure_tags = list(filter(lambda name: name != tag_name, adventure_tags))
         self.db.update_adventure(adventure_id, {"tags": adventure_tags})
+        self.db.remove_public_adventure_filters_indexes("tag", tag_name, db_adventure)
```
rix0rrr, line 110: Do this inside the database class. Make a new operation for it if you have to.

website/database.py
rix0rrr: By the way in the indexes, you probably need to remove tags/levels if those arrays got changed.

website/public_adventures.py
```diff
@@ -31,67 +31,166 @@ def __init__(self, db: Database, achievements: Achievements):
...
+        search = request.form.get("search", request.args.get("search", ""))
+
+        available_languages = set()
+        available_levels = set()
+        available_tags = set()
+
+        # Get all possible filters
+        field_filters = self.db.get_public_adventures_filters()
+
+        available_levels.update(field_filters.get("level", []))
```
rix0rrr, line 54: Why start with an empty set and then add elements to it?

```suggestion
        available_levels = set(field_filters.get("level", []))
```

website/public_adventures.py
```diff
@@ -31,67 +31,166 @@ def __init__(self, db: Database, achievements: Achievements):
...
+            if toReset:
+                # then it's the current selected tags, so remove current's tag
+                tags = [_tag for _tag in toReset.split(",") if _tag != tag]
+            else:
+                tags = tag.split(",")
+            tags = [t for t in tags if t]
+
+        # Get indexes
+        level_adventure_ids = self.db.get_public_adventures_indexes({"field_value": f"level_{level}"})
+        lang_adventure_ids = self.db.get_public_adventures_indexes({"field_value": f"lang_{language}"})
```
rix0rrr, line 72: It seems you are ignoring my suggestion to use a combined index for level and language, which would also allow pagination.

Can you put a short comment here for the people after you that you're not doing that, and the reasoning behind it? 

For example, a convincing motivation would look something like:

> Currently at XXX adventures, a search operation typically returns YYY MB. We will only start exceeding the 1MB download window at ZZZ adventures, which will probably take another AAA years. In the case of a combined (language, level) search, we won't be downloading more than BBBB bytes unnecessarily compared to a combined index, which adds at most CCC milliseconds. In comparison, the absence of having to write the combined index saves DDDD milliseconds for every adventure saved.

website/public_adventures.py
```diff
@@ -31,67 +31,166 @@ def __init__(self, db: Database, achievements: Achievements):
...
+        tag_adventure_ids = set()
+        for _t in tags:
+            tag_adventure_ids.update(self.db.get_public_adventures_indexes({"field_value": f"tag_{_t}"}))
+
+        lang_adventure_ids.intersection_update(level_adventure_ids)
+        tag_adventure_ids.intersection_update(level_adventure_ids)
+        adventure_ids = level_adventure_ids.union(lang_adventure_ids, tag_adventure_ids)
```
rix0rrr, line 79: I didn't know `union` takes 3 arguments. What does that mean?

Why are you doing a union here? Aren't we supposed to intersect all 3 filters?

rix0rrr: Also, where are you accounting for the possibilty that the filter can be empty? What if the user didn't pick a language, level, or set of tags?

website/public_adventures.py
rix0rrr: It would be nicer to hide all the details of how to come up with this information in the `DATABASE` class, instead of putting it here. Pass the filter into a method in that class, and let it figure out how to come up with the information you need.

website/database.py
```diff
@@ -608,6 +611,52 @@ def batch_get_adventures(self, adventure_ids):
...
+        PUBLIC_ADVENTURES_INDEXES.delete({"field_value": f"{field}_{value}",
+                                          "date_adventure_id": f"{adventure['date']}_{adventure['id']}"})
+
+    def get_public_adventures_indexes(self, key):
+        """This function returns adventure_ids for a specific index.
+        E.g., key={"field_value": "lang_en", "date_adventure_id": "..."}"""
+        result = PUBLIC_ADVENTURES_INDEXES.get_all(key)
```
rix0rrr, line 657: You probably want query them in reverse, so that the newest adventures are first.

Use the Dynamo feature (`reverse=True`).

create-public-adventures-filters.py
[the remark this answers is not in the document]
hasan-sh: Sounds great, shall i add a dodo command for them?

static/js/public-adventures.ts
```diff
@@ -101,5 +101,5 @@ document.addEventListener("updateTSCode", (e: any) => {
         initialize({lang: js.lang, level: parseInt(js.level), keyword_language: js.lang,
             javascriptPageOptions: js
             });
-    }, 1000);
+    }, 500);
```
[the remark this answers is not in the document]
hasan-sh: Rendering takes some time and if i initialize the js, some parts wouldn't work. The 1000 is just a longer, unneeded duration

website/database.py
```diff
@@ -608,6 +611,52 @@ def batch_get_adventures(self, adventure_ids):
...
+                filters["tag"].append(record.get("value"))
+            else:
+                lang, level = record.get("value").split("#")
+                filters["lang"] = filters.get("lang", [])
+                filters["lang"].append(lang)
+                filters["level"] = filters.get("level", [])
+                filters["level"].append(level)
+        return filters
+
+    def update_public_adventure_filters_indexes(self, adventure):
```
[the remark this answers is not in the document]
hasan-sh: True, but i use it in other places; e.g., when deleting a tag or cloning an adventure!

website/public_adventures.py
hasan-sh: > Also, where are you accounting for the possibilty that the filter can be empty? What if the user didn't pick a language, level, or set of tags?

Currently, the levels are a must and can't be ignored. If you make an ajax call, it'd be defaulted to level 1.

website/public_adventures.py
hasan-sh: 100% true

website/public_adventures.py
hasan-sh: > I didn't know `union` takes 3 arguments. What does that mean?
> 
> Why are you doing a union here? Aren't we supposed to intersect all 3 filters?

I intersect them first then take the union. Because if one is an empty set, the intersection would result in an empty set.

create-public-adventures-filters.py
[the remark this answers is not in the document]
rix0rrr: Not necessary, they don't really need to be run a lot, just once. The reason to commit them would be for reviewing and future reference. But that's a good idea to do!

static/js/public-adventures.ts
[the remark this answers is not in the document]
rix0rrr: Yeah, but fighting race conditions with sleeps isn't a great idea.

Imagine this website running on a slightly slower computer -- the init functions is going to run too quickly again, before all the initialization that you depend on has happened. You're never going to cover all computer speeds with one timeout.

It would be better to find a more deterministic point to do this:

- (Preferred) find some way to hook into the JavaScript that you need to run after, and make it call you when it's done. 
- (Not preferred but possible) Poll, as much as it sucks! Set an interval timer to run every 50ms, do some check against the DOM to see if an HTML element you depend on exists, and if it does, and do your initialization. Don't forget to cancel the timer! 

Now your initialization is going run ~more or less as quickly as it can.

website/database.py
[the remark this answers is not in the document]
rix0rrr: What I'm getting at is that you should hide the details of updating indexes as much as possible.

If the front-end doesn't need to know about the fact that there are indexes, if this can all remain in the `Database` class -- so much the better! Less for consumers to keep in mind, fewer chances of bugs.

Remember that there are other people working on this code base, who might not know the intricacies of "when you call function A, don't forget to also call function B". Ideally, they shouldn't have to! And don't underestimate how much you yourself will have forgotten about the details in 3 months' time as well 😉.

So, any operation that needs to touch the indexes should go through a public method of the `Database` class anyway; and that method can call this one whenever it needs to.

website/public_adventures.py
rix0rrr: I understand, but if both language and tags are given, your set is now:

`adventure_ids` is the set of adventures that have EITHER the given language, OR the given tags.

Is that what you meant? I thought it would have been:

`adventure_ids` is the set of adventures that have BOTH the given language, AND the given tags.

rix0rrr: > Currently, the levels are a must and can't be ignored. If you make an ajax call, it'd be defaulted to level 1.

That seems like a bummer though? If I just want to see all Spanish adventures, there's no way to do that?

website/database.py
rix0rrr: Well you know what, I guess order doesn't matter since you're throwing the results into a `set` afterwards anyway. 

But it could matter, and perhaps should matter, to preserve some sort of ordering and have the newest results on top.

static/js/public-adventures.ts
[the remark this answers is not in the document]
hasan-sh: > * (Preferred) find some way to hook into the JavaScript that you need to run after, and make it call you when it's done.

What could work is to listen on an event that's fired after the dom elements are replaced and then update the js. Will try that!

website/public_adventures.py
hasan-sh: > That seems like a bummer though? If I just want to see all Spanish adventures, there's no way to do that?

The main reason behind it is that the js needs to be initialized with a specific level in order to correctly run code in the editor!

hasan-sh: > `adventure_ids` is the set of adventures that have EITHER the given language, OR the given tags.

What OR? the union will combine these sets right. So, if i have:
```
lang_ids = {'3f8aea42eb324f08a16776671498dd1b'}                                                                                                                                                                       
tag_ids = set()
level_ids =  {'c9ccb61ae9334416aba972f5d5089939', '3f8aea42eb324f08a16776671498dd1b'}       

# resulting adventure_ids is:
set() 
```

But if i first have a base (the level IDs) and update the lang_ids and tag_ids with the intersection with the base, and then take their union, it'd result in:

```{'c9ccb61ae9334416aba972f5d5089939', '3f8aea42eb324f08a16776671498dd1b'}```       

Does that make sense or am i missing anything?

File: static/js/public-adventures.ts [hunk 20]
```diff
@@ -1,17 +1,20 @@
 import { ClientMessages } from "./client-messages";
 import { initialize } from "./initialize";
+declare const htmx: typeof import('./htmx');
```
rix0rrr, line 3: Is this used?

templates/public-adventures/body.html
```diff
@@ -11,7 +11,7 @@
                 value="{{currentSearch}}"
                 {% if currentSearch %}autofocus{% endif %}
                 hx-trigger="input changed delay:500ms, search" 
-                hx-post="/public-adventures/filter?level={{level}}&lang={{selectedLang}}&tag={{selectedTag}}&search={{currentSearch}}"
+                hx-put="/public-adventures/filter?level={{level}}&lang={{selectedLang}}&tag={{selectedTag}}&search={{currentSearch}}"
```
rix0rrr, line 14: `PUT` is an odd verb to use for this operation. Normally for read-only operations, we would use `GET`, or perhaps `POST`.

Why did you change it?

website/tags.py
rix0rrr: Still relevant.

website/database.py
```diff
@@ -608,6 +611,52 @@ def batch_get_adventures(self, adventure_ids):
         keys = {id: {"id": id} for id in adventure_ids}
         return ADVENTURES.batch_get(keys) if keys else {}
 
+    def get_public_adventures_filters(self):
```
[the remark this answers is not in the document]
rix0rrr: Still relevant.

website/database.py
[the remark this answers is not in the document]
rix0rrr: Still relevant.

website/public_adventures.py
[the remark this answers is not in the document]
rix0rrr: I would still expect this to be in a GET.

By the way, why are there 2 endpoints here? `GET /` and `PUT /filter` ?

website/public_adventures.py
```diff
@@ -119,35 +62,85 @@ def filtering(self, user, index_page=False):
...
+            tags = [t for t in tags if t]
+
+        # Get indexes
+        level_lang_adventure_ids, result = self.db.get_paginated_indexes(
+            {"field_value": f"level_lang_{level}_{language}"},
+            pagination_token=page, limit=20)
+        # In case we need want any language, we retrieve by the main level index.
+        if not level_lang_adventure_ids:
+            level_lang_adventure_ids, result = self.db.get_paginated_indexes({"field_value": f"level_{level}"},
+                                                                             pagination_token=page, limit=20)
```
rix0rrr, line 74: This would all be great detail to hide from the developer in `database.py`!

Together with all the stuff below where we filter on tags and then `batch_get` the remaining adventures. Hiding all that machinery so that the writer of the website can just say `self.db.get_public_adventures()` is the point of that class!

website/public_adventures.py
```diff
@@ -119,35 +62,85 @@ def filtering(self, user, index_page=False):
...
+        adventures = list(adventures.values())
+        for adventure in adventures:
+            if not adventure:
+                continue
+            if language and adventure.get("language", g.lang) != language:
+                continue
+            if tags and not any(_t in adventure.get("tags", []) for _t in tags):
+                continue
+            if search and not search.lower() in adventure.get("name").lower():
+                continue
```
rix0rrr, line 104: Filtering should already have happened earlier, otherwise the pagination is not going to do what you want.

Your page may end up with `0` items and a `Next >>` button, which looks very weird.

website/public_adventures.py
```diff
@@ -228,30 +223,8 @@ def clone_adventure(self, user, adventure_id):
         }
 
         self.db.update_adventure(adventure_id, {"cloned_times": current_adventure.get("cloned_times", 0) + 1})
+        self.db.update_public_adventure_filters_indexes(adventure)
```
rix0rrr, line 226: This should happen inside `update_adventure()`.

Felienne:
Converting this to a draft, there are a few good ideas in this PR but we will separate them out (no need to do anything more @hasan-sh)

hasan-sh:
> Converting this to a draft, there are a few good ideas in this PR but we will separate them out (no need to do anything more @hasan-sh)

Sure, it's actually done and I've satisfied all Rico's comments. Will keep an eye and see if anything needs to be clarified or changed as well:)

PR #5436, opened 2024-04-18
📜 Use `{play}` template instead of literal `play` in music adventure

bobvanderlinden:
Currently when opening the music adventure level 1 in Nederlands you'll get the following instruction:

> In dit level leer je het `speel` commando te gebruiken om een muzieknoot te spelen!

Though in the example right next to it, it says:

```
play C4
...
```

This example is English and not Nederlands. It should say:

```
speel C4
...
```

Other levels of the music adventure _do_ correctly use the `speel` command. There `{play}` template is indeed used. I applied the same template to level 1.

Let me know if I should also create an issue for this PR.

**How to test**

* Go to http://localhost:5000/hedy/1#music
* Change the language to "Nederlands"
* Verify that the example uses `speel` instead of `play` commands.

**Checklist**

### DIFF
--- a/content/adventures/ar.yaml
+++ b/content/adventures/ar.yaml
@@ -2573,14 +2573,14 @@ adventures:
                     في المستوى التالي ستتعلم كيفية تشغيل بعض الأغاني الموجودة.
                 example_code: |-
                     ```
-                    play C4
-                    play D4
-                    play E4
-                    play F4
-                    play G4
-                    play A4
-                    play B4
-                    play C5
+                    {play} C4
+                    {play} D4
+                    {play} E4
+                    {play} F4
+                    {play} G4
+                    {play} A4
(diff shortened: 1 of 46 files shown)


Felienne (approved):
Thanks for your first PR @bobvanderlinden, welcome to Hedy!! Well done diving into the yamls to find this and cross referencing with other levels!

Other languages, sadly, suffer from the same issue (they got copied over from the same source), see f.e. https://github.com/hedyorg/hedy/blob/1eac4aadf9efe92bbc7578d1c5e0adece3aac22d/content/adventures/cy.yaml#L2719

Would you mind fixing it in other languages too? 🙏

bobvanderlinden:
Good one! Didn't even notice before. I've search-replaced all instances of this example :+1:

PR #5446, opened 2024-04-22
💻 list feedback to super teachers

hasan-sh:
- **- admin creates super teachers - front and backend - with e2e test**
- **retrieve and show feedback to super teachers**
- **refresh current user after converting to super teacher**


Annelein:
@hasan-sh I've added this button which for now goes to /feedback but can become a 'super-teacher' page:
Screenshot 2024-04-22 at 15 13 49

And I've create a table for the feedback and the filters, can you do the backend work for the filters?

Issue #4029, opened 2023-02-13
Improve handling of server responses and response codes


jpelay:
An old PR (#3836) from @TiBiBa had the following objectives in mind

- [ ] > * Replace all jsonify() returns to make_response() with the corresponding response code
- [ ]  > * Remove all hard-coded response messages, even when not likely to be returned to user
-> #5558 
- [x] > * When return an empty succes, use 204 instead of 200 https://github.com/hedyorg/hedy/pull/5455
- [x] > * Improving the other error codes is handled in a separate issue https://github.com/hedyorg/hedy/issues/3634

That PR was closed because it was getting old and further refactors changed those files and made continuing working on that branch unfeasible. That's why I'm opening this issue to work on this later on!

Annelein:
> * [ ]   > * Remove all hard-coded response messages, even when not likely to be returned to user

@jpelay does this mean change all of these, into `return make_response(gettext("request_invalid"), 400)` ?


```
if not body:
        return "body must be an object", 400
if 'code' not in body:
        return "body.code must be a string", 400
if 'level' not in body:
        return "body.level must be a string", 400
if 'adventure_name' in body and not isinstance(body['adventure_name'], str):
        return "if present, body.adventure_name must be a string", 400
if 'is_debug' not in body:
        return "body.is_debug must be a boolean", 400
if 'raw' not in body:
        return "body.raw is missing", 400
```

Annelein:
@jpelay and should `return gettext('image_invalid'), 400` be changed to `return make_response(gettext('image_invalid'), 400)` ?

jpelay:
> @jpelay and should `return gettext('image_invalid'), 400` be changed to `return make_response(gettext('image_invalid'), 400)` ?

Yes to both of your questions!

rix0rrr:
> * Remove all hard-coded response messages, even when not likely to be returned to user

I don't have a ball in this game, but consider the following:

Every string you make translatable like this is a string that feels like an obligation to a translator to translate, if they want "100%".

If you're going to make error strings that are only shown when the frontend programmer has made a mistake against the backend contract, then you are making volunteer Arabic or Thai translators spend a lot of time translating strings that will only ever be seen by *you* in English, or perhaps Dutch.

So the question is, what is the benefit the project gets from changing this code? Consistency is the only one I can think of: now you don't have to think anymore which errors get surfaced to users and which don't. Consider what you're paying to get that, and whether it's worth it.

[...]

PR #5455, opened 2024-04-23
🧹 When returning an empty succes, use 204 instead of 200

Annelein:
Fixes a part of #4029 

**How to test**
1. All unittests should pass
2. Check if all new returns are correct


app.py
```diff
@@ -2265,7 +2265,7 @@ def change_language():
     # Remove 'keyword_lang' from session, it will automatically be renegotiated from 'lang'
     # on the next page load.
     session.pop('keyword_lang')
-    return jsonify({'succes': 200})
+    return jsonify({'succes': 204})
```
jpelay, line 2268: Why not using `make_response` her too? Since you've already changed it in other places

Annelein: Because it would give an unknown error..

jpelay (approved):

PR #5464, opened 2024-04-24
🚚 Don't run autopep8 on tests directory

rix0rrr:
`autopep8` has a bug on Python 3.12 where it may change the contents of f-strings, which breaks our tests which assert that some string is literally equal to another string. https://github.com/hhatto/autopep8/issues/744

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,10 @@ repos:
     rev: v2.1.0
     hooks:
     -   id: autopep8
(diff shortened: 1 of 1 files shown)


rix0rrr:
Wait, maybe it's something else after all.

rix0rrr:
Actually the bug was slightly different, it's this one: https://github.com/hhatto/autopep8/issues/744

rix0rrr:
Trying to fix the issue in the PR where we discovered it: https://github.com/hedyorg/hedy/pull/5463

PR #5470, opened 2024-04-25
📜 Rephrasing level 2 is command adventure

### LINKED ISSUE #5461: 📜 Improve phrasing in level 2 example
In Level 2, English, I see two potential confusions:

### First confusion

 ```
                    name {is} Hedy
                    age {is} 15
                    {print} name is age years old
```

The word `is` in the print statement was confusing to my kid, because it looks the same as the `is` in the previous two statements which have a completely different meaning in the language.

I don't know how to tackle this without being too verbose. For example, you could modify the print statement as
```
  {print} name happens to be age years old
```

### Second confusion

This one is simpler:

```
                    favorite_animals {is} _
                    {print} I like favorite_animals
```

Instead of plural `favourite_animals`, using singular would be better: `favourite_animal {is} ...`

The plural makes it appear as if a list of animals is expected.

MarleenGilsing:
favorite_animal instead of favorite_animals

fixes #5461

### DIFF
--- a/content/adventures/ar.yaml
+++ b/content/adventures/ar.yaml
@@ -353,8 +353,8 @@ adventures:
                     Copy your code from the previous tab and make the variables interactive by using `{ask}` commands.
                 example_code_2: |
                     ```
-                    favorite_animals is ask What is your favorite animal?
-                    print I like favorite_animals
+                    favorite_animal {is} {ask} What is your favorite animal?
+                    {print} I like favorite_animal
                     ```
             18:
                 story_text: The final change we will need to make to get Python code is changing `{ask}` into `{input}`.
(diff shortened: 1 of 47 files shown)


MarleenGilsing:
The issue has been resolved for all languages (for as far as it was still in English, or I could understand the language). I did notice one strange thing:

This was the Swedish translation of level 2: 
image

It states in the text that the example code uses the variable `favorite_animals`, but the example code doesn't use favorite_animals. It uses `favoritdjur`. This is confusing for students and happened in quite a lot of languages. I think this might happen because weblate only shows the translators a small part of the page to translate, so they don't see the context of their translations. Might be worth looking into, if this happens more often?

Felienne:
> The issue has been resolved for all languages (for as far as it was still in English, or I could understand the language). I did notice one strange thing:
> 
> This was the Swedish translation of level 2: image
> 
> It states in the text that the example code uses the variable `favorite_animals`, but the example code doesn't use favorite_animals. It uses `favoritdjur`. This is confusing for students and happened in quite a lot of languages. I think this might happen because weblate only shows the translators a small part of the page to translate, so they don't see the context of their translations. Might be worth looking into, if this happens more often?
> 
> In this case, and the others I found, i already fixed it of course.

Good find and yes that is certainly a limitation of Weblate that it does not show texts together :(

Felienne (approved):

PR #5472, opened 2024-04-25
 📜 Show music adventures

MarleenGilsing:
I noticed that the music adventures of level 15-18 weren't online yet, despite being ready, so I've added them to the hedy_content file.


Felienne (approved):
Whooops! Maybe we should add a test for this?

PR #5487, opened 2024-04-29
Resolve Weblate conflicts

rix0rrr:


Felienne:
Merging because otherwise weblate will remain stuck. Will deal with the e2e issue later

PR #5500, opened 2024-05-03
💻 show some roles of variables (constant, input constant, walker, container)

juliabolt:
Adds roles in the variable list, works for four roles now, constant, input constant, container and walker. 

**How to test**

* change the feature flag in debugging.ts called showRoles to true
* check that the variable list now shows roles of variables, in addition to the variable names and their values. (when using the Hedy debugger)
* run this example program in level 2 and see if it says naam is a constant
```
naam is Hedy
leeftijd is 10
leeftijd is 15
print naam is leeftijd jaar oud
```
* see if in level 2 the variables in this program are input constants
```
naam is ask Wat is je naam?
print Hoi naam
leeftijd is ask Hoe oud ben je?
print naam is leeftijd jaar oud
```
* run this example program in level 10 and see if it says that dier is a walker
```
dieren = hond, kat, papegaai
voor dier in dieren
  print dier ' is een leuk dier!'
```
* go to level 3 and see if dieren is a container in this example code
```
dieren is hond, kat, papegaai
print dieren at random
```
* change the feature flag back to false
* check that the variable list now looks like it normally does, with only the variable names and their values.

**Checklist**

- [x ] Describes changes in the format above
- [x ] Has a "How to test" section


juliabolt:
image
image

When testing the code with the additions in this PR, everything seems to work in level 1 to 5, but from level 6 on, random things seem not to work, like calculations in level 6 and the repeat command in level 7.
This already seems to go wrong, before ever getting to the function determine_roles that I added in hedy.py (when using the debugger and putting a breakpoint in that function).

juliabolt:
Issue in former comment is solved! thanks @jpelay :)

jpelay:
I fixed the parsing problem! But now it's not showing output when in some programs, for example this one of level 12:

```
a is 2
b is 3
for a in range 2 to 4
    a is a + 2
    b is b + 2
```

There's a response and the program is shown as executed, perhaps it has to do with the fact that the variables has a type of unknown?

image

juliabolt:
image
 I get this when I try your example code in Hedy level 12

juliabolt:
I don't get that error in alpha, but it is normal that there is no output in that program :P

jpelay:
> I don't get that error in alpha, but it is normal that there is no output in that program :P

HAHA you're right! oops

jpelay:
> The type unknown should not work any differently than the other types, it's just a string in a dictionary...
> 
> I still get issues trying to test this code, ModuleNotFoundError: iso639 (see app.py line 22)

Do the pip thing again!

PR #5510, opened 2024-05-08
🪲 saving programs consistently

### LINKED ISSUE #5507: 🪲 Trying a public adventure in the editor, saves the program
**Describe the bug**

Also while testing #5485 we found out that after tyring the adventure and editing it, the program was being saved and restored. We think that the behavior should be that these programs are not stored and the user shoukld be greeted with an empty editor every time.

After some digging Hasan found that the editor contents where being modified after the switch event.

hasan-sh:
Fixes #5507


jpelay:
Hi! I've been testing and, yes now the programs seem to be being saved more consistently. This is how a saved program from this page looks in the database:

```
    {
      "session": "fac24066d19f4fd8ac44508ac4a7fdae",
      "date": 1715266264375,
      "lang": "en",
      "version": "DEV",
      "level": 17,
      "code": "print 'hello' 1 + 2",
      "name": "Tik Tak Toe",
      "username": "teacher1",
      "error": false,
      "adventure_name": "Tik Tak Toe",
      "id": "720bcbb290384d7b9f5a018d168c8707",
      "username_level": "teacher1-17",
      "is_modified": true
    }
```

But I wonder if this is the solution? I think perhaps the best way is to not store programs in this page? I think we should only be storing programs that come from the /hedy page.

PR #5513, opened 2024-05-09
🪲 Running programs throws an error if logged in

rix0rrr:
When logged in, we automatically save a program as the user runs it.

In a recent change, we update the program counter only when a program is modified from the example code. But we wrote the entire 'program' record back to the database in an `update()` call, which includes the `id` field; Dynamo does not allow including the `id` field in an `update()` call.

This is also not necessary: a DynamoDB `update()` call only needs to include the fields we actually want to change, and not any of the fields that are unchanged (this may even lead to a race condition if any of those fields were changed in between the read and update calls, we would undo the intermediate change again).

Instead, only update the `is_modified` field, and update the Dynamo abstraction layer to check for this error condition.

**How to test**

Log in, and run a program.

### DIFF
--- a/app.py
+++ b/app.py
@@ -605,6 +605,7 @@ def parse():
             if transpile_result.has_music:
                 response['has_music'] = True
 
+            response['variables'] = transpile_result.roles_of_variables
         except Exception:
             pass
 
--- a/content/adventures/en.yaml
+++ b/content/adventures/en.yaml
@@ -617,7 +617,7 @@ adventures:
                     ```
             9:
                 story_text: |
-                    In a previous level you've created a calculator, in this level you can expand that code so it asks multiple questions.
+                    In a previous level, you've created a calculator. In this level, you can expand that code so it asks multiple questions.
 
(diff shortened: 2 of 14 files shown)


rix0rrr:
@Annelein this should have thrown an error locally before deploying and running against an actual DynamoDB, so my apologies that it didn't. But also, `update()` is supposed to only get the fields you want to change, while `put()` should be used if you really want to overwrite all fields of an entire record (and delete all non-included fields).

```
# Insert or replace an entire record (all fields not part of this write are gone)
table.put(entire_record)

# Update a single field (all other fields remain unchanged)
table.update(record['id'], { 'field': 'new_value', 'another_field': 'another_value' })
```

`update()` is safe against concurrent `updates()`, while `put()` will just overwrite all fields (even if another request changed any of the fields in the mean time).

Felienne (approved):

Annelein:
@rix0rrr thank you for the feedback, fix and explanation! Really helpful.

PR #5521, opened 2024-05-13
🪲 Old programs without 'is_modified' are shown

Annelein:
Also fixes that only modified programs are shown in 'My public profile'

**How to test**
Is it correct that we don't have saved programs in the test database? Then it's a bit difficult to test:
1. Go to programs.py and comment out line 101, 104 and 106
2. Save a few programs, check if they are shown in `/my-programs`
3. Make a few public and see if they are shown in 'My public profile'

### DIFF
--- a/app.py
+++ b/app.py
@@ -1042,7 +1042,9 @@ def programs_page(user):
         date = utils.delta_timestamp(item['date'])
         # This way we only keep the first 4 lines to show as preview to the user
         preview_code = "\n".join(item['code'].split("\n")[:4])
-        if item.get('is_modified'):
+        # When saving a program, 'is_modified' is set to True or False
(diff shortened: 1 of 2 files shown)


Felienne:
> Also fixes that only modified programs are shown in 'My public profile'
> 
> **How to test** Is it correct that we don't have saved programs in the test database? Then it's a bit difficult to test:

Good point! I will make a new issue to discuss that!

> 1. Go to programs.py and comment out line 101, 104 and 106
> 2. Save a few programs, check if they are shown in `/my-programs`
> 3. Make a few public and see if they are shown in 'My public profile'

I will be lazy and test on alpha :)

Felienne (approved):

PR #5523, opened 2024-05-13
💻 Add "previous page" buttons

rix0rrr:
This adds the feature to also go backwards in the result set to DynamoDB queries.

The same way there was a `results.next_page_token` before that you could pass in to get the next page of results, this adds a `results.prev_page_token`. If the `prev_page_token` is set, pass the token into the exact same query operation to return the previous page of results.

I've added a demonstration to the Admin "all users" page; I don't know where else "previous page" buttons should be added.

In addition, this adds a `get_page()` function which can be used for paginated result pages that need additional filtering. Filtering can be done on the server side (passing a `server_side_filter`) and on the client side (passing a `client_side_filter`). The `get_page()` function will repeatedly do server queries until it has filled up a page of the requested `limit` items (or until a timeout is hit), and make sure the `next_page_token` and `prev_page_token` values are set correctly. `get_page` should probably replace `QueryIterator`-based solutions -- `get_page` is easier to use.

@jpelay @Annelein @hasan-sh I could use some help with rolling this new feature out wherever it makes sense.

**How to test**

Go to the Admin page, click on "User overview", click "Next page". A "Previous page" button appears at the bottom of the screen, clicking it gets you back to the first page.


website/dynamo.py
```diff
@@ -689,6 +764,33 @@ def _decode(self, data):
         return {k: replace_decimals(DDB_DESERIALIZER.deserialize(v)) for k, v in data.items()}
 
 
+class PaginationKey:
```
hasan-sh, line 768: Could you share a doc where this is explained? I don't entirely get why we extract keys, and how is the `extract_ordered` method _ordering_ them? In what direction?

website/dynamo.py
```diff
@@ -1113,18 +1216,29 @@ def validate_only_sort_key(conds, sort_key):
...
+    'inverted' indicates whether the page token goes in the other direction
+    than the normally accepted query direction. If the query normally goes
+    forward, a query with an inverted page token would go backwards; but if the
+    query normally goes backwards (reverse=True), a query with an inverted page token
+    would go forward.
+    """
     if x is None:
         return None
-    return base64.urlsafe_b64encode(json.dumps(x).encode("utf-8")).decode("ascii")
+    return ('-' if inverted else '') + base64.urlsafe_b64encode(json.dumps(x).encode("utf-8")).decode("ascii")
```
hasan-sh, line 1231: This is just a trick for us to know in which direction we wanna retrieve results right?

website/database.py
```diff
@@ -234,8 +234,10 @@ def filtered_programs_for_user(self, username, level=None, adventure=None, submi
 
         # FIXME: Query by index, the current behavior is slow for many programs
         # (See https://github.com/hedyorg/hedy/issues/4121)
-        programs = dynamo.GetManyIterator(PROGRAMS, {"username": username},
-                                          reverse=True, batch_size=limit, pagination_token=pagination_token)
+        programs = PROGRAMS.get_many({"username": username},
+                                     reverse=True, limit=limit, pagination_token=pagination_token)
+        # programs = dynamo.GetManyIterator(PROGRAMS, {"username": username},
+        #                                   reverse=True, batch_size=limit, pagination_token=pagination_token)
```
rix0rrr, line 240: This doesn't do the same as it did before.

This will retrieve `limit` items, then filter them client-side (leaving you potentially with `<limit` items).

On the other hand, the `GetManyIterator` will keep on pulling items from the database until you stop pulling (which we will do only once we have exactly `limit` items, or have reached the end of the list).

website/database.py
```diff
@@ -455,23 +457,9 @@ def get_public_programs(self, level_filter=None, language_filter=None, adventure
...
-            if pagination_token is None:
-                break
-        del found_program_ids[limit:]  # Cap off in case we found too much
-
-        # Now retrieve all programs we found with a batch-get
-        found_programs = PROGRAMS.batch_get(found_program_ids)
-        return found_programs
+        page = PROGRAMS.get_many({'public': 1}, reverse=True, limit=limit,
+                                 pagination_token=pagination_token,)
+        return page
```
rix0rrr, line 462: Same here, and you completely removed the filtering.

website/dynamo.py
```diff
@@ -303,37 +324,70 @@ def get_many(self, key, reverse=False, limit=None, pagination_token=None, filter
...
         querylog.log_counter(f"db_get_many:{self.table_name}")
 
+        inverse_page, pagination_token = decode_page_token(pagination_token)
+        if inverse_page:
+            reverse = not reverse
+
         lookup = self._determine_lookup(key, many=True)
         if isinstance(lookup, TableLookup):
+            pagination_key = PaginationKey.from_table(self.key_schema)
             items, next_page_token = self.storage.query(
```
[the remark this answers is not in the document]
rix0rrr: We only reassign if we have a `limit`. If we don't, then we just leave the natural `next_page_token` that the underlying query returns.

website/dynamo.py
rix0rrr: Yep!

website/dynamo.py
```diff
@@ -303,37 +324,70 @@ def get_many(self, key, reverse=False, limit=None, pagination_token=None, filter
...
+            items.reverse()
+        else:
+            if limit:
+                # Set the next_page_token only if we retrieved N+1 items (there is an actual next page).
+                has_more_items = len(items) > limit
+                items = items[:limit]
+                next_page_token = pagination_key.extract_dict(items[-1]) if has_more_items else None
+            prev_page_token = pagination_key.extract_dict(items[0]) if items and pagination_token else None
+
+        return ResultPage(items, encode_page_token(next_page_token, False), encode_page_token(prev_page_token, True), pagination_key=pagination_key)
```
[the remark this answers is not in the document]
rix0rrr: Nice catch, thanks!

website/dynamo.py
rix0rrr: Good idea!

rix0rrr:
@hasan-sh I think this one is good to go!

templates/htmx-adventure-card.html
```diff
@@ -57,7 +57,7 @@
 
         {% if adventure["code"] %}
             <p class="text-gray-600 mt-1"><strong>{{_('number_lines')}}:</strong> {{ adventure.number_lines }}</p>
-            <pre level="{{ adventure.level }}">{{ adventure.code }}</pre>
+            <pre level="{{ adventure.level }}" class="text-xs">{{ adventure.code }}</pre>
```
hasan-sh, line 60: To my eyes the editor content is really small now! Any reason for that?!

website/dynamo.py
```diff
@@ -298,42 +348,163 @@ def get_many(self, key, reverse=False, limit=None, pagination_token=None, filter
...
+                next_page_token = decode_page_token(page.next_page_token)[1] if page.next_page_token else None
+
+            # The first element of the first page marks our prev_page_token
+            prev_page_token = page.pagination_key.extract_dict(
+                first_page[0]) if first_page and pagination_token else None
+
+        return ResultPage(
+            items,
+            encode_page_token(next_page_token, False),
+            encode_page_token(prev_page_token, True))
```
hasan-sh, line 507: would be nicer to switch the parameter's order!

templates/htmx-adventure-card.html
rix0rrr: To my eyes they were very large, not showing a lot of the program because of it.

Meet in the middle at `text-sm` ? 😜

website/dynamo.py
rix0rrr: Yeah, might be. I'm always a bit hesitant with obviously backwards incompatible changes, but I guess in this case it might be okay...

templates/htmx-adventure-card.html
hasan-sh: Hahaha, amazing!!

hasan-sh (approved):
All works great!

PR #5524, opened 2024-05-14
🖊️ Add support for boolean values

### LINKED ISSUE #5370: 🖊️Support False instead of 'False'
As Felienne says in #5333, it would be nice to support False instead of 'False'. 

When this is done, maybe I can introduce True and False in a seperate adventure.

boryanagoncharenko:
Fixes #5370
Adds support for boolean values with the following caveats:
- we support both `true` and `True` as correct boolean values. Since hedy is currently case sensitive, this PR uses a workaround to support both values and does not attempt to solve the case sensitivity issue.
- when printed, the boolean value will appear as a correct python value, e.g. `True`, regardless in which language it was originally written. In other words, the Bulgarian program `{print} {true}` will yield literally `True`. The problem cannot be solved with the existing mechanisms in Hedy, so a separate issue will be created to address the matter.
- highlighting is still work in progress

**How to test**
In levels 15-18, try the following programs:
```
a is true
print a
```
```
a is true

while a = True
    print 1
    a = false

print 2
```


static/js/lezer-parsers/tokens.ts
```diff
@@ -572,7 +576,11 @@ const keywordToToken: Record<number, tokenSpecilizer> = {
             "is": is15,
             "while": while15,
             "call": call15,
-            "with": with15
+            "with": with15,
```
jpelay, line 579: Hi, @boryanagoncharenko! I took the liberty to push a fix to the highlighting problems you commented to me on Discord. The problem was that for the tokenizer to use the keywords, they need to be added to the proper dictionary. If you set the variable as a  specialized, you need to add it to the `specialize` dictionary inside the import. The difference between specialize and extend is that with `specialize` every time the parser sees the string it will parse it as the keyword, whereas with extend it will decide weather to use the keyword or plain text based on context.

boryanagoncharenko: Thanks!

jpelay (approved):
Great work as always :D

PR #5530, opened 2024-05-14
🧹 Refactor cypress tests (part 2)

Annelein:
Fixes the following tests for #4027 :

- for-teacher page
   - [x]  class page

The last few tests, which are not in the class page folder. Have only a few edited lines, this is because those were id's changed in class page tests, so I changed all occurrences.

@jpelay I will take care of the conflicts that this will give for #5130


static/js/appbundle.js
```diff
@@ -103478,7 +103478,7 @@ def note_with_error(value, err):
         contentType: "application/json",
         dataType: "json"
       }).done(function(response) {
-        if (response.achievement) {
+        if (response && response.achievement) {
```
Annelein, line 103481: This check is necessary because deleting a class returns `make_response()` and this would give errors because then response doesn't exist. I'm not sure why.

jpelay (approved):
Lovely!!

PR #5532, opened 2024-05-14
🧹 Update deploy-to-prod.yml

Felienne:
Allows @jpelay to also deploy (I think, you never know with GitHub actions...)


Felienne:
> > Allows @jpelay to also deploy (I think, you never know with GitHub actions...)
> 
> So now I just have to go to Heroku and try to deploy?

Not even! You can just go this this action: https://github.com/hedyorg/hedy/actions/workflows/deploy-to-prod.yml and run it (feel free to try)

I only go to Heroku if I need to roll back (let me know if you need help to figure out how)

PR #5537, opened 2024-05-16
💻 Roles of variables: add stepper in for loop

juliabolt:
I've added the stepper role, it works with for loops, but other kinds of steppers are not working yet.

**How to test**

* change the feature flag showRoles in debugging.ts to true
* go to Hedy level 11 and run this program (or something similar) in the debugger:
```
for counter in range 1 to 10
    print counter
print 'Ready or not. Here I come!'
```
* check that in the variables view, the role that is shown for the variable counter is stepper. 

**Checklist**


juliabolt:
oops let me rebase this branch first

PR #5538, opened 2024-05-16
💻 add roles of variables: stepper in for loops

juliabolt:
I've added the stepper role, it works with for loops, but other kinds of steppers are not working yet. 

Also started using labels so that the variable role names can be translated into the different languages (added the English myself, with container now called list and input instead of input constant).

**How to test**

* change the feature flag showRoles in debugging.ts to true
* go to Hedy level 11 and run this program (or something similar) in the debugger:
```
for counter in range 1 to 10
    print counter
print 'Ready or not. Here I come!'
```
* check that in the variables view, the role that is shown for the variable counter is stepper. 

**Checklist**


translations/en/LC_MESSAGES/messages.po
```diff
@@ -939,6 +945,9 @@ msgstr "Link"
 msgid "list"
 msgstr "a list"
 
+msgid "list_variable_role"
+msgstr ""
```
jpelay, line 949: Hi @Julia! Now what you have to do is type the meaning of this label inside `msgstr`. Then you need to do `doit run devserver` and this will compile the labels and you should see the appropiate strings afterwards.

juliabolt: I did that, but it did not work... Somehow the msgstr things don't stay there. I put it in the messages.pot file, is that not the right one? Should I put it in the English file messages.po ?

jpelay: Yes, precisely, you should add it in the message.po file inside the `en` folder.

juliabolt:
I'll try to make a cypress test an make the feature flag a variable in the environment in a new PR, to prevent this PR from becoming too big :)

jpelay (approved):
Works 😄 
image

Felienne:
Ah interesting... this one passes but the Weblate one is still broken, so maybe it was not the same issue?

juliabolt:
yeah maybe it was another cypress test

PR #5539, opened 2024-05-16
🧪 Update test db

### LINKED ISSUE #5522: Add saved programs in test DB
> Also fixes that only modified programs are shown in 'My public profile'
> 
> **How to test** Is it correct that we don't have saved programs in the test database? Then it's a bit difficult to test:

Good point! I will make a new issue to discuss that!

> 1. Go to programs.py and comment out line 101, 104 and 106
> 2. Save a few programs, check if they are shown in `/my-programs`
> 3. Make a few public and see if they are shown in 'My public profile'

I will be lazy and test on alpha :)

_Originally posted by @Felienne in https://github.com/hedyorg/hedy/issues/5521#issuecomment-2108242386_

Annelein:
Fixes #5522 

**How to test**
See if teacher1 now has programs and a public profile


Annelein:
@Felienne this test fails: `second-teachers can NOT view each other's public programs` because teacher1 now has public programs and teacher4 can actually see them. So I don't think this was a properly working test before?

But did we change this, are second teachers allowed to view each other's public programs now? Or does this PR catch an issue?

Felienne:
> @Felienne this test fails: `second-teachers can NOT view each other's public programs` because teacher1 now has public programs and teacher4 can actually see them. So I don't think this was a properly working test before?

Ah good find!! The immediate effects of more test data :) 

> But did we change this, are second teachers allowed to view each other's public programs now? Or does this PR catch an issue?

I don't think this change was intentional, but maybe @jpelay or @hasan-sh knows?

hasan-sh:
> @Felienne this test fails: `second-teachers can NOT view each other's public programs` because teacher1 now has public programs and teacher4 can actually see them. So I don't think this was a properly working test before?
> 
> But did we change this, are second teachers allowed to view each other's public programs now? Or does this PR catch an issue?

This test isn't failing on master and passing as expected, but i see you added two public programs, not sure why, but then you need to adjust the tests accordingly. For instance, https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/public_programs/programs_page.cy.js#L141 wouldn't hold anymore!

tests/cypress/e2e/public_programs/programs_page.cy.js
```diff
@@ -131,17 +131,16 @@ describe("General tests for my programs page (with both custom teacher and built
                     })
     });
 
-    it("second-teachers can NOT view each other's public programs", () => {
+    it.only("second-teachers can NOT view each other's public programs after making them private", () => {
```
hasan-sh, line 134: we need to test it after removing the `.only` first!

hasan-sh (approved):

PR #5540, opened 2024-05-16
🧹 Add HedySelect custom element

jpelay:
One of the advantages of frameworks is that they allow you to define custom HTML elements and couple them with JavaScript behavior. Since a few years ago, vanilla JavaScript now supports the notion of [custom elements](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements) as part of the wider technologies used to support [web components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components). The basic notion here is that we can define a template that will hold the HTML code for our element, and then clone that template and attach it to our DOM, at the same time executing a function that will be called once the component is connected. 

There's another way to show the component, by using the shadow DOM. This will further encapsulate the element, by hiding it's inner components from the general application, however this has the disadvantage of not being able to use tailwind in this component, since it can't access the CSS stylesheets from the application. Due to this limitation I decided to just attach the element to the old fashioned DOM.

The basic structure of the components is the following:

* A template defined in `custom-element-templates.html`
* A class defined in `custom-elements.ts`
* The class is registered in the `customElements` registry.

After doing this we can just use the component within our HTML like we'd do with any other component. With the example of the selects, we used to have a `prepareDropdowns` function that was called on page load. Now, that function is called for every select once they're connected using the `connectedCallback` property. The usage of the `select` in the HTML is like this:

```html
<hedy-select id="select_id">
    <select multiple>
        <option value="1">1</option>
        <option value="2">2</option>
        <option value="3">3</option>
    </select>
</hedy-select>
```

That inner select will be used to generate the appropriate divs that will make the options, and will eventually be hidden from the document. 

To get the selected items, we just need to query our desired `hedy-select` and use the `selected` property; this will return a string array with the selected options. Notice that if the select is single, you can just select the first element of the array.

```typescript
(document.querySelector('#levels_dropdown') as HedySelect).selected
```

Going forward we can use this idea to replace other parts of our code base that can be made into single, reusable components. Do you guys have any ideas about this?

**How to test**

Everything should work normally


templates/public-adventures/body.html
```diff
@@ -19,137 +19,50 @@
...
+                  {% set selected = tag in selectedTag %}
+                  <option {% if selected %}selected{% endif %} 
+                  {% if selected %}
+                      hx-post="/public-adventures/filter?tag={{tag}}&reset={{selectedTag}}&level={{selectedLevel}}&lang={{selectedLang}}&search={{currentSearch}}"
+                    {% else %}
+                      hx-post="/public-adventures/filter?tag={{tag + ',' +  selectedTag}}&level={{selectedLevel}}&lang={{selectedLang}}&search={{currentSearch}}"
+                    {% endif %} 
+                    hx-target="#public_adventures" value="{{tag}}">{{tag}}</option>
+                {% endfor %}
+              </custom-select>
```
hasan-sh, line 59: If i click on the empty option that's supposed to remove other selected options, the following happens:

image

hasan-sh: also, i think it's best to merge this after the latest changes introduced in #5431, or at least be careful when merging them!

templates/explore.html
```diff
@@ -32,6 +32,24 @@
...
+                </custom-select>
+                <custom-select name="level" id="explore_page_level" data-autosubmit="true" class="w-1/3" data-type="single">
+                    <option hidden {% if not filtered_adventure %}selected{% endif %} value="">&mdash; {{_('level')}} &mdash;</option>
+                    {% for i in range(1, (max_level + 1)) %}
+                        <option hidden value="{{ i }}" {% if filtered_level and i|string == filtered_level %}selected{% endif %}>{{ i }}</option>
+                    {% endfor %}                
+                </custom-select>
+            </div>
+        </form>
+    </div>
```
hasan-sh, line 52: - still see the old dropdowns
- if i click on the option in the new select custom elements, the filtering isn't reflected on the ui
image

hasan-sh (changes requested):
Great work on the custom componenets. There's some points you can see i mentioned. Also, if i click outside of the dropdowns, they don't close, i think we need to tackle this as well!

hasan-sh:
@jpelay one point we also need to implement is when the select element is required; e.g,. in the level select in the customize-adventure page. Would that be possible?

templates/public-adventures/body.html
jpelay: > If i click on the empty option that's supposed to remove other selected options, the following happens:
> 
> image

But that is the "Select all" option, if you click it, it selects all of the other options, and if you deselect it, it will deselect every option.

templates/explore.html
jpelay: Yeah, thanks for pointing this out! Fixed it with HTMX 😄

jpelay:
> @jpelay one point we also need to implement is when the select element is required; e.g,. in the level select in the customize-adventure page. Would that be possible?

I actually handle it on case by case basis, on that particular case in the function `update_db_adventure` I check if there are no selected levels and notify the user if that is the case.

jpelay:
I addressed every comment!

Felienne (approved):

PR #5542, opened 2024-05-17
🧪generatingQLCs for level 1 test

alexpaascual:
** This code defines a function generate_question that analyzes a student's code to help them reflect on their work. The function takes an input string representing the student's code, identifies various commands within the code (such as print, ask, echo, play, forward, and turn) LEVEL 1, and generates a list of relevant questions. It then randomly selects and returns one of these questions to prompt the student to think about and self-explain their code. 
I am aware that it won't work as intended as I need to make more changes in other files for it to work properly :)**

 
**Checklist**

- [x ] Describes changes in the format above
- [x ] Has a "How to test" section


Felienne:
Marking this as draft unutil it is ready!

Felienne:
Hi @alexpaascual!

I am looking at your code and I see you are doing the analysis with string manipulation. This works but it is not very performant or elegant and it will be a lot of work for you. You can repurpose some of the work we have done!

We have a parser that already transforms a program into a tree (see [here](https://github.com/hedyorg/hedy/wiki/Hedy-Architecture#transpiling))

This can be used from program analysis! For example, in [this PR](https://github.com/hedyorg/hedy/pull/5500/files) you can see that Julia created a new function to determine roles, based on the transpiler. She used the function all_variables:

https://github.com/hedyorg/hedy/blob/0d0d159c8ed4eef733192f5aadc4b46ea8f80737/hedy.py#L1168

Tree traversal will be a bit more complex than what you are doing now, but it will be better and easier in the end. 

If you help me see what functions you need (by completing the list we talked about) I can help you along a bit more if needed!

alexpaascual:
Updated the analyze_code and generate_question functions to avoid using string manipulation. Instead of slicing strings, now using structured parsing and specific functions for each command(AST). This makes the code cleaner and easier to understand.

Felienne:
Hi @alexpaascual!

Since you are not going to implement, I am closing this for now (you can always reopen it if you want!)

Issue #5012, opened 2024-01-18
💻 Shorten our error messages


MarleenGilsing:
Our error messages might be longer than necessary.  For instance:
![image](https://github.com/hedyorg/hedy/assets/80678586/12dc6b6e-9002-4da7-a16a-3d8e3ce8fcd6)

This error message doesn't need 'at position 13'. I don't think students know what we mean by position 13 anyway and the mistake is highlighted nowadays, so the student will instantly see what they did wrong in line 1. 

I would even argue that 'The code you entered is not valid Hedy code.' is unnecessary information.

I think the shorter the message, the bigger change of a kid actually reading the message, so maybe we can take a look at the messages and tweak all the extra info.

Felienne:
Great remark! I have made similar remarks in https://github.com/hedyorg/hedy/issues/4729 Let's discuss this at the hackathon?

PR #5546, opened 2024-05-21
Shorten error messages

ArtV11:
This pull request is part of a larger project in which we are redesigning our error messages according to [these guidelines](https://github.com/hedyorg/hedy/wiki/Hedy-Language-Design#error-messages).

In this PR, I have worked on the redesigning topic (iii) Keeping Them Short where we aim to keep our error messages concise and short while providing sufficient information about the issue and its cause. Given that people often do not read the entire error message, critical information will be presented at the beginning.

Addresses #5012


Felienne:
Hi @jpelay!

One of the Cypress tests is failing on this one, could you take a look?

jpelay:
> One of the Cypress tests is failing on this one, could you take a look?

It's failing because the test is expecting the old error messages. I will update it and they should be ok!

PR #5552, opened 2024-05-22
🪲 Fix translation of colors

### LINKED ISSUE #5551: 🪲 Colors are not translated
**Describe the bug**
When the following program is translated to another language, the keyword `color` is translated correctly but the keyword for the selected color remains in English.

In levels 2-18:
```
color red
```
The translated program becomes `цвят red`, but it is supposed to be `цвят червено`.

Note that this bug is not caught in the level tests because the color keyword remains in English and is considered not yet translated.

boryanagoncharenko:
Fixes #5551
Colors are not properly translated and always remain in English. This PR adds colors to the translator, so that they are properly converted to the target language and back to the original.

**How to test**
- Automated tests are added, including translation tests. Note that existing level tests do not catch this error because when a keyword remains in English, it is assumed that no translation has been added yet.
- To manually test, translate to any language the following program in level 2 or up: `color red`


Felienne (approved):
Ah good catch, thanks!! Can't believe we never noticed!

PR #5556, opened 2024-05-22
🪲 Fix showing and hiding of variable view

### LINKED ISSUE #5553: 🪲 Variable view is wacky
**Describe the bug**
While debugging, after the variable view is shown, sometimes it opens and closes somewhat randomly.

jpelay:
The variable view was acting a little wanky, showing and hiding seemingly at random. After some digging, I found that this area was being hidden and shown by several pieces of code, so I tried to simplify the logic a bit; also instead of just hiding and showing the box, I added a little bit of animation by using JQuery's `slideToggle`.

Fixes #5553

**How to test**

Go to the code page, debug a program, and verify that the variable view is updated with each step, and that it can be hidden and shown without issues.

### DIFF
--- a/static/js/app.ts
+++ b/static/js/app.ts
@@ -13,7 +13,8 @@ import {
     incrementDebugLine,
     initializeDebugger,
     load_variables,
-    startDebug
+    startDebug,
+    toggleVariableView
 } from './debugging';
 import { localDelete, localLoad, localSave } from './local';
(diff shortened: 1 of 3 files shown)


boryanagoncharenko (approved):
Works smoothly and the animation is a nice touch! I love that this bug is fixed by primarily removing code :)

PR #5560, opened 2024-05-23
🪲 Fix stop program being shown indefinitely 

### LINKED ISSUE #5555: 🪲 Stop program button doesn't change to run program once program finishes execution
This happens from level 2 onwards, with every program. Maybe it has something to do with variables?

image

jpelay:
The problem was that we were trying to use the `theGlobalDebugger` variable while it was undefined. By using the `?` operator of typescript we guarantee that we won't be trying to call a method if the object is null.

Fixes #5555

**How to test**

I added a Cypress test., but also you can try running programs and see if the button works normally again!


Felienne (approved):
Solves the problem!

PR #5567, opened 2024-05-27
🪲 Numbers are not translated to Arabic when printed

boryanagoncharenko:
Fixes #5499
When the language of Hedy is set to Arabic, numbers should be displayed in Arabic numbers regardless of what alphabet was used in the input. This PR fixes the corner case in which when literal numbers are printed, they are not converted to Arabic.

** How to test**
Run Hedy locally, navigate to level 6 and switch the language to Arabic. Execute the following piece of code and ensure that both lines print the result `٢`:
```
print 1 + 1
print 2
```


wiki_images/s4.png
Felienne, line 1: Awwwwww love this overview!! Much needed extra documentation!

PR #5575, opened 2024-05-28
🧹 In cypress, use getBySel("") instead of get('[data-cy=""]')

Annelein:
This is a lot cleaner to read:
`cy.getBySel('view_class_link')`

instead of:
`cy.get('[data-cy="view_class_link"]')`


hasan-sh (changes requested):
There are still places where `.get("[data-cy='....` is used. For instance, in 
- https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/for-teacher_page/class_page/second_teachers_invites.cy.js#L55
- https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/for-teacher_page/general/preview-class.cy.js
- https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/hedy_page/quiz.cy.js 
-  https://github.com/hedyorg/hedy/blob/main/tests/cypress/e2e/public_programs/programs_page.cy.js 
- ...

Perhaps do a search for `data-cy` in all .cy.js files, so that you can change them everywhere!

Annelein:
@hasan-sh Thanks for the contribution, looks good! Can you fix the tests? Then it's ready I think

hasan-sh (approved):

PR #5577, opened 2024-05-28
🪲 Fix flaky programs test

### LINKED ISSUE #5578: 🪲 Flaky programs test
I introduced a flaky test on #5561 that is affecting some PRs. It is related to using the `loginForTeacher` function twice.

jpelay:
I added a flaky test by accident while fixing the filters. I did this by logging twice, which caused errors due to it having to wait, so it worked on local but not on GH Actions. 

To solve this, I removed the hardcoded waits and replaced them with intercepts. Everything should work as expected

Fixes #5578

### DIFF
--- a/app.py
+++ b/app.py
@@ -605,40 +605,12 @@ def parse():
             if transpile_result.has_music:
                 response['has_music'] = True
 
+            if transpile_result.has_sleep:
+                response['has_sleep'] = True
+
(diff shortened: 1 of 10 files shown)


Felienne (approved):
Thanks, let's hope this also unlocks the other stuck PRs!

Felienne:
Still flaky 😭 
image

PR #5582, opened 2024-05-30
🪲 Fix failing unit tests 

### LINKED ISSUE #5572: 🪲 Translation of booleans cause unit tests to fail
**Describe the bug**
Somehow we managed to get unit tests broken on main: https://github.com/hedyorg/hedy/actions/runs/9264577223/job/25484929454?pr=5570

I am not sure why has not our tests broken sooner

boryanagoncharenko:
Fixes #5572
Somehow 2 unit tests related to booleans are still failing.

**How to test**
Ensure that all unit tests for level 15 pass locally when the snippet_already_tested_with_current_hedy_version method always return False.


Felienne:
Not a big issue but why are there images added in this pr?

boryanagoncharenko:
> Not a big issue but why are there images added in this pr?

I am not sure what these are and why they are generated. I manually removed them now (and also in the previous commit). I will look at what is causing them to generate.

PR #5587, opened 2024-06-03
💻 Implemented "is" and sleep command for Micro:bit for level 2 

rmagedon97:
Added the assignment to a variable from level 2 in Hedy to Micro:bit
Added sleep functionality for Micro:bit

**How to test**

Open Hedy and choose level 2. Assign a variable and then print it. This will generate a hex file that can be uploaded here:
https://python.microbit.org/v/3/reference. 
You should see the correct python code after you upload the file. 


hedy.py
```diff
@@ -1889,13 +1890,19 @@ def make_print_ask_arg(self, arg, meta, var_to_escape=''):
...
+            # this regex splits words from non-letter characters, such that name! becomes [name, !]
+            p = r"[·\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}]+|[^·\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}]+"
+            res = regex.findall(p, arg)
+            return ''.join([self.process_variable_for_fstring(x, meta.line, var_to_escape) for x in res])
+        else:
+            if self.is_variable(arg, meta.line):
+                return arg
+            else:
+                # If the argument is not a variable, return it as a string literal with quotes
+                return f"'{arg}'"
```
boryanagoncharenko, line 1905: If the argument contains a single quote, this code would not transpile correctly. Luckily, we already have a function that escapes single quotes. Consider changing this line to:
`return f"'{process_characters_needing_escape(arg)}'"`

hedy.py
```diff
@@ -1932,31 +1939,56 @@ def play(self, meta, args):
     def assign(self, meta, args):
         variable_name = args[0]
         value = args[1]
-
-        if self.is_random(value) or self.is_list(value):
-            exception = self.make_index_error_check_if_list([value])
-            return exception + variable_name + " = " + value + self.add_debug_breakpoint()
+        if not self.microbit:
+            if self.is_random(value) or self.is_list(value):
```
boryanagoncharenko, line 1943: I like that the new microbit code path is non-intrusive. The code that needs to be different is copied in an if-microbit-else statement. My only concern is that the current code of the assign function needs to be refactored because it contains useless checks and repetition. I'd update the original code of the function in the following way: 
```
            exception = self.make_index_error_check_if_list([value])
            if self.is_variable(value, meta.line):
                # if the assigned value is a variable, this is a reassign
                value = self.process_variable(value, meta.line)
            else:
                # if it is not a variable and contains single quotes, escape them
                value = f"'{process_characters_needing_escape(value)}'"
            return exception + variable_name + " = " + value + self.add_debug_breakpoint()
```
This would also simplify the microbit part, I think. However, if this refactoring is too much for the current task at hand, please close this comment and I will add a note to refactor this piece of code at a later stage.

hedy.py
```diff
@@ -1932,31 +1939,56 @@ def play(self, meta, args):
...
+                    # if the assigned value is not a variable and contains single quotes, escape them
+                    value = process_characters_needing_escape(value)
+                    return variable_name + " = '" + value + "'" + self.add_debug_breakpoint()
         else:
-            if self.is_variable(value, meta.line):  # if the value is a variable, this is a reassign
-                value = self.process_variable(value, meta.line)
-                return variable_name + " = " + value + self.add_debug_breakpoint()
+            if self.is_random(value) or self.is_list(value):
+                exception = self.make_index_error_check_if_list([value])
+                return "    " + exception + str(variable_name) + " = " + value + self.add_debug_breakpoint()
```
boryanagoncharenko, line 1957: This line is problematic because the transpiled code of the exception is multiple lines (at least 4). The make_index_error_check_if_list method does not have any microbit logic, so this will just offset the first line of the snippet.

Perhaps I am missing context here but I wonder why with microbit the transpiled commands start with spacing? I suspect this may cause big issues when we transpile functions and definitions at a later stage (if this is in scope). Would it be an option to transpile and then indent the whole program?

hedy.py
```diff
@@ -1932,31 +1939,56 @@ def play(self, meta, args):
...
         else:
-            value = f'"{args[0]}"' if self.is_int(args[0]) else args[0]
-            if not self.is_int(args[0]):
-                self.add_variable_access_location(value, meta.line)
-            index_exception = self.make_index_error_check_if_list(args)
-            ex = make_value_error(Command.sleep, 'suggestion_number', self.language)
-            code = index_exception + \
-                textwrap.dedent(f"time.sleep(int_with_error({value}, {ex})){self.add_debug_breakpoint()}")
-            return code
+            if not args:
```
boryanagoncharenko, line 1981: For this command the microbit code is quite different than the original non-microbit one. In this case, there is not much gain from the fact that the 2 code snippets (the microbit and the non-microbit one) are right next to each other. Have you considered instead of using `if` statements, to subclass e.g. 
```
@v_args(meta=True)
@microbit_hedy_transpiler(level=2)
@source_map_transformer(source_map)
class MicrobitConvertToPython_2(ConvertToPython_2):
   def sleep(self, meta, args):
      ...
```
In this way the microbit code will be together and the code will not have if-microbit-else statements. However, the downside is that the microbit code might be overlooked when changes are made to the main transpiler.

If you have considered this option and rejected it, please feel free to close this comment. If you haven't and you want some help to explore it further, do not hesitate to ping me.

hedy.py
```diff
@@ -2830,6 +2814,168 @@ def print_empty_brackets(self, meta, args):
         return self.print(meta, args)
 
 
+@v_args(meta=True)
```
boryanagoncharenko, line 2817: After some thought and a brief micro:bit research, I think it will be best to have a separate stack of transpilers for microbit. The main motivation is that most commands will be transpiled in the same way, but those that need to be changed would be very different from the original code. For example, micro:bit does not have support for f-strings (right?), which means that the print and ask commands would need to be handled differently.

I already set up a framework for the new transpilers. Every microbit transpiler class for level N will inherit from the microbit transpiler of level N-1 and the original transpiler for level N. In this way, the new code would need to be minimal.

hedy.py
```diff
@@ -2830,6 +2814,168 @@ def print_empty_brackets(self, meta, args):
...
+            return f"sleep(1000){self.add_debug_breakpoint()}"  # Default 1 second sleep in milliseconds
+        else:
+            value = args[0]
+            if self.is_int(value):
+                # Direct conversion of seconds to milliseconds
+                milliseconds = int(value) * 1000
+            else:
+                # If the value is a variable, ensure it's used correctly
+                milliseconds = f"{value} * 1000"
+            return f"    sleep({milliseconds}){self.add_debug_breakpoint()}"
```
boryanagoncharenko, line 2854: I think you are adding whitespace in front of commands because the whole program will need to be placed in a `while True:` loop. Is this correct? If so, I am afraid adding whitespace here will not work. Instead, you need to transpile the whole program and then indent it right before you append it to the existing `while True:` program.

boryanagoncharenko (changes requested):
As agreed during the meeting with Teodor:
- I am converting both PRs to draft, since some changes are needed
- All added whitespacing needs to be removed from the microbit transpiler and added where the program is put together with the `while True:` cycle.
- Tests need to be added for the commands that are changed (the extent depends on the overall status of the project)

hedy.py
rmagedon97: After disusing this issue and choosing to have a separate transpiler for Micro:bit this functions no longer needs to be modified for Micro:bit.

hedy.py
rmagedon97: As we discussed in our meeting i was adding white spaces in the transpiler so that the output can be correctly formatted for micro-bit. However, we chose a different approach. The white spaces if needed will be handled in app.py in the save_transpiled_code_for_microbit function.

hedy.py
rmagedon97: As we discussed this will be the new approach with a separate transpiler for microbit.

hedy.py
rmagedon97: Thank you for pointing out this approach. As we discussed and agreed in our meeting I'll use it and continue with it. Thank you once again for writing the skeleton for the microbit transpiler. I'll continue to add more functionality to it.

hedy.py
rmagedon97: As discussed in out meeting I'll get rid of the white spaces in the transpiler and I'll handle them in Hedy.app instead.

jpelay:
While looking at your code I see that you generate a file in the backend  and then send it to the frontend, where it downloads. What would happen if there are concurrent calls to the endpoint? I'm pretty sure that will cause either slowdowns or race conditions. I think that a much better approach would be to send the transpiled code to the front-end and generate the file directly in the user's machine, this will prevent having to write to discs for every call.

Felienne:
> While looking at your code I see that you generate a file in the backend and then send it to the frontend, where it downloads. What would happen if there are concurrent calls to the endpoint? I'm pretty sure that will cause either slowdowns or race conditions. I think that a much better approach would be to send the transpiled code to the front-end and generate the file directly in the user's machine, this will prevent having to write to disc for every call.

Do we do that for dst files too (if not, that is probably a wise idea!) Maybe we can combine this into one part of the code that does that?

jpelay:
> Do we do that for dst files too (if not, that is probably a wise idea!) Maybe we can combine this into one part of the code that does that?

I just checked and for those files we generate a unique random filename, this would also prevent race conditions in this case!

hedy.py
```diff
@@ -2832,6 +2816,176 @@ def print_empty_brackets(self, meta, args):
...
+        if not args:
+            return f"sleep(1000){self.add_debug_breakpoint()}"  # Default 1-second sleep in milliseconds
+        else:
+            variable = args[0]
+            if self.is_int(variable):
+                # Direct conversion of seconds to milliseconds
+                milliseconds = int(variable) * 1000
+            else:
+                milliseconds = f"{variable} * 1000"
+                return f"sleep({milliseconds}){self.add_debug_breakpoint()}"
```
boryanagoncharenko, line 2876: This return statement should not be in the else clause but on the same level as the if-else clause Currently the transpiler for returns None for the program `sleep 2` and the tests are failing.

tests/test_level/test_level_02.py
```diff
@@ -557,6 +557,25 @@ def test_sleep(self):
...
+
+    def test_sleep_micro_bit_no_argument(self):
+        code = "sleep"
+        expected = "sleep(1000)"
+
+        self.multi_level_tester(code=code, expected=expected, max_level=5, microbit=True)\
+
+
+    def test_sleep_micro_bit_variable_argument(self):
+        code = " duration = 2\nsleep duration"
```
boryanagoncharenko, line 574: This isn't a valid Hedy program for 2 reasons: 1) there is a space in the beginning of the program and 2) the `=` should be changed to `is`

hedy.py
```diff
@@ -2832,6 +2816,176 @@ def print_empty_brackets(self, meta, args):
...
+    def sleep(self, meta, args):
+        if not args:
+            return f"sleep(1000){self.add_debug_breakpoint()}"  # Default 1-second sleep in milliseconds
+        else:
+            variable = args[0]
+            if self.is_int(variable):
+                # Direct conversion of seconds to milliseconds
+                milliseconds = int(variable) * 1000
+            else:
+                milliseconds = f"{variable} * 1000"
```
boryanagoncharenko, line 2875: If this is a variable (and not an int), you need to register its usage. Otherwise, the tests will fail with an unused variable. Use this code to register a variable usage: `self.add_variable_access_location(variable, meta.line)`

hedy.py
rmagedon97: /0Thank you for that I'll fix it :)

tests/test_level/test_level_02.py
rmagedon97: Great catch didn't even realized it. Thank you.

hedy.py
rmagedon97: Great job, now the three tests are passing all levels. Thank you for all the help.

hedy.py
```diff
@@ -2832,6 +2816,176 @@ def print_empty_brackets(self, meta, args):
...
+        variable_name = args[0]
+        value = args[1]
+
+        exception = self.make_index_error_check_if_list([value])
+        if self.is_variable(value, meta.line):
+            # if the assigned value is a variable, this is a reassign
+            value = self.process_variable(value, meta.line)
+        else:
+            # if it is not a variable, check if it is a number
+            if value.isdigit() or (value.startswith('-') and value[1:].isdigit()):  # check for integers
```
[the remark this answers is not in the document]
rmagedon97: Thank you for pointing that out. I'm now using the existing Hedy functions.

rmagedon97:
> While looking at your code I see that you generate a file in the backend and then send it to the frontend, where it downloads. What would happen if there are concurrent calls to the endpoint? I'm pretty sure that will cause either slowdowns or race conditions. I think that a much better approach would be to send the transpiled code to the front-end and generate the file directly in the user's machine, this will prevent having to write to disc for every call.

That's a good idea. I basically did the same was as we are doing the files for the turtle. I had some issues converting the file to hex that's why i used he same name for the file. After your comment i'll try to use a random file name.

tests/test_level/test_level_02.py
```diff
@@ -654,6 +673,12 @@ def test_assign(self):
 
         self.multi_level_tester(code=code, expected=expected, max_level=11, unused_allowed=True)
 
+    def test_assign_micro_bit(self):
+        code = "naam is Felienne"
+        expected = "    naam = 'Felienne'"
```
boryanagoncharenko, line 678: The test fails locally. I think there shouldn't be any leading whitespace. Please make sure that all tests are passing and when you are ready, convert the PR to ready for review. Then, I think we can move to the other PR

rmagedon97: Yes indeed i ran only the new tests. After i ran all tests locally there was one more failing test as well. Fixed that but will check few things before i convert it for review.

rmagedon97: Everything should be fine now until level 4

rmagedon97 (commented):
fixed

boryanagoncharenko (approved):

PR #5601, opened 2024-06-06
🧪 Run slow test suites for all PRs with caching

### LINKED ISSUE #5573: 🪲 Unit tests do not appear to be broken while they are
**Describe the bug**
The unit tests on main have been broken for 3 days but we only notice because they fail in a weblate PR. Running the tests locally also does not yield errors. However, only when we remove the test caching, the tests fail.

**How to reproduce**
Checkout commit 25d3d4f123b6cfe5318a22abff968932ac8f661a, commend out the snippet_already_tested_with_current_hedy_version content and run test_level_15.py.

boryanagoncharenko:
- Add all hedy_*.py files to the source code cache generation.
- Enable slow test suites to run for all PRs, with caching.

Fixes #5573
**How to test**
- Ensure that unit tests job on GitHub runs all tests. There should be 2 runs: 1) ~8000 tests and 2) ~23000 tests. Note that the first run will be very slow. All consecutive runs should be quicker.


jpelay (approved):
Thank you Boryana!

rmagedon97:
> l>  l- Too@@@[lop` ll
Ulpl

PR #5605, opened 2024-06-07
💻 Reduce achievements time

jpelay:
While working on another PR I found out that a new user that tries to create a class has to wait a lot of time for their first class to show. This happens because the `showAchievements` functions has some timeouts of 6 seconds, so the user, as a minimum has to wait this long if they get an achievement. 

Therefore I reduced the time of the achievements a bit.

**How to test**

* Log in as teacher3
* Create a class
* The class shouldn't take ages to load

Other way to test:

* Log in as teacher1
* Duplicate a class
* The for teachers page should reload quickly

You can try that on `main`  and see the difference (remember to clear the site data and restore the devdb in order for the achievement to be shown again)


static/js/app.ts
```diff
@@ -786,6 +786,7 @@ export async function delete_program(id: string, prompt: string) {
     // this function decreases the total programs saved
     updateProgramCount();
     const response = await postJsonWithAchievements('/programs/delete', { id });
+    console.log(response)
```
Annelein, line 789: Can be removed?

jpelay: Done!

Annelein (approved):

PR #5612, opened 2024-06-13
🪲 Fix error page

### LINKED ISSUE #5598: 🪲 Clicking a link to a program sometimes results in an error
Here is an example:

image

Id: afab870719be4310a8ed936d6365470a

(and as you can see, the error message is also broken, not a biggie but ugly)

jpelay:
The error page was failing because it was trying to retrieve an image that wasn't there.

Fixes #5598

**How to test**

* Go to http://127.0.0.1:5000/hedy/a24e2453366248fb9359fffa3089d209/view whilst logged out and see that the image and the message is shown.

### DIFF
--- a/templates/error-page.html
+++ b/templates/error-page.html
@@ -3,7 +3,7 @@
 {% block regular_content %}
   <div class="flex flex-col items-center">
       <div class="w-1/2 mt-8">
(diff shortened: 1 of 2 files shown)


Felienne:
Thanks for fixing!! I am puzzled though how did this got broken? Did we rename the image?

Felienne (approved):

PR #5614, opened 2024-06-14
Roles of variables

juliabolt:
Please replace all _____ before you create a pull request!

Start the name of the PR with one of the relevant prefixes in the title of the PR:

* 🖊️ -> changes related to grammars or the transpiler
* 🪲 -> solving a bug
* 🚚 -> changes unrelated to shipping, such as mergify scripts
* 🧹 -> refactorings (changes that do not change functionality and are only a cleanup of code)
* 💻 -> improvements and changes of the Hedy website
* 🧪 -> changes made to the test suite
* 📜 -> changes related to content

Fixes _______ (Always link the number of the issue or of the discussion that your PR concerns. If you use the word `fixes` before the issue number in this description, the related issue will automatically close then the PR is merged)

**How to test**

* _____ (See this https://github.com/hedyorg/hedy/pull/880#issue-1016304308 for an example)
* _____

**Checklist**

### DIFF
--- a/app.py
+++ b/app.py
@@ -2918,6 +2918,7 @@ if __name__ == '__main__':
     env_defaults = dict(
         BASE_URL=f"http://localhost:{config['port']}/",
         ADMIN_USER="admin",
+        SHOW_ROLES=False,
     )
     for key, value in env_defaults.items():
         if key not in os.environ:
--- a/static/js/debugging.ts
+++ b/static/js/debugging.ts
@@ -8,7 +8,7 @@ let TRADUCTION: Map<string,string>;
 
 //Feature flag for variable and values view
 let variable_view = true;
-let showRoles = false;
+let showRoles = true;
 let step_debugger = false;
 const fullLineCommands = [
   'print',
--- a/website/auth.py
+++ b/website/auth.py
@@ -159,6 +159,13 @@ def forget_current_user():
     session.pop("keyword_lang", None)  # Delete session keyword language if existing
     session.pop("profile_image", None)  # Delete profile image id if existing
 
+def show_roles_of_variables():
+    """Whether the roles of variables should be shown in the variable view.
+    (If not, only variable names and values will be shown in the list).
+    Checks the configuration in the environment variable $SHOW_ROLES    
+    """
+    can_show_roles = os.getenv("SHOW_ROLES")
+    return can_show_roles
 
 def is_admin(user):
     """Whether the given user (object) is an admin.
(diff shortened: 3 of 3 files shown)


website/auth.py
```diff
@@ -160,6 +160,15 @@ def forget_current_user():
     session.pop("profile_image", None)  # Delete profile image id if existing
 
 
+def show_roles_of_variables():
+    """Whether the roles of variables should be shown in the variable view.
+    (If not, only variable names and values will be shown in the list).
+    Checks the configuration in the environment variable $SHOW_ROLES    
+    """
+    can_show_roles = os.getenv("SHOW_ROLES")
```
jpelay, line 168: If `SHOW_ROLES` is not set you can also add a default value:

```suggestion
    can_show_roles = os.getenv("SHOW_ROLES", False)
```

app.py
```diff
@@ -2918,6 +2918,7 @@ def on_offline_mode():
     env_defaults = dict(
         BASE_URL=f"http://localhost:{config['port']}/",
         ADMIN_USER="admin",
+        SHOW_ROLES=False,
```
jpelay, line 2921: Great!!

jpelay (commented):
Now you can do something like this
```
response['variables'] = transpile_result.roles_of_variables
response['show_roles'] = show_roles_of_variables()
```

And then check for that in the front-end

website/auth.py
jpelay: Oh I just noticed this is in `auth.py` can you move it to `utils.py`?

static/js/debugging.ts
```diff
@@ -79,6 +79,7 @@ export function load_variables(variables: any) {
       if (variables[i][1]) {
         const variableName = variables[i][0].replace(/^_/, '');
         const role = programData?.variables[variableName];
+        const showRoles = programData?.show_roles;
```
jpelay, line 82: I think you're having a problem with this, because you need to add `program_data`to the list of variables!

juliabolt: I already had `const role = programData?.variables[variableName] ` in the code when I still used the feature flag in the frontend, then things worked fine...

juliabolt:
I said in the export function runPythonProgram (file app.ts) that the type of showRoles is a boolean, but now it needs be a string then?

jpelay:
> I said in the export function runPythonProgram (file app.ts) that the type of showRoles is a boolean, but now it needs be a string then?

If you just send the variable it'll be a string, you can do something like:

```
show_roles_of_variables() == "False" 
```

and that'll always send a boolean

jpelay:
> > `show_roles_of_variables() == "False"`
> > and that'll always send a boolean
> 
> If I do this, then the boolean will become true when the string is "False" right? I'm confused...

Yes, you're right! It's `== True` 😅 

PD: I edited your comment by accident, sorry!

juliabolt:
some tests keep failing though... Am I still doing things wrong with the env var?

PR #5616, opened 2024-06-16
📜 Improve error message structure and consistency

ArtV11:
This pull request is part of a larger project in which we are redesigning our error messages according to [these guidelines](https://github.com/hedyorg/hedy/wiki/Hedy-Language-Design#error-messages).

In this PR, I have worked on the redesigning topic (i) Structure. We aim to follow a consistent structure for all of our error messages. First sentence will inform our users of the issue, while the second sentence will offer a suggestion on how to resolve it. This approach helps users quickly understand where to look when something goes wrong and allows them to become accustomed to reading the error messages more efficiently due to their uniform structure. I added a suggestion on how to fix the problem for every error message while at it I also worked on (ii) Consistency. We aim to keep our error messages consistent by using the uniform wording and maintaining a similar sentence structure and length. Errors should follow the same format ensuring that the sentences start and end similarly. This consistency also aids in the readability and the structure of the error messages. because I was already creating new sentences while adding the suggestion and wanted to keep their wording consistent but the consistency aspect still needs some work for the future work.

Addresses https://github.com/hedyorg/hedy/issues/5012


translations/en/LC_MESSAGES/messages.po
```diff
@@ -19,133 +19,133 @@ msgstr ""
...
-msgstr "On line {line_number} you used an `{else}` but there is no `{if}` on the line before it."
+msgstr "Command `{else}` used before using an `{if}` on line {line_number}. Try writing `{if}` before writing `{else}`."
 
 msgid "Function Undefined"
-msgstr "You tried to use the function {name}, but you didn't define it."
+msgstr "Function {name} used without being defined. Try to define the function before use."
 
 msgid "Has Blanks"
-msgstr "Your code is incomplete. It contains blanks that you have to replace with code."
+msgstr "Code is incomplete. Try filling in blanks with code."
```
Felienne, line 34: Putting this on this one but it is a general remark, I don't believe this is better. It feels very abrasive and short, I miss the "human communication" that we had before, this feel very much like communication with a machine. Can we make it more conversational?

translations/en/LC_MESSAGES/messages.po
```diff
@@ -19,133 +19,133 @@ msgstr ""
...
-msgstr "It looks like your `{command}` is missing a variable at the start of the line."
+msgstr "Command `{command}` is missing a variable at the start of the line. Try writing the variable before the `{command}`"
 
 msgid "Misspelled At Command"
-msgstr "It looks like you might have misspelled the `{command}` command, instead you wrote `{invalid_argument}` in line {line_number}."
+msgstr "`{invalid_argument}` seems to have a spelling mistake on line {line_number}. Try writing `{command}` instead."
 
 msgid "No Indentation"
-msgstr "You used too few spaces in line {line_number}. You used {leading_spaces} spaces, which is not enough. Start every new block with {indent_size} spaces more than the line before."
+msgstr "Insufficent {leading_spaces} spaces used on line {line_number}. Try increasing by {indent_size} for each new block."
```
Felienne, line 88: Insufficient is a complex word! Maybe we should include target audience (10 and up) explicitly in the guidelines too and live by that?

Felienne:
I like the new structure, thanks! Will we do wording (like `Insufficient`) in this PR too? Otherwise it will block us from deploying.

Also (I am sure you know) tests are still failing, can you take a look?

jpelay (commented):
Thanks for making them more consistent! I like the direction this is heading. I suggested some changes to some of the messages.

translations/en/LC_MESSAGES/messages.po
```diff
@@ -19,133 +19,133 @@ msgstr ""
 "Generated-By: Babel 2.14.0\n"
 
 msgid "Access Before Assign"
-msgstr "Variable {name} used on line {access_line_number} before being set. Try setting the variable before use."
+msgstr "We detected that a variable {name} is being used on line {access_line_number} before being set. Can you set the variable before it is used?"
```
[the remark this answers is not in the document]
Felienne: Hi Jesus! 

Are we sure that tickmarks in the error message actually make the keywords highlighted (can you double check?) I though there was some kind of magic going on somewhere in the backend with some extra HTML being injected?

Felienne:
Hi @jpelay!

I am ok with all these changes, I am just not sure about the backtick doing what you think it does. If you can confirm it does, you may approve!

jpelay:
> I am ok with all these changes, I am just not sure about the backtick doing what you think it does. If you can confirm it does, you may approve!

Ohh sorry :c, I thought the backticks were necessary to show words as code in the error messages. What are they meant to do?

Felienne:
> > I am ok with all these changes, I am just not sure about the backtick doing what you think it does. If you can confirm it does, you may approve!
> 
> Ohh sorry :c, I thought the backticks were necessary to show words as code in the error messages. What are they meant to do?

I don't remember! ~Someone should read to code to figure it out :)~

Ok, I dove in a bit. I think you are right, partly, it does work like that, but there is also back-end magic involved as I thought, see here:

https://github.com/hedyorg/hedy/blob/063d197f5eb75b5c0505630129044b9a777f610f/hedy_error.py#L63

So it is not like just simple `` in markdown, which is what I remembered.

jpelay:
> Ok, I dove in a bit. I think you are right, partly, it does work like that, but there is also back-end magic involved as I thought, see here:
> 
> https://github.com/hedyorg/hedy/blob/063d197f5eb75b5c0505630129044b9a777f610f/hedy_error.py#L63
> 
> So it is not like just simple `` in markdown, which is what I remembered.

Ahhhhh, I see! Thanks for clarifying!

PR #5617, opened 2024-06-17
💻 Super teacher can delete public tags

Annelein:
- teacher1 is now always a super teacher in the dev_database
- when a tag was deleted from an adventure, the popularity wasn't decreased
- when you would add a tag to an adventure without a name, it would throw a 500 because the adventure wouldn't exist without a name
- some cleaning up

**How to test**
Add some tags in (public) adventures and delete them as a Super Teacher
Screenshot 2024-06-17 at 14 38 03


jpelay (approved):
Great work!

PR #5620, opened 2024-06-18
🧹 Tags: remove double get, sort tags

rix0rrr:
Two tiny changes ported over from #5431:

- Remove a duplicate database retrieval when looking up the tags of an adventure.
- Sort the tags (when they are saved to the database).

**How to test**

Unsorted tags should be sorted after adding a tag. Otherwise no functional difference.


Felienne:
Hi @rix0rrr!

This has a failing Cypress test that I suspects fails for good reasons, can you take a look?

boryanagoncharenko (approved):

PR #5625, opened 2024-06-21
🪲 Add curly braces around `add ... to` command

jpelay:
If you used the `add` command in one of the adventures, when adding the curly braces it was translated as `{add} el {toList} l` which caused the problem of the adventure not being shown correctly since `toList` is not the name of the keyword on our translation dictionaries. I changed the name of the keyword in the grammars.

**How to test**

I added a test for this scenario.


boryanagoncharenko (approved):
Works great! But I have to admit I had to reverse-engineer how the problem manifested.

jpelay:
> Works great! But I have to admit I had to reverse-engineer how the problem manifested.

Sorry for not being clear enough in the PR. In fairness I only realized because I used the script to fix the teacher manual in #5167 and a particular program failed to be properly formatted in every language.

PR #5626, opened 2024-06-22
🪲 DynamoDB queries fail on field with `#` in name

rix0rrr:
`#` is not allowed in a name in key expressions, so we need to remove it from the symbolic names.

Before:

```
KeyExpression = "#id#level = :id#level AND #week BETWEEN :week_min AND :week_max"
```

After:

```
KeyExpression = "#id_level = :id_level AND ..."
```

Also taking this opportunity to make the handling of conditions consistent.

**How to test**

This didn't use to work, but now should work:

```python
   data = PROGRAM_STATS.get_many({'id#level': id + '#' + str(level), "week": dynamo.Between(start_week, end_week)})
```


jpelay (approved):
Thank you!

PR #5630, opened 2024-06-25
🧹 Initialize database on app start-up

### LINKED ISSUE #5609: 🪲 Initialize DB tables on app inizialitation
**Describe the bug**
On #4862 I did a little hack by setting the  `DEBUG_MODE` variable before `main` is executed. This is not ideal and @rix0rrr suggested a better alternative:

> I think the better solution is probably to control the initialization order of all these things better, by making the tables class members of Database and initializing it when the app is created

jpelay:
Moves the database initialization to the constructor of the Database object, and then calling it on `__main__`, this way we have a bit more control in the order of execution of these steps.

Fixes #5609 

**How to test**

All tests should pass, and the database type checker should work as well.


rix0rrr (approved):
❤️

PR #5631, opened 2024-06-26
💻 Remove modal for make teacher as admin

Annelein:
As requested by @Felienne : remove modal for accepting teacher requests and remove reloading the page every time


website/admin.py
```diff
@@ -173,58 +174,48 @@ def get_admin_achievements_page(self, user):
             page_title=gettext("title_admin"),
         )
 
-    @route("/mark-as-teacher", methods=["POST"])
-    def mark_as_teacher(self):
+    @route("/mark-as-teacher/<username_teacher>", methods=["GET"])
```
jpelay, line 177: I think the problem it's happening because these calls were switched from GET to POST. And, I would prefer these calls to be POST, because we are not getting any data from the server, just modifying, which is exactly what post calls are meant for.

static/js/index.ts
```diff
@@ -13,6 +13,7 @@
  */
 import './htmx-integration';
 export * from './modal';
+export * from './admin';
```
jpelay, line 16: Could you remove this import? I don't see it being used.

jpelay (changes requested):

website/admin.py
Annelein: I've changed it back to post!

jpelay (approved):

PR #5634, opened 2024-06-27
💻 teacher mode for not logged in users

### LINKED ISSUE #5183: 💻 teacher mode for not logged in users
**Idea**
We want to make it possible for users to view our platform as an approved/signed in teacher would. To make this, 
we could create a teacher account that would be used as a previewing account that gets cleaned after each session. Perhaps any other approach would be fine if we achieve the goal. 

Moreover, we would like to show a bar similar to that we should when a teacher previews a class and inform them that they need to sign up in order to create persistent content. Whether what they'd created would be saved or not, we will come back once needed.
 
The user story would be something like:
- users click on "view teacher mode" within the landing page,
- (perhaps a modal to verify users' age?!)
- they can view teacher's page or perhaps start a tutorial right away,
- they would have the same rights as a teacher does.
- a message on the bar should be clear about the content with a sign up button.

@Felienne, feel free to add or edit anything!

hasan-sh:
Fixes #5183 

**How to test?**
- you should be logged out
- click on start teaching in the landing page
- test the teacher mode.

> We need to add a `PREVIEW_TEACHER_MODE_PASSWORD` env. variable.


Annelein (changes requested):
@hasan-sh really nice and awesome to have this, thanks! Only a few things that I think are not working properly yet:
- if I'm in teacher mode -> customize a test class -> add a student -> login as that student (in an incognito page) -> on the page where in teacher mode, don't exit teacher mode but **logout**  -> go back to the student adventure page, the customizations are still there. Only when I click on teacher mode again on the home page, the student is deleted from the 'preview mode' test class and the customizations are restored. Logging the student out and back in also doesn't work. It only works when clicking exit preview mode. But also: do we want to be able to add existing students to a class? Hope you understand this one, otherwise ping me on Discord!
- if I'm in teacher mode and I create a public profile, it never gets deleted. Not sure how we would want this. Maybe this teachermode account should just never be able to create a public profile?
- if I'm in teacher mode and I create a program or add information to my account (such as email, birthyear, etc) it also never gets deleted. I guess we also want this to be deleted, instead of people filling in personal information by accident and someone else being able to see it. Regarding the programs, it would probably get really messy if everyone adds random programs so might also be nice to delete those when exiting / loggin out the teacher preview mode?
- I can flag (public) programs, do we want this to be possible when in teacher preview mode?

hasan-sh:
@Annelein changed the approach a bit and added extra needed code to the `forget_user` which was missing, check it out.

Annelein (approved):

Issue #5386, opened 2024-04-08
💻 Programs count design 


Annelein:
https://github.com/hedyorg/hedy/pull/5278 implements that only programs that are _modified_ are counted as "programs saved".

But this "programs saved" is old legacy. Because now only _modified_ programs are counted, so it should be changed to something like "programs tried".

This is what `my_programs` looks like now:
<img width="1440" alt="Screenshot 2024-04-08 at 14 12 15" src="https://github.com/hedyorg/hedy/assets/48122190/42fc2548-fc4e-423a-b077-67db085ab361">

The search boxes and the "programs tried" part should look be coherent to `public_profile`. The only differents between the two pages would be the parts that have to do with a profile, because `my_programs` can be seen without a having your profile set to public.

<img width="1440" alt="Screenshot 2024-04-08 at 14 11 03" src="https://github.com/hedyorg/hedy/assets/48122190/af12944f-7956-4280-99c4-4927ce572c5d">

@gisellandrade can you propose a design for this? Let me know if there is something you don't understand!

gisellandrade:
Thank you so much for the explanation, I'll be working on this. I'm thinking about redesigning this ❤️

Annelein:
> Thank you so much for the explanation, I'll be working on this. I'm thinking about redesigning this ❤️

Yes a redesign would be amazing!

Felienne:
@AnneliesVlaar  and @MarleenGilsing: is this still needed or is this superseded by all the other new designs?

AnneliesVlaar:
It should be superseded by the new designs, but I didn't had it on my radar. I will assign myself to keep it on the radar.

PR #5642, opened 2024-07-01
🪲 fix discrepancy programs count

### LINKED ISSUE #5199: 🪲 Discrepancy in program count
One more from the same school.

It seems teachers see a really different number of programs than students. Here is an example where that happens for user Sofie1bs:

IMG_7851

Other kids in the class also had counts over a hundred. But the teacher sees this:

image

hasan-sh:
Fixes #5199 

### How to test?
- login to some user; `user3`
- see the number of created programs in `/users/user3` and `/landing-page`
- they should match.
- run some other programs and perhaps submit some.
- revist the pages, and the numbers should match.


Annelein (changes requested):
Hey @hasan-sh , looks good and it works for user3! 

But if I login with teacher1 in /programs I have 6 programs saved and in /user/teacher1 there are only 2, and in /landing-page there are 7....

hasan-sh:
> Hey @hasan-sh , looks good and it works for user3!
> 
> But if I login with teacher1 in /programs I have 6 programs saved and in /user/teacher1 there are only 2, and in /landing-page there are 7....

Hey @Annelein , in the landing page, we forgot to add the `is_modified` check. Just added it! So, it should be the same as in `/programs` now.

For the `/user/<username>` page, the count isn't necessarly the same as the other two, because here we only see public programs. Perhaps change the label? Or is it clear @Felienne ?

Annelein:
> Hey @Annelein , in the landing page, we forgot to add the `is_modified` check. Just added it! So, it should be the same as in `/programs` now.
> 
> For the `/user/<username>` page, the count isn't necessarly the same as the other two, because here we only see public programs. Perhaps change the label? Or is it clear @Felienne ?

hey! @hasan-sh looks good and yes you are right, those are public programs so they are different. But I guess it is confusing because the label indeed is the same. What about x public programs | x programs saved | x programs submitted ?
Let me know if you don't understand what I mean!

hasan-sh:
> hey! @hasan-sh looks good and yes you are right, those are public programs so they are different. But I guess it is confusing because the label indeed is the same. What about x public programs | x programs saved | x programs submitted ? Let me know if you don't understand what I mean!

It already says x programs saved. 

What do you think @Felienne? 
image

Annelein:
> Not sure if it's logical to show x programs saved AND public programs, because this page is a public one and users don't need to see how many private&public programs a user has made.
> 
Then I would say, lets change the label, and make it say 'x public programs', to avoid confusions.

Felienne:
> > Not sure if it's logical to show x programs saved AND public programs, because this page is a public one and users don't need to see how many private&public programs a user has made.
> 
> Then I would say, lets change the label, and make it say 'x public programs', to avoid confusions.

Yes makes sense to me!

hasan-sh:
Ok, after further investigation, the count is actually correct. The problem is that the property `program_count` in the users table for `teacher1` was not updated in `data-for-testing.json`. This is probably because we added programs for testing purposes without updating the `program_count` field! I'll update it in the local database.

Meaning what you see under `x programs saved` actually reflects the total programs saved by the user. You can test this by loging in as user with no programs, run programs and see the count in the `public-page` reflected accordingly. 

The question remains do you still need an extra label that counts the number of public programs, even though the page lists all public programs by the user?

Annelein:
@hasan-sh @Felienne

Allright so the `x public programs` label should be added for sure, because thats whats being shown in the page.

And then for the `x programs saved` label, I think we can still keep it, I think users might find it cool to see/show the amount of programs saved? Let me know what you think!

Felienne:
> And then for the `x programs saved` label, I think we can still keep it, I think users might find it cool to see/show the amount of programs saved? Let me know what you think!

I don't think I agree. Because what does "saved" mean, if users never save anything actively? It used to be an action with a click (before either of you joined, I think, so you might not know all history)

A better label would be "adventures explored" or "tried" because that is what is semantically comes down to.

Or, if we want to keep it simple, we can do "public programs" and simple "programs" <-- My preferred option, I think.

Annelein:
> > And then for the `x programs saved` label, I think we can still keep it, I think users might find it cool to see/show the amount of programs saved? Let me know what you think!
> 
> I don't think I agree. Because what does "saved" mean, if users never save anything actively? It used to be an action with a click (before either of you joined, I think, so you might not know all history)
> 
> A better label would be "adventures explored" or "tried" because that is what is semantically comes down to.
> 
> Or, if we want to keep it simple, we can do "public programs" and simple "programs" <-- My preferred option, I think.

I agree! But that's a different discussion as to keeping this counter or removing it, I guess? Because the counter in `/programs ` and `/landing-page` also has the label `x programs saved`. Thats why I compared it to keeping it in this page. So then we would have to change those labels too right?

 So:
- keep the counter in `/user/teacher1'
- change all labels in to 'adventures explored' or 'adventures tried' or something
- add the counter `x public adventures`

Felienne:
> > > And then for the `x programs saved` label, I think we can still keep it, I think users might find it cool to see/show the amount of programs saved? Let me know what you think!
> > 
> > 
> > I don't think I agree. Because what does "saved" mean, if users never save anything actively? It used to be an action with a click (before either of you joined, I think, so you might not know all history)
> > A better label would be "adventures explored" or "tried" because that is what is semantically comes down to.
> > Or, if we want to keep it simple, we can do "public programs" and simple "programs" <-- My preferred option, I think.
> 
> I agree! But that's a different discussion as to keeping this counter or removing it, I guess? Because the counter in `/programs ` and `/landing-page` also has the label `x programs saved`. Thats why I compared it to keeping it in this page. So then we would have to change those labels too right?
> 
> I want to work on this for #5386 but there is not design proposal yet...
> 
> So:
> 
> * keep the counter in `/user/teacher1'
> * change all labels in to 'adventures explored' or 'adventures tried' or something
> * add the counter `x public adventures`

Yes, all these three things afaiac!

hasan-sh:
> "programs" <-- My preferred option, I think.

This makes a lot of sense since they don't really save indeed. Will change the `amount_saved` to simply be `programs`. And any further changes can be done in your PR @Annelein ?

Annelein (approved):

PR #5660, opened 2024-07-09
🪲 fix level dropdown in custom-adventure page

hasan-sh:
fixing:

https://github.com/hedyorg/hedy/assets/20051470/fb007c32-43fc-4fc6-a04f-c8e5e2be5214

**How to test?**
- deselect all levels of an adventure,
- an alert about not selecting any level should appear

### DIFF
--- a/static/js/teachers.ts
+++ b/static/js/teachers.ts
@@ -218,7 +218,10 @@ export function remove_student(class_id: string, student_id: string, prompt: str
 function get_formatted_content(content: string, levels: string[], language: string) {
   const parser = new DOMParser();
   const html = parser.parseFromString(content, 'text/html');
-  const minLevel = Math.min(...levels.map((el) => Number(el)));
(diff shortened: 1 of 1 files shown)


Annelein:
@hasan-sh the console error is gone, thats good! But do we want it to throw an error if there is no level selected in an adventure? Is it not fine if there is no level selected?

And the error message is: _"This Hedy level is invalid."_ But thats not really the case because there is no level selected so how can it be invalid, "there is no level selected" should be the error message I guess?

Correct me if I'm wrong or you don't agree!

Annelein:
> Correct. This should be defintely handled in the front-end as well and this was the case when we used the `tailwind-elements` and should be present in the custom-hedy select.

@hasan-sh ok, so this PR is ready for now and its handled in #5540?

Annelein (approved):

PR #5662, opened 2024-07-10
📜 Translate level 1 to pt_BR

brunosantanati:
I noticed some spots inside level 1 were still in English. I translated all of them to pt_BR.

**How to test**

You can start the server locally ( <code>doit run devserver</code> ) and go through level one, tab by tab including the quiz. Here is one screenshot of one of those texts that were in English and now is in Portuguese.

Screenshot from 2024-07-10 02-14-20

All tests passed, so everything is good:

Screenshot from 2024-07-10 02-24-48


Felienne (approved):
Thanks!!!

PR #5664, opened 2024-07-12
📜 Translate level 2 to pt_BR

brunosantanati:
**Translation to pt_BR and minor adjustments to the texts and variables**

- I noticed some spots inside level 2 were still in English. I translated all of them to pt_BR.
- I fixed one typo
- I fixed variables names
- I changed some texts to address gender of Portuguese words (for example all texts were assuming that the Hedy parrot was a female animal (because Hedy is the same name of the programming language which would be considered a female thing), so I changed from "papagaio" (male) to "papagaia" (female). It's kind of tricky to explain these nuances of Portuguese, but as a native speaker I can guarantee that now there is more harmony when putting all texts together, it makes much more sense for the reader. All the pieces fit together more naturally now.
- I also changed some texts that don't work in Portuguese. That's also a tricky one to explain. For example:
<code>You don't need to \`{print}\`</code>
when we have commands configured to Portuguese, the phrase above does not work well. It's because  \`{print}\` would be translated to Portuguese imperative (imprima) and the phrase would feels awkward. So I made adjustments as below:
<code>Você não precisa imprimir usando \`{print}\`</code>
the translation would be something like: You don't need to print (here I used the verb in Portuguese) using \`{print}\` (here I used the command that translates to the imperative form). Here is the UI showing this phrase for example:
Screenshot from 2024-07-12 13-02-15
- I also noticed that there were missing parts of some texts comparing the Portuguese versions of them to the English versions. In those cases I based my translation on the English version that were more complete.

**How to test**

You can start the server locally ( <code>doit run devserver</code> ) and go through level 2, tab by tab including the quiz.

I ran <code>python -m pytest -n autoto</code> and all tests passed.


Felienne:
Thanks for the extensive details on the translations @brunosantanati!

> * I changed some texts to address gender of Portuguese words (for example all texts were assuming that the Hedy parrot was a female animal (because Hedy is the same name of the programming language which would be considered a female thing), so I changed from "papagaio" (male) to "papagaia" (female). It's kind of tricky to explain these nuances of Portuguese, but as a native speaker I can guarantee that now there is more harmony when putting all texts together, it makes much more sense for the reader. All the pieces fit together more naturally now.
> * I also changed some texts that don't work in Portuguese. That's also a tricky one to explain. For example:
>   `` You don't need to `{print}` ``
>   when we have commands configured to Portuguese, the phrase above does not work well. It's because  `{print}` would be translated to Portuguese imperative (imprima) and the phrase would feels awkward. So I made adjustments as below:
>   `` Você não precisa imprimir usando `{print}` ``
>   the translation would be something like: You don't need to print (here I used the verb in Portuguese) using `{print}` (here I used the command that translates to the imperative form). Here is the UI showing this phrase for example:
>   Screenshot from 2024-07-12 13-02-15


Awesome! If you made decisions there that could be of relevance to other translators, you might want to add that to the wiki?

> * I also noticed that there were missing parts of some texts comparing the Portuguese versions of them to the English versions. In those cases I based my translation on the English version that were more complete.

Yes, English is always the place that is most complete and recent (or Dutch, haha!)

Felienne (approved):

Issue #5322, opened 2024-03-27
🪲 Forward command failing


boryanagoncharenko:
**Describe the bug**
In levels 2 and 3, the forward command cannot transpile correctly variables that are also reserved keywords, e.g. `all`, `time` etc.

**Paste the Hedy code & level**
Level 2
```
time is 10
forward time
```

**Add a screenshot (optional)**
<img width="521" alt="Screenshot 2024-03-27 at 12 07 20" src="https://github.com/hedyorg/hedy/assets/3010723/1d4b4522-9e23-43bc-8f63-34d71c94a1c8">

karimsellami2024:
i Have encountered the same problem , the bug is still present . i notice that there is another opened issue that dates back to november . is someone working on it or would it be possible for me to look for and suggest a solution ?

boryanagoncharenko:
I did not realize there is another issue opened for this bug, so I created this one.
I am assigning the issue to you. Feel free to have a look and if you need any help, I'd be happy to assist.

boryanagoncharenko:
Addressed in #5676

PR #5676, opened 2024-07-26
🖊️ Preserve localization information for numbers and booleans

### LINKED ISSUE #5548: 🖊️ Preserve language of booleans and numbers in output
Language idea incl level
In level 15 and up, we have support for boolean values. Unlike other literal values, booleans can be printed. And since #5547, the output is always the translated boolean keyword in the selected language instead of the literal value that was actually used in the program:
```
# lang is Bulgarian
a = True
print a
# outputs 'вярно'
```

Similarly, when the current language is Arabic, numbers are automatically converted to even when the numbers in the original program are Latin:
```
# lang is Arabic
print 1 + 1
# outputs `٢`
```

Instead, we would like to be able to preserve the original value (`True` in the former snippet and `2` in the latter)

boryanagoncharenko:
Fixes #5548

**Goal**
 
The goal of this PR is make Hedy understand and preserve localization information when using numbers and boolean values. In other words, if the user types `1 + 1` the output should be `2` and if the input is `١ + ١` the output should be `٢`. Similarly, if the user types `a = True\nprint a` the output should be `True` and if the input is `a = Вярно`, the output should be `Вярно`. The numeral system or the textual representation of the boolean values should not be based on a static setting somewhere in the user profile. The localization information of the input and output should be dynamic.

**Brief implementation notes**

To make Hedy understand what numeral system or boolean values the user is using, I had to change the basic types of the transpiled code. So before, when the user executes a = ١, we would transpile it to a = 1. Now, the variable needs to be assigned to a type that not only holds the real value of 1 but also the localisation information, e.g. that the Arabic numeral system is used. The value is now transpiled to a = Value(1, num_sys=‘Arabic’). Similarly, for booleans we store information about the textual representation of the value: b = waar becomes b = Value(True, bools={True: ‘waar’, False: ‘vals’}).

Later, when the variable is referenced, we decide whether to use the real value or the presentation value based on the command. For example, if the variable is used in + operator, we definitely need the real value (1 and True). However, if the variable is used in a print or ask command, we need its presentation (١ and waar).

Let me give you a few more examples, so that you get a better understanding of the changes:
```
pi is 3.14
print pi

# now becomes
 
pi = Value(3.14, num_sys='Latin')
print(f'''{pi}''')
```
```
x is 1 + 2
y is x + 3

# now becomes

x = Value(1 + 2, num_sys='Latin')
y = Value(sum_with_error(x, 3, """Runtime Values Error"""), num_sys=get_num_sys(x))
```
```
field = '.', '.', '.'
field at 1 = 'x'

# now becomes

field = Value([Value('.'), Value('.'), Value('.')])
field.data[int(1)-1] = Value('x')
```

**Issues addressed as a side-effect**
In order to make this change possible, a few issues had to be addressed on the fly:
- In levels 6 - 11 the following code works successfully in Python (prints 'jahoor') but fails to print in Skulpt. The reason is that Skulpt does not recognize the non-Latin numbers as numeric (i.e. '٢'.isnumeric() returns False). The issue now resolved.
```
nummer1 is ٢
nummer2 is 2
if nummer1 is nummer2 print 'jahoor!'
```
- In level 12 we say quotes need to be used everywhere, including in list definitions but we could add a non-quoted value to a list. Since this introduces complexity in the transpiler and no example ever uses this 'exploit', I have rewritten the grammar of list addition to match the one for list creation.
- In level 14 an equality check with undefined variable succeeds. The issue is now addressed and quotes are required for non-variable strings:
```
naam = 'Hedy'
if naam == Hedy # This should raise an undefined var error
  print 'hedy'
```
- In level 2 the following code produces an unused var warning incorrectly (#5322). Now the issue is addressed:
```
all is 50
turn all
```
- In level 2 print does not escape variable names when they are reserved words. Thus, the following program yields `sum` instead of 2. However, if we change the name of the variable to `sun`, it outputs 2.
```
sum is 2
print sum
```
- In levels 6 and up, the play command cannot take non-Latin numbers. The issue is now resolved. 
```
play ٣١
play ٣١ + ٣١
n = ٣١ 
play n
```

**How to test**

Few of the unit tests can verify that the desired output is produced and, also, there are substantial differences between Python and Skulpt. Thus, the only way to ensure that the change does not lead to regression is to execute the main scenarios of all commands in the browser and manually validate their outcome.

Please check that no other functionality is broken, e.g. variable list appears correctly, debugging works as expected, etc.


jpelay (approved):
I love the change, and that  you addressed so many issues with it.

I tried every example and all of them worked! I think this is ready to ship it :D

PR #5680, opened 2024-07-30
🪲 Increase timeout for if_pressed programs #5678

### LINKED ISSUE #5678: 🪲 If pressed randomly fails
**Describe the bug**
Using the if-pressed functionality randomly leads to errors.

In level 7, try the following snippet. Ensure that you wait a few seconds before pressing 3 times a button:
```
repeat 3 times if x is pressed print 'x' else print 'another'
```
Note that the program terminates prematurely, sometimes on the second keypress.

Screenshot 2024-07-29 at 13 21 10

boryanagoncharenko:
Fixes #5678
Increased the timeout of if_pressed programs from 10 seconds to 10 minutes. The random failures turned out to be timeouts. While cleaning up the pygame code, I think we accidentally reduced the timeout for if_pressed programs. This surfaced as random failures especially when there are multiple if_pressed executions, e.g. in a `repeat` command.

**How to test**
Go to level 7 and run the following program:
```
repeat 3 times if x is pressed print 'x' else print 'another'
```
Wait for 10 seconds and then press a button. The program should not stop and there should be no errors reports in the network tab of the browser.

### DIFF
--- a/static/js/app.ts
+++ b/static/js/app.ts
@@ -1073,8 +1073,8 @@ export function runPythonProgram(this: any, code: string, sourceMap: any, hasTur
       // So: a very large limit in these levels, keep the limit on other ones.
       execLimit: (function () {
         const level = theLevel;
-        if (hasTurtle || hasMusic) {
-          // We don't want a timeout when using the turtle or music -> just set one for 10 minutes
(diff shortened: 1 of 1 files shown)


jpelay (approved):
Works!

PR #5683, opened 2024-08-01
🧪 Cypress part5

Annelein:
Fixes the following tests for https://github.com/hedyorg/hedy/issues/4027 :


Annelein:
@jpelay its ready, I only need help with the <custom_element> which doesn't render a data-cy which is needed for the general_settings Cypress test

jpelay (approved):

PR #5688, opened 2024-08-06
🪲 Improve unquoted error text

### LINKED ISSUE #3444: [BUG] Incorrect error message when an unquoted backslash is found in a print statement
**Describe the bug**
Trying to use a backslash as an operator is, of course, a syntax error, but Hedy seems to get confused about where exactly the error is. This situation can arise if a student, having recently been given access to division, forgets which way the slash should be slanted.

In the example code below, Hedy gives an error starting like this:
```
Be careful. If you ask or print something, the text should start and finish with a quotation mark.
```

That's fine so far, but the error becomes confusing in the next sentence:
```
You forgot that for the text divided.
```

The problem is that `divided` _was_ part of the quoted text.

I tried this on level 6.
```
print '5 divided by 5 is ' 5 \ 5
```

image

**Expected behavior**
I expected the error to say that the backslash needed to be quoted, or perhaps all of `5 \ 5`.

Ideally, though, Hedy would suggest using a forward slash instead, since that's probably what the student meant. I've seen some references to the Confusables package in other issues, and that might help in this case. Not sure, as I haven't looked into what that package does specifically.

  - Windows 10
  - Brave

boryanagoncharenko:
Fixes #3444 
This PR changes the text of the unquoted error to display only the segments which are not quoted correctly. Also, it removes the incorrectly doubled backslashes.

**How to test**
In level 6 run the following command:
```
print 'This is ' unquoted ' and '
```
The error message should only mention the word `unquoted` and not the parts which are correctly quoted.

Again, in level 6 execute the following command:
```
print '5 divided by 5 is ' 5 \ 5
```
Ensure that the error message points only to the `5 \ 5` part and that the message contains a single backslash (exactly as it appears in the code editor).


hedy.py
```diff
@@ -1413,6 +1413,42 @@ def process_characters_needing_escape(value):
...
+    "„": "“",  # inward double curved quotation marks or German quotes
+    "«": "»",  # guillemets or double angular marks or French quotes
+}
+
+
+def is_quoted(s):
+    return len(s) > 1 and s[0] in supported_quotes and s[-1] in supported_quotes[s[0]]
+
+
+def find_unquoted_segments(s):
```
jpelay, line 1430: Great! I love this.

jpelay (approved):
It works!

PR #5693, opened 2024-08-09
🪲 Stop highlighting unmatched quotes

### LINKED ISSUE #4874: [BUG] Unmatched quotes are highlighted
Got this email from a teacher pointing out it is confusing that this string is highlighted in purple but then does not parse (which is a great point)

image

jpelay:
The old way to handle strings, was to use a grammar rule that allowed everything between quotation marks. However, this had the problem that it highlighted even strings that were not complete, and even would mess the highlighting of the next lines. That's why I changed it here to a token rule, and now it properly handles strings that are not well formed.

Another change that I found out while working on this, was that list assignments from level12 onwards were not working properly because they still were expecting text without quotes.

This is the behaviour now:

image

This is how it looked before:

image

As you can see now, it's a lot clearer to know the exact line that is causing issues.

Fixes #4874

**How to test**

Write an unfinished string in any of the levels from 4 onwards.


boryanagoncharenko (approved):
Works great!

PR #5697, opened 2024-08-13
🖊️ Convert input from ask command to booleans

### LINKED ISSUE #5691: 🖊️ Add support for boolean values in the ask command
**Language idea incl level**
Boolean values are supported from level 15. However, if a boolean value is returned as input of the `ask` command, they appear as plain text. In level 12 input from ask is converted to an int or float but not a boolean.

```
a is true
b is ask 'True or False?'
if a == b
    print 'match ' a ' ' b
else
    print 'mismatch ' a ' ' b
```

boryanagoncharenko:
Fixes #5691 
Starting with level 12, we try to convert the input from the ask command to a number (first an int, then a float). If that conversion fails, we keep it as a string. Now that we support boolean values starting with level 15, it makes sense to try and convert the input to a boolean too.

**How to test**
Note that the added automated tests do not cover the full execution of the programs because we cannot fully test the interaction with the 'ask' command. To test this locally, run the following snippet in level 15:
```
a is true
b is ask 'True or False?'
if a == b
    print 'match ' a ' ' b
else
    print 'mismatch ' a ' ' b
```
When the input box appears put in the following: `true`, `false`, `True`, `False`, `15`, `not_a_boolean`. Note that the print statements should output exactly what you typed in, i.e. if you typed `true` you should not get `True`.
Change the language to Bulgarian and paste the following input `вярно` (which is True), `невярно` (which is False).


jpelay (commented):
Hi @boryanagoncharenko! This works great and I love it. I won't approve right away because you say:
> we cannot fully test the interaction with the 'ask' command. 

But we can! Using Cypress we can add a test to check if the behavior is the expected in the website. I can add the test if you want! We already do it with a couple of commands.

boryanagoncharenko:
> Hi @boryanagoncharenko! This works great and I love it. I won't approve right away because you say:
> 
> > we cannot fully test the interaction with the 'ask' command.
> 
> But we can! Using Cypress we can add a test to check if the behavior is the expected in the website. I can add the test if you want! We already do it with a couple of commands.

Excellent point! Cypress tests are slow so I added only a couple of tests to check the conversion to numbers in level 12 and booleans in level 15.
I saw that there are already tests that use the ask command but they seem to be testing debugging too. So I added a new file and I did some refactoring, but I am not confident of the naming or the tools section. Could you please have a look at these changes? Thanks in advance!

jpelay (approved):

PR #5737, opened 2024-09-02
🚚 Add sleeps to daily Weblate Update script

rix0rrr:
It seems that `wlc commit` followed by `wlc push` doesn't always create a PR (WeblateOrg/weblate#12400). This is the cause of the Weblate repository sometimes being locked without ever getting unlocked: unlocking happens on the PR-close workflow, but the PR never gets created.

In fact, since we also run a `wlc reset` to avoid future merge conflicts, the `wlc reset` destroys all pending translator changes, which has happened to some contributors and now their work is gone.

Add some sleeps to get a (presumably) async commit process to finish, and do an additional check for the `push` flag to reduce chances of data loss.

**How to test**

This is hard to test in isolation. We'll just have to monitor the GitHub Workflow runs over the next couple of days.

### DIFF
--- a/.github/workflows/update-weblate.yml
+++ b/.github/workflows/update-weblate.yml
@@ -64,11 +64,25 @@ jobs:
         done
 
         wlc pull
+
+        # 'commit' seems to be an async process, as sometimes nothing is pushed if we run this
+        # too quickly in succession. Add a short sleep to reduce chances of this happening.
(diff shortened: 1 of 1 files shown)


jpelay (approved):
I will try to translate something in Spanish and see if things work out!

jpelay:
I tested a bit and my comments are in the PR, so I think this worked?

PR #5738, opened 2024-09-02
🧹 Change deprecrated Mergify rules

jpelay:
@rix0rrr pointed out to my attention that Mergify was complaining about some rules we were using incorrectly. The documentation says that they're now part of `queue_rules`.

**How to test**

We'll have to check this run in the following days.


rix0rrr (approved):

.mergify.yml
```diff
@@ -16,10 +20,4 @@ pull_request_rules:
     actions:
       comment:
         message: Thank you for contributing! Your pull request is now going on the merge train (choo choo! Do not click update from main anymore, and be sure to [allow changes to be pushed to your fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).
-      queue:
-        name: default_queue
```
rix0rrr, line 20: Oops! In hindsight, I think we shouldn't have removed this part:

```
queue:
  name: default_queue
```

That's probably the command to enqueue the PR for actual merging.

Noticing this after noticing that the Weblate PR didn't merge itself.

jpelay: Sorry! Made a PR: #5741 !

PR #5740, opened 2024-09-03
🪲 Remove valid assignment uses from ask command errors

### LINKED ISSUE #3274: [BUG] Error when using `vraag` as a variable in level 3 (language Dutch, keywords English)
When I try to run this code in level 3 I get an error 
```
vraaga is vind je honden aardig, doe jij teveel gamen, je ging op pad
print vraaga at random
```

the error is:

`Dat was goede code hoor, maar niet op het goede level. Je schreef een ask voor level 1. Tip: Vanaf level 2 hoort ask met een variabele ervoor. Bijv: naam is ask Hoe heet jij?`

I think there is some translation problem because the variable name `vraaga` is part of a Dutch command `vraag` (ask). I use Dutch as the main language and English as the keyword language. 

image

boryanagoncharenko:
Fixes #3274
In levels 1-6 the `ask` and `assign` commands sometimes overlap and sometimes return an error incorrectly. With this PR, some correct usages are removed from the common error scenarios. Specifically:
- in levels 2-3 assignments can have variable names starting with `ask`, e.g. `asking is 5`, `ask_output is test`
- in levels 4-6 assignments can start with the word `ask`, e.g. `var is asking you a question`

However, the incorrect uses of `ask` should still return an error:
- in levels 2-3 `ask what is it?` should return a wrong level exception
- in levels 4-6 `var is ask what is it?` should return a missing quotes exception

**How to test**
```
# in level 2
ask missing var
# this should return an error

asks is Felienne
# this should defined a variable ‘asks’
```
```
# in level 3, language NL
vraaga is vind je honden aardig, doe jij teveel gamen, je ging op pad
print vraaga at random
# this should not yield an error
```
```
# in level 4 and 6
var is asking me a question?
print var
# this should not yield an error
```


Felienne (approved):
Awesome, this was such a longstanding issue!

PR #5741, opened 2024-09-03
🚚 Add queue name to `pull_request_rules`

jpelay:
The mergify script is not working properly and we might think it can be related to this:

> Oops! In hindsight, I think we shouldn't have removed this part:
> ```
> queue:
>   name: default_queue
> ```
> That's probably the command to enqueue the PR for actual merging.
> 
> Noticing this after noticing that the Weblate PR didn't merge itself.

Original comment made by Rico in: https://github.com/hedyorg/hedy/pull/5738/files#r1742077765


jpelay:
Before approving this we might want to wait for #5725 to finish running the tests to see if that one gets merged, if it doesn't then I think we did messed up with the configuration

rix0rrr (approved):

PR #5746, opened 2024-09-04
Fix keypress modal not showing 

### LINKED ISSUE #5745: 🪲 Keypressed modal is not displayed
When the if-pressed command is used there should be a modal in the output that indicates that the program is waiting for input. Now it is not being displayed at all.

boryanagoncharenko:
Fixes #5745

**How to test**
Open level 5 and execute the following code. Check that in the output window there is a modal that indicates when the program is waiting for input. Once you press a button, the modal should disappear.
```
if y is pressed forward 15
else turn 90
```


jpelay (approved):
Works! The problem was introduced when we replaced the `-` with `_` and forgot to replace the references to some if the ids.

image

PR #5747, opened 2024-09-04
💻 Stop aproving teacher accounts

### LINKED ISSUE #5724: Stop approving teachers accounts
As discussed in the meeting yesterday, we will simply auto-approve all teachers' accounts. 

We can, for now, leave the interface as we have it now, so that we can *revoke* access though the admin panel if needed.  In the future we might want to have a better search option, because now it is paginated and annoying!

jpelay:
Changes the process of becoming a teacher by skipping the approving step. Now users can simply make a teacher account by choosing the option when creating an account or simply choosing to do so from the profile menu.

Fixes #5724

**How to test**

* Create a new account choosing the teacher option. You can use this function to make your life easier:
```
function fill_fields() {
    let username = `teacher_${Math.random()}`;
    $('#username').val(username)
    $('#email').val(username + '@p.ai');
    $('#password').val('123456')
    $('#password_repeat').val('123456')
    $('#language').val('en')
    $('#birth_year').val('1999')
    $('#gender').val('f')
    $('#country').val('VE')    
}
```
* After creating the account you should be able to create classes and such.

Then, you can log into a normal account, go to the profile page and click this button:
image

Now your account should be a teacher account.


jpelay:
@Felienne Please do let me know once we ship this, when is a good moment to turn all pending teacher request into teacher accounts!

Felienne:
> @Felienne Please do let me know once we ship this, when is a good moment to turn all pending teacher request into teacher accounts!

I think we could already do that now? I will have a look at a query today!

website/auth_pages.py
```diff
@@ -209,6 +211,65 @@ def signup(self):
...
+                    locale=user["language"], template="welcome_teacher", email=user["email"], username=user["username"]
+                )
+            except Exception:
+                print(f"An error occurred when sending a welcome teacher mail to {user['email']}, "
+                      "changes still processed")
+
+        session.get('user')['is_teacher'] = True
+        session['welcome-teacher'] = True
+
+        # I want to redirect the user to a tutorial!
```
boryanagoncharenko, line 270: This comment is slightly confusing. Is the teacher is sent to the For-Teachers section. Is this the goal?

boryanagoncharenko (commented):
Looks great! I tested all scenarios I could think of and they worked.

I noticed that the user.teacher_request is still used in multiple places (profile.html, support.html, etc.). Do we want to get rid of these usages now or is that a clean up for later? Or is it that we want to keep these parts?

website/auth_pages.py
jpelay: Yes you're right, it was a goal. I want to update the tutorial and send them there, I think things are stable enough now to update the teacher tutorial.

boryanagoncharenko (approved):

PR #5751, opened 2024-09-06
🪲 Attempt to fix weblate error when adding a new language

boryanagoncharenko:
Fixes #5750
For some reason the title element is missing from the content/page/zun.yaml file. This should fix the error but I have no clue why this happens.


Felienne (approved):

Felienne:
> Fixes #5750 For some reason the title element is missing from the content/page/zun.yaml file. This should fix the error but I have no clue why this happens.

Yeah, this is a known issue. I also don't know why it happens, when it does, I always add the file by hand. Would be nice to fix and not hard, you'd think (Weblate does not make this mistake for all the other folders, after all) but on the other hand, Weblate is a mystery, it might take you hours and new langs don't happen often.

Felienne:
Ok, sadly this attempt did not work, so Weblate is now blocked:

image

To "fix" this, you can run the fix weblate Action, that undoes all uncommitted changes:https://github.com/hedyorg/hedy/actions/workflows/resolve-weblate-conflict.yml

I just ran it, but there might be some things lost. 

That is not normal with a new language (contrary to the pages issue) so I guess other things were changed in the meantime leading to issues. There is brittleness all over the place with Weblate.

PR #5754, opened 2024-09-06
🚚 Include files in template directory in tailwind task

jpelay:
I noticed that when I added new tailwind classes in HTML files in the `templates` directory, the dodo task didn't compile them, and therefore didn't work. When I added the classes in files inside subfolders, the task was completed succesfully. Therefore in this PR I'm including a new glob pattern for files directly inside the  `templates` directory.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -103,6 +103,7 @@ def task_tailwind():
     return dict(
         file_dep=[
             *glob('templates/**/*.html'),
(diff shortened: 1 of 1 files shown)


rix0rrr:
I for sure thought that `**/*.html` should include `*.html`.

But I guess not…?

rix0rrr (approved):

Issue #5681, opened 2024-07-30
🪲 For-list command stops when placed in an if_pressed statement


boryanagoncharenko:
**Describe the bug**
In level 10, the following code should output all elements of the list if the key x is pressed. However, the for loop outputs the first element of the array and then the program stops:
```
lijstje is kip, haan, kuiken
if x is pressed
    for dier in lijstje
        print dier
else
    print 'onbekend dier'
```

**Add a screenshot (optional)**
<img width="1021" alt="Screenshot 2024-07-30 at 13 13 35" src="https://github.com/user-attachments/assets/4ac6eba1-b422-4be8-a6c9-45a315761305">

boryanagoncharenko:
The issue seems to be that Skulpt stops the program when a time.sleep() is encountered (and there is one added at the end of the for-loop body). This seems to have been an issue before https://github.com/skulpt/skulpt/issues/764 but it should have been already addressed. Not sure if the issue persists in our fork.

We could:
- Try syncing with the latest version of Skulpt and/or attempting to fix the matter there.
- Just remove all time.sleep statements from the body, even though that might lead to some confusion in the output.

Felienne:
> The issue seems to be that Skulpt stops the program when a time.sleep() is encountered (and there is one added at the end of the for-loop body). This seems to have been an issue before [skulpt/skulpt#764](https://github.com/skulpt/skulpt/issues/764) but it should have been already addressed. Not sure if the issue persists in our fork.
> 
> We could:
> 
> * Try syncing with the latest version of Skulpt and/or attempting to fix the matter there.

I am ok with trying that but we should totally check with @jpelay whether/how that would impact us since we did a lot of work on top of Skulpt with the debugger.

> * Just remove all time.sleep statements from the body, even though that might lead to some confusion in the output.

Do you mean only in the `if_pressed` or in general? It is potentially a bit confusing but if it solves this problem, I am ok with it (how often, in practice are kids going to put a loop within an if statement) But if you mean in general, that would be a no for me because the small wait is very helpful!

boryanagoncharenko:
> I am ok with trying that but we should totally check with @jpelay whether/how that would impact us since we did a lot of work on top of Skulpt with the debugger.

Actually, I just checked that syncing the fork would not resolve the issue. The problem still exists in Skulpt and can be reproduced with the following programs:

```
import turtle

t = turtle.Turtle()

for c in ['red', 'green', 'yellow', 'blue']:
    t.color(c)
[...]

PR #5763, opened 2024-09-10
🪲 Fix bug in custom Skulpt module extensions

### LINKED ISSUE #5729: 🪲Turtle commands in if-pressed statements terminate unexpectedly
**Describe the bug**
If the body of an if-pressed statement contains multiple turtle commands, only the first gets executed

Level 12:
```
if x is pressed
    forward 90
    turn 90
    forward 90
else
    forward 90
    turn -90
    forward 90
```

Screenshot 2024-08-29 at 17 06 38

The issue seems to be related to #5681. We decided to go with the not-so-great solution of removing the time.sleep statements. Obviously, we cannot drop the forward or turn command here.

boryanagoncharenko:
This PR addresses the following issues:
- When the body of an if-pressed command contains more than one turtle functions, only the first one will be executed.
- When the body of an if-pressed command contains time.sleep(), the execution would terminate when reaching this line.

Fixes  #5729 #5681

**How to test**
Run locally, go to level 15 and run the following scenarios:
1. Check that the code below outputs all items in the list. Note that they should not appear at once, but with a tiny delay in between:
```
lijstje is "1", "2", "3", "4", "5"
if x is pressed
    for dier in lijstje
        print dier
else
    print 'onbekend dier'
```
2. Check that when pressing x, all statements are executed. Note that the waiting-for-key-press modal should appear and then disappear while the action is being executed. Note that pressing the x button while the turtle is moving should not trigger a new run. If an error occurs, the keys should not be animated anymore.
```
i = 0
while i < 20
    if x is pressed
        turn 10
        color 'blue'
        turn 90
        forward 50
        color 'red'
        turn 90
        forward 50
        color 'orange'
        turn 90
        forward 50
        color 'green'
        turn 90
        forward 50
    else
        turn -15
        color 'blue'
        turn -90
        forward 50
        color 'red'
        turn -90
        forward 50
        color 'orange'
        turn -90
        forward 50
        color 'green'
        turn -90
        forward 50
    i = i + 1
```

### DIFF
--- a/static/vendor/skulpt-stdlib-extensions.js
+++ b/static/vendor/skulpt-stdlib-extensions.js
@@ -21,49 +21,71 @@ var $builtinmodule = function (name) {
           keyElement.remove()
         }, 1500);
       }
-    } 
+    }
+
+    var ongoingIfPressedCall = false;
+
+    function callIfPressedFunc(name, resolve, reject) {
+      var f = Sk.misceval.loadname(name, Sk.globals);
+      var currentProgram = window.sessionStorage.getItem("currentProgram");
+
+      Sk.misceval.asyncToPromise(() =>
+        Sk.misceval.callOrSuspend(f), {}, currentProgram).then(() => {
+          resolve();
+        }).catch((e) => {
+          reject(e);
+        }).finally(() => {
+          ongoingIfPressedCall = false;
+        });
+    }
 
     function keyBoardInputPromise(if_pressed_mapping) {
+      ongoingIfPressedCall = true;
       $('#keybinding_modal').show();
       return new Promise((resolve, reject) => {
         window.addEventListener("keydown", (event) => {
-          let pressed_mapped_key = false;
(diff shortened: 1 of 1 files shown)


static/vendor/skulpt-stdlib-extensions.js
```diff
@@ -21,49 +21,71 @@ var $builtinmodule = function (name) {
...
-    } 
+    }
+
+    var ongoingIfPressedCall = false;
+
+    function callIfPressedFunc(name, resolve, reject) {
+      var f = Sk.misceval.loadname(name, Sk.globals);
+      var currentProgram = window.sessionStorage.getItem("currentProgram");
+
+      Sk.misceval.asyncToPromise(() =>
```
jpelay, line 32: Ahh it makes sense! AsyncToPromise is the one to use

jpelay (approved):
It works!

PR #5764, opened 2024-09-10
💻 Add checkbox in view programs page

### LINKED ISSUE #5752: 💻 Add tickmark in student exercise
I can already add tickmarks fro the overview page, but could we add a tickmark box inside the assignment of the student, then I don't have to go back, that saves me a lot of clicks, so here:

image

And, can we show that to the students too, if I have ticked it off, they get a tickmark icon next to the tab?

That would be amazing!!

jpelay:
Adds a checkbox in the view programs page, only visible if the user is the teacher of the creator of the program. By ticking the checkbox, the program is stored in the database as reviewed. By going backwards to the table, it'll trigger a reload and the page will be updated. At last, I also added a tick mark in the adventure page, that lets the student know if the teacher reviewed this program.

Fixes #5752 

**How to test**

1. Log in as a student and run a program in several tabs (can't be the examples, because otherwise it won't be shown in the table)
2. Now log in as the teacher and go to the class page where the student belongs. Click the eye icon and then tick the checkbox.
3. The checkbox for this program should be ticked.

### DIFF
--- a/static/js/teachers.ts
+++ b/static/js/teachers.ts
@@ -767,7 +767,6 @@ export function initializeClassOverviewPage(_options: InitializeClassOverviewPag
                            ( typeof window.performance != "undefined" && 
                                 window.performance.navigation.type === 2 );
     if ( historyTraversal ) {
-      console.log('There was reload')
       window.location.href = window.location.href
     }
   });
--- a/static/js/teachers.ts
+++ b/static/js/teachers.ts
@@ -759,6 +759,18 @@ export function initializeClassOverviewPage(_options: InitializeClassOverviewPag
   });
 
   initializeGraph()
+  // An ugly hack, but if someone goes back trhough the page, the cache
+  // causes the old version of the page to be shown
+  // So we hard reload it
+  window.addEventListener( "pageshow", function ( event ) {
+    var historyTraversal = event.persisted || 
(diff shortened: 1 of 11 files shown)


static/js/teachers.ts
```diff
@@ -759,6 +759,18 @@ export function initializeClassOverviewPage(_options: InitializeClassOverviewPag
...
   initializeGraph()
+  // An ugly hack, but if someone goes back trhough the page, the cache
+  // causes the old version of the page to be shown
+  // So we hard reload it
+  window.addEventListener( "pageshow", function ( event ) {
+    var historyTraversal = event.persisted || 
+                           ( typeof window.performance != "undefined" && 
+                                window.performance.navigation.type === 2 );
+    if ( historyTraversal ) {
+      console.log('There was reload')
```
boryanagoncharenko, line 770: I am not sure whether this line is left intentionally

boryanagoncharenko (commented):
Everything works as expected! Not approving yet, because I think a console.log statement might have been left in unintentionally

static/js/teachers.ts
jpelay: You're right!

boryanagoncharenko (approved):

PR #5791, opened 2024-09-18
🪲 Add support for more quotes 

### LINKED ISSUE #5785: 🪲Add support for Polish quotation marks
**Describe the bug**
Currently, we do not have support for some combinations of quotation marks:
- Poland uses „…”
- Sweden and Finland use  ”…”
- Korea uses 《…》
- Japan uses 「...」

boryanagoncharenko:
Fixes #5785
We already support multiple types of quotes for strings, but we were missing a few. A Polish translator got the build failing because they used a variation of double quotes that we do not support.
This PR adds support for:
- Polish variation of double quotes: „”
- Swedish/Finish variation of double quotes:  ””
- Korean quotes:  《》
- Japanese quotes: 「」

**How to test**
In level 4, the following print statements should work without an error:
```
print „pol”
print ”swe”
print 《kor》
print 「jap」
```


grammars/keywords-zun.lark
Felienne, line 1: Is this change related to the quotes?

boryanagoncharenko: this file caused translation tests to fail (when the cache is switched off locally). I will investigate why there is no failure on GitHub, but I suspect it is caching.

Felienne (approved):

PR #5812, opened 2024-09-25
🪲 Remove a source of client exceptions

rix0rrr:
A lot of client exceptions were originating from the embedded editor, trying to call `removeAttribute()` on an object that happened to be `null`.

The reason was an assertion that the result of `getElementById()` would never be `null`, where in fact, it was.

Replace with a null coalescing function call `?.removeAttribute()` so that the function call falls to the floor if the element doesn't exist.

**How to test**

Visit http://localhost:8080/embedded/1/?lang=nl&run=true&readOnly=false&fullWidth=false&program=IyBTY2hyaWpmIGpvdXcgY29kZSBoaWVy%22 and run any program. Observe that no exceptions get logged to the JavaScript console.

(cc @jpelay, I spied you [here](https://github.com/hedyorg/hedy/pull/5673) 😛 this is why `!` is dangerous, even you *think* this can never happen there's always a case you didn't think of -- better to program defensively or fail real hard with an explicit error message)

### DIFF
--- a/static/js/debugging.ts
+++ b/static/js/debugging.ts
@@ -60,7 +60,7 @@ export function toggleVariableView(){
   if($('#variables #variable_list li').length == 0){
     $('#variable_button').hide();
   }
-  else{    
+  else{
     $('#variable_button').show();
     $('#variable_list').show();
     document.getElementById('variables_arrow')!.classList.remove('fa-angle-up');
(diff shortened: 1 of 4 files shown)


jpelay (approved):
> (cc @jpelay, I spied you https://github.com/hedyorg/hedy/pull/5673 😛 this is why ! is dangerous, even you think this can never happen there's always a case you didn't think of -- better to program defensively or fail real hard with an explicit error message)

Dang it! Sorry, you're right, should've the ? instead.

Issue #5187, opened 2024-02-28
💻 stop copying over English content in Weblate


Felienne:
It is not always clear to teachers untranslated content is shown when they see English content, it would be nice if we can show an icon or something (or a different color) if a content block is equal to English (tested with a text equivalence)

Felienne:
Hi all (esp @boryanagoncharenko and @jpelay are relevant for this)!

My proposal is as follows, stepwise

* We change the default setting of Weblate to copy over content, and figure out what happens then (empty files? no files? placeholders?)
* In our code, detect if there is no file or whatever weblate does and catch it. For now, we can just show our English version. This does not change a lot right away, but is immediately better than what we have now, because in the current situation, if we update English, the copied files do not get updates, so a lot of languages have some outdated, previously copied over content in them.

When that is done, we can decide if we, as originally suggested in this issue, want to highlight the untranslated content, only for teachers or for everyone.

But let's get this piggy washed first! (this is a Dutch saying)

boryanagoncharenko:
The way to copy content from one language to another is done through using an add-on called [Prefill translation with source](https://docs.weblate.org/en/latest/admin/addons.html#addon-weblate-generate-prefill). When it is disabled and a new language is added, Weblate creates empty yaml files and a complete and correct .po file without any translations (see image below). As a side note, now we know why in the past we got generated PRs with empty yaml files - it happened because the prefill add-on did not do its job on time.

<img width="1149" alt="Screenshot 2024-09-18 at 14 07 32" src="https://github.com/user-attachments/assets/d9ce3c09-f56a-4885-bf54-2e51a07d49a7">

Another aspect of the missing prefill add-on is that the components cannot be translated in the Weblate interface. Since they are empty yaml files, no content is offered to translators. See the image below:

<img width="1496" alt="Screenshot 2024-09-18 at 14 00 10" src="https://github.com/user-attachments/assets/8089319a-8a86-46bf-ab63-d52651df6a7e">

Based on the current behavior, I guess we need to have a template yaml file that contains keys, but not values. If this file is used in the prefill, translators would see what they need to translate but untranslated the values would not be filled.

boryanagoncharenko:
Regarding the first issue, it seems that Weblate has a setting "Template for new translations" which can be set to the Monolingual base language file https://docs.weblate.org/en/latest/admin/projects.html#component-new-base. When a new language is added, the keys will be extracted from this file but only the present translations will be added as values.

[...]

PR #5823, opened 2024-10-01
🚚 Stop copying yaml content when a new language is added

boryanagoncharenko:
This is a draft PR meant to collect feedback for a solution of issue #5187.

**Description of the problem**
When a new language is added in Weblate, the prefill add-on copies the English content into the new language yaml files. This means that when the English content changes, we manually need make the change in all language that do not have the content translated. Instead of this, we would like to keep in the yaml files only the content that has actually been translated.

**Yaml files**
All newly added content/yaml files are not prefilled with the current English translation and only contain the translated value for the corresponding language. Currently, only the om language has partial content to illustrate the solution. The content of each yaml file is merged with the en.yaml counterpart when the yaml is loaded.
- Schema validation. The schema definitions have explicit required fields and only the en.yaml files are checked against them. A python script generates new schema definitions (*.generated.schema.json) that lift the required fields constraints. The generation is part of the doit backend task and the schemas are gitignored. The generated schemas are used for all non-en.yaml files.
- Correctness check added in #4020. We have a script that checks if there are different array lengths between English and other languages. This check is amended to ensure the structure of the language yaml file is a subset of the English file. Now it check that:
  - The nodes in the 2 yaml files are of the same type, specifically arrays and dictionaries.
  - An array in a language file should not have more elements than the English file.
  - A dictionary in the language file should not have keys that are not present in the English file.
- Note that the merging mechanism silently discards the content of the language file if any of the correctness rules are not met. The assumption is that the English file will have the correct structure. This meant to be inline with the silent empty dictionary if an IOError occurs while loading the yaml file. However, we could have a different approach here: fail and just load the fallback? Or at least log that something went wrong?

**Po files**
Just like the yaml files, all newly added .po files will not be prefilled with English and, again, the om language is used as an example. The merging of every language with English happens through a custom function that defines a fallback. The regular gettext function is substituted with the custom one.

**Deduplication of existing languages**
The om language is added to this PR for exemplary purposes and is not an officially supported language. Since all current languages have a version of en.yaml files, the merging will not yield immediate results. It will take effect only when a new language is added. Perhaps it is wise to run a basic form of deduplication of the current files and test immediately. Note that complete deduplication of the yaml files is currently very hard to achieve because every yaml file copies a different version of the Enligsh content. However, we could simplify the problem by only removing the duplication that matches the current English version.


content/slides/id.yaml
```diff
@@ -130,7 +130,7 @@ levels:
                 {print} What song would you like to hear?
                 {ask} I like that song too!
                 {print} Next up... {echo}
-            debug: true
+            debug: 'True'
```
rix0rrr, line 133: Is this intentional?

gettext_with_fallback.py
```diff
@@ -0,0 +1,8 @@
+from flask_babel import gettext
+from website.flask_helpers import gettext_with_fallback
+
+
+# Explicitly substitute the flask_babel function gettext with our own implementation that
+# adds a fallback language. Note that we need to monkey-patch instead of use the gettext_with_fallback
+# directly because the Babel extract function works if we use gettext with literal strings.
```
rix0rrr, line 7: I'm not entirely following what this is trying to say.

This file seems to be a re-export of a symbol elsewhere, so that we can do:

```py
from gettext_with_fallback import gettext
```

Instead of doing:

```py
from website.flask_helpers import gettext_with_fallback as gettext
```

Is that right?

translations/en/LC_MESSAGES/messages.po
```diff
@@ -1577,7 +1577,7 @@ msgid "star"
 msgstr "a star"
 
 msgid "start_learning"
-msgstr "Start learning"
+msgstr "Start learning EN"
```
rix0rrr, line 1580: Is this to see that you can see the change? 😄

rix0rrr (approved):
Love it! ❤️

gettext_with_fallback.py
boryanagoncharenko: Thanks for the feedback, I should update the comment then! I wrote it down as a reminder that I cannot use directly in the python `gettext_with_fallback('msg')` instead of `gettext('msg')` because the extract will not work (although this would be ideal because we decrease the chance of someone using the gettext without a fallback in the future). We have a hack with the client messages to workaround that. I will think of a better way to phrase it.

About the import, you are totally right! The idea was to have a single line of code and I did not realize I can to the `import` and `as` on the same line. So, yes, this is exactly my intention.

app.py
```diff
@@ -443,6 +444,10 @@ def before_request_https():
 Compress(app)
 Commonmark(app)
 
+# Explicitly substitute the flask gettext function with our custom definition which uses fallback languages
+app.jinja_env.globals.update(_=gettext)
```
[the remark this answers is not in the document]
boryanagoncharenko: Agree. I am currently adding tests to ensure that the merging of yaml will not fail at runtime if the correctness check is skipped. I will add this to the list of things to test.

PR #5835, opened 2024-10-04
Update nl.yaml

Felienne:
Note to self: dit moet ook nog voor andere talen!


Felienne:
Ah the pink tabs never have assignments in them!  In that case we need to move this concet a bit a later in the level, this is too complex

PR #5838, opened 2024-10-04
🖊️ Speed up turtle

### LINKED ISSUE #5837: 🖊️ Speed up turtle?
Turtle is quite slow, if we would implement our own (probably a lot of work, haha) we could fix that?

jpelay:
Speeds up the turtle a bit, from the old value of 3 (mapped in the turtle as slow), to 10 (fast). It's faster but not by much I think.

New:
https://github.com/user-attachments/assets/232fbe14-4c12-4e99-a3d2-56ceb453b8ef

Old:
https://github.com/user-attachments/assets/0423c906-6a52-41de-a26e-c774fa6fe6b8

Fixes #5837 
**How to test**

* Go to level 12 and paste this program 

```
define square
    repeat 4 times
        turn 90
        forward 40
call square
forward 50
call square
forward 50
call square

```


Felienne:
ow this is so awesome!! I had alway assumed that this was not possible, somehow.

Felienne:
Guess this opens the door to exposing the speed command in Hedy too (I can to that myself!)

boryanagoncharenko (approved):

PR #5840, opened 2024-10-06
📜 add foundation info

Felienne:
Adds info on the NGO to the site (mandatory if we want to apply for ANBI status)


Felienne:
Ola @boryanagoncharenko!

Can you approve this (so we know it does not mess up Weblate)

Also, if I never have to do this annoying copy-paste of 5 different versions again, my life would be som much better!

Felienne:
Soon, I hope!

boryanagoncharenko (approved):

PR #5854, opened 2024-10-15
🪲 Dont try translating program with blanks

### LINKED ISSUE #5833: 🪲 Program page gives an error
Posted from school, will update later

This link gives an error: https://hedy.org/hedy/1cefaa3bcb3742ea814fee01ffae35ca/view

jpelay:
Fixes the problem by checking if the program has blanks before trying to translate it. Also, I changed the in-site calculation if a program was modified, for the flag that comes from the DB

Fixes #5833 

**How to test**

* Log in as student3 and then go to level 1 and execute this program
```
print Welcome to Hedy's restaurant 🍟
_ What would you like to order?
echo So you would like to order
print Thank you for your order!
print It's on its way!
echo So you would like to order
print Thank you for your order!
print It's on its way!
echo So you would like to order
print Thank you for your order!
print It's on its way!
```
* Now log in as teacher1, change your keyword language to anything other than English and go check the program uploaded by student1, it should work normally.


Felienne (approved):
Yes, does the job indeed!

image

Thanks for the detective work!!!

PR #5864, opened 2024-10-21
💻 Create student accounts in bulk 

### LINKED ISSUE #3936: 💻 Upload of a CSV file to create student accounts for a class
### Discussed in https://github.com/hedyorg/hedy/discussions/3933

<sup>Originally posted by **stevecopley** January 13, 2023</sup>
When I'm using Scratch with a class, to avoid too much faffing about, I usually create the student account ahead of the first lesson. Scratch has the ability to upload a **CSV file** with a list of usernames/passwords to mass-create the accounts and link them to the current class. Something similar for Hedy would be great.

(Would be good to have a downloadable template CSV file with a couple of examples in to give teachers clear guidance as to the fields/formats you required.)

</div>

boryanagoncharenko:
This PR redesigns the screen that allows teachers to create multiple student accounts for a class. As discussed during the contributors' meetings the page should allow teachers to:
- create students and let Hedy auto generate passwords
- create students and provide their own passwords using a semicolon as a separator
- paste a two-column spreadsheet and expect Hedy to transform the input accordingly
- copy to clipboard and print the successfully created accounts

Fixes #3936 

Things to note
- When an error occurs, the error message does not go away until you remove it through its close button. This is done intentionally because the error message could point out which rows of your input are incorrect. So, if it just goes away after x seconds, I think this could be very inconvenient.
- The loading indicator which is displayed when the accounts are being created should take the whole page, because we do not want people to go away from the page while accounts are being created. The loading indicator is currently in the layout.html as an attempt to make it reusable but maybe when it takes the whole page, it will not be reusable.

**How to test**
Features to test:
- You should be able to create one or multiple student accounts with and without passwords.
- When you print from Excel, the separators should be automatically changed to semicolons. Note that this should work if you use Ctrl/Cmd + V or right-click and chose Paste from the context menu.
- You should also be able to successfully create student accounts when you are a second teacher. In the devdb teacher4 is a second teacher of CLASS1.
- You should be able to print and copy to clipboard the successfully created accounts. Note that sometimes when you try to print the accounts information, the borders of the table will not be shown in the print-preview screen. I do not know why this happens, but the they are correctly printed. I use 'Save to PDF' functionality and the pdf files look great.

Errors to test:
- You forgot to put in any usernames.
- You tried to create usernames with invalid characters. Currently these are `@` and `:`.
- You got the auto generate passwords toggle checked, but you provided usernames and passwords correctly separated with semicolon. In this case you should get a specific error that directs you to the toggle. 
- You got the auto generate passwords toggle unchecked, but you provided only usernames and no passwords. In this case you should get a nice error directing you to the toggle.
- You got the auto generate passwords toggle unchecked and you provide multiple usernames and passwords. However, only 1 row is missing a semicolon. You should get an error that directs you to this row.
- You got the auto generate passwords toggle unchecked but some of your passwords are less than 6 characters.
- You supplied the same username twice, with auto generated passwords on and off.
- You supplied a username that is already stored in the database. You should get an error and a way to resolve it.
- You tried to create more than 100 usernames.


MarleenGilsing:
Works perfectly! 
While testing it om my local host I couldn't read the error messages, but it did seem like the correct error messages happend in the correct places. 

Only thing I did find: I tried to make accounts using a space instead on an enter and I was allowed to do this: 
image

It's not that much of a problem, because if the teacher does this they can just go back and do it properly. Would it be easy to detect if a teacher uses multiple spaces? If so, maybe we can ask them if they are sure that they want to make only one (or x amount) account(s). 

 PS. I  tried what would happen if I copied a text into this field, so naturally I copied the Wikipedia page for kroketten, and now I have a student with the username 'een eierbal is een grote bitterbal met een gekookt ei in het midden.' 😂

static/js/teachers.ts
```diff
@@ -547,102 +548,129 @@ export function add_account_placeholder() {
...
+          padding-left: 10px;
+          padding-right: 10px;
+          padding-top: 5px;
+          padding-bottom: 5px;
+          font-size: 24px;
+          border: 1px solid gray;
+        }
+      }
+    </style>`;
+    newWindow.document.write(table!.outerHTML + css);
```
jpelay, line 589: I think here instead of a bang we can use optional chainning with the `?` operator.

static/js/teachers.ts
```diff
@@ -709,6 +727,21 @@ function setLevelStateIndicator(level: string) {
...
+export interface InitializeCreateAccountsPageOptions {
+  readonly page: 'create-accounts';
+}
+
+export function initializeCreateAccountsPage(_options: InitializeCreateAccountsPageOptions) {
+  const accountsInput = document.getElementById("accounts_input") as HTMLFormElement;
+  if (accountsInput) {
+      // Apparently we need the setTimeout to get the whole text of the textarea because the event provides only
+      // the text that was in the clipboard which does not work in if the user is appending text and not replacing it.
+      accountsInput.addEventListener('paste', function() {
```
jpelay, line 739: Instead of the if, we could also use the ? operator here, it would be a bit sleaker I think!

It'd be something of the sort of:

```ts
accountsInput?.addEventListener('paste', function()
```

tests/cypress/e2e/for-teacher_page/class_page/add_students/create_accounts.cy.js
```diff
@@ -1,58 +1,53 @@
...
-      ({classname, students} = createClassAndAddStudents());
-      // download login credentials
-      cy.readFile('cypress/downloads/accounts.csv');
+    it('Is able to create student accounts with usernames', () => {
+      const className = createClass();
+
+      navigateToClass(className);
+      cy.wait(500);
+
+      cy.get('body').then($b => $b.find('[data-cy="survey"]')).then($s => $s.length && $s.hide())
```
jpelay, line 18: This line is actually not needed anymore!

tests_e2e.py
```diff
@@ -1584,16 +1584,19 @@ def test_invalid_create_accounts(self):
             self.post_data('for-teachers/create-accounts', invalid_body, expect_http_code=400)
 
     def test_create_accounts(self):
-        # GIVEN a new teacher
+        # GIVEN a new teacher with a new class
         self.given_fresh_teacher_is_logged_in()
+        class_ = self.post_data('class', {'name': 'class1'})
```
jpelay, line 1589: Yes, this is great! We are also testing on Cypress, so this is finel.

jpelay (changes requested):
Hello Boryana! I tested all the scenarios you presented in the PR and they worked great. I just left a couple of comments, just minimal changes really. And also, I took the liberty to add a bit of styling to the table so that it would adhere to our standard.

AnneliesVlaar:
Maybe the pop-up "You will create x accounts. Are you sure?" can provide a preview of the accounts that will be created, for example: 
create 31 accounts

AnneliesVlaar:
I really like the "copy to clippy" button and that when you past it in Word it is a table that you can adjust to your own needs 😄

AnneliesVlaar:
I did see the error message, but there was no row specified. The 'x' to close the message was next to the word row, I first thought there was an error on row x. 

Scherm­afbeelding 2024-10-25 om 10 51 46

Why do we block the use of '@' and ':'?

AnneliesVlaar:
Something small that I noticed was the font of the passwords. In the preview on the website they are in Sans-Serif, when pasting in Word it is also a Sans-Serif font, and the print preview is Serif. In the Sans-Serif font I found it very hard to see the difference between a small letter L (l) or the capital letter i (I). Maybe we should change the font in the previews to Serif or avoid using difficult characters (also like zero (0) and capital o (O)).

AnneliesVlaar:
Can we add some 'Hedyness' to the print preview? Maybe something like this?
Hedy

PR #5918, opened 2024-11-07
📜 Slides update

### LINKED ISSUE #5028: 📜 Slides sound "learner facing"
The content on the slides is meant to be used in lessons, but sounds a bit like it is speaking directly to an (individual) student, f.e. here in level 13:

image

Let's try to make this a bit more systematic?

### LINKED ISSUE #5051: 📜 New concepts missing in slides
**Describe the issue**
I expected that the slides for a level introduce all new concepts for that level, but I find this is not the case.
I use the slides when preparing for a lesson and during lessons, so I sometimes find out during a lesson that something new is available in the level that I didn't introduce to the kids. 

Examples I found:

Level 2: (turtle) angle (instead of only left and right)
Level 4: (turtle) color
Level 5: pressed (key pressed)
Level 6: is -> =
Level 12: not really any concept missing, but very concise, probably not enough to use in teaching
Level 16: the slide tells us 'random' is changing, but doesn't show in which way it changes
Level 18: very concise... But that's probably not a problem, by the time students reach that level, they should be able to read the instructions on the adventure tabs.

### LINKED ISSUE #5805: 📜 Add teacher explanation to the slides
Add a new field in the yaml with explanation that we can show as notes in Powerpoint (when this is done, we will also need to show it).

MarleenGilsing:
fixes #5805 
fixes #5051 
fixes #5028 

Started working on updating the slides. I added notes for the teachers as well. I've done level 0, 1 and 2 now. 

@Felienne and/or @AnneliesVlaar, I'd love to hear what you think of the notes (then I'll take that into account before I continue with the rest of the levels).


MarleenGilsing:
@AnneliesVlaar They are presented like the speakers notes you can add to a powerpoint. So the teacher can see them when presenting (if they are using a seperate screen to presnet from). Or they can just look at the notes before giving the lesson of course. 

It doesn't need to be detailed feedback (unless you feel like it of course, haha), but more 'over all feedback'. Like, if you were a teacher without much programming experience, would this be helpful? Or would you like more/less info? Would you like a whole script of what to say, or are these short notes enough?

content/slides/en.yaml
```diff
@@ -60,6 +89,8 @@ levels:
 
                 `{print}` is used to show text on the screen.
             code: '{print} Show text on the screen'
+            notes: |-
```
AnneliesVlaar, line 92: Maybe we can do a small recap of the first presentation by asking 'what will the output be' or 'what do you need to do to make the computer follow the instructions'

content/slides/en.yaml
```diff
@@ -78,57 +111,121 @@ levels:
             code: |-
                 {print} Welcome to your own rock scissors paper!
                 {ask} What will you choose?
+            notes: |-
+                Show that the computer now asks you a question. Fill in an answer and also show that once you fill in the answer, nothing happens.
+                Make your students aware that the computer will only follow your instructions, it will not do anything by itself. This means that if you want the computer to print the answer, you have to give that instruction.
```
AnneliesVlaar, line 116: I think these kind of instruction can be really helpful to grasp programming in Hedy when they don't have much experience yet.

content/slides/en.yaml
```diff
@@ -78,57 +111,121 @@ levels:
...
             editor: /hedy#print_command
+            notes: |-
+                Lastly show the students that they can copy the example code into their programming field by pressing the yellow button.
+        8:
+            header: Let's get started!
+            text: |-
+                This is a good time to try the Parrot, Haunted House, Story and Rock, Paper Scissors adventures.
+            editor: /hedy/1
+            notes: |-
+                These are the first adventures to start with. Once (most of) the students are done, you can introduce the drawing turtle and the `{play}` command.
```
AnneliesVlaar, line 146: Whe don't mention the debug slide here. When is good moment to show debugging?

content/slides/en.yaml
```diff
@@ -78,57 +111,121 @@ levels:
...
+                {forward} 100
+                {turn} {left}
+                {forward} 50
+                {turn} {left}
             debug: 'True'
-        13:
-            header: Let the programming fun begin!
-            text: Enjoy the adventures in level 1!
+            notes: |-
+                The debugger can come in handy when making a drawing. The students can now continue with the rest of the adventures of this level.
```
AnneliesVlaar, line 204: I think the debugger is useful to see the turtle executing the code step by step even if there is not a bug. Maybe it is good to elaborate on that here.

content/slides/en.yaml
```diff
@@ -78,57 +111,121 @@ levels:
...
+
+                Which adventure did you like least? Why?
+
+            notes: |-
+                Ask the students for their opinion.
+        17:
+            header: Show and tell!
+            text: Would anyone like to show the class a program they've made?
+            notes: |-
+                Let some students show their work to the class.
```
AnneliesVlaar, line 223: Would it be helpful to give a note on how to do that (if students in a class on the class page you can show adventures of students)? Maybe that is too much and we don't want that kind of information on different places.

content/slides/en.yaml
```diff
@@ -164,13 +270,19 @@ levels:
...
         5:
             header: Sleep
             text: You can also use the `{sleep}` command to pause the code for a while.
             code: |-
                 {print} My favorite colour is...
                 {sleep} 2
                 {print} green!
+            notes: |-
+                Show the students the sleep command. Ask them what they think the 2 stand for.
+                Change the 2 into another number to show that they can decide how long the code will be paused.
```
AnneliesVlaar, line 285: "Change the 2 into another number to show that they can decide **for how many seconds** the code will be paused"

content/slides/en.yaml
MarleenGilsing: I thought about doing the debugger first as well, but then I thought the instructions would take way too long and it would be better to first get the students to work and later show the debugger. They don't really need it in the first level anyway because the programs are only a couple of lines.

MarleenGilsing:
I've updated the slides upto level 8. They now include notes for the teachers. The workbooks are also referenced. 
I'll update the other levels when we have the workbooks for those levels as well.

MarleenGilsing:
Hi @ Ruth 
There is a lot of work to do on the slides! Thank you for helping out! I have made a list of what needs to be done. Please start with the first taks (throwing out old content), because else you'd waste time updating it and then throwing it out anyway, haha. 

Can you do this for **all languages** that have slides **except for English (English is to to date!)**? 

**Throw out old content**
* Can you throw away all old English content. This means that if you come across English content in another language file, this can be thrown out. This is the old English content and doesn't need to be in the file. It doesn't matter if there is content missing, because all missing content will be supplemented with the English most up-to-date version. 

**Change the numbers of the levels:**
Level 0 through 4 stay the same
Level 5 stays the same, but has slide 6 ('another new command') removed, this means that the numbers of the slides need to change too (6 is removed, so 7 is the new 6, 8 is the new 7 etc etc) 
Level 6 becomes 7 (this means there is no level 6 for now, I will create content later, this is not a problem)
Level 7 becomes 8
Level 8 becomes 9
Level 9 becomes 10
Level 10 becomes 11
Level 11 is thrown away
Level 12 stays the same, but has slide 2 and 3 thrown away (so the slides numbers need to change)
Level 13 through 18 are thrown away (this means there are no level 13,14,15 and 16, content needs to be created for this later). 

**Other changes:** 
* Each first slide might say 'welcome to level 1' (or 2 etc) can you make sure they match the new levels?
* Each level also might end with 'Enjoy the adventures in level 1!' Can you make sure they match the new levels?

**Update the Dutch translation**
* Can you compare the Dutch file to the English one and make sure the Dutch file is up to date as well? 

**Add content for the higher levels**
* Level 13 - 16 do not have have any content yet. Could you create content for these levels?

If you have any questions, you can find me on discord too.

wgzwemer:
**Progress on slides so far:**
 - I threw out all old content and filled the empty files with brackets.
 - I changed the numbers of the slides in all languages and adjusted the 'welcome' and 'good-luck' page accordingly.
 - I updated the Dutch translation until level 11.

**What I will do next:**
- Update English slides of level 12
- Make new content for level 13-18.
- Translate this new content to Dutch?

Issue #5924, opened 2024-11-08
💻 Remove Parsons


jpelay:
Commented by Marleen:

> In the parsons problem it's not really clear that you can do 2 exercises and that you have to click the 2. (Same as the old design though).

AnneliesVlaar:
I think the easiest solution would be to separate the puzzles in different tabs/exercises. Puzzle 1; Puzzle 2

Felienne:
Parsons should be removed, they are not well integrated in the trajectory and not well-designed (very different from the quiz, for example) The idea of parsons might as well be done with paper sheets.

boryanagoncharenko:
The comment for this issue left from the meeting on 6 Dec is to check how often parsons are used.

Based on my research, users press the `Run code` button on a parson puzzle 190 times per day on average. This has been measured on the amount of calls to the /store_parsons_order endpoint which is called whenever the user attempts to execute a parson program.

@AnneliesVlaar @MarleenGilsing please let me know whether you need further information on the usage or you would like me to proceed with deleting parsons.

MarleenGilsing:
190 seems like a big number at first, but i guess if we compare it to normal adventures it isn't much, right? Especially if 190 clicks is all the 34 parsons puzzles combined? And one child might click multiple times as well?

boryanagoncharenko:
On average, users click the `Run code` button for non-parson programs 5820 times a day and only 190 times for parson programs.

It is expected that one user clicks on the `Run code` button multiple times when solving a problem regardless if it is a parson or not. I am afraid it is hard to provide a good estimate on the amount of unique users and the specifics of their interactions. To get a grasp, I examined a smaller subset of the logs and noticed that:
- on weekdays we usually have about 50 different users who use parsons. On weekends the amount of users using parsons goes down to 10.
- about 3/4 of the parson users click the `Run code` button up to 5 times. The rest of the users (1/4) click it 6 or more times, sometimes going up to 100 times in their active session.

Let me know if you need further information.

MarleenGilsing:
Alright, seems like they are not that popular. The layout definitely needs an update if we would keep them, so maybe for now we can just remove them and think of ways to ingerate them into the workbooks.

MarleenGilsing:
Yes, agree. @AnneliesVlaar Do you agree too? 

I think we should also: 
* Check the teacher manual for mentioning of parsons puzzles and change content if nessecary. 

I could do this too.

AnneliesVlaar:
[...]

Issue #5923, opened 2024-11-08
🪲When finishing a quiz you're thrown to the next level quiz


jpelay:
This happens both when you're logged in or out

PR #5926, opened 2024-11-08
💻 Improvements to redesign of adventure page

jpelay:
- [x] The clear button is gonna create some very angry students who accidentally removed all their work. Solutions: Can we have an 'are you sure you want to remove all your code' pop-up? An 'undo' button would work too, or ctrl+Z? Or, last option: Do we even need the clear button? How many times does a student want to clear the whole field, and even if they do, it's not much effort to do so without having a clear button...?

- [ ] We need Dutch translations for some new buttons. => Maybe someone can add it here?  Or in Weblate?

- [ ] The button says 'show editor' while I think we have been using the term 'input screen'. I think editor might be better though, so maybe it doesn't need to change? => @Felienne I dont have preferences either way, do you agree?

- [x] There are nearly no margins at the sides (maybe because I don't have a very big screen) which makes it look a bit crowded. Can we add some margins? Same goes for the text of the adventures, I think it would look better if tehre was some white space on the sides of the text field.

- [x] I don't really like the underling of 'level 1 - introduction', I think it might look better without the underlining, but that might just be my taste. The underlining also makes it feel clickable, which it is not. => I made it a little thicker, but I'm keeping it to solve another point further down.

- [ ] I really love the 'see more adventures' menu! I do think it's slightly counter intuitive that the button is on the right, while the menu appears on the left. => I don't think it's that counterintuitive due to us reading from left to right, in the case of languages that read from right to left, it appears on that side. 

- [ ] Extra idea for the 'see more adventures' menu: Would it be possible to show in the menu which of the adventures you have completed? Like with a little tickmark or something behind the adventure? => This makes a lot of sense!! @Felienne I think we'd want here the tickmarks we are now using in the tabs? Shall we also add a tickmark next to the title? 

- [ ] (low priority) why is the 'e' in our font a smaller size? It looks kind of strange? => Haha, no way for me to fix this sorry! Maybe I can change the font!

- [x] (low priority) Before you could see if the adventure was just another adventure, or if you were learning a new command because of the tabs colours. We don't really have that anymore, exept for in the menu, but kids probably won't use that much. Maybe we can think of another way to show the students that they will learn a new command in this adventure. => This is why I didn't remove the underline, you can see now the type of adventure in the different colors of the underlining.

- [ ] In the parsons problem it's not really clear that you can do 2 exercises and that you have to click the 2. (Same as the old design though). => Opened #5924

- [ ] BUG: If you finish the level 1 quiz and click 'go to level 2' you'll be sent to the quiz of level 2, instead of the introduction. => Opened #5923

**How to test**

In regards of the margins, you can set the size of the screen in the responsive view mode to: 1077x524


MarleenGilsing:
Looks great Jesus! I like the idea of making the underlining functional!

MarleenGilsing (approved):

PR #5939, opened 2024-11-14
💻 Merge first line of error messages with details

### LINKED ISSUE #4729: 💻 Rewrite first line of error messages
We have a project open for the phrasing of error messages (#3450) but I think there are a few other, smaller, things we can also do. I am interested in picking this up myself now that other stuff is working well.

I will open two issues today for discussion later on.

We need to critically assess the error message layout! Errors now always that with the same message in bold:
image

This was really not a clear design decision. I think someone (maybe even me!) early on added this when we had 0 users and we never got back to it because there were so many other things to work on like teacher features.

Now that we have more time to think about this, I really believe we should remove this message, since: 

* it does not add information (is always the same)
* it takes up space 
* it grabs the attention because it is bold

Any thought?

jpelay:
Refactors the first line of error messages. When the caption doesn't add  anything it removes it, and when it does, its shown in the details of the error message.

These are the new error messages:

Connection_error
image

Execute_error
image

Other_error
image

Transpile_error This was entirely removed
image

Warning: removed as well
image

Fixes #4729 
**How to test**


AnneliesVlaar (approved):
The error and warning messages looks much better now that the captions are removed or merged with the message. I do think that some captions can be shorten, but I will open a different issue for that.

PR #5943, opened 2024-11-15
📜 #5899 explaining double quotes

### LINKED ISSUE #5899: 📜Single or double quotation marks
While translating Feliennes worksbooks I saw a lot of exercises that contain either single or double quotation marks. We don't use double quotes in the adventures, only singles, but I think it makes a lot of sense to use the double ones, or at least teach the students that you can use doubles to prevent problems with contractions. Should we add this to the adventures?

MarleenGilsing:
fixes #5899 

Explains to use double quotes instead of changing the spelling when there's an apostrophe. 

Am I correct to assume that I don't have to change anything in other languages now because of @boryanagoncharenko 's awesome update?


boryanagoncharenko:
Yes, you are correct. There is no need to copy-paste content. The English version will appear for all other languages until they get a translated version.

boryanagoncharenko (approved):

PR #5960, opened 2024-11-22
📜 Cleaning up old English content from the adventures

MarleenGilsing:
Removes all the english content from non english adventures yamls


boryanagoncharenko:
@MarleenGilsing, thanks for removing old English content. Great work!
Since this may cause merge conflicts with the daily Weblate updates, I will merge this PR as is now. If there are more changes that need to be done, e.g. more languages, let's make them in a separate PR.

PR #5971, opened 2024-11-25
Validate content

### LINKED ISSUE #5935: 🪲 Translatable content is parsed as HTML
**Describe the bug**
Currently, Hedy has content available for translation that contains HTML and is parsed as such. For example, [learn-more-sections->[1]->text](https://hosted.weblate.org/translate/hedy/webpages/en/?checksum=fe0f106c95380ee5&sort_by=-priority,position) contains an <iframe> tag and slides_info in messages.po contains multiple <a> tags. At the same time we have a fully automatic way of accepting content contributions without reviews. This means that a translator could add HTML that could potentially cause harm to users.

boryanagoncharenko:
Fixes #5935

**How to test**


website/flask_helpers.py
```diff
@@ -70,3 +75,84 @@ def dumps(self, obj, **kwargs):
...
+        href_times = i.lower().count(prefix.lower())
+        whitelist_keys_present = [link in i for link in whitelisted_keys if link.startswith(f'#{prefix.upper()}_')]
+        return href_times == 1 and any(whitelist_keys_present)
+
+    # If the content contains something which remotely reminds of an HTML or markdown comment, it is suspicious.
+    if re.search(r'<!--', input_) or re.search(r'-->', input_) or re.search(r']:', input_):
+        return False, escape_special_chars(input_)
+
+    # Find all occurrences which look like an opening HTML tag
+    matches = re.findall(r'<\s*[^/][^>]+', input_)
```
jpelay, line 137: I'm not sure about this particular regex, I've been trying it, and I'm not sure that the `[^/]` is doing anything, since it matches it anyways when it gets to the `[^>]` part

website/flask_helpers.py
```diff
@@ -70,3 +75,84 @@ def dumps(self, obj, **kwargs):
...
+    'FOR_TEACHERS': 'https://hedycode.com/for-teachers',
+    'DISCORD_SERVER': 'https://discord.gg/8yY7dEme9r',
+    'MAIL_TO_HELLO_HEDY': 'mailto: hello@hedy.org',
+    'DISCORD_VIDEO': 'https://www.youtube.com/watch?v=Lyz_Lnd-_aI',
+    'CUSTOMIZE_ADVENTURE': 'https://github.com/hedyorg/hedy/assets/80678586/df38cbb2-468e-4317-ac67-92eaf4212adc',
+}
+
+whitelisted_keys = {v for k in whitelisted_links.keys() for v in {f'#HREF_{k}#', f'#MD_{k}#', f'#SRC_{k}#'}}
+
+whitelisted_html_tags = ['a', 'b', 'em', 'strong', 'img']
```
jpelay, line 96: We also use an `iframe` for Felienne's conference in strangeloop: 
https://github.com/hedyorg/hedy/blob/7fdaf40397ce3240a2028fc2a83d9bcbd6f5a783/content/pages/en.yaml#L69-L70

translations/en/LC_MESSAGES/messages.po
```diff
@@ -349,7 +349,7 @@ msgid "classes"
 msgstr "Classes"
 
 msgid "classes_info"
-msgstr "Create a class to follow the progress of each student in dashboard, and to customize the adventures your students see, and even adding your own! You can create as many classes as you like, and each class can have multiple teachers each one with different roles. You can also add as many students as you want, but mind that each student can only be in one class at a time. You can find more information about classes in the <a href=\"https://hedy.org/for-teachers/manual/preparations#for-teachers\">teacher manual</a>."
+msgstr "Create a class to follow the progress of each student in dashboard, and to customize the adventures your students see, and even adding your own! You can create as many classes as you like, and each class can have multiple teachers each one with different roles. You can also add as many students as you want, but mind that each student can only be in one class at a time. You can find more information about classes in the <a #HREF_TEACHERS_MANUAL_FOR_TEACHERS#>teacher manual</a>."
```
jpelay, line 352: I think we should surround these in `{}`, because if they change this it'd mess up their translation. You're not supposed to translate things that are surrounded with those!

jpelay (commented):
I think this works great, and yeah the only issue would be to replace all of the link we already have.  I like that even if English content is bad, it still doesn't show it as a link:

image

Not ideal still, in the case someone does put the URL there and we give them free SEO.

website/flask_helpers.py
boryanagoncharenko: The `[^/]` part is there so that we do not match the closing tags, which I want to disregard when looking for html content. It is still a very naive solution.

website/flask_helpers.py
boryanagoncharenko: I am not sure why this iframe is part of the .yaml file and not in the .html file. My intention is move it to the page itself.

translations/en/LC_MESSAGES/messages.po
boryanagoncharenko: Excellent idea, thanks!

PR #5972, opened 2024-11-25
🧹 Remove the `awscli` dependency

rix0rrr:
This dependency is causing some dependency installation problems on Windows, so get rid of it.

**How to test**

Hopefully this unblocks the Hedy Offline build when the next release comes around. If not, we'll dig deeper again with the error log that we have at that point.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,6 @@ ruamel.yaml==0.18.6
 docopt==0.6.2
 pydantic==1.10.13
 lazy==1.4
-awscli==1.32.77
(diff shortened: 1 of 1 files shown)


rix0rrr:
> This seems to create a new problem :(

Ah yeah. `awscli` used to bring in `colorama`, and we seem to be using `colorama` without having our own dependency on it. So now we're not installing it anymore. I've added the dependency back.

Felienne (approved):
Thanks!!

PR #5985, opened 2024-11-30
🪲 Don't crash the "view program" page if the program has a syntax error

rix0rrr:
It used to be that the "view program" page would crash if a program has syntax errors. The reason is that we try to translate programs from the language of the author to the language of the viewer, but that operation may fail due to parse errors, which would cause the page to fail loading.

We used to only check for blanks and avoid translating a program because we knew that programs with blanks would fail to parse; but there are many more reasons that would case a parse failure, so instead of trying to predict them I'm now just catching the error.

In this PR I want to address a failure case as quickly as possible. There are some open issues/questions that someone else could/should have a look at later:

- I have not written a test yet to make sure this doesn't re-occur.
- We do translation in 2 steps: first to English, then to the target language. I don't know if that's on purpose or not; I imagine so, it's what the original code was doing and it must be for a reason.
- The original code used to always normalize to English first, then translate to the target language. I've changed it to only translate if the source and target languages are different in order to save compute time. I hope that's doesn't break any use cases I'm not aware of.
- We have no way to surface the parse error the user. That may be fine because they can try running the program and then seeing the parse error, but just something to note.

**How to test**

Write a syntactically invalid program and save it.

Use the `/hedy/.../view_program` link to view it, and observe that the page loads successfully.


jpelay:
Hello Rico, thanks for the PR. 

> * I have not written a test yet to make sure this doesn't re-occur.


I've just added a test for this case! Tried in main and it fails, and in this branch it succeeds.

> * We do translation in 2 steps: first to English, then to the target language. I don't know if that's on purpose or not; I imagine so, it's what the original code was doing and it must be for a reason.

We can't translate two arbitrary languages with each other, so we use English as a bridge, we can update the code, but I don't feel that's worth it right now.

> * The original code used to always normalize to English first, then translate to the target language. I've changed it to only translate if the source and target languages are different in order to save compute time. I hope that's doesn't break any use cases 

I'm not aware of.
This is great! Thanks!

> * We have no way to surface the parse error the user. That may be fine because they can try running the program and then seeing the parse error, but just something to note.

I think it's apparent that this program doesn't work, but indeed we are not letting the user know that there's a problem here. I think we can send a flag to the front end that the program contains an error and then show an alert somewhere. What do you think?

image

rix0rrr:
> We can't translate two arbitrary languages with each other, so we use English as a bridge, we can update the code, but I don't feel that's worth it right now.

If this is the case: `translate_keywords` has a `from_lang` and a `to_lang`. 

* Does it check that at least one of those two arguments has the string `"en"` and throw an error otherwise? It should (but a quick look at the code didn't show that)
* If it can do that, it can also call itself recursively twice if it detects that NOT one of the arguments is `"en"`. That moves the burden of being aware of what `translate_keywords` can do from the *consumer* of `translate_keywords()` to a performance question of the *implementor* of `translate_keywords()`.

> What do you think?

I think it's not worth it to invest time into. The error will show itself if the viewer clicks "Run program".

tests/cypress/e2e/hedy_page/invalid_program.cy.js
```diff
@@ -50,7 +51,7 @@ describe('Error code gives correct error', () => {
         const error_message = `We detected that forward doesn't work with lalala because it is text. Can you try changing lalala to a number or input from ask?`;
         goToHedyPage();
 
-        cy.get('#editor > .cm-editor > .cm-scroller > .cm-content').click();
+        codeMirrorContent().click();
```
rix0rrr, line 54: ❤️

rix0rrr:
Thanks for the test!

jpelay:
> Thanks for the test!

No worries, I'm actually just now seeing that we can translate any arbitrary languages indeed (I think we couldn't before, but may be I misremember) so I'm adapting that code as well.

rix0rrr:
Lovely, thanks! Feel free to merge when you feel good about it.

rix0rrr (commented):
Dont add waits it’s never the answer

jpelay:
> Dont add waits it’s never the answer

😭 I added a nice wait for an intercepted link and it's not working, so I'm debugging the issue!

jpelay (changes requested):
Dont merge until i figure out the test

jpelay:
Instead of a wait I added a timeout, still not the best, but a bit better I think!

jpelay (approved):

PR #6005, opened 2024-12-02
📜 Correct the message for keypresses

### LINKED ISSUE #3856: [BUG] Change "Waiting for a button press..." into "Waiting for a key press..." when appropriate
**Describe the bug**
While waiting for a key press or a button press, the message below the output screen will always say "Waiting for a button press.." Even when waiting for a key press.

For example in level 5:
```python
if x is pressed print "hello world"
```

**Add a screenshot**
image

boryanagoncharenko:
This PR changes the message displayed to the user `Waiting for a button press` to `Waiting for a keypress`. This is needed because we do not have buttons and all content refers to keys, not buttons. Also note that keypress is one word.

Fixes #3856

**How to test**
Go to level 5 and write any if-pressed program (e.g. `if x is pressed print 'x' else print 'other'`) and run it. Check that the message in the output screen is updated.


jpelay (approved):
Works!
image

PR #6013, opened 2024-12-04
🪲 www.hedycode.com not redirected

rix0rrr:
When you search Google for Hedy, `www.hedycode.com` is still the first hit instead of `hedy.org`. The reason is we were only redirecting `hedycode.com` without the `www` properly, but not the www-domain.

Also redirect the `www`-domain.

**How to test**

Cannot be tested offline. Deploy first, then visit `www.hedycode.com` and observe that you are redirected.


boryanagoncharenko (approved):

jpelay:
Just tested this on production, and it works!

PR #6014, opened 2024-12-04
🪲 Small reorganization of the for-teachers page 

### LINKED ISSUE #5967: 💻 Confusing UX in teacher overview
If, as a teacher, you have classes opened, but then open adventures, nothing chances as the looks of it, as adventures open below.

Hope this makes sense, if not, I can show in the meeting next week!

Steps to reproduce are:

* Open classes and make sure you have a few so that the later ones are not in view:
image

* Then open adventures and it looks like nothing happened, or that all your adventures are deleted (that is what I though, for a minute there!)

image

boryanagoncharenko:
This PR proposes minimal changes to the layout of the for-teachers page, in order to avoid the issue of "where did all my adventures go?"-situation. Specifically:
- The tables that show specific content, e.g. classes, adventures and slides are now displayed immediately the button that determines their visibility.
- Because the tables have less space, I made the text of their inner buttons on smaller screen size. Probably this is something that we should do regardless if the tables are in their separate boxes or not.
- Note that I specifically did not implement an auto-close feature (if one of the tables is displayed, the rest get collapsed). With the current layout changes I think it makes more sense to not have auto-close because we set the visibility through buttons nested in the right side of panes, which is already non-standard. Ofc I may be wrong; if you feel it should be there, will add it :)

Fixes #5967

Here is how it looks:
Screenshot 2024-12-04 at 17 02 13
Screenshot 2024-12-04 at 17 02 20

This is how the action buttons collapse to icons without text when the screensize is small:
Screenshot 2024-12-04 at 17 02 35
Screenshot 2024-12-04 at 17 02 46


rix0rrr:
I don't mind this redesign, and this is definitely more clear, but I thought the tiles were on purpose.

Getting rid of the tiled layout seems like a bigger, and also unnecessary, change at this moment?

In any case I'll ship it because it seems strictly better. Good thought on the icon labels in the tables! 

## Unimportant aside about table buttons

On that: we really need some data set with a lot of data in it to test layouts; in a table with a lot of rows the buttons really start to overwhelm. Check this out:

image

Look at all that redundant and duplicated space dedicated to actions that a user will only rarely use! And look at how overwhelming it looks. I think ultimately we'd want to move to something more like this:

image

(And get rid of the zebra stripes... those are a bit 90s design 😉)

But that's not relevant for this PR! Just sharing my thoughts

rix0rrr (approved):

jpelay:
image
> (And get rid of the zebra stripes... those are a bit 90s design 😉)

This makes a lot of sense yes!

PR #6025, opened 2024-12-06
🪲 Display ask and keypressed modals below the output pane

### LINKED ISSUE #5946: 💻 Show "Waiting for a button press.." under the output field
The text "Waiting for a button press.." when using the 'if key pressed' functionality is placed on top of the output screen. This prevents from seeing the text:

Scherm­afbeelding 2024-11-16 om 13 29 25

Suggestion is to place the text "Waiting for a button press.." under the output screen by showing a smaller output screen.

boryanagoncharenko:
This PR ensures that the ask modal and the key-pressed modal displayed in the output pane, appear below the output editor instead of on top of it.

Fixes #5946

**How to test**
The output appears in 3 different places and each of them have to be tested:
- output on the /hedy page
- output on the /tryit page
- output in teacher slides (embedded editor) 

I tested with the following code in level 15. Make sure to test consecutive displaying of only the ask and only the key-press modals, and then alternating between the two.
```
for i in range 1 to 10
    print 'Some stuff'
for i in range 1 to 10
    if x is pressed
        print 'ya'
    else
        print 'nee'
    a is ask 'what'
    print a
```

It is essential to check that:
- no modal is displayed on top the output.
- the output window always scrolls to the bottom. It should not be the case that when a modal appears beneath the output, its content gets scrolled up because the latest content will remain hidden (specifically test this for if-pressed).
- check if there are any changes for the turtle canvas
- note that showing the variable list seems to be broken and not related to this change


Felienne (approved):
Nice, thanks!!

image

Love that you also tested the new editor

image

PR #6031, opened 2024-12-09
🚚 Support 'begins_with' condition in Dynamo

rix0rrr:
Add support for the DynamoDB 'begins_with' condition. This condition can be used on sort keys for efficient lookup, or on non-key fields for inefficient (but convenient) filtering.

Also in this PR:

- Rename `filter` to `server_side_filter` for `get_many`; this doesn't *remove* the old argument but it is considered deprecated starting today.
- Rewrite the documentation about server side filtering to be a bit more clear.

**How to test**

No functional changes, so nothing to test.


jpelay (approved):
Thanks a lot Rico!

PR #6039, opened 2024-12-12
💻 Remove developer mode 

### LINKED ISSUE #5987: 💻 Remove developer mode
I feel it is a distraction and does not add a lot. It would be better to maybe, one day, add a sandbox if we have decided how to deal with the higher levels. For now, remove

boryanagoncharenko:
Removes:
- the developer's mode toggle under the code editor
- the mandatory developer's mode option in the class customization

Fixes #5987

**How to test**
- As a user, you should not be able to see the toggle for programmer's mode. Even if you are part of a class that has this a mandatory developer's mode customization. Whatever combination of adventure/quiz/puzzle sequence you open, the editor should have correct size. Use the parson puzzle on level 6 because it has 5 blocks which makes it higher than usual.
- If you are a teacher, you should not be able to see the mandatory developer's mode option in the class customization.


Felienne (approved):
Thanks!! Makes the interface a lot cleaner!

PR #6070, opened 2024-12-22
:denmark: Add some translations for Danish

Raphexion:
Add some translations for Danish without using Weblate. Using custom scripts.


Felienne:
Thanks for your first Hedy PR @Raphexion! Let us know if you need help!

Raphexion:
Merry Christmas @Felienne :christmas_tree: :santa: 

First and foremost, thank you for you amazing work with Hedy. It is an amazing tool.

I plan to use Hedy next year to teach 10 year-olds programming. I am a volunteer for the Danish Coding Pirates. But, I am not Danish myself :upside_down_face: . So needed try a different approach than manual translation.

I wrote a script to extract the messages in English, Dutch and Spanish and them them to a LLM (mistral). And replaced them inplace for the da .po file. But before commit I wanted to verify the quality. So, now during Christmas, I've asked some relatives if they would accept the translation. 

Unfortunately, almost all translations have some "problem". So my automatic approach is not working. 

I will come back in the future with another approach. I only want to add good quality translation.   

I will close this PR.

Felienne:
Thanks for the very kind words @Raphexion and happy New Year to you!! Always great to hear that people find value in Hedy!

And also thanks for letting us know about the LLM, that is in fact very helpful! We have over the years considered LLMs for translations and we were always too worried about the quality and your experiences give us more clarity about that!

Do let us know, here or on Discord, if there is anything else we can do to support you to start teaching or translating!

PR #6073, opened 2024-12-27
🪲 Fix programs page redirect

jpelay:
Right now the programs page and my profile page is failing in Alpha. When looking at the logs this is what comes up:

```
werkzeug.routing.exceptions.BuildError: Could not build url for endpoint 'programs_page' with values ['page']. Did you mean 'app.programs_page' instead?
```

Investigating a bit, this might be caused because the url is not registered yet, like [here](https://stackoverflow.com/questions/51234212/werkzeug-routing-builderror-could-not-build-url-for-endpoint)


jpelay:
Well this didn't work 🤡

PR #6075, opened 2024-12-27
🪲 Fix pagination links

rix0rrr:
The pagination links in various pages were using `url_for('programs_page')`, which builds a URL to a route on the global `app` object.

Since #6052 we don't have a global app object anymore. Instead, we use a global Blueprint called `app` instead.

In this PR, make all `url_for` calls in the same `app.py` file Blueprint-relative by prepending a period: `url_for('.programs_page')`, and make the `url_for` in the teachers page that points to a user's programs page (probably legacy and unused?) absolute by prepending the full Blueprint name: `url_for('app.programs_page')`.

Add tests to make sure that the page renders successfully.

**How to test**

Automatic tests have been added, so if they pass we should be good.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1126,9 +1126,9 @@ def programs_page(user):
     sorted_adventure_programs = hedy_content.Adventures(g.lang) \
         .get_sorted_adventure_programs(all_programs, adventure_names)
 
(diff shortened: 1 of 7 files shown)


jpelay (approved):
Thanks!

PR #6080, opened 2024-12-31
📜 Add tags to mailchimp users

boryanagoncharenko:
This PR adds the following functionality:
- when a user, who is subscribed to the newsletter, changes their country, the change is reflected in the Mailchimp subscriber profile.
- when a user creates or customizes a class, a corresponding tag is added to their Mailchimp subscriber profile.

Fixes partially #3995

**How to test**
Ensure you have access to the Mailchimp account of Hedy.
- Create a new teacher account and subscribe to the newsletter. Check that the Mailchimp subscriber has the tags 'teacher' and the country you selected.
- Create a class and do not customize it. Check that the Mailchimp subscriber has a 'created_class' tag.
- Customize the class and check that the Mailchimp subscriber has a new tag: 'customized_class'
- Go to the profile page of the teacher and change their country. Check that the Mailchimp subscriber has the country tag updated accordingly. The old country, if any, should be removed and the new country, if any, should be added.
- Go to the profile page of the teacher and change their email. Check that there is a Mailchimp subscriber with the new email address and that all tags from the old subscriber are present. Check that the old Mailchimp subscriber is archived.

Note that currently you cannot subscribe to the newsletter after you signed up. So you cannot check if a teacher with existing customized classes gets the right tags when they subscribed.

To consider:
All functions for managing subscriptions are blocking but we do not need their output to continue with the response and the calls to Mailchimp are not retried. Is it worth investing in these for this feature?


website/auth.py
```diff
@@ -36,42 +32,6 @@
 
 env = os.getenv("HEROKU_APP_NAME")
 
-MAILCHIMP_API_URL = None
```
rix0rrr, line 39: Nice!

website/classes.py
```diff
@@ -50,6 +51,7 @@ def create_class(self, user):
         }
 
         self.db.store_class(Class)
+        add_class_created_to_subscription(user['email'])
```
rix0rrr, line 54: I don't know how fast or reliable the mailchimp API is, but in case it's slow or buggy: any way we can avoid doing this if it's already been done? (By keeping flags in Dynamo or sth?)

Also are we catching exceptions for low-value API calls like this?

rix0rrr: It's a bit of a bummer this feels weird to be part of `database.py`, no?

Otherwise that would seem like a much better place for all these repetitive calls...

website/newsletter.py
```diff
@@ -0,0 +1,149 @@
...
+        "Authorization": "apikey " + os.getenv("MAILCHIMP_API_KEY"),
+    }
+
+
+def run_if_mailchimp_config_present(f):
+    """Similar to 'requires_login', but also tests that the user is an admin."""
+    @wraps(f)
+    def inner(*args, **kws):
+        if MAILCHIMP_API_URL:
+            return f(*args, **kws)
```
rix0rrr, line 32: Super nitpick: probaby we shouldn't be doing `return f`: `f` can only ever return `None` because it may be that we're not calling it at all, in which case we will *definitely* return `None` so you can't rely on a return value anyway?

website/classes.py
boryanagoncharenko: The calls to Mailchimp are defenseless at this point - no exception handling, retries, timeouts, and they are blocking the response while they do not need to. I don't think retries are needed in this case - I confirmed this with the POs. We definitely need exception handling (so that we do not fail the whole call) and timeouts (perhaps maximum 5 seconds?). I am not sure about making the call non-blocking. This could be a fire-and-forget call but I am on the fence on the cost vs benefit. What do you think?

I did not think of storing the data in the DB to reduce the Mailchimp calls. These operations are not frequent and if we have timeouts + exception handling, would it be necessary?

I was tempted to put this in the `database.py` but it truly does not belong there semantically. I am also a bit afraid moving this to the database.py file could lead to potential bugs in the future because other maintainers would probably not expect it either.

rix0rrr: > other maintainers would probably not expect it either.

Yeah that's extremely fair. And I also hate to introduce another domain modeling layer just for the occasional call that needs to do both database and something else.

As for the calls themselves: I think tossing them onto a background thread and catching and logging errors is probably a nice middle ground. This should not be a huge lift.

Something like:

```python
def fire_and_forget(f):
  def run_this_and_catch_errors():
    try:
      f()
    except Exception:
      logger.exception('Exception in background thread')

  t = threading.Thread(target=run_this_and_catch_errors)
  t.daemon = True
  t.start()
```

boryanagoncharenko: Thanks for the pointers. I have seen the threading code but tbh I was a bit hesitant to spawn threads like that since they could go out of hand. Since the calls for creating or updating a subscription are not a lot, probably this will be ok and this is just me overthinking. So, let's go this way.

Btw I thought nowadays people use asyncio and I have no experience with it, so this is what I meant by cost of implementation.

rix0rrr: > Since the calls for creating or updating a subscription are not a lot, probably this will be ok and this is just me overthinking

That's what I'm assuming. If we're at all concerned, we should probably use this: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor, set it to a reasonable number of workers (could be 1), and just send jobs to that (that we never wait for). Maybe that's just the safer default anyway...

> Btw I thought nowadays people use asyncio and I have no experience with it, so this is what I meant by cost of implementation.

async io is quite a trend... and the benefits are also that it is trendy, not that it is better. Or at least, I'm not convinced that it is in common scenario's.

boryanagoncharenko: Let me check that I understand your suggestion. If we swap the thread.start() with creating a new ThreadPoolExecutor on every request, that too can go out of hand. So probably you mean that we should create a global ThreadPoolExecutor on app startup and use it whenever we need to run and not wait for a function. Is this correct? If this is the case though, based on the docs, a ThreadPoolExecutor does not seem like the right thing to use:

> All threads enqueued to ThreadPoolExecutor will be joined before the interpreter can exit. Note that the exit handler which does this is executed before any exit handlers added using `atexit`. This means exceptions in the main thread must be caught and handled in order to signal threads to exit gracefully. For this reason, it is recommended that ThreadPoolExecutor not be used for long-running tasks.

I did not realize asyncio is coroutines. I'd also not go for that in our case.

rix0rrr: I don’t think that docstring precludes using it, but the individual threads are probably also fine for our use case

rix0rrr (approved):

PR #6086, opened 2025-01-06
📜 Fix daily Weblate errors

boryanagoncharenko:
Fixes #6084

**How to test**
All automatic checks must pass


boryanagoncharenko:
@TamilNeram I am afraid I have to revert your contribution in Tamil because it causing a very large volume of errors and breaks the build. Since your commit contains a lot of changes, let's work out a way to resolve the errors. Would you be able to join our Hedy Discord server, so that we can discuss the issues there? Thank you advance!

PR #6091, opened 2025-01-07
💻 Remove developer's mode in the new design

### LINKED ISSUE #5987: 💻 Remove developer mode

boryanagoncharenko:
First, some context is needed for this PR. It has been decided that to avoid a thousand paper cuts, we will release all big changes together later this year. We already have a /tryit page which is an altered copy of the existing /hedy page and allows us to receive feedback on the design. We will add all changes to this new /tryit page while keeping the current /hedy page as is. For all features that goes out of the /tryit page (e.g. removing the enforce-developer-mode in the class customization setting does not reside on the /tryit page), we will use a flag.

This PR:
- removes the developer mode from the /tryit page while preserving it on the /hedy page.
- removes the enforce-developer-mode class customization option behind a tryit flag.

Fixes #5987

**How to test**
Ensure that everything on the /hedy page works without changes:
- With developer mode disabled, you should be able to switch between parsons, quizzes and code adventures and the editor should be the correct height. The developer toggle should disappear when you are on a parsons or quizzes page. Make sure you test this with a parson problem with 5 lines (e.g. level 5, parson 2).
- With developer mode enabled, you should be able to switch back and forth between parsons and code adventures (using the buttons below the code output) and the editor should have the correct height. 
- When you go to the class customization page, you should be able to see the 'mandatory developer mode' checkbox.
- Setting its value should enforce the developer mode for the enrolled students.

Ensure the new features are working:
- The class customization page does not have the 'mandatory developer mode' option when you add `?tryit=yes` to the end of the url in your browser. Removing the suffix or changing it to `?tryit=no`, should make the option appear.
- The /tryit page should not display the developer toggle. The height of the editor should be correct when switching between adventures and quizzes. Please do not test the parsons on the /tryit page because they will be removed next and it is ok if they are temporarily not working as expected.
- When a class has the developer mode enforced, the /tryit page should not hide the explanations.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -169,6 +169,20 @@ def is_testing_request(request):
     return not is_heroku() and bool('X-Testing' in request.headers and request.headers['X-Testing'])
 
 
+def is_try_it(req):
+    """ Many changes need to be released together, so they are done behind a flag - the TRY_IT env variable.
+    To facilitate testing, the flag can be overridden by adding a tryit=[0,1,false,true,yes,no] get param to
+    the request. In this way all users can switch between the 2 designs without the need to change an env variable.
+    Note that the get param override is added only because the specific features behind the flag are static and do not
+    require the flag to be passed down to further requests.
+    """
+    if req:
+        override = request.args.get("tryit", False)
+        if override:
+            return False if override in ['0', 'false', 'no'] else bool(override)
+    return os.getenv('TRY_IT', '0') == '1'
+
+
 def extract_bcrypt_rounds(hash):
     return int(re.match(r'\$2b\$\d+', hash)[0].replace('$2b$', ''))
 
--- a/utils.py
+++ b/utils.py
@@ -169,18 +169,17 @@ def is_testing_request(request):
     return not is_heroku() and bool('X-Testing' in request.headers and request.headers['X-Testing'])
 
 
-def is_try_it(req):
-    """ Many changes need to be released together, so they are done behind a flag - the TRY_IT env variable.
-    To facilitate testing, the flag can be overridden by adding a tryit=[0,1,false,true,yes,no] get param to
-    the request. In this way all users can switch between the 2 designs without the need to change an env variable.
-    Note that the get param override is added only because the specific features behind the flag are static and do not
-    require the flag to be passed down to further requests.
-    """
-    if req:
-        override = request.args.get("tryit", False)
-        if override:
-            return False if override in ['0', 'false', 'no'] else bool(override)
-    return os.getenv('TRY_IT', '0') == '1'
+def is_redesign_enabled():
+    return REDESIGN_2025_ENABLED_BY_DEFAULT or is_opted_in_to_preview()
+
+
+def is_opted_in_to_preview():
+    opt = request.args.get("tryit", False)
+    return False if opt in ['0', 'false', 'no'] else bool(opt)
+
+
+# Flip this to True when we are ready to release
+REDESIGN_2025_ENABLED_BY_DEFAULT = False
 
 
 def extract_bcrypt_rounds(hash):
--- a/static/js/app.ts
+++ b/static/js/app.ts
@@ -230,7 +230,8 @@ export function initializeCodePage(options: InitializeCodePageOptions) {
 
   const validAnchor = [...Object.keys(theAdventures), 'parsons', 'quiz'].includes(anchor) ? anchor : undefined;
   let tabs: any;
-  if (options.page == 'tryit') {
+  const isTryItPage = options.page == 'tryit';
+  if (isTryItPage) {
     tabs = new IndexTabs({
       // If we're opening an adventure from the beginning (either through a link to /hedy/adventures or through a saved program for an adventure), we click on the relevant tab.
       // We click on `level` to load a program associated with level, if any.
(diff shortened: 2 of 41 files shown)


rix0rrr:
> The class customization page does not have the 'mandatory developer mode' option when you add ?tryit=yes to the end of the url in your browser. Removing the suffix or changing it to ?tryit=no, should make the option appear.

Instead of a URL argument, can I suggest the following:

- Get the flag from the `session` instead
- If a user visits the `/tryit` page, set the value in `session` to True.
- Only if the flag is set, add a banner to the top of every page (add in `layout.html` I suppose) that says something like "since you have visited /tryit, you are now looking at a preview version of this page (click here to disable)", and make it a link that hits the server with HTMX to unset the variable from `session`, then afterwards reloads the page with some Hyperscript (we have some examples of that around the page).

utils.py
```diff
@@ -169,6 +169,20 @@ def is_testing_request(request):
     return not is_heroku() and bool('X-Testing' in request.headers and request.headers['X-Testing'])
 
 
+def is_try_it(req):
+    """ Many changes need to be released together, so they are done behind a flag - the TRY_IT env variable.
```
rix0rrr, line 173: Why the environment variable in addition to the query string argument?

boryanagoncharenko: We agree that big-bang releases are risky. Basically we will have 2 branches of the code base - the old and the new, and I am using an environment variable to denote which is the default branch. However, I would like to let the product owners (and developers) test each branch on every stage without having to flip the env variable. I think providing an easy way to switch will slightly reduce the risk of having one branch not working and I thought this is a way to achieve it. This is also why putting the value in the session is not exactly an option - you actually want to let users switch or follow simple instructions on how to switch. But maybe this is a long shot and will not reduce the risk of errors. Let me know if you prefer just to have it gone.

rix0rrr: When you say "branch", you don't mean "git branch" but more like "if branch", right?

Yeah, agree with being able to test both code paths side by side. I thought the query string/session var would be to let users explore the different page behaviors?

* Since our audience is non-technical, or they might look at the `/tryit` page on the Heroku-hosted versions, environment variables are probably not going to be used by them to switch. Instead, we should offer them some easily manipulable UI affordance to see the preview version of a page. 
  * You already have the query string, which is great, and if you think it's sufficient I'm happy to ship it.
  * I'm a little wary of relying on the query string alone. It might suffice, but for example without remembering to render it into all URLs everywhere it will not survive navigating to a different page, or HTMX requests. Whereas a "sticky" variable in the session will just automatically propagate everywhere, so there's less risk of it breaking in unexpected ways.
  * Of course, the problem of making it sticky means there also must be a user-visible affordance to unstick it, which would be the UI top bar that I mentioned. So yeah, that's additional work.
* We've established that the env var is not for users, to explore different versions of the same page, because users will use a UI element. If it's not for them, it's probably to "officially release" the new behavior in a couple of months, right? To make it the default?
  * If that's the case, we might as well just make it a global variable in code, that we flip to `True` come release time.


```py
def is_new_behavior():
  return NEW_HEDY_PAGE_DEFAULT or is_opted_in_to_preview()

def is_opted_in_to_preview():
  return session.get('tryit')

# Flip this to True when we are ready to release
NEW_HEDY_PAGE_DEFAULT = False
```

rix0rrr: I wouldn't call it `try_it` everywhere because when it becomes the default, does it make sense that `TRY_IT=1` means "we are now live on the new feature"? 😉

boryanagoncharenko: Oh, sorry, I forgot the word "branch" is overloaded! I meant if-branches, not git branches.

Your assumptions are exactly what I had in mind.

I am slightly more inclined to stay with the query parameter, even though I admit it is an awkward solution. The reason is that 1) it does not require extra effort, i.e. offering the user to "unstick" and 2) none of the features to be deleted require to propagate the flag value, so a simple query string will suffice. Of course, I might be wrong and then the moment we encounter a feature which needs to send the flag value to multiple pages/calls, we will be forced to use the session. However, if you prefer to do the session from now, I am totally fine with it.

I used an environment variable because it allows switching without code changes which in turn eliminates the chance for potential pipeline blockages and reduces the time to rollback in case of an emergency. But this is an implementation detail really, I can change it to a global variable if you prefer.

`tryit` is the name of the new hedy page, so I decided to use the same name for the flag. But you are right - it does not have be used everywhere. I'd be very grateful for suggestions :)

utils.py
```diff
@@ -169,6 +169,20 @@ def is_testing_request(request):
     return not is_heroku() and bool('X-Testing' in request.headers and request.headers['X-Testing'])
 
 
+def is_try_it(req):
```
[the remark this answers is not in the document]
boryanagoncharenko: thanks!

utils.py
rix0rrr: Something obvious, like this?

```py
REDESIGN_ENABLED_BY_DEFAULT = False
```

rix0rrr: >  I can change it to a global variable if you prefer.

I think I do. It's just more obvious in my mind.

PR #6092, opened 2025-01-07
💻 New public adventures page 

rix0rrr:
This PR changes the UI of the teachers "public adventures" page. It does away with the tabs and replaces it with a scrollable list on the left side. It's also faster and doesn't rely on all requests coming in to a single Python process (which will not be true on Heroku).

Behavioral changes w.r.t the previous version:

* A cloned public adventure starts out as private (used to be public). With the old behavior, after cloning, we would just have 2 public copies of the exact same adventure, which seems to not add any value. If a teacher changes it (and puts `print my boss is a poopy-head!` in the adventure), those changes are inadverently public  It seems to make more sense that a user deciding to republish and adventure should be a conscious decision.
* It's no longer possible to see adventures from all languages at the same time: selecting a language is required. I figure that real users are interested in a single language only, and if you are a Spanish speaker adventures in Chinese or Telugu don't help you much.

This is what it currently looks like:

image

**How to test**

Log in as a teacher, and search and use the interface. Use clone, use flag, use delete if you are `teacher1`.


rix0rrr:
I'm sure the Cypress tests will be broken because I changed a lot about the page. Haven't gotten to those yet.

static/js/public-adventures.ts
```diff
@@ -1,35 +1,12 @@
...
 document.addEventListener("updateTSCode", (e: any) => {
-    setTimeout(() => {
-        const js = e.detail;
-        updateURL();
-        initialize({
-            lang: js.lang, level: parseInt(js.level), keyword_language: js.lang,
-            javascriptPageOptions: js
-        });
-    }, 1000);
+    console.log('updateTSCode', e);
```
boryanagoncharenko, line 4: This probably was added for testing?

rix0rrr: Ah yes, thanks

boryanagoncharenko:
I am ready to ship it, but I think we might need to ask the POs to check if they are ok with the changes. @MarleenGilsing @AnneliesVlaar would you like to take a look? The UI has changed quite a bit. Also, cloned adventures are not automatically marked as public anymore.

boryanagoncharenko:
Btw when there are many adventures, I do not see a Next Page button in the bottom left corner (as on the screenshot). I just get a scrollbar:

https://github.com/user-attachments/assets/3741b4d2-ae55-4a33-8118-e93198d817f4

rix0rrr:
> Also, cloned adventures are not automatically marked as public anymore.

Yes, this was on purpose. I don't think that cloning a public adventure to be automatically public as well is expected behavior. 

In the usual case, we would just have 2 public copies of the exact same adventure, which seems to not add any value.

rix0rrr:
> Btw when there are many adventures, I do not see a Next Page button in the bottom left corner (as on the screenshot). I just get a scrollbar:

Do you have more than 20?

boryanagoncharenko:
> > Btw when there are many adventures, I do not see a Next Page button in the bottom left corner (as on the screenshot). I just get a scrollbar:
> 
> Do you have more than 20?

I have 10, so this is why there is no button. But I have to admit something is a bit off as user experience. The scrollbar appears if there are more than 5 adventures, but the button shows up if there are more than 20. Maybe we can increase the height of the adventures container so that it can display more than 5 adventures at a time?

Screenshot 2025-01-13 at 15 44 57

Actually, I am sure the POs would have a preference.

rix0rrr:
> Maybe we can increase the height of the adventures container so that it can display more than 5 adventures at a time?

I had that at first. I think loading 5 adventures at a time is annoying for pagination, but having the list stretch the entire height makes it way too long.

A simple compromise is probably to put the NEXT PAGE button *inside* the scrollable container, and making it very obvious that there is scrolling to do?

rix0rrr:
I did the things I said, hope this is better

boryanagoncharenko (approved):
This is a strictly better version of the public adventures page both from a technical and UI perspective. I am approving this before the POs have reviewed this functionally because the PR has already caused merge conflicts. The feedback of the PO review will be addressed in a separate issue.

AnneliesVlaar:
I like the new design! It is more clear and I love the little icon in the list that says “This adventure has a solution example” <3

I do agree with Boryana that the scrolling through the list of adventures feels a bit weird. I do think the button within the scrolling container makes it better, although the button doesn't work for me, when I click on "next page" It shows "There are on public adventures matching these criteria. Try searching for something else."

Maybe I don't understand the problem/solution of the scrolling container and how many items are loaded, but is it possible to change the "next page" button into a "show more adventures" button and when clicking that button show for example 10 adventures more (so 30 in total)?

rix0rrr:
> I do agree with Boryana that the scrolling through the list of adventures feels a bit weird. I do think the button within the scrolling container makes it better, although the button doesn't work for me, when I click on "next page" It shows "There are on public adventures matching these criteria. Try searching for something else."

That's odd. Can you share your search criteria? And does it fail immediately after the first page? 

> Maybe I don't understand the problem/solution of the scrolling container and how many items are loaded, but is it possible to change the "next page" button into a "show more adventures" button and when clicking that button show for example 10 adventures more (so 30 in total)?

Aha I see. You'd prefer an infinite-scroll type interface, instead of pagination?

AnneliesVlaar:
> > I do agree with Boryana that the scrolling through the list of adventures feels a bit weird. I do think the button within the scrolling container makes it better, although the button doesn't work for me, when I click on "next page" It shows "There are on public adventures matching these criteria. Try searching for something else."
> 
> That's odd. Can you share your search criteria? And does it fail immediately after the first page?
> 

I don't have specific search criteria, 'English', 'all levels', 'all tags' and indeed it fails immediately I have never seen a second page.

> > Maybe I don't understand the problem/solution of the scrolling container and how many items are loaded, but is it possible to change the "next page" button into a "show more adventures" button and when clicking that button show for example 10 adventures more (so 30 in total)?
> 
> Aha I see. You'd prefer an infinite-scroll type interface, instead of pagination? Yeah I can do that.

To infinity and beyond!

PR #6097, opened 2025-01-10
📜 add answers for adventures

Felienne:
This PR adds two fields to the yamls so we can keep track of correct answers and give teachers hits and tips. We will have to think about how to also shows these to teachers but that is of later concern


boryanagoncharenko:
> You do not yet have to do anything with this info, but a question to you: will this mess up Weblate? :)

Weblate should be OK with this change. I do not foresee any problems.

MarleenGilsing:
Great! This looks nice and simple to add. 

Would it be easiest if @LucasPeters010 works with pycharm and github desktop as well? He doesn't have this set up yet.

LucasPeters010 (commented):
Answer for adventure 'story' level 1 till 8

PR #6122, opened 2025-01-22
🧹 Remove warnings during tests

rix0rrr:
We were persistently printing 2 deprecation warnings during unit tests. These warning originated from old versions of the Lark parser and the Pydantic library, that were using deprecated Python APIs.

Upgrade both libraries to newer versions; the warnings are now gone.

Because we cache Lark parsers on disk, we need to make sure to mix the Lark version into the cache key so that we don't load old parsers into a new version of the library.

We can't seem to go higher than `1.1.9`, because when we do we run into the following parsing problem:

```
E           exceptions.InvalidArgumentTypeException: Invalid Argument Type {'command': 'forward', 'invalid_type': Markup('text'), 'allowed_types': 'a number or input from `ask`', 'invalid_argument': '100', 'line_number': 1}

We detected that `forward` doesn't work with `100` because it is text. Can you try changing `100` to a number or input from `ask`? at line [1]
```

**How to test**

If all tests pass, this is good to go.


boryanagoncharenko (approved):

rix0rrr:
@mergifyio requeue

PR #6124, opened 2025-01-22
💻 Replace public adventures pagination with infinite scroll

rix0rrr:
The public adventures page had a pagination mechanism, with "next page"/"prev page" buttons.

A more modern idiom these days is to use "infinite scroll": once the user scrolls to the bottom of the container, we fetch the next bit of results, and so on.

**How to test**

Go to the public adventures page, scroll to the bottom of the result list, try to catch a glimpse of the spinner before the next set of results loads.

image


AnneliesVlaar (approved):
Looks good!

PR #6126, opened 2025-01-23
🪲 Set highlighter language and level 

### LINKED ISSUE #6112: 🪲 code highlighting in slides does not work properly for translations on Alpha
**Describe the bug**

I am testing my translations in slides for possible errors in code fragments and saw this issue. I have also tested this on a few other languages.

Highlighting for English are fine. Translations uses plain grey for everything until level 4. 
From then on, variables, quoted string, numbers are colored but mostly in different colors from English version. see screenshots below.
Some command keywords even change color depending where they are used, but they are mostly grey.

All slides

screenshot from level 1 slides
Image

screenshot from level 13 slides
Image

Windows 10, MS Edge latest

boryanagoncharenko:
PR #6085 makes uses of `data-level` and `data-kwlang` but these are never set. As a result, the language of the highlighter always defaults to English regardless of the current language. This PR makes some minimal changes to address the issue.

Fixes #6112

**How to test**
As an anonymous user go to /hedy and change the language to Turkish (or any other language). In the editor paste `yazdır test` and check that the first word is highlighted as a command. Do the same check for the /tryit page and a logged in user.


rix0rrr (approved):

static/js/cm-editor.ts
```diff
@@ -210,8 +210,8 @@ export class HedyCodeMirrorEditor implements HedyEditor {
             state: state
         });
 
-        const levelStr = $(element).closest('[data-level]').attr('data-level');
-        const lang = $(element).closest('[data-kwlang]').attr('data-kwlang') ?? 'en';
+        const levelStr = $('[data-level]').attr('data-level');
```
rix0rrr, line 213: The old code tried to find the closest enclosing element that has a `data-level` attribute.

The new code finds any `data-level` attribute anywhere on the page, whether it has a relation to the editor or not.

Why did the old code not work?

templates/hedy-page/editor-and-output.html
```diff
@@ -15,7 +15,8 @@
       <!-- The actual editor -->
       <div id="editor" data-cy="editor" style="background: #272822; font-size:0.95em; color: white; direction: {{ g.dir }};" lang="en" blurred='false'
         {% if editor_readonly %}data-readonly="true"{% endif %}
-        class="w-full flex-1 text-lg rounded min-h-0 overflow-hidden fixed-editor-height"></div>
+        class="w-full flex-1 text-lg rounded min-h-0 overflow-hidden fixed-editor-height"
+           data-kwlang="{{ current_language().lang }}" data-level="{{ level }}"></div>
```
rix0rrr, line 19: An enclosing scope should be good enough, and I think it should be on there already, no?

templates/incl/editor-and-output.html
```diff
@@ -25,7 +25,8 @@
         <!-- The actual editor -->
         <div id="editor" style="background: #272822; font-size:0.95em; color: white; direction: {{ g.dir }};" lang="en" blurred='false'
           {% if editor_readonly %}data-readonly="true"{% endif %}
-          class="w-full flex-1 text-lg rounded min-h-0 overflow-hidden fixed-editor-height"></div>
+          class="w-full flex-1 text-lg rounded min-h-0 overflow-hidden fixed-editor-height"
+             data-kwlang="{{ current_language().lang }}" data-level="{{ level }}"></div>
```
rix0rrr, line 29: Same here, if we can put this attribute on a `<div>` that contains all editors, it should work already?

rix0rrr (changes requested):

static/js/cm-editor.ts
boryanagoncharenko: There is no data-level or data-kwlang on any element. This is why it did not work. I added the attributes to the editor but then ofc `closest` does not work. We could either put the attributes on an enclosing element indeed which is a bit weird or keep them in the editor and fetch them directly from the editor `$('#editor').attr('data-*')`. What was your intention?

rix0rrr: Putting the level and kwlang directly on the editor works for the main editor because we control the HTML for it, but doesn't work very well for example for all the syntax highlighted preview editors in running text, because those are `<pre>` tags that are generated from Markdown. So I wanted a place to put it where it's still in context, but not the editor itself, so it become "some enclosing `<div>`".

I'd like to keep the logic consistent between the read/write editor and the preview-only editors.

I for sure though `x.closest()` would return `x` if it matched the selector, but perhaps it doesn't. The document seems to bear that out though, I recall looking this up:

image

https://api.jquery.com/closest/

rix0rrr (approved):

rix0rrr:
I ended up putting them in `level-page.html`, which contains a `<div>` that is a shared common ancestor of the editor and the adventure tabs with preview text. It seems to work.

I did change the source back to `g.keyword_lang` though.

> If the user is logged in and their profile language is set to Turkish but their keyword language is set to English, should we highlight Turkish keywords if they are used? I think we should, so the language there should not be the g.keyword_lang, it should be the session language.

Well... if not that, then what does `keyword_lang` even mean?

It may be that there is no distinction between keyword language and interface language; but if that's the case, I think I'd rather do that by setting `g.keyword_lang = g.lang` somewhere in the Python, rather than having a mix of language interpretations in the HTML templates.

rix0rrr:
Current status:
image

boryanagoncharenko:
I am afraid using the `level-page.html` does not cover all scenarios. The following screenshots are done with Teacher1 when the language and the keyword language are both set to Bulgarian:

Viewing a public program:
Screenshot 2025-01-27 at 11 00 48

Previewing a custom adventure:
Screenshot 2025-01-27 at 11 01 13
 
I think the common denominator is the `layout.html` which is too wide of a context.

rix0rrr:
Do we have a list of places where common UI elements are used? I feel like this "not catching of all cases" comes up all the time... ☹️

boryanagoncharenko:
> Do we have a list of places where common UI elements are used? I feel like this "not catching of all cases" comes up all the time... ☹️

I know :(

I traced that the editor-and-output.html is used in the following places:
- the code-page which uses `layout.html > level-page.html`. Btw the code-page can be tested via `/hedy`, `/tryit`, `/slides/{level_id}`, `/hour-of-code`. The rest of the usages will be removed soon.
- the customize-adventure page which uses `layout.html > auth.html`
- the view-program page which uses directly `layout.html`
- the public-adventures page (specifically the htmx-preview-adventure.html) which uses `layout.html > auth.html`

But I noticed we also have an embedded-editor.html which is returned by `embedded/{level_id}`, which I forgot to check at all.

Btw I think the keyword language is meant to define in which language to display the keywords in the program at hand. The highlighting, however, should work for the current language and English. I base this assumption on the way the grammars of the language (the lark files, not the highlighting grammars) are created.

rix0rrr:
Thanks, that list is extremely helpful!

> Btw I think the keyword language is meant to define in which language to display the keywords in the program at hand. The highlighting, however, should work for the current language and English.

I think highlighting will automatically work for `<language> + English`.

But that brings up a good point: let's say keyword language is set to German and display language is set to Dutch (for whatever reason). Shouldn't we then highlight "German + English" (i.e., the keyword language) rather than "Dutch + English" (the UI language) ?

boryanagoncharenko:
> But that brings up a good point: let's say keyword language is set to German and display language is set to Dutch (for whatever reason). Keywords should be rendered in German if we do that, right? Shouldn't we then highlight "German + English" (i.e., the keyword language) rather than "Dutch + English" (the UI language) ?

In your profile, if you set a language other than English, you can only choose your keyword language to be either that same language or English. So, if my language is Turkish, I can pick to have as a keyword language either English or Turkish. I think this was done to simplify the problem of grammar generation.

boryanagoncharenko:
However, I have to point out that we did explore the idea of having multiple languages at the same time, e.g. Dutch, Frisian and English, or Dutch, French and English. This would work for countries with multiple official languages. It sounds like a very nice feature but it got de-prioritized.

rix0rrr:
I added the attributes in other missing places. It's not elegant, but it'll do... ? We can start paying more attention in the future.

PR #6134, opened 2025-01-25
💻 Show the reason why you can see adventures

### LINKED ISSUE #6133: 🪲 Annelies can see Felienne's public adventures
**Describe the bug**

Image

rix0rrr:
You can see adventures of teachers that you share classes with: both  if you are the second teacher, and if they are the second teacher.

This is surprising at first, so this adds an icon that indicates why you can see a given adventure.

image

**How to test**

Log in as a user that shares a class with another teacher. Observe that you can see their adventures.

Closes #6133.


AnneliesVlaar:
The icon really helps! 

I would expect to only see the custom adventures that are used in the shared class, would that be possible?

When clicking on a adventure from another teacher, you do go to the customization page, but it's not possible to edit the adventure (which is the expected behavior I think). 

Can we remove the link behind the name of the adventure that goes to the customization page, and disable the remove button. Then the only thing one can do is preview the adventure?

rix0rrr:
> I would expect to only see the custom adventures that are used in the shared class, would that be possible?

Everything is possible 😆, it's just a matter of time. On first glance it *seems* like it shouldn't take too long, but I don't have the data model super clear in my head so it might be that there are dragons.

> Can we remove the link behind the name of the adventure that goes to the customization page, and disable the remove button. Then the only thing one can do is preview the adventure?

I haven't checked, but I think the current explicitly implemented behavior is that if you share a class with someone, you can do everything to their adventures, including editing and deleting. (The reason I'm saying this is because I've seen calls to `get_second_teacher_adventures()` in many different places in the `for_teachers` file, including in the "deleting" code path).

While I agree with you that it seems unexpected, it seems to have been "expected behavior" by someone. It might make sense as well: if you share a class with someone, you trust them to teach with you, and they might want to make a small improvement to some adventure. It would be annoying if they then had to wait for you to make the change.

These are good questions, that we might need to ponder a bit. Also, I don't think the current PR makes anything worse, you would just like to see *more* improvements. I propose we do the following:

* If you think this PR makes the situation better, merge it as-is
* Create 2 new issues for the 2 changes you would like to see (showing only adventures that are used in a class, second teachers only allowed to preview an adventure and nothing else)
* Have a discussion on the additional issues and see if this is what we really want, and what the reason is for the current behavior being what it is
  * This could be a Discord discussion, and it could be informed by going back into the pull request history of the project, finding the PR that in introduced the current behavior, and seeing who made it (ask them) and seeing if there's any discussion on it

rix0rrr:
I'll teach you a trick as well!

If you know what source file to look at (I happen to know that most of the teacher code that deals with adventures is in `for_teachers.py`), you can go to that file in GitHub:

2025-01-27 at 10 23

Then click "Blame" at the top:

2025-01-27 at 10 23

And then you can see for every line of code, who last changed it and in what PR.

I will then just Ctrl-F search in the file to look for things to do with "second adventures" (Ctrl-F for "second" or something to that effect), and then I come across this:

2025-01-27 at 10 24

And I can click through to the PR that introduced this code:

https://github.com/hedyorg/hedy/pull/4644

Written by Hasan. He's unfortunately not around anymore to ask, but he left some breadcrumbs:

2025-01-27 at 10 28

So I was wrong:

* It's not possible to edit/delete an adventure you don't own.
* But I *think* all adventures are visible so that a second teacher has the ability to add new adventures to the class that haven't been added to the class yet.

website/for_teachers.py
```diff
@@ -1874,3 +1878,10 @@ def _create_customizations(db, class_id):
     }
     db.update_class_customizations(customizations)
     return customizations
+
+
+def render_why_class(adventure):
+    return safe_format(
+        gettext('see_adventure_shared_class'),
+        class_name=adventure.get("why_class"),
+        creator=adventure.get('creator')),
```
boryanagoncharenko, line 1887: Currently, this produces the following message:
Screenshot 2025-01-27 at 11 43 45

I suspect the comma is not intentional and the message should appear without the parenthesis?

boryanagoncharenko:
This PR is ready to go. @AnneliesVlaar if you agree with it, please feel free to approve it yourself.

AnneliesVlaar (approved):

PR #6147, opened 2025-01-30
💻 Disable the teachers' survey 

### LINKED ISSUE #6143: 💻 Disable the questionnaire
Disable the questionnaire referred in issue #6051. This issue is created because disabling the questionnaire has higher priority than redesigning it.

boryanagoncharenko:
This PR temporarily disables the survey presented to teachers on the class overview page. This is the first step from the survey redesign process. This is why most of the code is not deleted but commented out. Ideally, it will be used again soon, when the redesign is done.

Fixes #6143

**How to test**
- Run Hedy locally and log in as a teacher.
- Go to For-Teachers and click on an existing class. When the class overview page opens, there should be no Survey link in the upper right corner (on the left of the 'Customize class' green button)
- Create a new class and check that the Survey link is not displayed.


jpelay (approved):
Great! Thanks for working on this one :)

PR #6151, opened 2025-01-31
💻 Delete tutorials

### LINKED ISSUE #5991: 💻 Remove tutorial
I suspect this is not used a lot, and it would probably need to be redesigned for the new UI? It is also buggy.

boryanagoncharenko:
This PR removes tutorials including the content and the reference from the teacher's manual.

Fixes #5991

**How to test**
- All pages should load without an error especially the /hedy and /tryit pages
- The teacher's manual should not have a reference to the tutorial
- The localhost:8080/tutorial#default should not trigger the tutorial but return a 404


static/vendor/reveal/js/config.js
```diff
@@ -18,10 +18,6 @@ export default {
 	// Display presentation control arrows
 	controls: true,
 
-	// Help the user learn the controls by providing hints, for example by
```
jpelay, line 21: I don't think this was an intended change? This code belongs to the slides library

static/vendor/reveal/js/controllers/controls.js
```diff
@@ -154,40 +154,6 @@ export default class Controls {
 
 		}
 
-		if( this.Reveal.getConfig().controlsTutorial ) {
```
jpelay, line 157: This one as well

jpelay (changes requested):
Everything seems cool, but I don't think the changes to the slides library were intentional

static/vendor/reveal/js/config.js
boryanagoncharenko: You are right, thanks for pointing it out!

jpelay (approved):

PR #6159, opened 2025-02-03
💻 Allow existing users to subscribe to the newsletter

### LINKED ISSUE #6103: 💻 Allow existing users to subscribe to the newsletter
**Context**
Currently, a user can subscribe to our newsletter only when they create their account. This has 2 consequences:
- the code that subscribes the users assumes they are brand new accounts with no classes, customizations or history and this includes the changes in PR #6080, so the code does not check if it should add any relevant tags.
- the existing audience in Mailchimp does not have tags that indicate if the user has created classes, customizations etc.

**Scope**
Create a subscribe endpoint that users can use after they have created their account. This means that this endpoints should check whether the user has done all actions of interest and should update their Mailchimp tags accordingly. Also, this means we might need to revise the triggers of the journeys.

Add an elegant (or at least not annoying) way to offer the users to subscribe. Probably we want to offer this only to logged in teachers and in a non-intrusive way. We could offer this option on the Hedy landing page and in a separate section in the user profile page.

AnneliesVlaar:
This PR adds "Newsletter" and "Danger zone" sections to the the profile page.

Before approving:
- [ ] Revise the client-facing messages
- [ ] Translate the client-facing messages to Dutch

Fixes #6103

**How to test**
- Log in as a student and go to the profile page. Note that there is no Newsletter section. The Delete profile button is now under its own Danger zone section.
- Create a new teacher account without subscribing to the newsletter at signup.
     - Go to the profile page and check that the Newsletter section is expanded by default. It should offer you to subscribe. Do not subscribe yet.
     - Log in as another teacher and invite the new teacher account you created as a student to a class (via username). Go back to the profile of the new teacher and check that the Invitations section of the profile is expanded by default but not the Newsletter. Once you accept/reject the invitation, the Newsletter section should get expanded by default.
     - Change line 69 in file newsletter.py to `return True`, so that you can simulate a successful subscription and restart the application. Click subscribe to the button, the UI should update without reloading the page. 
     - Change line 2275 in the app.py file to `subscription_status="subscribed"` to simulate that there is a successful subscription. If you reload the profile page now, the Newsletter section should be disabled by default. When you open it, you should see a message confirming that you are subscribed.
     - Change the 2275 in the app.py file to `subscription_status="unsubscribed"`. If you reload the profile page now, the Newsletter section should be expanded by default. It should ask you for feedback and let you resubscribe via the Mailchimp form.


AnneliesVlaar:
Good idea to add a delete profile section

boryanagoncharenko:
> If a user subscribed to the newsletter and then via the link in the newsletter unsubscribed, do we know who is subscribed and unsubscribed?

Yes, we do:
- if the user has never subscribed, they are not in the Mailchimp audience
- if the user has subscribed, we see that their status is "subscribed"
- if the user has unsubscribed, their status is "unsubscribed"

Btw I did not cover how this section will look for the people who are already subscribed. I guess the section should not be there at all?

AnneliesVlaar:
Ah that's great. For people who already subscribed I would suggest to have something similar to the unsubscribed:

**You're subscribed to our newsletter**
_Run to your mailbox to see if the new issue already has been delivered!_

(Or something like that)

boryanagoncharenko:
@AnneliesVlaar would you be able to:
- [ ] check this PR for usability
- [ ] see whether you agree with the written text, maybe some of the messages are too wordy?
- [ ] translate the new messages to Dutch

Thanks!

build-tools/heroku/tailwind/styles.css
```diff
@@ -791,6 +791,10 @@ p.close-dialog {
   @apply bg-blue-200 px-8 py-2 rounded-lg text-3xl cursor-pointer;
 }
 
+.section-header-important {
```
jpelay, line 794: Maybe `.section-header-danger` is a better name?

website/newsletter.py
```diff
@@ -48,14 +48,31 @@ def run():
...
+    """ Fetches the subscription status of the user. Values can be 'subscribed',
+    'unsubscribed' or '' if the user has never subscribed """
+    if email and MAILCHIMP_API_URL:
+        response = get_mailchimp_subscriber(email)
+        if response and response.status_code == 200:
+            return response.json().get('status')
+    return ''
+
+
+def create_subscription(email, country, teacher=True, created_class=False, customized_class=False):
```
jpelay, line 61: Do you think in the future we might add more tags to this functions or these two will be enough? I mention it, because this function might quickly have tons of parameters, it has happened to us in the past.

jpelay (commented):
I think this is fine!! I tested it using the steps above and it worked :) I left a couple of comments, but they are nitpicks really.

I won't approve because I think this still misses the dutch translations? So I'll let Annelise handle that :)

jpelay:
@AnneliesVlaar all are translations done? Because if so we can merge this :)

AnneliesVlaar:
> @AnneliesVlaar all are translations done? Because if so we can merge this :)

Yes done!

jpelay (approved):

PR #6161, opened 2025-02-05
🪲 Always return empty JSON object on successful response, fixes #6160

### LINKED ISSUE #6160: 🪲 Students joining a class via link succeeds with an error message
The following bug is present on production and `main`.

When a student uses a teacher link to join a class, and clicks the button to join the class, the request to do so succeeds, but the student gets an error message about a network error. Subsequent presses on the button result in an error message explaining that the class has already been joined. Refreshing the page reveals that the student has successfully joined the class.

The error occurs because the response of [`join_class` in `classes.py`](https://github.com/hedyorg/hedy/blob/a3683e7674c6617cacd9026595d03728f3fbe3df/website/classes.py#L179) return an empty string `text/html` response, while [`join_class` in `teachers.ts`](https://github.com/hedyorg/hedy/blob/a3683e7674c6617cacd9026595d03728f3fbe3df/static/js/teachers.ts#L126) expects a JSON response. Because the request actually returns a `200`, it shows a general network error, and the end-to-end tests only verify the `200`.

This bug seems to have been introduced in https://github.com/hedyorg/hedy/pull/5558 in May 2024, which switched from `jsonify` to `make_response`. Or perhaps it was fixed shortly after that and recently reintroduced, because that's almost a year ago and this bug is so visible I'd think it would've been noticed sooner.

It's likely this bug affects other actions in a similar way.

Image

confiks:
Fixes #6160, and potentially similar bugs as well. 

⚠️ I changes all occurrences of `'', 200` in `make_response`, but I only tested that `join_class` works as intended and #6160 is fixed. I don't know enough about the application to test the other occurrences effectively. This should be tested for the other functionality as well by a maintainer.


boryanagoncharenko (approved):
Thank you so much for your contribution. When we adopted make_response, we introduced a bunch of bugs caused by the string response. We addressed most of them but, unfortunately, some slipped. I tested all scenarios you changed and they all work great.

PR #6163, opened 2025-02-05
🚚 Update weblate merge-resovling script

boryanagoncharenko:
Currently, the merge-weblate-resolving-conflicts script executes a `doit run _autopr` which generates code and then immediately does a `git commit`. If a new language is added though, some of the newly generated files remain untracked and are not included in the commit. This causes the following merge command to fail:

```
+ git merge -s recursive -X ours origin/main --no-edit
error: The following untracked working tree files would be overwritten by merge:
	grammars/keywords-jbo.lark
Please move or remove them before you merge.
```

Evaluate whether adding a `git add .` before the commit would have unwanted side effects.


rix0rrr:
How are we going to test this? Did you run it locally? Can we prevent automatic merge on the next “real” run so that we can confirm it doesn’t add unwanted files?

rix0rrr (approved):
I’m okay to give it a go

rix0rrr (approved):

PR #6184, opened 2025-02-21
📜 Fix typos and command highlighting in english adventure

bobidle:
I believe i found some more typos and tried to fix them. Also added some missing command highlighting (like quotes and brackets) for easier translations. Let me know if something should be differently or better fixed.


AnneliesVlaar (approved):
Thanks for these fixes!

PR #6210, opened 2025-03-06
💻 Add a switch to remove Parson and Quiz adventures

### LINKED ISSUE #5924: 💻 Remove Parsons

boryanagoncharenko:
This PR introduces the following changes:
- When the redesign flag is disabled, the Quiz and Parson adventures and the class customization options related to these adventures work exactly as before.
- When the redesign flag is enabled, the Quiz and Parson adventures no longer appear on the /hedy page. Additionally, on the class customization page, the adventures are removed from the sorted adventures and the options to hide parsons and quizzes are hidden.
- The Quiz and Parson adventures are removed from the /tryit page regardless of the redesign flag value.

Fixes #5924

**How to test**
Quiz and Parson adventures appear in many places so it is hard to provide a concrete list of steps for testing. To enable the redesign flag navigate to the `hedy/utils.py` file and change the return value of the `is_redesign_enabled()` method to `True`. If you would like to disable the redesign mode, change the value back to `False`. The testing aspects are:
- The pages: the adventures appear at least on the /hedy page, the /tryit page and the /customize-class page
- The type of user: please test with an anonymous user, a logged in student who is not enrolled in a class, a student who is enrolled in a class with and without customizations, and a teacher
- It is important to test enabling the redesign flag, i.e. if you create a class and customize it so that the Quiz adventure is second in order and then you enable the redesign, the class should be able to load without issues.


AnneliesVlaar:
When redesign_enabled is set to True parsons and puzzels do not appear in any of the cases, not on the Hedy page, not on tryit and not in customize class. That's great! 

I'm not sure if it is relevant, but if I have a class with puzzles and quizes (redesign_enabled -> FALSE), I set the flag to TRUE and then back to FALSE, the checkboxes for disable quizes/puzzles come back but the quiz/puzzle adventures don't. If I add these adventures manually to the class these adventures work again like expected. But when I push the button 'Remove customization' the quizes and parsons adventures are removed. Again I don't think it is important, but if we merge this we get different behavior for the 'Remove customization' button.

hedy_content.py
```diff
@@ -348,6 +357,9 @@
     ]
 }
 
+ADVENTURE_ORDER_PER_LEVEL = {level: [a for a in adventures if a not in ['parsons', 'quiz']]
```
jpelay, line 360: This variable is used a lot across app.py and for_teachers.py, in this last file its often used as `hedy_content.ADVENTURE_ORDER_PER_LEVEL`. You also need to replace it there, becase at this moment is like the parsons and quizes were removed, for example, in the restore customizations, like Annelies was suggesting.

jpelay:
Apart from that comment, things seem to be working fine :)

hedy_content.py
boryanagoncharenko: Thank you very much for pointing this out. I have no clue how I managed to not replace it everywhere.

AnneliesVlaar (commented):
The switch from FALSE to TRUE still works as intended and when in FALSE modus the quizes and puzzles do come back when using 'remove customization'

jpelay:
I think everything is working, it's just important to note that if I set the redesign flag to True, quizes and parsons dissapear from the class customizations on DB, needing to be added manually again or use the restore customizations button.

jpelay (commented):
I think everything is working, it's just important to note that if I set the redesign flag to True, quizes and parsons dissapear from the class customizations on DB, needing to be added manually again or use the restore customizations button. If you think that's not that big of a deal, then I'll approve.

jpelay (approved):
Then all good to go!

PR #6217, opened 2025-03-13
🪲 Fix race condition on customization's update

### LINKED ISSUE #6176: 🪲 Second teacher has issues opening levels
My co-teacher in class (@averburglmc) (https://hedy.org/for-teachers/class/2003976384a84343b856976cddc011a5) reports that every time he opens a level, he gets the popup that personalization has succeeded but it has not. He can only solve this by going back and forth a few times and retrying, which is annoying. He says it happens like this every week but for me as first teacher it seems not to happen?

Would be nice if this is picked up!

jpelay:
The problem in the issue, as Boryana pointed out, was a race condition caused by making two calls at roughly the same time:  to /update-customizations, and to /get-customization-level. This last call also performs some updates in the customization object, so if this one is done first, the customization object only gets the old update.

To address this I first tried just not triggering that second update, but then a few tests failed. So my new approach is the following:

1. I deleted the parameter from the autosave that triggers that second call. It doesn't really make sense to make that call if we are not sure that the first update was completed, so I moved it to the `save_customizations` function.
2. I modified the `get_class_info` to only update the customizations in case of a first load. If we are accessing this function through HTML calls, it should not be doing updates.

Fixes #6176 

**How to test**

* Updating the other options in customizations should work.


static/js/autosave.ts
```diff
@@ -34,18 +28,6 @@ export function autoSave(formId: string,
         }
           // Now we can simply trigger the submit event on the parent form element.
           formElement.requestSubmit()
-
-          // Sometimes different components depend on each other, that is component A needs to be updated if component B
```
boryanagoncharenko, line 38: This is probably me not understanding how this works. But why was this code added, i.e. why did we have to update partial-sortable-adventures when customize-class was updated? And why are we removing it?

boryanagoncharenko (commented):
There is something that I noticed going wrong with the sorting of levels. I disabled level 1 and ordered the adventures like this: Intro, RockPaperScissors, Parrot, Haunted House etc:
Screenshot 2025-03-14 at 14 32 33

Then reloaded the page, which for some reason changed the selected level in the dropdown to 2. When I navigated back to level 1, I get a level which is enabled and the order of adventures is incorrect:

Screenshot 2025-03-14 at 14 32 53

I think this is a regression with the current changes because I am not able to replicate this on main.

boryanagoncharenko:
> There is something that I noticed going wrong with the sorting of levels. I disabled level 1 and ordered the adventures like this: Intro, RockPaperScissors, Parrot, Haunted House etc: Screenshot 2025-03-14 at 14 32 33
> 
> Then reloaded the page, which for some reason changed the selected level in the dropdown to 2. When I navigated back to level 1, I get a level which is enabled and the order of adventures is incorrect:
> Screenshot 2025-03-14 at 14 32 53
> 
> I think this is a regression with the current changes because I am not able to replicate this on main.

Actually, I take it back - I do not know whether this is a regression. This functionality seems so weird, I cannot tell whether it is a regression or not.

boryanagoncharenko (approved):

PR #6220, opened 2025-03-13
📜 Fix links to news entries

bobidle:
Align the date formatting of the news entries and link to an external page as the internal download is unavailable.

Is the local download differently fixable?


jpelay (approved):
Thanks a lot for this fix!!

PR #6223, opened 2025-03-14
💻 Fix variable list not showing after level 6

### LINKED ISSUE #6026: 🪲 The variable list does not appear in the output after level 5
**Describe the bug**
After the execution of a program we display a list will all variables used in the program in the output pane. This is how it looks:
Screenshot 2024-12-06 at 15 26 30

However, this variable list is not displayed after level 5. Probably the reason is that in level 6 we start wrapping all variables in `Value()`. 

**Steps to reproduce**
Open level 6 and execute the following program. Notice that after it is executed, there is no variable button or list displayed:
```
a is 15
print a
```

boryanagoncharenko:
The variable list displayed after a program execution stopped showing up in levels 6 and up when we introduced localization. The root cause is that the localization mechanism required creating more complex Skulpt objects which contain localization information.

Fixes #6026

**How to test**
Go to multiple levels from 6 and up and check that the variable list shows up populated correctly after executing a program. Make sure that you test variable assignments of integers, strings, lists, floats and input. Also, test composite variable assignments, e.g. expressions. For example, in level 6:
```
a is 'str'
b is 1
c is 2
d is b + c
e is 1, 2, 3
f is ask 'test'
```


jpelay (approved):
Awesome it works

PR #6232, opened 2025-03-17
🪲 fixes 6215 - introduces a new password generator that doesn't use characters that are easy to misread

henkm:
We use a new random_password_generator function that uses a subset (base58) of the available characters to generate passwords. The reason for this change is that some characters are easy to misread, and this can lead to confusion.

**How to test**

* Invite new students to a group
* Print the generated passwords
* Confirm that none of the passwords contain the characters 0, O, l, and I

**Checklist**

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -312,6 +312,16 @@ def random_id_generator(
     return ''.join(random.choice(chars) for _ in range(size))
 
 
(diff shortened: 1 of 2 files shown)


boryanagoncharenko (approved):
Thank you for your contribution

PR #6264, opened 2025-04-04
💻 Add adventure solutions

boryanagoncharenko:
With this PR teachers can view the solutions of adventures if such solutions exist.

Fixes # 6261

**How to test**
Currently, no content exists for solutions, so to test this, you will need to add solutions. For example, add the following snippets to the content/adventures/en.yaml file:
```
  answer_code: |
      ```
      print this is the solution code 
      ```
  answer_note: |
      this is an explanation of the solution
  answer_code_2: |
      ```
      print this is the solution code 2
      ```
  answer_note_2: |
      this is an explanation of the solution 2
  answer_code_3: |
      ```
      print this is the solution code 3
      ```
```
Note that `answer_code*` is a required parameter, but `answer_note*` is not. Once the data is there, go to the /tryit page and test at least the following scenarios:
- [ ] teachers and second teachers should see the solution tabs. Students should not see the tabs.
- [ ] switching between the Adventure and Solution tabs of an adventure
- [ ] switching between adventures that have solutions and such that do not have solutions


jpelay (changes requested):
It works well when adding the content to the yaml, and also the changing between the content and the solution works. However, we need to also include the solutions from the teacher adventures themselves. This is a field that's stored in them and we also need to be able to show it.

boryanagoncharenko:
> It works well when adding the content to the yaml, and also the changing between the content and the solution works. However, we need to also include the solutions from the teacher adventures themselves. This is a field that's stored in them and we also need to be able to show it.

Oh no, I am missing something big here! Please let me know where I am going wrong.

Solutions are only meant for teachers. And custom adventures (displayed in the /hedy or /tryit page) are only meant for students who are in a specific class of the teacher that has customizations to include a teacher's adventure. However, under what circumstances does a teacher see these custom adventures on the /hedy or /tryit pages? Right now, if I log in as a teacher, I do not see the custom adventures or any class customizations because I am not part of a class - I am the teacher of that class. And I cannot be enrolled in my own class because I could have more than one classes (and I cannot be enrolled in all of them).

The only functionality that could potentially show the solutions for custom adventures is when a teacher previews their own class. However, I'd argue that solutions should not be displayed because if solutions appear in the preview, we will give the false impression that students would see the solutions tabs.

I easily adjusted the backend to include the solutions code for the custom adventures but what are the steps to display them on the /hedy or /tryit pages?

jpelay:
> > It works well when adding the content to the yaml, and also the changing between the content and the solution works. However, we need to also include the solutions from the teacher adventures themselves. This is a field that's stored in them and we also need to be able to show it.
> 
> Oh no, I am missing something big here! Please let me know where I am going wrong.
> 
> Solutions are only meant for teachers. And custom adventures (displayed in the /hedy or /tryit page) are only meant for students who are in a specific class of the teacher that has customizations to include a teacher's adventure. However, under what circumstances does a teacher see these custom adventures on the /hedy or /tryit pages? Right now, if I log in as a teacher, I do not see the custom adventures or any class customizations because I am not part of a class - I am the teacher of that class. And I cannot be enrolled in my own class because I could have more than one classes (and I cannot be enrolled in all of them).
> 
> The only functionality that could potentially show the solutions for custom adventures is when a teacher previews their own class. However, I'd argue that solutions should not be displayed because if solutions appear in the preview, we will give the false impression that students would see the solutions tabs.
> 
> I easily adjusted the backend to include the solutions code for the custom adventures but what are the steps to display them on the /hedy or /tryit pages?

Hmmmm, yes you're correct I didn't think this through. I remember Annelies mentioning it in a meeting and that's why I brought it up. I think you're correct therefore let me approve this!!

jpelay (approved):

AnneliesVlaar:
> The only functionality that could potentially show the solutions for custom adventures is when a teacher previews their own class. However, I'd argue that solutions should not be displayed because if solutions appear in the preview, we will give the false impression that students would see the solutions tabs.

I agree that it would give a false impression if we would show the answer tab in the class preview. But we still have a problem to solve, if a teacher wants to check the answers of custom adventures what is the (quickest) way to get to these answers? Let's think about this when we're working on the for teacher pages

PR #6277, opened 2025-04-15
🪲 fixes #6276

### LINKED ISSUE #6276: 🪲HTML can be passed as a class name
**Describe the bug**
When a teacher creates a class, they can provide HTML content as the class name instead of plain text

https://github.com/user-attachments/assets/0ce650b3-d546-4b94-a70b-c273ac522076

FBastiaan04:
This issue was fixed by adding a `escapeHTML` function to `utils.ts` and using it in `custom-elements.ts` to escape the html of the options.

Please replace all _____ before you create a pull request!

Start the name of the PR with one of the relevant prefixes in the title of the PR:

* 🖊️ -> changes related to grammars or the transpiler
* 🪲 -> solving a bug
* 🚚 -> changes unrelated to shipping, such as mergify scripts
* 🧹 -> refactorings (changes that do not change functionality and are only a cleanup of code)
* 💻 -> improvements and changes of the Hedy website
* 🧪 -> changes made to the test suite
* 📜 -> changes related to content

Fixes #6276 (Always link the number of the issue or of the discussion that your PR concerns. If you use the word `fixes` before the issue number in this description, the related issue will automatically close then the PR is merged)

**How to test**

* Follow along with the video presented in #6276
* Note that the option is no longer a link, but instead contains the properly escaped name of the class.

**Checklist**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@
 .dockerignore
 Dockerfile
 License.txt
+node_modules
--- a/static/js/custom-elements.ts
+++ b/static/js/custom-elements.ts
@@ -1,4 +1,5 @@
 import { ClientMessages } from "./client-messages";
+import { escapeHTML } from "./utils";  
 
 export class HedySelect extends HTMLElement {
     constructor() {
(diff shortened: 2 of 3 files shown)


Kache:
Posted my own [findings](https://github.com/hedyorg/hedy/issues/6276#issuecomment-2893147174) in the issue, in short:

 * The BE is already properly escaping the HTML -- a flaw in `HedySelect` is adding an extra round of rendering
 * With the above in mind, IMO it makes more sense to remove the extra rendering rather than escaping a second time

In addition, I disagree with "pre-escaping" user input before storing in the DB. To illustrate, consider other potential destinations for user input, like the URL (or SQL queries). It doesn't make sense to preemptively escape for all formats where it might be used. Instead, escaping should be done contextually at usage time by default, e.g. templating system that escapes HTML by default, SQL queries that are parametrized to escape inputs by default, etc.

PR #6283, opened 2025-04-18
💻 Improvements and redesign of adventure page

### LINKED ISSUE #6130: 💻Changes to new adventure design
With the deletion list in mind some buttons will go away:
- cheatsheet
- programmersmode

In the design below the next/previous adventure button, the completion bar and the more adventures button are removed. To move to a next adventure, the user can click on the adventure name, that opens up the dropdown menu and select another adventure. In this way the users always get's an idea on where in the list of adventures they are (at the beginning or almost finished)
Image

The list of adventures are in the current situation distinguished with a small colored vertical line in front. Maybe we can use a lightbulb to mark the explanation adventures to make the distinction more visible:
Image

jpelay:
This PR addresses the comments from the POs regarding the new adventure page. These changes have been the following:

* The "See more adventures" button that opened the side pane has been removed and a new button that opens a dropdown has been added. This button retains the same functionality as the side pane, with a few improvements.
image

* In the dropdown, instead of using colored bars to indicate the type of the adventure, new symbols have been added: a lightbulb for adventures that explain commands, a pen in a box to indicate teacher adventures.
image

* In the previous image you can see also checkmarks to indicate that a program has been reviewed by the teacher in the grading page. Also, a paper plane appears when the program has been submitted. The checkmark has priority over the paper plane. Comment for @AnneliesVlaar and @MarleenGilsing, I couldn't add the icon for "Unsubmitted" because that would require further changes in the backend, can we leave this for a further PR?

*  You now can only have one level open in the dropdown, meaning that you can only see the adventures for a single level. This is necesarry because the dropdown is rather small, and opening several levels could lead to getting lost.

* When opening the adventures for a level, the dropdown automatically scrolls to show those adventures. When opening the dropdown itself, it scrolls automatically to the opened level.

* The cheatsheet button has also been removed.

* The share button have also been removed which leads me to another question for @AnneliesVlaar and @MarleenGilsing. The name of the program is something you can set in the share modal (pictured in the image below) with the removal of this section, how can students be able to change the names of their programs? If this a feature we want to remove completely? For now I just set that input field as hidden and students cant change the names of their programs but can submit them, without this inout the storing of programs would fail.

image

* the go to next adventure and previous adventure buttons have been removed and also the progress bar.

Fixes  #6130 (Always link the number of the issue or of the discussion that your PR concerns. If you use the word `fixes` before the issue number in this description, the related issue will automatically close then the PR is merged)

**How to test**

Please test extensively these features! Specially the ones around submitting programs, blocking levels on customizations and see the effect in the dropdown, and general usability!


app.py
```diff
@@ -1639,7 +1658,12 @@ def tryit(level, program_id):
     adventures = load_adventures_for_level(level)
     load_customized_adventures(level, customizations, adventures)
     load_saved_programs(level, adventures, loaded_program)
-    adventures_for_index = load_all_adventures_for_index(customizations)
+    adventures_for_index = load_all_adventures_for_index(customizations, user_programs)
+
+    if current_user()['username']:
+        user_programs = g_db().all_last_programs_for_user(current_user()['username'])
+        print(user_programs)
+        print(adventures_for_index)
```
boryanagoncharenko, line 1666: the `print` statements are probably for testing?

boryanagoncharenko (commented):
From a usability perspective, it is weird that the menu does not close if I pick an adventure from my current level, but it does close automatically when I chose an adventure from another level.

https://github.com/user-attachments/assets/ad8e2a8a-0226-440c-8686-643ba4d6ab95

I'd also expect the menu to close when I shift the focus to another element on the page, e.g. the menu should autoclose when I click the Hand-in button. However, I understand this might be hard to achieve.

jpelay:
> From a usability perspective, it is weird that the menu does not close if I pick an adventure from my current level, but it does close automatically when I chose an adventure from another level.
> 
>  Screen.Recording.2025-04-20.at.15.45.11.mov 
> I'd also expect the menu to close when I shift the focus to another element on the page, e.g. the menu should autoclose when I click the Hand-in button. However, I understand this might be hard to achieve.

Hi! Yes totally, it might be a bit difficult to achieve, let me time box this to see if I can fix this quickly!

app.py
jpelay: For sure yes, thanks for pointing it out

boryanagoncharenko:
I made a minimalist commit which removes the `print` statements and corrects the function name and documentation in the database.py file. I hope it will not clash with any local changes you have. With that said, there are a couple of issues I discovered:
1. I cannot hand in programs. I am trying to submit a program for the Story adventure of level 1 and no matter what I do, I always get the message "This program does not have an id".
2. The hand-in button seems to stretch (and go to another row) when the screen side changes. It looks a bit off but I am not sure what the goal of the new design is. This is how it looks:

https://github.com/user-attachments/assets/d49280f7-d9e0-4366-b6ce-f499e2741c97

A small point for consideration for the POs is that right now, the `ticked` state overwrites the `submitted` state. So when a program is already ticked and I submit it again (nothing stops me from doing so) I still see the program as ticked.

boryanagoncharenko:
I stopped being able to replicate the issue with submitting a program. Probably it occurred because of a combination of the browser local storage and the dev db which I am afraid I cannot reproduce anymore.

boryanagoncharenko (approved):

AnneliesVlaar:
@jpelay and @boryanagoncharenko Really nice work!! I can only find small stuff (I will continue testing next week)

• Pen in box for teacher adventures, gives the impression that it is something different, but from the student perspective it is just another adventure like the rest. I would suggest to have no symbol for custom adventures.

• @jpelay Unsubmitted icon not possible yet. When a teacher unsubmits a program the paperplane disapears, that is also a visual indication for the student to find the program that needs to be made again. So indeed let's leave this for the future. 

• @jpelay Giving names to programs, I think we can remove this option because most programs are solutions to excersises. But for the free play adventures it would make sense to give it a name. I like the way how it displayed now without a name. 

• Read aloud, it's not clear if and what voice is selected. How to stop reading aloud, there is no option 'silent' or something similar. When I click the name of the voice again the checkmark doesn't disappear. And for me it doesn't work on alpha..?

• Hand in modal has cross and close button, double up? Remove close button and only have a cross?

• What is the idea of the 'give the following link to your teacher'? Is that still necessary?
Scherm­afbeelding 2025-04-25 om 10 39 43

• If I submit a program, I don't see a paper plane in the side pane. Only when I go to **another adventure in another level** the symbol appears (or reload the page, but then I go back to the introduction adventure). I would expect to see it when I go to another adventure in the same level. Is that possible?

• I do see the checkmark :D <3

boryanagoncharenko:
Thanks @AnneliesVlaar! I created the following issues:
- #6296 we will try to address this on Monday so that you can retest
- #6297 this we will not be able to address until next week. However, we could make an iteration soon after the release next week? I will leave the priority of this issue to you.

Issue #6288, opened 2025-04-21
📜wrong text arabic content


stoufix:
**Describe the issue**
on level 2 , arabic language , on the quiz section question 7 , 
the question was (i will explose in 3 seconde) , it should be (the bomb will explose in 3 seconde) .

![Image](https://github.com/user-attachments/assets/3baccc80-7e41-44b4-bdeb-d838ca514f6b)

**Expected text**
                {print}  ستنفجر القنبلة  خلال ٣ ثوانٍ!

I hope it's a coincidence that this bomb quiz  is only in Arabic .

AnneliesVlaar:
Hi @stoufix as you can check in the english version we intentionally use “I will explode” (as in expressing an emotion: https://dictionary.cambridge.org/dictionary/english/explode ) 

![Image](https://github.com/user-attachments/assets/00d78252-5d92-4bd8-a733-101336e03aa7)

If you find typos you can fix them via weblate, for more information see: https://github.com/hedyorg/hedy/wiki/Hedy-Translation-Tutorial

PR #6289, opened 2025-04-21
[issues:6288] : fix arabic content , level 2 question 7

stoufix:


AnneliesVlaar:
see comment on issue 📜wrong text arabic content #6288

PR #6299, opened 2025-04-29
💻 Improvements and fixes to level page

### LINKED ISSUE #6296: 🪲 Address feedback for the new hedy page
This issue is opened for addressing the changes requested in #6283. Specifically:
- Remove the "pen in box" icon for teacher adventures.
- The read-aloud button is not working properly:
    - It's not clear if and what voice is selected.
    - How to stop reading aloud, there is no option 'silent' or something similar.
    - When I click the name of the voice again the checkmark doesn't disappear. 
    - It doesn't work on alpha
    - I should be able to use the button without opening the dropdown
- Remove the cross in the upper-right corner from the hand-in modal
- Remove the 'give the following link to your teacher' interface from the hand-in modal
- Make the submit button displayed in the dropdown immediately after a program is submitted.

jpelay:
Addresses comments regarding the redesign of the level page.

* The pen-in-box has been removed
* When submitting a program, the paper plane is now immediately seen in the dropdown.
* The cross button has been removed from the pop up
* The copy link UI element has also been removed

Question regarding read outloud: what do you think about returning to the old select?

image
  

Fixes #6296


boryanagoncharenko:
I think I have completely misunderstood how the read aloud button works. I thought it should do 2 separate things: allow you to pick a voice AND act as a normal button so that when you click it it reads aloud. The purpose of the button is that you can choose silence or read-aloud option with a specific voice. This implies that you should be able to see your current selection (so that you know whether the read-aloud option is even on). Which leads basically to the old design indeed: a label that says "Read aloud" followed by a dropdown through which you can turn it on and off. We should just style it according to the new theme. @AnneliesVlaar @MarleenGilsing do you agree?

AnneliesVlaar:
> I think I have completely misunderstood how the read aloud button works. I thought it should do 2 separate things: allow you to pick a voice AND act as a normal button so that when you click it it reads aloud. In reality the button lets you choose silence or read-aloud option with a specific voice. This implies that you should be able to see your current selection (so that you know whether the read-aloud option is even on). Which leads basically to the old design indeed: a label that says "Read aloud" followed by a dropdown through which you can turn it on and off. We should just style it according to the new theme. @AnneliesVlaar @MarleenGilsing do you agree?

Yes I totally agree, the old read aloud worked good it only needs the new style

AnneliesVlaar:
> Question regarding read outloud: what do you think about returning to the old select?

I do like it in the new design that it is somehow more compact. 
Maybe we can do something like this, when it is not reading aloud it shows the shush symbol with the word 'silence' 
Level 1 (Read aloud) (Read aloud)

And when there is a voice selected it shows this voice with the speaker in front. 

Level 1 (Share becomes hand in) (Share becomes hand in)

Would this be possible?

AnneliesVlaar:
> When submitting a program, the paper plane is now immediately seen in the dropdown.

Super nice!! 🚀

AnneliesVlaar:
> The cross button has been removed from the pop up

The close-button and the 'I ll work on it a little longer' have the same result. We do want the user to make a choice between hand in or work on it longer. I suggest we also **remove the close button**, even more because clicking outside the modal also closes it. So there are enough ways to get rid of it. 

Scherm­afbeelding 2025-04-29 om 09 35 18

AnneliesVlaar:
After a user hands in an adventure there is a new message/modal shows "This is a submitted program and can't be altered." 
Which is not information I would expect or need at that moment. It is also shown when you already submitted a program and click again on the submit button. 

Scherm­afbeelding 2025-04-29 om 11 45 24

We could give some feedback after submitting a program but instead of "This is a submitted program and can't be altered." we could say something like "This program has (already) been submitted successfully."
Because it is more an information message would it be an idea to show it as a green message, just like the once we show when you change settings in your personal settings?

Scherm­afbeelding 2025-04-29 om 11 52 13

Maybe we don't even need a message because after submitting the yellow box above the editor appears, telling the user that the program is submitted and can't be altered. 

Scherm­afbeelding 2025-04-29 om 12 17 30

But even if the user submitted the program, the user can still click on the button. We can then show the message "This program has (already) been submitted successfully." Or maybe we should disable the button, I've shown above how that could look like.

jpelay:
Hello @AnneliesVlaar! Thanks for the review:

1. I think that is better if we keep the close button, some people browse using screen readers and having an explicit close button might be better, is an analogy that most people are used to.
2. I took upon your idea of disabling the hand in button when submitting a program! Now its disabled if the program has been submitted. I also slightly edited the message.
3. I removed the options from the customizations page.

boryanagoncharenko:
Technically the changes seem fine. @AnneliesVlaar if you agree with them, please approve the pull request.

AnneliesVlaar:
> 2. I took upon your idea of disabling the hand in button when submitting a program! Now its disabled if the program has been submitted. I also slightly edited the message.

This is great!

boryanagoncharenko (approved):

PR #6306, opened 2025-05-01
🪲 Address problems with the url of the tryit page

### LINKED ISSUE #6305: 🪲 Switching between adventures causes an #undefined url fragment
**Describe the bug**
When switching between levels the current url changes from `/hedy` to `/tryit`. When switching adventures in the same level,  the hash fragment of the url becomes `#undefined` instead of the usual adventure id, e.g. `#print_command` or `#parrot`

Currently the issue can be reproduced locally and on alpha. Steps to reproduce:

1. Open the /hedy page, note that the fragment of the current url is #6095 
2. Open the levels dropdown and select the parrot adventure from level 1. Note that when you select it, url changes its anchor to `#undefined`. If you cannot replicate this with the first try, select another adventure within the level. 
3. Open another level and notice that the url changed from `/hedy` to `/tryit`

boryanagoncharenko:
When switching between levels the current url changes from /hedy to /tryit. When switching adventures in the same level, the hash fragment of the url becomes #undefined instead of the usual adventure id, e.g. #print_command or #parrot

Fixes #6305

**How to test**

- Open the /hedy page, note that the fragment of the current url is `#default`
- Open the levels dropdown and select the parrot adventure from level 1. The anchor of the url should be changed to `#parrot` and not to `#undefined`
- Open an adventure in another level and check that the url is `/hedy` and not `/tryit`


jpelay (approved):
Works!

Issue #6309, opened 2025-05-02
💻 Make the turtle-download-button appear on the same row as the run program button


boryanagoncharenko:
After a turtle program is executed, a download-turtle-program button appears near the run program button. Currently the button uses an icon and it should use text instead. Also, it should appear on the same row as the run code and debug buttons, but now it appears below them.

<img width="448" alt="Image" src="https://github.com/user-attachments/assets/55504c4f-f087-43d6-bcc4-12720f116a4c" />

jpelay:
Hi @AnneliesVlaar, since you've shown the same design here as in #6310, Im going to work on both of these issues in the same PR.

Issue #6311, opened 2025-05-02
🪲 Show handed-in programs which are example-identical to the teacher


boryanagoncharenko:
Currently, if a student hands in a program which is identical to the example code, this program will not be displayed in the grading table of the teacher. We explicitly filter and show only modified programs, which differ from the adventure code. This leaves the program completely in a locked state that cannot be resolved by the student or the teacher: the student cannot change the program because it is handed in and the teacher does not see the submission, so that they could approve or reject it.

The way to solve this is to allow teachers to see programs that have a modified flag set to false. The current grading page should not display information that the program is identical to the example code. This will be addressed in the redesign.

PR #6322, opened 2025-05-12
💻 Reduce size of turtle window

### LINKED ISSUE #6310: 🪲 Fix the variable pane in the output when running a turtle program
When executing a turtle command, the blue pane showing the variables appears as a super thin line. You cannot open it, but it is there:

Image

@AnneliesVlaar I am not sure what the desired fix entails. Shall we make the variable pane go away completely or shall we display it as in non-turtle programs? I assume turtle programs can have variables, so it makes sense to display the pane.

jpelay:
On this PR I fix two thing:

1. I reduce the size of the turtle output window from 300px to 250px, thus allowing the variable pane to be shown below the turtle.
2. I place the save drawing button next to the hand in button, and change te emoji to a font awesome icon.

Fixes #6310 #6309 

**How to test**

Go to a level greater than 2, and try a code that uses both the turtle and a variable.


AnneliesVlaar:
> Fixes #6310 #6311

I think you mean #6309 instead of #6311?

AnneliesVlaar:
It seems the padding around the download button (and the other buttons on the left) is different then from the expand out button. The buttons on the left change when the download button appears. 

Scherm­afbeelding 2025-05-13 om 15 05 44


Can you make the button have the same appearance as debug/hand in/expand out? And is it possible to put it below the output? Since the output is downloaded I think it makes more sense to put it there.

jpelay:
Thanks for your input Annelies! I know that these buttons are not priority, but I couldn't help but to dedicate some time to improving their responsiveness.

Here I attach some videos where I highlight some improvements I made. Particularly there is a certain screen size between 1024px and 1300px where the space is very crammed and the buttons would overflow and look really big. What I did was to hide their labels in this range, thus allowing for more space. I also added tooltips, so users can still know the purpose of the buttons, beyond just the icons.

First: the save turtle and expand editor buttons. When the screen size is between the aforementioned range the space would be very small and thus not be able to fit the buttons and their lablels in the same line. 

https://github.com/user-attachments/assets/27d99047-a90f-477c-b110-e109b06d5b06


I did the same thing for the debug buttons, now I just show their icons and not the labels. In the next video you can see the previous behaviour.

https://github.com/user-attachments/assets/3dc2fb1e-8a7b-41ac-b66e-ac034137719d



https://github.com/user-attachments/assets/5f9fe9a7-83c0-4ac9-b59c-9619ae5f3f9f



I think that these fixes help a lot specially in smaller desktop screens, like the ones you would find in a budget laptop.

jpelay:
I have fixed the issue where the buttons were bigger than they needed to be. Thanks a lot for catching it!

boryanagoncharenko:
The PR looks ok from a technical perspective but I am a bit concerned about usability. Right now if I execute a turtle program with 2 variables, the blue pane shows only 1 variable and nothing indicates that I can scroll inside the blue pane. 
Screenshot 2025-05-19 at 16 01 49

This immediately leaves me to believe that the second variable is not created at all. The blue pane has a button that opens and collapses it, so I think it would be surprising to our users that it only opens for the first variable and scrolls for the rest.

I know this goes beyond the scope of the task at hand, but why is the variable pane have the variables on separate rows. Currently, it has so much blank space. And why is it there at all? Wouldn't it make more sense to appear not on top of the output but somewhere near the debug button or in the input editor? Perhaps there was an improvement opportunity that we missed with the redesign of the tryit page.

@AnneliesVlaar the PR is good technically. If you disagree with my UI concerns or you think the current solution is good enough, please approve the PR or let me know and I will approve it.

AnneliesVlaar:
> This immediately leaves me to believe that the second variable is not created at all. The blue pane has a button that opens and collapses it, so I think it would be surprising to our users that it only opens for the first variable and scrolls for the rest.

I agree with your findings Boryana and there might be a better way to show the variables. But I think we should leave at as it is for now. Because it is only a problem with the turtle and the variables are hidden but not inaccessible. So let's put it on the wish list for when we have no more priority thinks to do ;-)

boryanagoncharenko (approved):

jpelay:
> The PR looks ok from a technical perspective but I am a bit concerned about usability. Right now if I execute a turtle program with 2 variables, the blue pane shows only 1 variable and nothing indicates that I can scroll inside the blue pane. Screenshot 2025-05-19 at 16 01 49
> 
> This immediately leaves me to believe that the second variable is not created at all. The blue pane has a button that opens and collapses it, so I think it would be surprising to our users that it only opens for the first variable and scrolls for the rest.
> 
> I know this goes beyond the scope of the task at hand, but why is the variable pane have the variables on separate rows. Currently, it has so much blank space. And why is it there at all? Wouldn't it make more sense to appear not on top of the output but somewhere near the debug button or in the input editor? Perhaps there was an improvement opportunity that we missed with the redesign of the tryit page.
> 
> @AnneliesVlaar the PR is good technically. If you disagree with my UI concerns or you think the current solution is good enough, please approve the PR or let me know and I will approve it.

This is Safari right? Apple loves to hid scroll bars, on Chrome it looks oh right. I still believe the best solution for this is to give the turtle its own space. 350 pixels for everything is too little! 

image

PR #6325, opened 2025-05-15
🪲 Show submitted unmodified programs to teachers

### LINKED ISSUE #6311: 🪲 Show handed-in programs which are example-identical to the teacher

boryanagoncharenko:
The PR allows for displaying unmodified programs to teachers. Note that this PR focuses only on removing the situation in which a program is locked because it is handed in by a student and not visible by the teacher. This PR does not try to tell the teacher that the program is unmodified. This visualization will be done in the new teachers' interface.

Fixes #6311

**How to test**

- Log in as a student, who is enrolled in a class, and navigate to an adventure that has a working sample code (some levels have incomplete programs and you need a sample snippet that would just run without changes), e.g. Level 3, adventure Music. Paste the sample adventure code and execute it without changing it. Now hand in the program.
- Log in as the teacher of the class and check that the teacher can actually see the submitted program. It should appear in the "Student's Adventures" table and it should also appear in the "Class performance graph".

### DIFF
--- a/website/for_teachers.py
+++ b/website/for_teachers.py
@@ -343,8 +343,7 @@ class ForTeachersModule(WebsiteModule):
                 adventures_tried += 1
                 name = adventure_names.get(program['adventure_name'], program['adventure_name'])
                 customized_level = class_adventures_formatted.get(str(program['level']))
-                if next((adventure for adventure in customized_level if adventure["name"] == name), False)\
-                        and program.get('is_modified'):
+                if next((adventure for adventure in customized_level if adventure["name"] == name), False):
                     student_adventure_id = f"{student}-{program['adventure_name']}-{level}"
                     current_adventure = self.db.student_adventure_by_id(student_adventure_id)
                     if not current_adventure:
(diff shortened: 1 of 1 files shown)


jpelay:
HI @boryanagoncharenko! I notice that this PR not only shows submitted unmodified programs, but it shows all unmodified saved programs, is that what we wanted?

boryanagoncharenko:
> HI @boryanagoncharenko! I notice that this PR not only shows submitted unmodified programs, but it shows all unmodified saved programs, is that what we wanted?

Oh no, unfortunately, you are right! The grading grid on the teacher's page also shows saved programs which were not handed in. So basically, if I am a student and I execute a program (without handing it in), it appears as submitted on the teacher's page. Correct me if I am wrong but this is not a regression done in the PR. This seems to be the behavior on main too, is that right? I completely agree that we should perhaps revisit this idea.

@AnneliesVlaar it looks kind of obvious that we should display only handed-in programs but since it is a different issue than the original one, I would like to double-check with you on the solution.

boryanagoncharenko:
I looked a bit further in the issue and I got more confused. I am not sure it is so obvious that the teacher's grading table is supposed to show only submitted programs. No tests, in-memory data, or github issues suggest that this is meant for submitted programs only. And to make things even more confusing, what programs should we show in the Class performance graph (below the grading table). While I did change the graph to display unmodified programs, now I am wondering whether this is a good idea.

AnneliesVlaar:
@boryanagoncharenko Let's discuss this in the meeting on Friday to ensure we're talking about the same things all the time (I think there are multiple ways to show programs of users to teachers 🫤)

AnneliesVlaar:
As discussed in the meeting, let's leave it as it is for now and make sure we address this properly in the new Teacher UI (meaning, showing only handed in programs in the grading environment and mark them when they are (almost) identical)

PR #6331, opened 2025-05-16
🪲 Teachers cannot be invited as second teachers (part 1)

### LINKED ISSUE #6321: 🪲 DB inconsistency of users is_teacher property type
**Describe the bug**
Newly-created teachers cannot be invited as second teachers. Here is how to replicate this:
- On alpha/production create a new teacher
- Try inviting this newly created teacher as a second teacher of a class. Note that you cannot find the username of the teacher at all.

The root cause has to do with the fact that DynamoDB creates the user with `is_teacher=true` of type boolean. While the rest of the code base (including the table definition) expects the type of the field to be a number `is_teacher=1`. Once the value is changed from `true` to `1`, the username appears in form filtering users for inviting second teachers.

The conflict occurred on 6 Sep 2024 which implies there any many users created with an incorrect value.

boryanagoncharenko:
This PR is a part of 2-step fix. It changes the value of the `is_teacher` attribute for newly created teachers from `true` to `1`. Correcting the wrong attribute values for existing users will be done separately of this PR through a one-time-script.

As a side effect, it addresses the following problem: a teacher with `is_teacher=true` cannot be invited as a second teacher because they cannot appear in the invitation modal.

Fixes #6321

**How to test**
This issue cannot be replicated locally - it requires dynamo db. Once this is deployed to alpha, create a new teacher and try inviting them as a second teacher of a class. Their username should appear in the search pop up.


jpelay (approved):
Now I see I introduced this bug 🙈

PR #6341, opened 2025-05-20
🪲 Two color command bugs

### LINKED ISSUE #6316: 🪲Turtle color not working from level 12 and higher
**Describe the bug**
Color (in or outside of a function) is not working in level 12 and higher

```
color green
forward 50
```

boryanagoncharenko:
This PR addresses two regressions introduced with the last large refactoring of the transpiler:
- when passing color keywords directly to the `color` command quotes became required. Instead of `color red`, the correct syntax became `color 'red'` which made all sample snippets fail at runtime.
- the `color` command could not use parameters of a function as variable names. I checked that the scope of this issue is only the color command.

@AnneliesVlaar however, I have to bring your attention to some asymmetrical syntax that we have regarding colors. With this PR, quotes are not needed when using literals:
`color red`
but they are needed when using variables:
```
c = 'red'
color c
```
Quotes are also required when they are passed as function parameters:
```
def move with color
    color color
call move with 'red'
```

Wouldn't it actually be better to keep the quotes around color literals, so that the correct syntax is `color 'red'` after level 12? Alternatively, we could make colors these as reserved that can be used anywhere without quotes but I am afraid this will be a massive change that would move away from Python instead of towards it. Let me know your

Fixes #6316

**How to test**
In level 12 and above, the following code should work:
```
color red
```
Also, try this in different languages
```
kleur geel
```

Additionally, in level 13 and above the color command can use function parameters without producing an error:
```
define move with color
    color color
c is 'blue'
call move with c
```


AnneliesVlaar:
> Wouldn't it actually be better to keep the quotes around color literals, so that the correct syntax is `color 'red'` after level 12?

As discussed in the meeting, we agree that `color 'red' ` would be better fit from level 12 on, so that it is more in sync with variables and parameters

jpelay (changes requested):
The code works great, but can we improve this message to say that people are using an unused variable instead of saying this error about the value? It's a bit confusing imo. 

image

boryanagoncharenko:
> The code works great, but can we improve this message to say that people are using an unused variable instead of saying this error about the value? It's a bit confusing imo.
> image

Excellent point! The error message needs to be improved indeed.

jpelay (approved):
Thanks a lot for the fix 😄

image

AnneliesVlaar:
> To ease the transition for students, perhaps adventures can provide an extra line in the examples?
> 
> Before 12:
> 
> ```
> color red
> ```
> 
> In (select) examples in lvl 12:
> 
> ```
> red = 'red'  # provide this line
> color red    # looks the same, but it's a variable, now
> ```
> 
> Or maybe it might confuse more? 🫤

I think it might be even more confusing. I'm glad we have syntax highlighting :)

PR #6347, opened 2025-05-22
💻 Show loading spinner when opening code page

### LINKED ISSUE #6312: 💻 Clicking the navigation button on the /hedy page causes an error
When the /hedy page loads, the default adventure is loaded first and after the javascript loads, the currently selected adventure  is loaded. If the user clicks on the navigation button before javascript is loaded, an error in the console is produced. This is especially visible for users with slow internet connections.

jpelay:
When users open our page, and they are on a slow connection, there's no indiciation for them to know that the page is loading. Therefore it seems that page is unresponsive and there's an error.

This PR addresses that by not showing content directly to the user when the code page loads, but rather a stub of an adventure with an animation, indicating that the javascript is loading. To prevent the behaviour of clicking a button and showing an error, I disabled the buttons while the page loads.

Here's a video of how it looks:

https://github.com/user-attachments/assets/aea486b0-ca10-4a68-b882-8fbfd2a4b912

Fixes #6312

**How to test**

Open your browser's dev tools by pressing F12. Then go to the network tab and click on disable caching, on Throttling click on slow 4G. Reload the page and after a bit you should see the new spinner.

image


static/js/app.ts
```diff
@@ -1715,7 +1722,7 @@ function updatePageElements() {
     if (isSubmitted) {
       hand_in_btn?.setAttribute('disabled', 'disabled');
     } else {
-      hand_in_btn?.removeAttribute('disabled');
+      // hand_in_btn?.removeAttribute('disabled');
```
boryanagoncharenko, line 1725: The whole `else` statement should be removed, right?

boryanagoncharenko (commented):
This PR definitely improves the user experience in case of slow connection and it gets rid of the preloading-the-initial-adventure issue. However, it also goes a bit against htmx as it relies on using page initialization functions.

static/js/app.ts
jpelay: Nooo quite the opposite! Thanks for the catch!

boryanagoncharenko: You are right, it is needed.

Kache:
It seems this PR contains changes from #6337, was that inadvertent?

b/c #6337 was still under review, e.g. #6352 adds a few additional fixes on top of it

boryanagoncharenko:
> It seems this PR contains changes from #6337, was that inadvertent?
> 
> b/c #6337 was still under review, e.g. #6352 adds a few additional fixes on top of it

Thanks for pointing this out. This is an error on my end. I reviewed #6337 and the changes can to be merged safely. Definitely not ok that it happened sneakily in this PR though. I opened #6352 which makes small adjustments on top of your commit.

PR #6348, opened 2025-05-23
🪲 Fix HedySelect HTML injection

Kache:
`HedySelect` added an extra round of rendering by using [`innerText`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText), so use `innerHTML` instead, trusting the raw HTML from the server

The raw HTML is already escaped for rendering (evident /w javascript disabled) because the template system escapes by default for safety

Points from contributors meeting:

 * Removing the extra rendering is safe, but loses the "feature" of user-styled `HedySelect` options. This behavior can be safely reintroduced in the future by supporting markdown instead
 * Preemptively HTML escaping user input before db storage is too presumptive. User input could feasibly be used in other places like URLs or SQL queries, so relevant escaping at usage makes more sense.
 * It can still improve usability to allow/block certain characters in user input, but that can be done separately

Testing: [repro video](https://github.com/hedyorg/hedy/issues/6276#issue-2995857046)

Issue: #6276


jpelay (approved):
Thanks a lot on your first contribution @Kache! I just tested this, and works great :) On main it renders the html as is, while with this change it properly renders it as text.

Kache:
rebased b/c conflicts in compiled JS

PR #6352, opened 2025-05-29
🪲 Fix language switcher

### LINKED ISSUE #3937: 📜  Language switcher dropdown issues
"to" appears twice in the Hedy language. Once in "add to" and once in combination with range. Both can have different translations and the wrong one is shown in Dutch for the "to" in level 9 (it should be "toe" not "tot")

image

So we need to find the right one and not (as I suspect happens here) the first one.

boryanagoncharenko:
Based of the existing PR #6337, this PR corrects the `to` and `to_list` usage and ensures the keyword `pressed` appears correctly in the lang switcher table.

Fixes #3937
**How to test**
- Make sure the language switcher translates keywords and phrases correctly in all levels, specifically in the upper levels where `add to` and `range to` are used.


jpelay (commented):
~Everything works great, but `if_pressed` it's still displayed as `is_pressed`, is that how you did it or is there an error right now?~

Im dumb, was trying in the wrong page lol

jpelay (approved):
Works perfectly :)

PR #6354, opened 2025-06-06
 🧹 Clean code related to already removed features

### LINKED ISSUE #6318: 🧹Clean up the code of the old adventure page
Now that we have a lovely new design (which lived on the tryit page for a long time) it is a good idea to clean up some old code. We put it on the low priority list, so future code-archeologists ... We really had the intention to clean up ;-)

jpelay:
Now that we were able to ship our new design for the adventures page, we can begin to remove from the code all of the features that we removed in version, most notably quizzes and parsons.

So, a summary:

* It removes some of the html files that weren't needed. Some of them I kept, to maintain two endpoints called /adventures and /render-code. These two will be fixed when we get onto refactoring the teacher's page.
* Removes all of the content for the quizzes and parsons.
* Removes all mentions to quizzes and parsons in the code, except for the ones in the section for teachers, that one will be refactored shortly.
* It moves the hour-of-code endpoint to the new design. In this vein, I also removed the capability to pass a subset of adventures through the URL, something that was not really used and is fixed through customizations.

This changes shouldn't affect the way the site works, since all of these was unused code! Still I'd like if someone could do a little round of testing around and check that things are indeed working :)

Fixes #6318

**How to test**

* Everything should be working as expected


AnneliesVlaar:
@jpelay 
I tested al the obvious things and these work as normal 👍 
Two things that did not work.

- When the language of the teacher is not English, the code editor in the slides is not working.
- If a teacher is grading a program of a student it is not possible to run the code because all the buttons are disabled. 
Scherm­afbeelding 2025-06-24 om 13 27 01

AnneliesVlaar (approved):

PR #6368, opened 2025-07-16
💻 Added speakers notes to slides

### LINKED ISSUE #5805: 📜 Add teacher explanation to the slides

FBastiaan04:
Fixes #5805 

**How to test**
Currently, no content exists for solutions, so to test this, you will need to add solutions. For example, add the following snippets to the content/slides/en.yaml file:
```yaml
        1:
            header: Welcome to Hedy!
            text: Hedy is a programming language.
            notes: |-
                Ask your students to write down some words that they associate with programming. Do a quick round to hear the answers.
        2:
            header: Programming
            text: |-
                Programming is giving instructions to a computer, using the language of the computer.
                Such a language is called a "programming language".
            notes: |-
                Unfortunately, computers don't understand normal people language, like English, but they speak in their own language.
                Don't panic, you don't have to type in a string of zeros and ones, but we are going to learn to give the computer instructions in a programming language.
```

`notes` is an optional parameter. Verify that clicking `s` still opens the preview and stopwatch, even without notes for the current slide.

* Go to [/slides/1](http://localhost:8080/slides/1).
* Press the `s` key on your keyboard (this will open a popup that might take a few seconds to load).
* Speakers notes should appear.
* Check this for other slides you added notes to in the yaml.
* Check that everything also works if there are no notes for the current slide.

**Checklist**


jpelay:
Hello, Baastian! Sorry for the late review. I pressed s and indeed you can see a speaker's notes screen when you press s. However, I can't access level 1 slides, can you check that? There are also failing tests, can you look into those? Thanks!

FBastiaan04:
I fixed most of the problems that caused the checks to fail. The issue however is that the `content/slides/en.yaml` is copy-pasted directly form https://github.com/hedyorg/hedy/blob/slides-update/content/slides/en.yaml. This file changes some of the ordering for the slides, and this causes a failure in the tests since the files for the other languages are not updated. The specific issue is that in the `en` version, level 1, slide 12 contains the `debug` property in the old version, but not in the new version.
Should a `debug` prop be added to this slide as a quick fix to pass the tests, should the other language files be updated or should the changes in ordering be reverted?

AnneliesVlaar:
> , but it does make the "how to test" section obsolete

You can do something like this: https://github.com/hedyorg/hedy/pull/6264

jpelay (approved):

PR #6373, opened 2025-07-25
🪲 Fix code editor in slides in non-english languages

### LINKED ISSUE #6123: 🪲 Some slides can't use code editor, others do, depending on the page language
**Describe the bug**
I am currently checking my translations in slides and testing if program fragments have any issues.

The problem is in the use of the code editor in slides. Starting from level 3 slides, the editor panel gives 404 error.

This happens in 1 situation out of 3 possible states:
- ~O - users can see the editor if they are logged in~
- O - users can see the editor if the ~page~ keyword language is English, ~even if they are not logged in~
- X - the editor part gives 404 error if ~the user is not logged in AND page~ keyword language is not English

EDIT: I checked again and see being logged in is not related directly. I missed I last changed my keyword language to English while trying to check other changes in my translations

I prefer being able to access to the editor in slides in my language and without logging in, as it is useful to use slides in temporary environments. Considering it can still be used without logging in just by setting the language to English suggests me the security is not a concern here.

However, if you choose to require a logged in user, then you should also fix this for English language. In that case, just check the slides with the numbers other than the ones I listed below.

Please check the issue and fix the problem.
YILMAZ

Editor is giving 404 error in slides of these levels:
- 3
- 5
- 7
- 9  to 17

Steps to check the issue:
- open https://hedy.org and change page and/or keyword language other than English
- open https://hedy.org/slides/3#/2/0/0
- see the 404 page
- from within the slide, change the page/keyword language to English
- refresh and see the editor working

(use page language switcher if you are not logged in, and keyword language switcher if you are logged in. both works the same and gives 404 if language is not English)

https://github.com/user-attachments/assets/7e55e8e9-f290-4baf-8ad1-e3c03ff1507c

https://github.com/user-attachments/assets/8c5cb850-1b06-4264-9464-238353424d5a

Windows 10, MS Edge

FBastiaan04:
Fixes #6123

**How to test**

* Select a language other than english
* Go to [slide 3](http://localhost:8080/slides/3#/2/0/0) and go to the slide with the code editor
* Verify that the code editor works in the selected language

**Checklist**


app.py
```diff
@@ -1404,7 +1404,7 @@ def index(level, program_id):
     max_level = hedy.HEDY_MAX_LEVEL
     level_number = int(level)
     return render_template(
-        "hedy-page/code-page.html",
```
jpelay, line 1407: This will cause problems in many other pages, the mentions to this template are not accidental, they are intentional. The reason for the duplciated code-page.html template is due to the public adventures still needing the old one.

app.py
```diff
@@ -2004,6 +2009,7 @@ def get_slides(level):
         add_used_slides_to_subscription(email)
 
     slides = SLIDES[g.lang].get_slides_for_level(level, keyword_language)
+    print("HERE ", slides[2])
```
jpelay, line 2012: Can you remove this debug print?

app.py
jpelay: I think the problem in the new template is that is missing a variable, can you try fixing it instead?

jpelay (changes requested):

jpelay:
Hi, Bastiaan! Im still getting issues with the PR, look at this image:

FBastiaan04:
This may be a classic case of "it works on my PC". This error also appears to be different from the one I saw before. That had an error in the code for the slides themselves, whereas this looks like an error parsing the spanish code.
I was unable to reproduce this error. I tried the following steps:
1. I ran hedy by starting the dev container and running "python app.py" within that container.
2. I opened localhost:8080 in microsoft edge.
3. I changed my language and command language to spanish, as it appears as though this is the language in your screenshot.
4. I opened localhost:8080/slides/3 in the browser.
5. I clicked the "next slide" button until I saw the slide about random.
image

I also tried to view this while logged in (as a student account). Also in spanish. This had the same result.
Could you please help me to reproduce this error?

Lastly, while trying to reproduce this I also found an error where in all language files in /content/slides/ level 3, slide 7 has `editor` instead of `code`, causing an error. I could put a fix for that in this PR, as it's an error in a similar place. However, this will change all files in slides and is therefore likely to cause merge conflict. Should I make a new issue and PR?

PR #6384, opened 2025-08-14
💻 Read out loud redesign

### LINKED ISSUE #6314: > Question regarding read outloud: what do you think about returning to the old select?
> Question regarding read outloud: what do you think about returning to the old select?

I do like it in the new design that it is somehow more compact. 
Maybe we can do something like this, when it is not reading aloud it shows the shush symbol with the word 'silence' 
Level 1 (Read aloud) (Read aloud)

And when there is a voice selected it shows this voice with the speaker in front. 

Level 1 (Share becomes hand in) (Share becomes hand in)

Would this be possible?

_Originally posted by @AnneliesVlaar in https://github.com/hedyorg/hedy/issues/6299#issuecomment-2837900882_

brianpeiris:
Fixes #6314

Adds a mute button for the code editor's read out loud feature. The button mutes and unmutes the voice. When muting, any currently queued speech is also cancelled. The first option in the voice dropdown was changed to display "Choose a voice", using a new `choose_a_voice` translation string, instead of the silence emoji

image

image

Note that I used Font Awesome's ["volume-xmark"](https://fontawesome.com/icons/volume-xmark?f=classic&s=solid) icon for the mute state, since the ["volume-slash"](https://fontawesome.com/icons/volume-slash?f=classic&s=solid) icon is not included in the free set of icons.

----

**How to test**

* Visit http://localhost:8080/hedy/7#repeat_command
* Change `repeat 3` to `repeat 10`
* Choose a voice
* Click "Run code"
* Click the mute button and see that the speech is stopped
* Click the mute button again
* Click the "Run code" button
* See that the code is read out loud
* These are also good examples to test with: http://localhost:8080/hedy/4#parrot, http://localhost:8080/hedy/4#clear_command

**Checklist**

### DIFF
--- a/messages.pot
+++ b/messages.pot
@@ -593,6 +593,9 @@ msgstr ""
 msgid "expiration_date"
 msgstr ""
 
+msgid "facebook_page"
+msgstr ""
+
 msgid "favorite_program"
 msgstr ""
 
@@ -837,6 +840,9 @@ msgid "levels"
 msgstr ""
 
 msgid "link"
+msgstr ""
+
+msgid "linkedin_page"
 msgstr ""
 
 msgid "list"
--- a/templates/main-page.html
+++ b/templates/main-page.html
@@ -68,8 +68,8 @@
         </div>
         {% endfor %}
     </div>
-    <div class="bg-pink-800 text-sm text-white py-16">
-        <div class="flex w-100 max-w-screen-md mx-auto text-links-plain">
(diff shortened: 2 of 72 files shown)


jpelay (approved):
OMG Brian! Thank you so much, this works great and ties itself neatly with the rest of the page. This is perfect to go. I think itll go into production tommorrow or the day after!

PR #6405, opened 2025-09-12
📜More content adventures new level structure

MarleenGilsing:
New level 6 story adventure 

Follows the new level structure, so it seems like a lot of changes but it's only the english adventure yaml en hedy_content.py.


Felienne:
Ha @MarleenGilsing, I tried to rescue what I could but you might want to check if I did not delete too much (I only add the new level 6 text in default)

PR #6406, opened 2025-09-12
💻 Enhance print styles for workbooks

rix0rrr:
Made the following changes:

- It is now possible to add (printed) page breaks by adding `new-page` instead of exercise blocks into the workbook files (see `en.yaml`). The page breaks will only appear in print mode (press Ctrl-P).
- The menu bar is no longer printed in print mode
- The code blocks now have grey outlines in print mode.
- Exercise blocks with multiple choice options next to them are now styled in HTML, so that the text won't fold anymore.

**How to test**

Visit http://127.0.0.1:8080/for-teachers/workbooks/1 (in English) and hit Ctrl-P. Observe the changes mentioned above.


rix0rrr:
Downside of this decision: because the structure of all YAML files must be the same for all languages, we are now also forcing page breaks into the same places for all languages. 

This means that even though page layout may be different (because of different sentence lengths) the page breaks must appear in the exact same place.

Maybe just theoretical, but something to consider.

rix0rrr:
I have another useful PR stacked on top of this one. Would be nice to get this merged if possible.

Felienne (approved):

rix0rrr:
@mergify requeue

PR #6411, opened 2025-09-30
🚚 Add a script to print a language's workbook to a PDF

rix0rrr:
This adds a page that prints all the workbooks of a single language one after the other, and a script that uses Puppeteer to visit that page and export the page to a PDF.

This makes it possible to generate a printable PDF for any workbook with little effort.

The script currently works against a local server, so you need to run a copy of the devserver on your machine before running the script.

**How to test**

* Run a local copy of the server (`doit run devserver`)
* In a different terminal window, run `npm run print-workbook en` or `npm run print-workbook nl` (or `npm run print-workbook en nl`)
* Observe that `workbook-en.pdf` and `workbook-nl.pdf` files appear in the current directory.


Felienne (approved):

package.json
```diff
@@ -54,8 +55,10 @@
     "jquery-ui-dist": "^1.13.1",
     "jszip": "^3.10.1",
     "postcss": "^8.5.26",
+    "puppeteer": "^24.20.0",
```
sstorck, line 58: Should these be `devDependencies` instead?

tools/print-workbook.ts
```diff
@@ -0,0 +1,51 @@
+import puppeteer from 'puppeteer';
```
sstorck, line 1: Does this need to be `puppeteer` instead of `puppeteer-core`? `puppeteer` will install its own Chrome browser to the cache.

https://pptr.dev/guides/installation

PR #6417, opened 2025-10-10
📜Fixing content mistakes in new grammar

MarleenGilsing:
adding extra content (that got lost in the merge) and fixing small mistakes

Initially did some changes in Dutch as well, but moved them to a seperate PR #6418  (in case weblate gets mad). So this PR is English only.


MarleenGilsing:
@jpelay Fixed hospital level 14 and story level 6 :)

PR #6426, opened 2025-10-22
fix: rename 'commands' to 'keywords' in codebase

Jhanvi-code23:
This PR renames all occurrences of 'commands' to 'keywords' in the codebase, as requested in issue #XYZ. 
Only code references were updated; UI text for kids was left unchanged.


Felienne:
In any case it is very unlikely this can be merged in October, so if this is for Hacktober it won't work sadly.

PR #6440, opened 2025-11-19
Fix problems on the grammars

jpelay:
Please replace all _____ before you create a pull request!

Start the name of the PR with one of the relevant prefixes in the title of the PR:

* 🖊️ -> changes related to grammars or the transpiler
* 🪲 -> solving a bug
* 🚚 -> changes unrelated to shipping, such as mergify scripts
* 🧹 -> refactorings (changes that do not change functionality and are only a cleanup of code)
* 💻 -> improvements and changes of the Hedy website
* 🧪 -> changes made to the test suite
* 📜 -> changes related to content

Fixes _______ (Always link the number of the issue or of the discussion that your PR concerns. If you use the word `fixes` before the issue number in this description, the related issue will automatically close then the PR is merged)

**How to test**

Follow these steps to verify this PR works as intended:

* _____ (See this https://github.com/hedyorg/hedy/pull/880#issue-1016304308 for an example)
* _____

**Checklist**
Done? Check if you have it all in place using this list: (mark with x if done)

- [ ] Contains one of the PR categories in the name
- [ ] Describes changes in the format above
- [ ] Links to an existing issue or discussion
- [ ] Has a "How to test" section

If you're unsure about any of these, don't hesitate to ask. We're here to help!

PR #6448, opened 2025-12-02
🧪 Update Cypress tests for level refactor

jpelay:
Fixes the cypress tests and further problems with the website in the context of the level redesign


jpelay:
Depends on #6440 ( that one has the fixes related to the grammars and this one for the cypress tests)

PR #6489, opened 2026-01-23
Add a limit of 100 lines to the code editor

### LINKED ISSUE #6441: 💻Add a limit of 100 lines to the code editor
Students can write program with infinite lines at the moment, but only the first 100 lines will work, because we've put a limit at how many lines should be run. It would be better if the students simply couldn't type more lines than 100 (except for the free play adventure taht we want to create).

ToniSkulj:
Fixes #6441

### Summary
This PR changes the current behavior so that users receive immediate feedback while typing if a program exceeds the allowed length, instead of only getting an error after sending it to the server.

Current result after running program:

image

### Bug in current approach
Currently you are allowed to type 101 lines whilst the error returns that only 100 are allowed.

image

### Max line per level
It's now possible to set the max lines per level or range.

## How to test
From level 1 to 9: It should **not** be possible to go over **100** lines in the code editor when typing.
From level 9 and above:  It should **not** be possible to go over **200** lines in the code editor  when typing.

Whenever the user tries to add a new line character the user should get an informative warning message:

image


static/js/cm-editor.ts
```diff
@@ -86,6 +87,27 @@ export class HedyCodeMirrorEditor implements HedyEditor {
 
     constructor(element: HTMLElement, isReadOnly: boolean, editorType: EditorType, __: string = "ltr") {
         let state: EditorState;
+
+        let lineLimitFilter = EditorState.transactionFilter.of(
```
jpelay, line 95: Nice!

tests/cypress/e2e/lezer-tests/level_01.cy.js
```diff
@@ -1,112 +1,130 @@
-import { multiLevelTester, singleLevelTester } from "../tools/lezer/lezer_tester"
-
-describe('Lezer parser tests for level 1', () => {
```
jpelay, line 3: Why remove these? Or is it an unintended change?

jpelay (approved):

PR #6540, opened 2026-04-03
🚚 Hedy Offline automatically opens in browser on startup

OlivierSwaak:
Makes Hedy Offline automatically open in browser on startup.

Fixes no existing issue.

**How to test**

* Build Hedy Offline
* Launch the executable file to see the webbrowser is automatically opened.

I am a bit on the fence on whether or not to let it open `localhost` or to just pick the first ip adress we print. However, in the official documentation we also refer to `localhost` and I think that the edge case of a student laptop hosting multiple programs is small, so I think I'll just keep it at `localhost`.


OlivierSwaak:
I added a small fix to this PR as sometimes the ip-list would be empty (for example if it was only able to host on `127.0.0.1`) which made it impossible to launch Hedy if the autolaunch didn't work (or was already done earlier) or if you did not know it hosted the website. 

Because of this, I added `http://localhost/` to the list (with a disclaimer that it will only work for this machine) such that if hosting is successful, the list will never be empty:

image

jpelay:
> I added a small fix to this PR as sometimes the ip-list would be empty (for example if it was only able to host on `127.0.0.1`) which made it impossible to launch Hedy if the autolaunch didn't work (or was already done earlier) or if you did not know it hosted the website.
> 
> Because of this, I added `http://localhost/` to the list (with a disclaimer that it will only work for this machine) such that if hosting is successful, the list will never be empty:
> 
> image

Great fix! I was just testing and realized this. Good job

jpelay (approved):

PR #6560, opened 2026-05-21
🚚 Make Hedy build in a Docker container

rix0rrr:
In preparation of potentially moving Hedy to a different platform, make it build in a Docker container.

There is now a `Dockerfile` and `.dockerignore` at the top-level, and matching `doit` commands:

```
doit run build_container
doit run container
```

To build and start the container. Needs Docker, or a Docker-compatible CLI that has been aliased to the `docker` command (like `finch`).

Inside the container, we use `gunicorn` to run the server, just like we do on Heroku.

The Python dependencies are now split into `requirements.txt` (development dependencies) and `requirements-prod.txt`, which only contains the dependencies necessary to run the server. That way, we can install only the prod dependencies into the container and save space.

Also move some NPM `dependencies` to  `devDependencies`.

Environment configuration was starting to become complex, so has been made more explicit in `envs.py`: the various different bits of behavior now have an explicit name that is controlled in one place, so we no longer have random `is this a dev server? Do behavior X, otherwise Y` lying around the code. We will now have explicit "should I cache files over HTTP" or "should we allow testing requests", etc.

Remove some files that should never have been in source control to begin with.


jpelay:
Hello, rico! Where are we planning to move Hedy to?

rix0rrr:
> Hello, rico! Where are we planning to move Hedy to?

Ultimately, if I can get it all done in my spare time 😉, self-hosting at [Hetzner](https://www.hetzner.com/). We can do it for about 1/10th the price of what we pay Heroku there.

jpelay:
> Ultimately, if I can get it all done in my spare time 😉, self-hosting at [Hetzner](https://www.hetzner.com/). We can do it for about 1/10th the price of what we pay Heroku there.

That makes sense yes, Heroku is very expensive. I think for Hedy's scale we can make due with a single server running on a VPS. If you need anything or if could help let me know

PR #6572, opened 2026-05-28
🪲 Guard autosave against non-DOM event targets (fixes #6561)

### LINKED ISSUE #6561: 🪲Can't save any adventure changes - Error in autosave.ts
We do a Saturday coding class for kids. Thanks for making Hedy

The Customize Adventure page is broken. In the browser console there are errors like,

```
Uncaught TypeError: Cannot read properties of undefined (reading 'dataset')
    at autosave.ts:32:37
    at autosave.ts:14:7

Uncaught (in promise) Error: logged
    at JVe (comm.ts:71:25)
```

It makes network calls to https://hedy.org/client_exception, which returns just "logged".

This is easy to repro in Chrome with all extensions disabled:

1. https://hedy.org/for-teachers
2. Open an adventure
3. Adventure tab
4. Insert any text
5. Autosaves start failing in browser console

Image

The changes are not saved.

vamshikrishnaramasamy:
Closes #6561.

In `static/js/adventure.ts` we pass the CKEditor `change:data` callback's argument straight into `autoSave(...)`. That argument is a CKEditor `EventInfo`, not a DOM `Event`, and its `target` is the editor's document model — which has no `dataset`. The `(e.target as HTMLElement).dataset["autosaved"]` line in `autosave.ts` therefore throws `TypeError: Cannot read properties of undefined (reading 'autosaved')` every time the editor fires `change:data`, so the autosave never runs and the customize-adventure page can't save.

The fix is just to guard the cast at runtime with optional chaining. Real DOM events from the `input`/`change` listeners still hit the same early-return path; the CKEditor case now falls through to `formElement.requestSubmit()` like it was meant to.

### Verifying

1. Open a custom adventure with the editor
2. Type into the body — before the fix, console shows the TypeError and nothing saves; after the fix, the form submits and changes persist


jpelay (approved):
Thank you very much for the PR!

PR #6574, opened 2026-05-29
🪲 Normalize tabs to spaces when loading code into the editor (fixes #6525)

### LINKED ISSUE #6525: 🪲 Tab instead of spaces doesn't work
**Describe the bug**
A tab instead of spaces doesn't work for indentation. A tab in the code editor is automatically converted to 4 spaces. But a tab in example code from a custom/public adventure remains a tab and is not recognised as proper indentation.

Level 14
```
a = 1
b = 2
if a == b:
	print "Yahtzee!"
else
	print "Geen Yahtzee!"
```

Image

vamshikrishnaramasamy:
Fixes #6525.

A tab pressed in the editor is turned into the indent unit (4 spaces) by `indentWithTab`, so you can't end up with a literal tab by typing. But code that comes from a custom/public adventure or an example is loaded straight through the `contents` setter, and any tabs in it are inserted verbatim. Hedy treats those tabs as not-indentation, so a snippet like the level 14 if/else in the issue won't run after you paste it in.

The fix normalizes tabs to the indent unit inside the `contents` setter. That setter is the single point every load path goes through — example preview, the paste-example-code button, and switching adventure tabs — so loaded code now indents the same way typed code does.

I picked the setter rather than the individual call sites so the normalization can't be missed by a future load path, and reused the existing `indentSize` constant the editor already uses for `indentWithTab` and `indentUnit`.

Verified `npx tsc --noEmit` passes. The transformation itself is a deterministic replace of `\t` with the 4-space indent unit, matching what typing a tab already produces.


jpelay (approved):
Thank you!

PR #6578, opened 2026-05-29
🚚 Add a bunch of unit tests

rix0rrr:
Add a large number of tests to get the coverage of Python code up. These assert that
the code runs, and in some cases assert something about the values that are sent into
Jinja templates. These tests are (unfortunately) automatically generated, but this is an easy way of getting our coverage numbers up.

They are not asserting anything about the rendered output.

Measure coverage on tests.


rix0rrr:
pre-commit.ci autofix

jpelay (approved):
Nice work!

PR #6584, opened 2026-06-12
📜 Tidying up some content mistakes

### LINKED ISSUE #6431: 📜Haunted House adventure, level 1
The level explanation says we're supposed to ask the player about the monster, but that's not what we're doing in the exercise. The only question is about the door.

Image

Perhaps it's assumed that the programmer should only name or describe the monster behind the door instead of "asking". Or maybe the example question should be changed.

MarleenGilsing:
Fixes #6431 and some other small mistakes


MarleenGilsing:
@jpelay This is ready to go!

jpelay (approved):

Issue #5910, opened 2024-11-01
🖊️change 'echo' into a variable


AnneliesVlaar:
**Language idea incl level**
In level 1 the echo command could be a variable that stores the answer from 'ask'.

**Additional context**
The echo command is complex, it has two functions in one. It first prints text that is put beyond the command and also prints the answer to a question at the end of the printed text at the place which is invisible (shown in the purple circle). 

![Scherm­afbeelding 2024-11-01 om 14 36 32](https://github.com/user-attachments/assets/b97ad1e0-03d8-42be-be7f-39dc9d775edc)

If the answer to a question with 'ask' would always be stored in a variable with the name 'echo', the command 'echo' would only repeat the answer:

![Scherm­afbeelding 2024-11-01 om 14 44 30](https://github.com/user-attachments/assets/81b90235-ad0c-42fc-aebc-7ec6dc0b2e85)

Felienne:
Nice idea, but we now have it as it is and we are used to it :)

AnneliesVlaar:
idea of @MarleenGilsing to use 'answer' instead of 'echo' and to reuse 'answer' as a variable name in the first adventure(s) of level 2

MarleenGilsing:
<img width="1374" height="347" alt="Image" src="https://github.com/user-attachments/assets/658e4889-1a83-4842-a3fa-be431ea41ac0" />

**For the hackathon**

This is what the answer variable should look like. 
The command echo will be removed. The variable answer will take its place. answer will be a variable, not a keyword, so the students should still use print to print the answer (contraty to echo, where you didn't need print). 
answer should be highlighted light blue, like the variables in level 2.
Please note how nothing changes with ask in level 1, so you should still be able to use ask without a variable before it.

jpelay:
Fixed on #6594  thank you so much @koenelsdijk for y our help!

PR #6594, opened 2026-06-25
💻  Highlight level 1 answer keyword as variable

jpelay:
In support to #5910 we are going to highlight the `answer` keyword as a variable on level 1m but only if there's an `ask` command before.

So in this case:

```
ask what is your name
print answer
```

Answer would be highlighted. But in this case:

```
print answer! pretty name
```

It would not be highlighted. Also, it's worth pointing out that since this is a keyword, the translation of the variable will also be highlighted. For example in dutch both answer and antwoord would be highlighted.

**How to test**

There a few test cases I added, to test locally:

1. Go to level 1 and write this program:

```
ask what is your name
print answer
```

Answer should be highlighted. Take into consideration that since we haven't merged https://github.com/hedyorg/hedylang/pull/11 the actual interpolation will not yet be available.


jpelay:
I'm adding a feature flag to only enable this on alpha and local development for now.

PR #6597, opened 2026-06-27
Pythonisch level 13 16

AnneliesVlaar:
make all code in level 13 till 16 look pytonisch in the same way (no space between print/input and bracket, always use colons, change ask to input, always use = instead of is or == when it's a comparison, def instead of define)


AnneliesVlaar:
merged this branch into https://github.com/hedyorg/hedy/pull/6591

PR #6602, opened 2026-07-06
🪲 Offline: font defaults to arial since Balsamiq Sans is included via an unreachable online stylesheet (when no internet)

OlivierSwaak:
Style discrepancy due to downloaded stylesheet not accessible when offline.

Now still redundant due to inclusion of online stylesheet https://fonts.googleapis.com/css2?family=Balsamiq+Sans:ital,wght@0,400;0,700;1,400;1,700&display=swap at line 50 of `templates/layout.html`. If this successfully replaces this online stylesheet, this line can **and should** be removed!

Fixes no existing issue.

**How to test**

* Run `Build offline Hedy`
* Make sure your internet is turned off (otherwise the online download works anyways)
* Launch `run-hedy-server.exe`
* Compare the font on for example the main page (e.g. the large Hedy in the middle of the screen) with the online build.


static/fonts/balsamiqsans/balsamiqsans.css
```diff
@@ -0,0 +1,35 @@
...
+  src: url(../fonts/balsamiq_sans/BalsamiqSans-BoldItalic.ttf);
+  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+}
+
+@font-face {
+  font-family: 'Balsamiq Sans';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: url("../fonts/balsamiq_sans/BalsamiqSans-Regular.ttf");
```
jpelay: These are not working if we remove the online stylesheet, they have to be changed to 

```suggestion
  src: url("./BalsamiqSans-Regular.ttf");
```

jpelay (commented):
Hi Olivier,

Im not currently near a windows computer, so I couldnt test the offline build, but I did look into your question about removing the online stylesheet, and for that to work we need to change the paths in the balsamiqsans.css.

static/fonts/balsamiqsans/balsamiqsans.css
OlivierSwaak: Oh, you're right, these paths are still left over from when I had these declarations as a part of the `build-tools/tailwind/styles.css`. I shall fix this ASAP!

And apparently I also used the used a deprecated pathname, I have become a bit sloppy ;).

OlivierSwaak: Have updated this in the new commit!

jpelay: Awesome! Thanks!

jpelay:
What else do you want to do in this PR?

OlivierSwaak:
I think if we're positive this offline inclusion safely replaces the online stylesheet we used before (and does not mess anything up for the offline build) then this is everything within the scope of this PR. 

The only consideration I might have wrt maybe also keeping the online version of this font is if for some reason the font would be updated in the future, the online version will be kept up to date, while this one will 'deprecate' and will have to be updated manually. 

However, this did not seem to be a problem for the `fontawesome` fonts which are/were included similarly. 

Finally, I have looked at if any other fonts were missing but this does not seem to be the case, as the other fonts used (like `roboto`) seem to be standardly included.

OlivierSwaak:
I have removed some redundant lines which were necessary for the inclusion of the online font. So, if we're sure this can replace this online stylesheet, then this PR is ready to be merged!

jpelay:
> The only consideration I might have wrt maybe also keeping the online version of this font is if for some reason the font would be updated in the future, the online version will be kept up to date, while this one will 'deprecate' and will have to be updated manually.

I have also thought of this, but I think it's ok if we are left with a slightly out of date version. We can keep an eye, but is not crucial.

Also, I have tried building the offiline Hedy with the command:

```
doit run _offline
```

And the program it generates (without an exe, I have to add it myself) doesnt run. How do you test?

OlivierSwaak:
Oh, that's odd. Each time I try to test with both the build generated by the `Build Hedy Offline` CI and with `doit run _offline`. I'll have a look. Do you get any traceback? Sometimes, when I have an error in building, I just bin the entire `dist` folder and everything works afterwards.

OlivierSwaak:
For me the doit command seems to build successfully (at least on a Windows machine). Let me know if the problem persists, then I'll take another look at it!

jpelay (approved):
All good Olivier thanks for the fix!

PR #6617, opened 2026-07-17
📜 removes the content that tells the teacher which adventures to do

MarleenGilsing:
small content changes to the slides. 

I think i might have accidentally pushed this to main instead of a branch of my own. Hopefully this doesn't f-up anything... It seems like its okay... 😅

If the tests are okay, this is ready to go @jpelay


jpelay:
> I think i might have accidentally pushed this to main instead of a branch of my own. 

Hahahaha not a problem dont worry

PR #6632, opened 2026-07-29
🪲 Fix quoted list highlighting

### LINKED ISSUE #6623: 🪲Syntax highlighting with quotes in lists not consistent
**Describe the bug**
The highlighting of values in a list is inconsistent in case quotes are being used. 

This concerns level 11, probably some other levels as well.
See screenshot below for code and output when code is ecxecuted

Image

Hedy version 26.7.2

**Expected behavior**
Either one of the two following options seems logical to me:
1. An error "In this level quotes are not allowed when using lists" is displayed for line 3 and 4
2. The values for the list in line 3 and line 4 should be white, just like in line 1 and 2. The quotes should be white as well, as they are considered part of the value in the list. The komma in line 3 and 4 should be red, just like it is in line 1 and 2.

**Severity**
I have not seen any of my students asking questions about this, so I don't consider this a big problem. 
You may want to give this a low priority or not fix at all.

vamshikrishnaramasamy:
Quoted list values now parse consistently in levels 11 and 12. Includes a regression test and regenerated parsers.

Fixes #6623


highlighting/lezer-grammars/level11.grammar
```diff
@@ -8,7 +8,7 @@ Print { print ( String | ListAccess | Expression )+ }
 Ask { Text (is+ | Op<"=">) ask+ ( String | ListAccess | Expression )+ }
 
 // Commas in every language are allowed
-AssignList { Text (is+ | Op<"=">) (Text+ | Number ) ~ambig  (Op<',' | '،' | '，' | '、' > (Text+ | Number ))+ } 
+AssignList { Text (is+ | Op<"=">) (String | Text+ | Number ) ~ambig  (Op<',' | '،' | '，' | '、' > (String | Text+ | Number ))+ }
```
jpelay, line 11: Hello, @vamshikrishnaramasamy!

Thank you for the PR, but we are not going to move forward with this solution. If you look at the grammars, strings are not required for assigning text to a variable from up until level 13. Therefore, we dont need to add a String node to the AssignList rule.

What we actually need to do is the opposite and maybe you are interested in fixing this: stop highlighting strings on assignments and other places where they are not allowed.

jpelay (changes requested):

PR #6653, opened 2026-08-26
🪲 Make opening a level stick, also for second teachers

### LINKED ISSUE #6098: 🪲 Second teacher can't (always) open levels?
**Describe the bug**
Today we ran into the issue that a second teacher couldn't open a level, he tried it a bunch of times but it did not stick (despite seeing the popup "customization saved". I am the first teacher of the class, and for me it worked right away (class id: 2003976384a84343b856976cddc011a5)

jpelay:
Fixes #6098

Toggling a level on the configure class page posted the state of every checkbox, so a page left sitting open would revert whatever another teacher had changed in the meantime. Both teachers saw "customization saved". The toggle now sends only the level that moved.

Two more ways the same screen could claim a level was open when it wasn't:

- A level with an opening date scheduled in the future stayed closed, and the redesigned pages have no way to show or clear that date. Toggling a level now clears its date. Dates on levels you don't touch are kept — previously every save blanked `opening_dates`, `other_settings`, `level_thresholds` and `dashboard_customization`.
- A class that was never customized has no customizations record, so every level rendered as closed (they were all open) and the first toggle returned a 500. The page creates the record now, which also removes the dead block behind `migrate_customizations = False`.

Also drops two debug `print`s; the one in `datetotimeordate` ran on every `/hedy/<level>` request.

The old whole-list payload is still accepted so browsers on a cached bundle keep working. It can go a release or two from now.

### Tests

New cypress spec `class_configure_open_levels.cy.js`, 5 tests. With the fix reverted, 4 of the 5 fail.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jpelay:
pre-commit.ci run

Issue #6647, opened 2026-08-19
[UI idea] Improve accessibility for adventures


sstorck:
**Idea incl level**
I ran some basic Lighthouse accessibility checks and found a handful of issues.

**Additional context**
Accessibility checks can be run from the DevTools of most browsers.

sstorck:
Lighthouse report for print_command: https://googlechrome.github.io/lighthouse/viewer/?gist=8368d4b3d9215a987964c2a35f1d9c17

PR #6662, opened 2026-09-01
🚚 Improve `doit extract` for TypeScript localization

sstorck:
Related to #6647, this makes it easier to update strings that are only used in TypeScript.

- Gathers all missing `gettext` strings for client-messages.txt into an easily copy/pasteable block.
- Adds a dependency on client-messages.txt for `doit extract`.

**How to test**

1. Add multiple strings to `ADDITIONAL_GETTEXT_KEYS` in generate-client-messages.py.
2. Observe output of `doit run extract`.
3. Copy the output of the command into client-messages.txt.
4. Observe that `doit run extract` will extract all missing messages.

**Checklist**


jpelay (approved):
Nice! Thanks!

PR #6666, opened 2026-09-04
📜 Give every level of the teacher manual a "concepts and changes" setion

jpelay:
Given the changes we want to make to the for teachers manual, I propse adding a new section like this and modifying the current common errors one so that it can go nicely in the teacher manual and the fir teachers section,


MarleenGilsing:
Added a concepts and changes section to each level. Edited and added to the common mistakes sections. 

In English everything is up to date now in these sections. In Dutch it's done upto where the translations stopped (halfway level 7), so there is more translating work to do (for another day). 

I'll take another look at the teacher manual (What's Hedy etc etc) next week

MarleenGilsing:
Hi @jpelay 

I am done with all the content changes 😄  Everything has been translated into Dutch as well! 

I was done after the commit named 'whoop whoop' (I know I know, I was done naming them hahaha). My local host worked, which usually means all my mistakes are gone, but then my Github Desktop tells me to fetch and when i do these quotation marks reappear where they are not needed. And my local hosts tells me I made a mistake (the quotation marks) again.. Could you take a look? Thank youuuuu

PR #6667, opened 2026-09-04
💻 Add ARIA landmarks

sstorck:
#6647

Adds localized ARIA landmarks to the exercises, editor, and output to facilitate screen reader navigation.

**How to test**

Open a code-page and inspect the accessibility tree. Note the new regions.

**Checklist**

### DIFF
--- a/templates/incl/menubar.html
+++ b/templates/incl/menubar.html
@@ -1,4 +1,4 @@
-<nav aria-label="Website" class="px-4 lg:px-16 h-14 lg:h-16 w-full max-w-screen-2xl mx-auto print:hidden">
+<nav aria-label="{{ _(Website)}}" class="px-4 lg:px-16 h-14 lg:h-16 w-full max-w-screen-2xl mx-auto print:hidden">
     {% block menu %}
     <ul class="h-full w-full list-none flex flex-row items-center gap-x-2 xl:gap-x-6">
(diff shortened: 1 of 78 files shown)


templates/incl/menubar.html
```diff
@@ -1,4 +1,4 @@
-<nav aria-label="Website" class="px-4 lg:px-16 h-14 lg:h-16 w-full max-w-screen-2xl mx-auto print:hidden">
+<nav aria-label="{{ _(Website)}}" class="px-4 lg:px-16 h-14 lg:h-16 w-full max-w-screen-2xl mx-auto print:hidden">
```
jpelay: You're missing the quotation marks around website here! Also I know is not like a hard followed rule, but I would prefer if moving forward you keep the labels lowercase

jpelay (changes requested):
Thanks for the PR, sstork! Just a small comment related to a bug.

templates/incl/menubar.html
sstorck: Thanks for the catch! Looks like there was already a lowercase message for this.

jpelay (approved):
Thanks! Im sure this will improve Hedy for our users.

sstorck:
Any suggestions on how to proceed with the merge? I can't see how this change could have broken the python test.

PR #6691, opened 2026-09-18
🐛  Keyword translation fix

sstorck:
This script and corresponding YAML changes are the result of a discussion on Discord regarding bulk editing translation bugs.

The original bug was that `turn` was incorrectly escaped in [`adventures->turtle->levels->2->story_text_2`]((https://hosted.weblate.org/search/hedy/adventures/?q=adventures-%3Eturtle-%3Elevels-%3E2-%3Estory_text_2%0D%0A&sort_by=-priority%2Cposition&checksum=) using only backticks rather than backticks and curly braces. @jpelay pointed out that there may be the same issue with other keywords. While this was the only instance in `/content/adventures`, there were a lot of issues in the other projects.

Initially, I started from the `en.yaml` files to only edit the ones where the source was wrong, but I noticed that some languages had a lot more instances, e.g. `el.yaml`, so I decided to fix it across all of the files.

There are a few other opportunities to bulk edit these, but they'll require more effort to determine whether the keyword is currently in a code block and also prompt the user with options:
- For each keyword that is _not_ in a code block, optionally add the backticks around it. e.g. `{answer}` in [adventures->turtle->levels->2->story_text_2](https://hosted.weblate.org/translate/hedy/adventures/en/?checksum=3623e9a7d97b29e8)
- For each keyword in a code block that has no escaping, optionally add the curly braces. e.g. `{turn} right`.

NOTE: The script is hardcoded to work in GitHub Codespaces using `/workspace/` as the root.

**Checklist**


sstorck (commented):
FYI: I moved the CRLF fix into its own PR: https://github.com/hedyorg/hedy/pull/6695

